PR #4, opened 2021-08-20
Make devtools resizable

SimonBrandner:
Fixes https://github.com/vector-im/matrix-video-chat/issues/3


robertlong:
LGTM, thanks again @SimonBrandner! Also thanks for keeping the dependencies light. There's already `react-use-gesture` in the project so I appreciate that this doesn't use another external library for that.

PR #22, opened 2021-08-24
WebRTC Datachannels

robertlong:
This is a demo using datachannels in the room. It's still pretty basic, but I'd like to expand on it further and hopefully, eventually get it into matrix-js-sdk


robertlong:
This is now in the matrix-js-sdk group calls PR https://github.com/matrix-org/matrix-js-sdk/pull/1902

PR #166, opened 2021-12-20
Fix docker build

dbkr:
and add catchall route


robertlong:
Looks good. I haven't used nginx in a bit, but I assume this this returns 200s instead of 404s for the catchall by default? Can't remember. It should be doing 200s for now.

dbkr:
copying from the room: yep, it just serves up index.html with a 200

PR #206, opened 2022-02-07
Rename groupcall.txt -> groupcall.json.

michaelkaye:
This will stop groupcall.txt looking like it's a text log file and instead indicate it's an json artifact containing current state (which it is)

The file will still be stored on the rageshake server but the extension will indicate it's actually json data. We're already putting it in the "file" part of the rageshake, not the "logs" or "compressed-logs" parts, so it's already not meant to be handled as a log file.

(This is just a neatening up to make it easier to find the json data later)

### DIFF
--- a/src/settings/rageshake.js
+++ b/src/settings/rageshake.js
@@ -189,7 +189,7 @@ export function useSubmitRageshake() {
             body.append(
               "file",
               new Blob([JSON.stringify(json)], { type: "text/plain" }),
-              "groupcall.txt"
+              "groupcall.json"
             );
           }
         }


robertlong:
Awesome, thanks!

@michaelkaye I named the file `.txt` due to [these docs](https://github.com/matrix-org/rageshake#post-apisubmit)
> and ext is one of jpg, png, or txt

Should json be added to that list or is any extension allowed?

robertlong:
Update: I tested this locally after merging and I'm getting the following error in the`details.log`

> Attachment upload failures:
    Error saving groupcall.json: Invalid upload filename

robertlong:
Gonna revert for now

michaelkaye:
Urgh, that's annoying; sorry about that. I'll see if we can make the rageshake server grok json so it can describe what it has; i thought it could already.

robertlong:
Thanks for pushing on this 👍

PR #220, opened 2022-02-16
Add button identifiers.

pixlwave:
This PR adds a couple of identifiers to the mic and camera buttons for use in [pixlwave/MatrixVOIP](https://github.com/pixlwave/MatrixVOIP/) to help fix https://github.com/pixlwave/MatrixVOIP/issues/1. This would allow the web view to call `document.getElementById("microphoneButton").click()` instead of toggling the device capture state. (Also means that the additional overlay in the app can be removed and everything would behave a lot more "natively").

There are very likely better ways to achieve this (I have zero experience with React), happy to update in whatever way makes sense - in the future it would be cool if the possibility of observing the mic/camera states existed to toggle hardware LEDs etc.

### DIFF
--- a/src/button/Button.jsx
+++ b/src/button/Button.jsx
@@ -82,7 +82,7 @@ export const Button = forwardRef(
 export function MicButton({ muted, ...rest }) {
   return (
     <TooltipTrigger>
-      <Button variant="toolbar" {...rest} off={muted}>
+      <Button variant="toolbar" {...rest} off={muted} id="microphoneButton">
         {muted ? <MuteMicIcon /> : <MicIcon />}
       </Button>
       {() => (muted ? "Unmute microphone" : "Mute microphone")}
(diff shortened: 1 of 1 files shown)


robertlong:
In the room lobby and in the room itself, you should have access to the `groupCall` global.

You can do the following to toggle microphone / video mute:

```js
groupCall.isLocalVideoMuted();
groupCall.isMicrophoneMuted();
groupCall.setLocalVideoMuted(true);
groupCall.setMicrophoneMuted(true);
groupCall.setLocalVideoMuted(!groupCall.isLocalVideoMuted());
groupCall.setMicrophoneMuted(!groupCall.isMicrophoneMuted());
```

I think these methods will be more stable and better than emulating clicks on the buttons. Let me know if you'd like access to anything else for hotkeys.

PR #303, opened 2022-04-26
Add support for to-device messages via OLM

robertlong:
Depends on https://github.com/matrix-org/matrix-js-sdk/pull/2322

### DIFF
--- a/src/matrix-utils.js
+++ b/src/matrix-utils.js
@@ -3,6 +3,7 @@ import {
   GroupCallIntent,
   GroupCallType,
 } from "matrix-js-sdk/src/browser-index";
+import IndexedDBWorker from "./IndexedDBWorker?worker";
 
 export const defaultHomeserver =
   import.meta.env.VITE_DEFAULT_HOMESERVER ||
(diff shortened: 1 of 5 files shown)


src/matrix-utils.js
```diff
@@ -25,12 +28,69 @@ function waitForSync(client) {
...
+    const script = document.createElement("script");
+    script.setAttribute("src", src);
+    script.onload = resolve;
+    script.onerror = reject;
+    document.body.appendChild(script);
+  });
+}
+
 export async function initClient(clientOptions) {
+  await addScript(olmJsPath);
```
dbkr, line 42: Is this necessary as well as the hardcoded script addition in index.html? I think you should be able to import Olm as a module to avoid both (https://github.com/vector-im/element-web/blob/develop/src/vector/init.tsx#L87).

src/matrix-utils.js
```diff
@@ -25,12 +28,69 @@ function waitForSync(client) {
...
+
+  let indexedDB;
+
+  try {
+    indexedDB = window.indexedDB;
+  } catch (e) {}
+
+  const storeOpts = {};
+
+  if (indexedDB && localStorage && !import.meta.env.DEV) {
```
dbkr, line 53: Fwiw I think the bit you need is the crypto store below - this shouldn't really be necessary to get crypto working (although it will speed up your subsequent loads).

robertlong: Yeah Matthew and I discussed adding this to Element Call so I figured it was good to do it here in addition to the crypto store.

src/matrix-utils.js
robertlong: Ok, I've made a change to import it that way. One gotcha I found is that `@matrix-org/olm` expects `OLM_OPTIONS` to exist in the global context. Not sure why.

robertlong: I've added an issue to the Olm repository here https://gitlab.matrix.org/matrix-org/olm/-/issues/10

dbkr: Thanks - needing OLM_OPTIONS defined is a bit weird - we don't seem to need it in element-web - not sure why though.

dbkr (approved):

robertlong:
So we have a staging environment setup on main now. Should we merge this and test it out?

PR #307, opened 2022-04-29
Fix toggle button toggling

dbkr:
Just use isSelected directly rather than makking the button have its
own state. Also, the isPressed from useToggleButton looks like its
whether the user has the mouse button down on it or not rather than
whether the toggle switch is on, which was making the state wrong.


robertlong:
Yeah, this seems like a good workaround for what react-aria was doing 👍

PR #308, opened 2022-04-29
Put PTT behind 'feature flag'

dbkr:
AKA does the URL hash start with '#ptt'

This will let us merge PTT back to the main branch

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
-  "editor.formatOnSave": true,
+  "editor.formatOnSave": false,
   "editor.insertSpaces": true,
   "editor.tabSize": 2
 }
--- a/src/home/RegisteredView.jsx
+++ b/src/home/RegisteredView.jsx
@@ -13,6 +13,7 @@ import { JoinExistingCallModal } from "./JoinExistingCallModal";
 import { useHistory } from "react-router-dom";
 import { Headline, Title } from "../typography/Typography";
 import { Form } from "../form/Form";
+import { shouldShowPtt } from "../shouldShowPtt";
 
 export function RegisteredView({ client }) {
   const [loading, setLoading] = useState(false);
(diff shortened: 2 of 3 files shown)


src/home/RegisteredView.jsx
```diff
@@ -13,6 +13,7 @@ import { JoinExistingCallModal } from "./JoinExistingCallModal";
 import { useHistory } from "react-router-dom";
 import { Headline, Title } from "../typography/Typography";
 import { Form } from "../form/Form";
+import { shouldShowPtt } from "../shouldShowPtt";
```
robertlong, line 16: I believe this is missing from the PR?

.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-  "editor.formatOnSave": true,
+  "editor.formatOnSave": false,
```
robertlong, line 2: I wish we could leave this on. If you're using VSCode with prettier enabled it should "just work", I'd like to get to the bottom of this. As discussed we can likely move to the tooling I've set up for Third Room.

robertlong (changes requested):
I think the shouldShowPtt file wasn't checked in. Otherwise looks good to me, looking forward to being able to test this out on prod 👍

.vscode/settings.json
dbkr: Oops - I didn't mean to commit this. My vote would probably be to not commit this though, as these are often down to people's individual preference. If we move to prettier, this should make vscode's format-on-save redundant anyway.

src/home/RegisteredView.jsx
dbkr: oops - fixed

.vscode/settings.json
robertlong: Yup sounds good. I see the other linting PR which should fix this.

robertlong:
I made one more change, I moved to using a hook that uses `useLocation`. If you just edit your url without a refresh the `#ptt` hash change will be detected.

robertlong (approved):
Lgtm 👍

PR #311, opened 2022-05-04
Fix mouseup/down behaviour of PTT button

dbkr:
Handle mouseup events anywhere so the button releases if you move
the cursor out of the button & release. Likewsie for window losing
focus.


robertlong (approved):
Looks good 👍  I think the document mouseup event handler makes sense

PR #313, opened 2022-05-05
Return to normal state when time limit reached

dbkr:
Based on https://github.com/vector-im/element-call/pull/311

Fixes https://github.com/vector-im/element-call/issues/297

### DIFF
--- a/src/room/usePTT.js
+++ b/src/room/usePTT.js
@@ -17,6 +17,8 @@ export function usePTT(client, groupCall, userMediaFeeds) {
     };
   });
 
+  const [unmuteError, setUnmuteError] = useState(null);
+
   useEffect(() => {
     function onMuteStateChanged(...args) {
       const activeSpeakerFeed = userMediaFeeds.find((f) => !f.isAudioMuted());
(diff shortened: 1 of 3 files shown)


src/room/usePTT.js
```diff
@@ -47,29 +49,35 @@ export function usePTT(client, groupCall, userMediaFeeds) {
     };
   }, [userMediaFeeds]);
 
-  const startTalking = useCallback(() => {
+  const startTalking = useCallback(async () => {
+    setState((prevState) => ({ ...prevState, pttButtonHeld: true }));
+    setUnmuteError(null);
```
robertlong, line 54: I might store the `unmuteError` in `state` instead of it's own `useState`.

robertlong (approved):
I'd make that one small change, otherwise looks great 👍

src/room/usePTT.js
dbkr: OK - I flipped back & forth between staying consistent and splitting it out since merging with prevState each time seemed quite boilerplate-y, but since it's already setting other state in one case anyway maybe consistency is better.

PR #314, opened 2022-05-05
Set formatter to prettier

dbkr:
Although this will only work with the extension installed


robertlong (approved):
Yup, makes sense. Thanks for tracking that down!

PR #316, opened 2022-05-06
Initial round of typescripting

dbkr:
This switches to using my group-call branch with js-sdk develop merged in: https://github.com/matrix-org/matrix-js-sdk/pull/2355 . This will need more tests before it can be considered for merging into js-sdk. The merge was distinctly nontrivial so it will need decent testing before going live.


robertlong:
Looks good 👍 I added one commit which swaps out the manual vite/svgr types and imports them from their respective libraries.

dbkr:
Ah, thanks for adding the proper types for the imports - all my googling failed to find the proper way to do that!

PR #317, opened 2022-05-10
Add GHA to deploy main branch to Netlify

dbkr:


robertlong:
Deploying to netlify from GHA is much simpler than I thought it'd be. Thanks for doing this! I'll likely do the same for Third Room. Also with the new [job summaries feature](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) this is honestly preferable so that I can add additional info like a link to the storybook.

robertlong (approved):

PR #318, opened 2022-05-11
Add redirects to netlify deployment

dbkr:


ara4n:
only comment is that it feels a bit weird that a redirect config file is sitting in /scripts (which would typically be executable scripts, no?)

dbkr:
yeah - wasn't sure where else to put it: a '/support' directory or somewhere?

ara4n:
perhaps `config` (and move default.conf into there too?)

ara4n:
lgtm otherwise.

dbkr:
oh yeah, that's probably nicer

PR #319, opened 2022-05-11
Add push-to-talk sound effects

dbkr:
With some hastily made placeholder sounds. The 'blocked' sound is probably a bit over-aggressive but they're placeholders so I'll let it go and leave it to an actual sound designer!

Fixes https://github.com/vector-im/element-call/issues/296


src/room/usePTT.ts
```diff
@@ -147,8 +188,8 @@ export const usePTT = (
     }
 
     window.addEventListener("keydown", onKeyDown);
-    window.addEventListener("keyup", onKeyUp);
-    window.addEventListener("blur", onBlur);
+    //window.addEventListener("keyup", onKeyUp);
```
robertlong, line 191: Are these meant to be left commented out? Releasing spacebar no longer works without this.

robertlong (changes requested):
Yeah the call blocked sound effect is pretty aggressive 😅 but the sound effects seem to work alright. However I noticed a few lines are commented out that break the PTT functionality.

src/room/usePTT.ts
dbkr: Oops - good catch. This is how I'd been testing the admin talkover without two physical computers, but forgot to put it back afterwards. 🤦

dbkr:
Assuming this is good now & merging.

PR #325, opened 2022-05-13
Mute local mic if blocked and fix races on mute / unmute

dbkr:
* Mute our own microphone if we're blocked from speaking by another user
 * Serialise mute/unmute operations on a promise chain to fix races

Based on https://github.com/vector-im/element-call/pull/322

### DIFF
--- a/src/room/usePTT.ts
+++ b/src/room/usePTT.ts
@@ -119,7 +119,11 @@ export const usePTT = (
       activeSpeakerUserId === client.getUserId() &&
       activeSpeakerFeed?.userId !== client.getUserId()
     ) {
-      // We were talking but we've been cut off
+      // We were talking but we've been cut off: mute our own mic
+      // (this is the easier way of cutting other speakers off if an
+      // admin barges in: we could also mute the non-admin speaker
+      // on all receivers, but we'd have to make sure we unmuted them
+      // correctly.)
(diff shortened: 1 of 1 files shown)


src/room/usePTT.ts
```diff
@@ -39,6 +63,24 @@ export const usePTT = (
...
+  // Used to serialise all the mute calls so they don't race. It has
+  // its own state as its always set separately from anything else.
+  const [mutePromise, setMutePromise] = useState(Promise.resolve());
+
+  // Wrapper to serialise all the mute operations on the promise
+  const setMicMuteWrapper = useCallback(
+    (muted) => {
+      setMutePromise(
+        mutePromise.then(() => {
+          groupCall.setMicrophoneMuted(muted).catch((e) => {
```
robertlong, line 75: Are you supposed to be returning this promise here so that it properly awaits the `setMicrophoneMuted` promise?

robertlong (changes requested):
Looks like the `setMicMuteWrapper` function might have a bug? Otherwise looks good 👍

src/room/usePTT.ts
dbkr: Yes! Good catch.

robertlong (approved):

robertlong:
Looks good 👍

PR #326, opened 2022-05-13
Fixes for touch interface on push-to-talk button

dbkr:
* Avoid also getting a 'mousedown' event by making the event listener
   non-passive so the preventDefault() works
 * Remember the touch that pressed the button so we only un-press
   when that touch ends, otherwise the button gets released if the
   user taps the screen anywhere else.


robertlong (approved):
Looks good 👍

PR #327, opened 2022-05-13
Add sound when speaker stops speaking

dbkr:
And also a slightly nicer blocked sound (ok, I couldn't let it go).

Also unifies a few types on null rather than undefined.

Based on https://github.com/vector-im/element-call/pull/325


robertlong (approved):
Seems good 👍 Also I like the new blocked sound more 🙂

PR #333, opened 2022-05-17
Bump js-sdk dependency

dbkr:

PR #337, opened 2022-05-18
Bump js-sdk dependency to encrypt to-device messages

dbkr:
Fixes https://github.com/vector-im/element-call/issues/257

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "color-hash": "^2.0.1",
     "events": "^3.3.0",
     "lodash-move": "^1.1.1",
-    "matrix-js-sdk": "github:matrix-org/matrix-js-sdk#acef1d7dd0b915368730efabee94deb42b2e4058",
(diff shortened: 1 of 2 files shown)


dbkr:
Probably want tov wait for https://github.com/matrix-org/matrix-js-sdk/pull/2383 and then bump this to include that

robertlong:
Now I'm confused. Which version of js-sdk is in this PR https://github.com/vector-im/element-call/pull/333

We definitely should wait to roll out E2E if it breaks when the other members don't have E2E keys.

dbkr:
#333 was to pull in the fix for races when muting / unmuting - I've closed it in favour of this one (which is a bit of an arbitrary choice since this one will want to be changed to a new sha once that js-sdk PR is merged).

robertlong (approved):
Looks good 👍 Let's test more on staging.

PR #340, opened 2022-05-18
Fix soft crash when opening invite modal in lobby

robintown:
It was exploding on `roomId` being `undefined`.


dbkr (approved):
Looking forward to having types on these :)

PR #346, opened 2022-05-22
Add linux/arm64 Docker image

Kalissaac:
Notes: none

Uses QEMU and Buildx since GitHub Actions doesn't have support for ARM-based runners. Unfortunately, as a result the build time increases to about 20 minutes.


dbkr:
Ideally there'd be a way to do this that just builds one extra, platform-independent layer and adds it on top of the different platform images for nginx. Failing that, we could use the "cross-compilation" method (https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/) (ie. basically just add `--platform=$BUILDPLATFORM` to the `FROM node:16-buster as builder` line, I think) although I assume this would still run the vite build n times rather than once when building for n platforms. Let's not run the vite build under emulation though, please.

dbkr:
Also, don't forget to add DCO signoff.

Kalissaac:
I was able to get it running with the cross-compilation method, from what I can tell it runs the vite build once and then copies the files over to each of the nginx images. I'll add the commit sign offs and push again, sorry about that.

dbkr (approved):
Thanks!

PR #358, opened 2022-05-30
Convert matrix-utils to typescript

dbkr:


src/matrix-utils.ts
```diff
@@ -1,34 +1,49 @@
...
   });
 }
 
-export async function initClient(clientOptions) {
+export async function initClient(
+  clientOptions: ICreateClientOpts
+): Promise<MatrixClient> {
   // TODO: https://gitlab.matrix.org/matrix-org/olm/-/issues/10
   window.OLM_OPTIONS = {};
   await Olm.init({ locateFile: () => olmWasmPath });
```
robintown, line 49: GitHub won't let me comment on the `let indexedDB;` two lines below this, but can it be typed?

src/matrix-utils.ts
```diff
@@ -161,16 +180,16 @@ export async function createRoom(client, name, isPtt = false) {
   console.log({ isPtt });
```
robintown, line 180: While you're here, this should probably be an actual log message rather than a raw object

src/matrix-utils.ts
```diff
@@ -161,16 +180,16 @@ export async function createRoom(client, name, isPtt = false) {
...
   await client.createGroupCall(
-    room_id,
+    createRoomResult.room_id,
     isPtt ? GroupCallType.Voice : GroupCallType.Video,
     isPtt,
     GroupCallIntent.Prompt
   );
 
-  return room_alias || room_id;
+  return createRoomResult.room_id;
```
robintown, line 189: What's the reason for this change? I think this means you'll now no longer get the room alias in the URL when creating a room

dbkr: Good catch - I took this out as it wasn't in the types (it's unspecced) but yeah, it does need the room alias. I'm not sure why it fell back to the ID though, so have removed that.

dbkr:
There are quite a few more changes here now, sorry!

robintown (approved):

PR #359, opened 2022-05-30
Lower timeout on js-sdk API call to 5s

dbkr:
Based on https://github.com/vector-im/element-call/pull/358


src/matrix-utils.ts
```diff
@@ -80,6 +80,9 @@ export async function initClient(
     ...storeOpts,
     ...clientOptions,
     useAuthorizationHeader: true,
+    // Use a relatively low timeout for API calls: this is a realtime
```
robintown, line 83:
```suggestion
    // Use a relatively low timeout for API calls: this is a realtime application
```

robintown (approved):

PR #360, opened 2022-05-30
Sort call feeds consistently when choosing active speaker

dbkr:
As hopefully explained by comment

For https://github.com/vector-im/element-call/issues/331

### DIFF
--- a/src/room/usePTT.ts
+++ b/src/room/usePTT.ts
@@ -31,7 +31,11 @@ function getActiveSpeakerFeed(
   const activeSpeakerFeeds = feeds.filter((f) => !f.isAudioMuted());
 
   // make sure the feeds are in a deterministic order so every client picks
-  // the same one as the active speaker
+  // the same one as the active speaker. The custom sort function sorts
+  // by user ID, so needs a collator of some kind to compare. We make a
+  // specific one to help ensure every client sorts the same way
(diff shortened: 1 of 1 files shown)


src/room/usePTT.ts
```diff
@@ -30,6 +30,17 @@ function getActiveSpeakerFeed(
...
+  // make sure the feeds are in a deterministic order so every client picks
+  // the same one as the active speaker
+  const collator = new Intl.Collator("en", {
+    sensitivity: "variant",
+    usage: "sort",
+    ignorePunctuation: false,
+  });
+  activeSpeakerFeeds.sort((a: CallFeed, b: CallFeed): number =>
+    collator.compare(a.userId, b.userId)
+  );
```
robintown, line 42: Why isn't `activeSpeakerFeeds.sort()` deterministic on its own?

robintown (commented):
a question

src/room/usePTT.ts
robintown: Nevermind, was blind to the fact that we're sorting by the `userId` key

robintown (approved):

src/room/usePTT.ts
dbkr: To be fair, this could be better commented - I'll add one.

PR #361, opened 2022-05-30
Convert useGroupCall to TS

dbkr:


src/room/useGroupCall.ts
```diff
@@ -18,10 +18,57 @@ import { useCallback, useEffect, useState } from "react";
...
+  GroupCall,
 } from "matrix-js-sdk/src/webrtc/groupCall";
+import { MatrixCall } from "matrix-js-sdk/src/webrtc/call";
+import { CallFeed } from "matrix-js-sdk/src/webrtc/callFeed";
+import { RoomMember } from "matrix-js-sdk/src/models/room-member";
+
 import { usePageUnload } from "./usePageUnload";
 
-export function useGroupCall(groupCall) {
+export interface UseGroupCallType {
```
dbkr, line 29: If anyone has any ideas or knows of conventions on what to call this type, I'd love it to have a better name.

src/room/useGroupCall.ts
```diff
@@ -41,20 +88,25 @@ export function useGroupCall(groupCall) {
...
+    localScreenshareFeed: null,
+    localDesktopCapturerSourceId: null,
     requestingScreenshare: false,
     participants: [],
     hasLocalParticipant: false,
   });
 
-  const updateState = (state) =>
+  const updateState = (state: Partial<State>) =>
     setState((prevState) => ({ ...prevState, ...state }));
```
robintown, line 110: Given that all updates to state happen through this function, this really feels to me like a case where we should just use individual state hooks

robintown (approved):
Looks good either way

src/room/useGroupCall.ts
robintown: `GroupCallInfo` is all I can think of

src/room/useGroupCall.ts
dbkr: Yeah, or a least a few different grouped ones.

PR #363, opened 2022-05-30
Show when connection is lost on PTT mode

dbkr:
https://github.com/vector-im/element-call/issues/362 filed for better UI

### DIFF
--- a/src/room/usePTT.ts
+++ b/src/room/usePTT.ts
@@ -15,10 +15,11 @@ limitations under the License.
 */
 
 import { useCallback, useEffect, useState } from "react";
-import { MatrixClient } from "matrix-js-sdk/src/client";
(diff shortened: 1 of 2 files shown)


src/room/usePTT.ts
```diff
@@ -211,6 +217,17 @@ export const usePTT = (
...
+  const [connected, setConnected] = useState(true);
+
+  const onClientSync = useCallback(
+    (syncState: SyncState) => {
+      setConnected(syncState !== SyncState.Error);
+    },
+    [setConnected]
+  );
+
   useEffect(() => {
```
robintown, line 231: I think we should split this `useEffect` up, so it's not removing and re-adding the sync listener every time the button gets held down etc.

robintown (approved):
Looks good otherwise

PR #366, opened 2022-05-31
Fix crash when setting audio output on Chrome for Android

robintown:
According to https://github.com/vector-im/element-call/issues/364 Chrome only offers a single choice of audio output, so it's probably still fine to lie to the user and show the dropdown?

Closes https://github.com/vector-im/element-call/issues/364


dbkr (approved):
Yeah, it seems relatively unlikely that a browser would offer multiple outputs and the *not* implement setSinkID!

PR #367, opened 2022-05-31
Add a VU meter-style animation to radio mode

robintown:
https://user-images.githubusercontent.com/48614497/171291580-48286ae7-c1d9-48d2-a79a-fd2a9c859aa2.mp4

### DIFF
--- a/src/room/PTTButton.tsx
+++ b/src/room/PTTButton.tsx
@@ -16,6 +16,7 @@ limitations under the License.
 
 import React, { useCallback, useEffect, useState, createRef } from "react";
 import classNames from "classnames";
+import { useSpring, animated } from "@react-spring/web";
 
 import styles from "./PTTButton.module.css";
 import { ReactComponent as MicIcon } from "../icons/Mic.svg";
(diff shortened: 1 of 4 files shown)


src/room/PTTButton.tsx
```diff
@@ -130,12 +133,32 @@ export const PTTButton: React.FC<Props> = ({
...
+  const { shadow } = useSpring({
+    shadow: (Math.max(activeSpeakerVolume, -70) + 70) * 0.5,
+    config: {
+      clamp: true,
+      tension: 300,
+    },
+  });
+  const shadowColor = showTalkOverError
+    ? "rgba(255, 91, 85, 0.2)"
+    : "rgba(13, 189, 139, 0.2)";
```
dbkr, line 146: Could we use a CSS variable? I think they should be supported natively on all our target browsers, so I think it ought to be fine? Otherwise, maybe just have them as consts with a comment saying what they are?

dbkr (approved):
Otherwise looks great!

PR #370, opened 2022-06-01
Fix app when built in production mode

dbkr:
The recent typescripting appears to have caused the typescript
compiler to get confused about dependency references and start
refwrencing things like CRYPTO_ENABLED in the js-sdk before it's
defined them.

This avoids using things from the (javascript) browser-index import
and instead pulls everything in from the typescript files, then
fixes the resulting type failures, (in some cases with hacks).

Based on https://github.com/vector-im/element-call/pull/358

### DIFF
--- a/src/auth/useInteractiveLogin.ts
+++ b/src/auth/useInteractiveLogin.ts
@@ -47,7 +47,7 @@ export const useInteractiveLogin = () =>
 
     // XXX: This claims to return an IAuthData which contains none of these
     // things - the js-sdk types may be wrong?
-    /* eslint-disable camelcase */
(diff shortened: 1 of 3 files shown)


src/auth/useInteractiveLogin.ts
```diff
@@ -15,41 +15,55 @@ limitations under the License.
...
+            user: username,
+          },
+          password,
+        }),
+      stateUpdated: null,
+      requestEmailToken: null,
+    });
+
+    // XXX: This claims to return an IAuthData which contains none of these
+    // things - the js-sdk types may be wrong?
```
dbkr, line 49: Filed as https://github.com/vector-im/element-call/issues/371

robintown (approved):

PR #372, opened 2022-06-01
Organize colors

robintown:
This organizes and renames all the color variables to the names used on [Compound](https://www.figma.com/file/X4XTH9iS2KGJ2wFKDqkyed/Compound?node-id=557%3A0), and replaces hardcoded values with variables in a few spots.


SimonBrandner:
Yes, please, let's not make the same mistake as in element-web! :heart:

dbkr (approved):
Definitely - good call.

PR #373, opened 2022-06-02
'Webcam' → 'Camera'

robintown:
I'm pretty sure this is a design call we can just make, since the well-defined parts of the designs only ever use the word 'camera', never 'webcam'.

Closes https://github.com/vector-im/element-call/issues/344


SimonBrandner (approved):
I am not in the group to approve, am I? :sweat_smile:

robintown:
Nope, but I'm sure you could be, maybe ping someone in EC internal?

PR #387, opened 2022-06-08
typescript src/settings

toger5:
ModalProps fixes
LogEntry interface
missing return promise


src/settings/SettingsModal.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
```
toger5, line 1: This is for `e.target.checked`. EventTarget does not have `checked` defined. So `// @ts-ignore` is used.

src/settings/rageshake.ts
```diff
@@ -37,37 +38,58 @@ limitations under the License.
...
+        };
+      }
+    );
   }
 
-  log(level, ...args) {
+  // these functions get overwritten by the monkey patch
+  error(...args: unknown[]): void {}
+  warn(...args: unknown[]): void {}
+  info(...args: unknown[]): void {}
```
toger5, line 90: is there a better way?

src/settings/SettingsModal.tsx
```diff
@@ -91,7 +100,8 @@ export const SettingsModal = (props) => {
               type="checkbox"
               checked={spatialAudio}
               description="This will make a speaker's audio seem as if it is coming from where their tile is positioned on screen. (Experimental feature: this may impact the stability of audio.)"
-              onChange={(e) => setSpatialAudio(e.target.checked)}
+              // @ts-ignore
+              onChange={(event: Event) => setSpatialAudio(event.target.checked)}
```
dbkr, line 104: I think you should be able to find an event type that matches - `React.ChangeEvent<HTMLInputElement>` or something, probably.

src/settings/rageshake.ts
dbkr: I'm not immediately seeing why these need to be defined here, but this file is basically taken from react-sdk so I'd just update it to steal the typescript version: https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/rageshake/rageshake.ts

src/settings/useMediaHandler.tsx
```diff
@@ -177,35 +207,36 @@ export function MediaHandlerProvider({ client, children }) {
     [client]
   );
 
-  const setAudioOutput = useCallback((deviceId) => {
+  const setAudioOutput: (deviceId: any) => void = useCallback((deviceId) => {
```
dbkr, line 210: Also string I think

dbkr (commented):
You might have picked one of the harder ones to start with since this has the rageshake code! It's nearly there though.

src/settings/rageshake.ts
toger5: It is defined so that `this.logger.error(err);` does not raise an error:
`Property 'error' does not exist on type 'ConsoleLogger'.`

toger5: Okay. Thanks for sharing the react-sdk/rageshake file. It seems it should not use this.logger (the `ConsoleLogger` but instead the imported `PrefixedLogger` (without `.this`).
I will use most of the react-sdk version then.

src/settings/useMediaHandler.tsx
```diff
@@ -72,7 +95,9 @@ export function MediaHandlerProvider({ client, children }) {
     );
 
     return {
+      // @ts-ignore
```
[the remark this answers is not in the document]
toger5: This is for ignoring the private check. `audioInput` and `videoInput` are private members and I was not sure If I should change that in the js-sdk?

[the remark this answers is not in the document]
toger5: Maybe there is a better solution to the ignore. For now I added a comment so it is at least clear why I added the ignore.

src/settings/useMediaHandler.tsx
```diff
@@ -72,7 +96,9 @@ export function MediaHandlerProvider({ client, children }) {
     );
 
     return {
+      // @ts-ignore, ignore that audioInput is a private members of mediaHandler
       audioInput: mediaHandler.audioInput,
+      // @ts-ignore, ignore that videoInput is a private members of mediaHandler
```
dbkr, line 101: thanks for the comment - it's good to mark what's going on with these things so we can later fix either the call or the library, depending on which we think is correct.

dbkr (approved):

dbkr:
@toger5 you can merge this now it's approved (you should have write access to do so)

PR #391, opened 2022-06-09
Add warning if incompatible versions are being used

dbkr:
This will probably be overly sensitive until we start timing out
member events (ie. https://github.com/matrix-org/matrix-js-sdk/pull/2446
lands) because lots of calls might have old member events from people
who've joined previously.

Screenshot 2022-06-09 at 21 55 11

Requires https://github.com/matrix-org/matrix-js-sdk/pull/2447
Fixes https://github.com/vector-im/element-call/issues/383


src/IncompatibleVersionModal.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { Modal, ModalContent } from "./Modal";
+import { Body } from "./typography/Typography";
+
+interface Props {
+  userIds: Set<string>;
+  room: Room;
+}
+
+export const IncompatibleversionModal: React.FC<Props> = ({
```
robintown, line 28: The V should be capitalized

src/index.css
```diff
@@ -29,6 +29,7 @@ limitations under the License.
   --accent-20: #0dbd8b33;
   --alert: #ff5b55;
   --alert-20: #ff5b5533;
+  --warning: #e8bf37;
```
robintown, line 32: This isn't in our design systems from what I can see, supposedly Alert is to be used for all "content requiring user attention"

src/Header.jsx
```diff
@@ -84,3 +87,25 @@ export function RoomSetupHeaderInfo({ roomName, avatarUrl, ...rest }) {
...
+
+  const onDetailsClick = useCallback(() => {
+    modalState.open();
+  }, [modalState]);
+
+  if (users.size === 0) return null;
+
+  return (
+    <span className={styles.versionMismatchWarning}>
+      Incomaptible versions!
```
robintown, line 102: You have a typo

src/button/Button.module.css
```diff
@@ -207,3 +207,10 @@ limitations under the License.
 .lg {
   height: 40px;
 }
+
+.linkButton {
+  background-color: transparent;
+  border: none;
+  text-decoration: underline;
+  cursor: pointer;
+}
```
robintown, line 216: I think what our design systems have to say about this is that link buttons should be green with no underline, like in Element Web

robintown (changes requested):

src/icons/AlertTriangleFilled.svg
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg fill="none" version="1.1" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
+<path d="m12 2.8965c-0.6991 0-1.3483 0.36517-1.7109 0.96289l-8.4687 14.141c-0.35536 0.6154-0.35777 1.3729-0.0058594 1.9902 0.3519 0.6174 1.0043 1.002 1.7148 1.0098h16.941c0.7106-0.0078 1.3629-0.39237 1.7148-1.0098 0.3519-0.6173 0.34954-1.3748-0.005859-1.9902l-8.4688-14.141c-0.3625-0.59772-1.0118-0.96289-1.7109-0.96289zm0 5.6035a0.5 0.5 0 0 1 0.5 0.5v4a0.5 0.5 0 0 1-0.5 0.5 0.5 0.5 0 0 1-0.5-0.5v-4a0.5 0.5 0 0 1 0.5-0.5zm0 7.5a1 1 0 0 1 1 1 1 1 0 0 1-1 1 1 1 0 0 1-1-1 1 1 0 0 1 1-1z" fill="#000"/>
+</svg>
```
robintown, line 4: This seems to be different from the icon in Compound?

Screenshot 2022-06-09 at 17-58-24 Figma

src/index.css
dbkr: True - I stole it from one of the e2e badges, although probably better to stick with the colour palette.

src/icons/AlertTriangleFilled.svg
dbkr: Ah, for some reason I only looked at the icons we were using in element-web rather than looking in compound - replaced.

robintown (approved):
Looks good!

PR #397, opened 2022-06-11
typescript `src/button`

toger5:


toger5:
the current linter issues are realted to `node_modules/@react-aria/button/dist/types.d.ts` There are indeed imports missing in that file but it is not part of this repo? What would be a reasonable solution?

dbkr (approved):

PR #399, opened 2022-06-13
Make AEC work with spatial audio on Chrome

robintown:
Closes https://github.com/vector-im/element-call/issues/381


dbkr (approved):
otherwise lgtm

PR #418, opened 2022-06-28
Catch an exception & add log line on setsinkid

dbkr:
Ported over from the postmessage PTT branch


SimonBrandner (approved):
Otherwise LGTM

PR #435, opened 2022-07-03
Produce a more informative error when running in an insecure context

robintown:
So that people are more likely to succeed at answering their own support questions like https://github.com/vector-im/element-call/issues/302

### DIFF
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -36,6 +36,14 @@ initRageshake();
 
 console.info(`matrix-video-chat ${import.meta.env.VITE_APP_VERSION || "dev"}`);
 
+if (!window.isSecureContext) {
+  throw new Error(
+    "This app cannot run in an insecure context. To fix this, access the app " +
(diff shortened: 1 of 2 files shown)


src/main.tsx
```diff
@@ -36,6 +36,14 @@ initRageshake();
 
 console.info(`matrix-video-chat ${import.meta.env.VITE_APP_VERSION || "dev"}`);
 
+if (!window.isSecureContext) {
+  throw new Error(
+    "This app cannot run in an insecure context. To fix this, access the app " +
+      "via a local loopback address, or serve it over HTTPS.\n" +
+      "https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts"
```
SimonBrandner, line 43: The indent here seems a little of

SimonBrandner (approved):

src/main.tsx
robintown: I didn't put it there by choice, Prettier did :(

PR #438, opened 2022-07-04
Add config param to disable e2e for signalling

dbkr:
Deliberately not making it an option on the room since then we'd have to worry about server doing downgrade attacks.

Requires https://github.com/matrix-org/matrix-js-sdk/pull/2492

### DIFF
--- a/src/matrix-utils.ts
+++ b/src/matrix-utils.ts
@@ -14,6 +14,7 @@ import {
   GroupCallType,
 } from "matrix-js-sdk/src/webrtc/groupCall";
 import { ISyncStateData, SyncState } from "matrix-js-sdk/src/sync";
+import { logger } from "matrix-js-sdk/src/logger";
 
 import IndexedDBWorker from "./IndexedDBWorker?worker";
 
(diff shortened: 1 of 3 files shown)


src/matrix-utils.ts
```diff
@@ -79,13 +80,27 @@ export async function initClient(
...
   }
 
+  // XXX: we read from the URL search params in RoomPage too:
+  // it would be much better to read them in one place and pass
+  // the values around, but we initialise the matrix client in
+  // many different places so we'd have to pass it into all of
+  // them.
+  const params = new URLSearchParams(window.location.search);
+  // disable e2e only if enableE2e=false is given
+  const enableE2e = !(params.get("enableE2e") === "false");
```
SimonBrandner, line 90:
```suggestion
  const enableE2e = params.get("enableE2e") !== "false";
```

Looks more clear

SimonBrandner (commented):
See https://github.com/matrix-org/matrix-js-sdk/pull/2492#pullrequestreview-1027860028

src/matrix-utils.ts
```diff
@@ -79,13 +80,27 @@ export async function initClient(
...
   }
 
+  // XXX: we read from the URL search params in RoomPage too:
+  // it would be much better to read them in one place and pass
+  // the values around, but we initialise the matrix client in
+  // many different places so we'd have to pass it into all of
+  // them.
+  const params = new URLSearchParams(window.location.search);
+  // disable e2e only if enableE2e=false is given
+  const enableE2e = params.get("enableE2e") !== "false";
```
SimonBrandner, line 90: What happens if someone tries to join an encrypted call without encryption?

dbkr: There aren't really 'encrypted calls' as such: clients with this flag set to false will still decrypt any encrypted signalling they receive, so (I think) it should all still work. However, this is very deliberately hidden in an undocumented URL param, so anyone playing with it gets to keep both pieces if it breaks. :)

SimonBrandner: Ah, I see

SimonBrandner (approved):

Issue #324, opened 2022-05-13
PTT – Input device selections remain empty in EW / EA integration


Johennes:
**Describe the bug**

Testing the EW / EA integration in https://matrix.to/#/#ecwidgettest:matrix.org, the input device selections in EA remain completely empty.

**Expected behavior**

Dropdowns should either be disabled or not empty.

**Screenshots**

![Screenshot_20220513-125046](https://user-images.githubusercontent.com/1137962/168282384-38dc6bac-f4e2-48e2-8cac-a845ae9a9022.png)
![Screenshot_20220513-125053](https://user-images.githubusercontent.com/1137962/168282393-7be20a09-75fc-4309-9132-837d78d22e09.png)
![Screenshot_20220513-125101](https://user-images.githubusercontent.com/1137962/168282397-3c05410f-af5e-46ee-9461-663c578f7639.png)

robintown:
Sometimes browsers return blank labels for all devices, depending on how the user granted permissions. I don't know if there's a way to work around this in the EA integration in particular, but I think the only reasonable thing we can do in this case is disable/hide the device switcher dropdown since we don't have enough info about the devices.

Johennes:
If there's just one input device, I think hiding is fine. If there are more than one (e.g. front and rear camera), maybe we could simply enumerate them à la "Camera 1", "Camera 2", ...? At least the user would still be able to pick one that way.

PR #449, opened 2022-07-07
Prevent empty device labels

Johennes:
Fixes: #324

This isn't very elegant but at least prevents the empty dropdowns on Android. Possibly should move into `MediaHandlerProvider`? Open to feedback. 🙂


SimonBrandner (approved):
Looks sane

dbkr:
NB. I have 3 different devices in the list, all blank, and I think Matthew had similar, so I think this will end up with 3 "Default Microphone"s.

Johennes:
> NB. I have 3 different devices in the list, all blank, and I think Matthew had similar, so I think this will end up with 3 "Default Microphone"s.

Ok, sorry I wasn't aware. Changed it to enumerate the devices now.

PR #457, opened 2022-07-08
yarn upgrade

dbkr:
Fixes https://github.com/vector-im/element-call/issues/456 as
the various react libraries had got out of sync (also we were well
overdue a dependency update).


SimonBrandner (approved):
Looks like a `yarn upgrade`

PR #465, opened 2022-07-13
Auto-register if displayName URL param is given

dbkr:
Also moves some more logic out of views.

Fixes https://github.com/vector-im/element-call/issues/442

### DIFF
--- a/src/auth/generateRandomName.ts
+++ b/src/auth/generateRandomName.ts
@@ -19,7 +19,6 @@ import {
   adjectives,
   colors,
   animals,
-  Config,
 } from "unique-names-generator";
 
 const elements = [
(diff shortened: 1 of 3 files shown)


src/auth/useRegisterPasswordlessUser.ts
```diff
@@ -0,0 +1,59 @@
...
+import { randomString } from "matrix-js-sdk/src/randomstring";
+
+import { useClient } from "../ClientContext";
+import { useInteractiveRegistration } from "../auth/useInteractiveRegistration";
+import { generateRandomName } from "../auth/generateRandomName";
+import { useRecaptcha } from "../auth/useRecaptcha";
+
+export interface UseRegisterPasswordlessUserType {
+  privacyPolicyUrl: string;
+  registerPasswordlessUser: (string) => Promise<void>;
```
robintown, line 27: The parameter needs a name, otherwise TS interprets it as `(string: any) => Promise<void>`

dbkr: oh, good spot, thanks

PR #468, opened 2022-07-14
Add local volume control

SimonBrandner:
Requires https://github.com/matrix-org/matrix-js-sdk/pull/2525

<hr>

During calls it can often happen that the audio coming from one person is louder than the audio coming from another. This can of course be fixed on the sending side but it's often easier to just change the volume levels locally. This is especially helpful during calls with screen-sharing where audio is also shared to control if the user wants to hear the screen-sharing audio (or how loud they want it to be)

<hr>

This needs product to take a look

<hr>

Peek 2022-07-15 11-16


robintown:
This is of course up to Product, but I'd suggest the ability to go above 100% as well to deal with quiet speakers (just mentioning it here for completeness)

jakewb-b:
This is great - I'm definitely in favour.

> This is of course up to Product, but I'd suggest the ability to go above 100% as well to deal with quiet speakers (just mentioning it here for completeness)

I agree with being able to turn it up. Referring to it as 'above 100%' *feels* a bit awkward to me and might be confusing, although I assume it means artificially boosting the volume. Maybe we're better off just starting the slider in the middle and then either not having numbers as you move it left and right, or having it something more arbitrary like sliding left goes down to -10, and sliding right goes up to +10. Ideally that's something @gaelledel could take a look at.

Can I also just confirm, is the volume setting for a given user maintained if:
- That user drops from this call and rejoins it almost immediately (e.g. within 5 minutes)?
- Rejoins the same call on a later date (i.e. it's a weekly call in the same room, with the same people)
- Joins another call with me?

In the first case, it might be good to have the setting maintained, but in other cases I'd er towards re-setting it to the middle each time.

SimonBrandner:
> Can I also just confirm, is the volume setting for a given user maintained if:
> 
> * That user drops from this call and rejoins it almost immediately (e.g. within 5 minutes)?
> * Rejoins the same call on a later date (i.e. it's a weekly call in the same room, with the same people)
> * Joins another call with me?
> 
> In the first case, it might be good to have the setting maintained, but in other cases I'd er towards re-setting it to the middle each time.

Currently, the volume is only maintained during the call. It should be possible to persist the volume per-user per-call though I am not sure the time reset is the way to go - it would be a little difficult to implement and I think other VoIP software (such as Discord) persists the value independent of time. Though that is of course just my POV on this

jakewb-b:
Yeah, I'm a little undecided on whether we'd want it to persist per-user per-call. On the one hand, if a given user is *always* too quiet, and I *always* have to turn them up, that would be annoying. 
On the other hand, if I've once turned a user all the way down because they were too loud, it would be easy to forget that and think their audio was broken when they joined next week's call. Or, conversely, have them suddenly blasting audio at high volume because last call I turned them all the way up. 

Let's leave that for now and stick with re-setting each time. I think we could re-visit this when it gets more design attention, as maybe the solution is more of a visual indicator on a user if they've had their audio adjusted away from the default in either direction.

jakewb-b (approved):
I'm happy with this from the product side. Once Gaelle has capacity, it would be good for her to review this and make any recommendations for further changes, but for a fairly minor piece of setting UI I think we can do that in the future once it's live.

src/button/Button.tsx
```diff
@@ -237,3 +238,20 @@ export function InviteButton({
     </TooltipTrigger>
   );
 }
+
+export function OptionsButton({
+  className,
+  ...rest
+}: {
+  className?: string;
+  [index: string]: unknown;
```
robintown, line 247: Is it possible to make this any more specific? Such as `Omit<"variant", ButtonProps>` or something

src/video-grid/VideoTileSettingsModal.module.css
```diff
@@ -0,0 +1,76 @@
+.content {
+  margin: 27px 34px;
+}
+
+.localVolumePercentage {
+  width: 40px;
```
robintown, line 6: I would suggest specifying this in `ch` units so it scales as expected no matter the font or font size

src/video-grid/VideoTileSettingsModal.module.css
```diff
@@ -0,0 +1,76 @@
...
+  appearance: none;
+
+  cursor: pointer;
+  width: 100%;
+}
+
+.localVolumeSlider[type="range"]::-moz-range-track {
+  -ms-appearance: none;
+  -moz-appearance: none;
+  -webkit-appearance: none;
```
robintown, line 22: Are the prefixed versions necessary? They could do with a comment if so

src/video-grid/useMediaStream.ts
```diff
@@ -84,6 +85,13 @@ export const useMediaStream = (
     }
   }, [audioOutputDevice]);
 
+  useEffect(() => {
+    if (!mediaRef.current) return;
+    if (localVolume === null || localVolume === undefined) return;
```
robintown, line 90:
```suggestion
    if (localVolume == null) return;
```

This is probably just a personal preference so feel free to ignore

src/video-grid/useMediaStream.ts
```diff
@@ -187,15 +195,17 @@ export const useSpatialMediaStream = (
   audioOutputDevice: string,
   audioContext: AudioContext,
   audioDestination: AudioNode,
-  mute = false
+  mute = false,
+  localVolume: number
 ): [RefObject<Element>, RefObject<MediaElement>] => {
   const tileRef = useRef<Element>();
   const [spatialAudio] = useSpatialAudio();
```
robintown, line 202: If spatial audio is enabled, the volume control won't have any effect since audio gets routed separately from the media element via the Web Audio API. You'd need to add a [GainNode](https://developer.mozilla.org/en-US/docs/Web/API/GainNode) into the processing chain below

src/video-grid/VideoTileSettingsModal.tsx
```diff
@@ -0,0 +1,73 @@
+/*
+Copyright 2022 Matrix.org Foundation C.I.C.
```
robintown, line 2: This should be New Vector Ltd instead - the copyright in most other files is currently wrong

src/video-grid/VideoTileSettingsModal.tsx
```diff
@@ -0,0 +1,73 @@
...
+          step="0.01"
+          value={localVolume}
+          onChange={onLocalVolumeChanged}
+        />
+      </FieldRow>
+    </>
+  );
+};
+
+interface Props {
```
robintown, line 61: Shouldn't this be `extends` something for the modal props?

src/video-grid/VideoTileSettingsModal.tsx
```diff
@@ -0,0 +1,73 @@
...
+  feed: CallFeed;
+}
+
+export const VideoTileSettingsModal = (props: Props) => {
```
robintown, line 65:
```suggestion
export const VideoTileSettingsModal = ({ feed, ...rest }: Props) => {
```

Just so the feed doesn't get mixed in with the modal props

src/button/Button.tsx
robintown: Or rather, turn the type into an interface that extends ButtonProps

src/video-grid/VideoTileSettingsModal.tsx
SimonBrandner: Yes, but I don't think we have that yet - `Modal` is still JS

robintown (approved):
:shipit:

gaelledel:
@SimonBrandner Please view https://www.figma.com/file/AjvSw4lI9iGO5aTe4WAl4r/Element-Call?node-id=2406%3A65588 for local video

PR #471, opened 2022-07-15
Wait for the created room to come down sync before placing a group call

robintown:
Partially fixes https://github.com/vector-im/element-call/issues/470

The other thing we need to do to fully fix that issue, is place the group call on load if it for some reason doesn't exist.


SimonBrandner (approved):
Great!!

PR #473, opened 2022-07-16
Add support for audio sharing

SimonBrandner:
Requires https://github.com/vector-im/element-call/pull/468
Requires https://github.com/matrix-org/matrix-js-sdk/pull/2530


robintown (approved):
The formatter is unhappy, but otherwise looks good

PR #479, opened 2022-07-20
Fix 'cannot find room' error

dbkr:
We weren't waiting for rooms to arrive down the sync stream after
joining them but before trying to use them.

More regression details in linked issue.

Fixes https://github.com/vector-im/element-call/issues/477


ara4n (approved):
looks plausible to me.

PR #487, opened 2022-07-29
typescript `src/input`

toger5:


src/input/AvatarInputField.tsx
```diff
@@ -36,20 +48,24 @@ export const AvatarInputField = forwardRef(
...
 
-      fileInputRef.current.addEventListener("change", onChange);
+      currentInput.addEventListener("change", onChange);
 
       return () => {
-        if (fileInputRef.current) {
-          fileInputRef.current.removeEventListener("change", onChange);
+        if (currentInput) {
+          currentInput.removeEventListener("change", onChange);
         }
```
robintown, line 69:
```suggestion
        currentInput?.removeEventListener("change", onChange);
```

src/input/AvatarInputField.tsx
```diff
@@ -19,13 +19,25 @@ import React, { useEffect } from "react";
...
+  displayName: string;
+  onRemoveAvatar: () => void;
+  // TODO: add used parameters for HTMLInputElement
+  [index: string]: unknown;
+}
+
+export const AvatarInputField = forwardRef<HTMLInputElement, Props>(
   (
     { id, label, className, avatarUrl, displayName, onRemoveAvatar, ...rest },
     ref
```
robintown, line 43: I can't highlight it, but please add a type to the `objUrl` `useState` line

src/input/AvatarInputField.tsx
```diff
@@ -19,13 +19,25 @@ import React, { useEffect } from "react";
...
-export const AvatarInputField = forwardRef(
+interface Props {
+  id: string;
+  label: string;
+  className: string;
+  avatarUrl: string;
+  displayName: string;
+  onRemoveAvatar: () => void;
+  // TODO: add used parameters for HTMLInputElement
+  [index: string]: unknown;
```
robintown, line 37: Extend `HTMLAttributes<HTMLInputElement>` instead?

src/input/Input.tsx
```diff
@@ -14,12 +14,19 @@ See the License for the specific language governing permissions and
...
 import { ReactComponent as CheckIcon } from "../icons/Check.svg";
 
-export function FieldRow({ children, rightAlign, className, ...rest }) {
+interface FieldRowProps {
+  children: JSX.Element;
+  rightAlign?: boolean;
+  className?: string;
+}
+
+export function FieldRow({ children, rightAlign, className }: FieldRowProps) {
```
robintown, line 29: Please add a return type

src/input/Input.tsx
```diff
@@ -33,11 +40,28 @@ export function FieldRow({ children, rightAlign, className, ...rest }) {
   );
 }
 
-export function Field({ children, className, ...rest }) {
+interface FieldProps {
+  children: ReactNode;
+  className?: string;
+}
+export function Field({ children, className }: FieldProps) {
```
robintown, line 47:
```suggestion
}

export function Field({ children, className }: FieldProps) {
```

robintown, line 47: Return type here too

src/input/Input.tsx
```diff
@@ -99,6 +118,6 @@ export const InputField = forwardRef(
   }
 );
 
-export function ErrorMessage({ children }) {
+export function ErrorMessage({ children }: { children: JSX.Element }) {
```
robintown, line 121: return type

src/input/Input.tsx
```diff
@@ -66,13 +90,8 @@ export const InputField = forwardRef(
...
         {prefix && <span>{prefix}</span>}
         {type === "textarea" ? (
-          <textarea
-            id={id}
-            {...rest}
-            ref={ref}
-            type={type}
-            disabled={disabled}
-          />
+          // for review: can i remove type and ref here?
```
robintown, line 93: I don't think so, the types will hopefully work out if you change the type of `ref` to `HTMLInputElement | HTMLTextAreaElement`

src/input/Input.tsx
```diff
@@ -33,11 +40,28 @@ export function FieldRow({ children, rightAlign, className, ...rest }) {
...
+  id: string;
+  label: string;
+  className: string;
+  type: string;
+  checked: boolean;
+  prefix: string;
+  suffix: string;
+  description: string;
+  disabled: boolean;
+  [index: string]: unknown;
```
robintown, line 61: could extend `HTMLAttributes<HTMLInputElement | HTMLTextAreaElement>` instead

src/input/SelectInput.tsx
```diff
@@ -15,16 +15,21 @@ limitations under the License.
...
 import styles from "./SelectInput.module.css";
-import classNames from "classnames";
 import { ReactComponent as ArrowDownIcon } from "../icons/ArrowDown.svg";
 
-export function SelectInput(props) {
+interface Props extends AriaSelectOptions<object> {
+  className: string;
+}
+
+export function SelectInput(props: Props) {
```
robintown, line 32: return type

src/input/Toggle.tsx
```diff
@@ -15,22 +15,30 @@ limitations under the License.
...
 
-export function Toggle({ id, label, className, onChange, isSelected }) {
+interface Props {
+  id: string;
+  label: string;
+  className: string;
+  onChange: (selected: boolean) => void;
+  isSelected: boolean;
+}
+export function Toggle({ id, label, className, onChange, isSelected }: Props) {
```
robintown, line 31:
```suggestion
}

export function Toggle({ id, label, className, onChange, isSelected }: Props) {
```

robintown, line 31: return type

src/input/Toggle.tsx
```diff
@@ -15,22 +15,30 @@ limitations under the License.
...
   const buttonRef = useRef();
```
robintown, line 32: type parameter for `useRef`?

src/input/Input.tsx
toger5: I did not add the return types to components for most of the classes since they are determined automatically.
Should we change that in all files?

src/input/Input.tsx
toger5: That is what I tried first. It throws the following error though:
```
Type 'ForwardedRef<HTMLInputElement | HTMLTextAreaElement>' is not assignable to type 'LegacyRef<HTMLTextAreaElement>'.
  Type 'MutableRefObject<HTMLInputElement | HTMLTextAreaElement>' is not assignable to type 'LegacyRef<HTMLTextAreaElement>'.
    Type 'MutableRefObject<HTMLInputElement | HTMLTextAreaElement>' is not assignable to type 'RefObject<HTMLTextAreaElement>'.
      Types of property 'current' are incompatible.
        Type 'HTMLInputElement | HTMLTextAreaElement' is not assignable to type 'HTMLTextAreaElement'.
          Type 'HTMLInputElement' is missing the following properties from type 'HTMLTextAreaElement': cols, rows, textLength, wrapts(2322)
index.d.ts(143, 9): The expected type comes from property 'ref' which is declared here on type 'DetailedHTMLProps<TextareaHTMLAttributes<HTMLTextAreaElement>, HTMLTextAreaElement>'
```

toger5: I am doing an explicit cast for the ref now. I think that should work.

src/input/AvatarInputField.tsx
toger5: Thanks!

src/input/AvatarInputField.tsx
```diff
@@ -15,42 +15,56 @@ limitations under the License.
...
 import styles from "./AvatarInputField.module.css";
 
-export const AvatarInputField = forwardRef(
+interface Props extends HTMLAttributes<HTMLInputElement> {
+  id: string;
+  label: string;
+  avatarUrl: string;
+  displayName: string;
+  onRemoveAvatar: () => void;
+  //hack because extending from HTMLAttributes<HTMLInputElement> seems to not work
```
toger5, line 35: This is still bothering me. Am I doing something wrong? HTMLAttributes<HTMLInputElement> should contain the `name` property

src/input/Input.tsx
```diff
@@ -33,11 +40,41 @@ export function FieldRow({ children, rightAlign, className, ...rest }) {
...
+  checked?: boolean;
+  className?: string;
+  description?: string;
+  disabled?: boolean;
+  required?: boolean;
+  // this is a hack. Those variables should be part of `HTMLAttributes<HTMLInputElement> | HTMLAttributes<HTMLTextAreaElement>`
+  // but extending from this union type does not work
+  name?: string;
+  autoComplete?: string;
+  autoCorrect?: string;
```
toger5, line 67: And this is the same situation

toger5:
All the Linter issues are fixed now.
@robintown do you think we should add the explicit return types?
Maybe we can still find a solution for the `HTMLAttributes<HTMLInputElement>` issue where the interface extension does not really work...

robintown:
For the `HTMLAttributes` problem, try installing @types/react-dom perhaps?

toger5:
`name` was part of `AllHTMLAttributes` so that could be used to make it work.
in the situation of the `textarea` / `input` we still have issues because textarea an input have different types for `onChange` so we would need to check for the type first and than explicitly  cast to the correct one.

src/SequenceDiagramViewerPage.tsx
```diff
@@ -57,7 +57,7 @@ export function SequenceDiagramViewerPage() {
         <SequenceDiagramViewer
           localUserId={debugLog.localUserId}
           selectedUserId={selectedUserId}
-          onSelectUserId={setSelectedUserId}
+          onSelectUserId={setSelectedUserId as (key: Key) => unknown}
```
toger5, line 60: is there a better solution for this?
At some point we need it to be a `(key: Key)=>any` because that is required by `<SelectInput onSelectionChange={onSelectInputChangedFunctionOfTypeKeyKeyAny} />`

robintown: Can't we use the `Dispatch<(prevState: undefined) => undefined>` type that's in GroupCallInspector.tsx?

robintown (changes requested):
I still think we should either have explicit return types on our components, or type the whole thing as `FC<...>`

src/SequenceDiagramViewerPage.tsx
toger5: That is what I tried initially but it raises the follwoing error:
```
(property) SequenceDiagramViewerProps.onSelectUserId: (key: React.Key) => unknown
Type 'Dispatch<(prevState: undefined) => undefined>' is not assignable to type '(key: Key) => unknown'.
  Types of parameters 'value' and 'key' are incompatible.
    Type 'Key' is not assignable to type '(prevState: undefined) => undefined'.
      Type 'string' is not assignable to type '(prevState: undefined) => undefined'.ts(2322)
 ```
 So, at some point in time we need to convert it to  '(key: Key) => unknown` so that it is accepted by the `onOnSelectUserId`

robintown: Ah, okay, in that case I would recommend changing the type of `onSelectUserId` to `(userId: string) => void` and do the typecast inside SequenceDiagramViewer instead, so that the component asserts in a single location that it always outputs strings

robintown (commented):
See my reply about `onSelectUserId`; otherwise this looks good to go

robintown (approved):

src/SequenceDiagramViewerPage.tsx
toger5: That is a really nice solution. Thanks!

PR #488, opened 2022-07-29
typescript `src/popover`

toger5:


src/popover/Popover.tsx
```diff
@@ -14,14 +14,23 @@ See the License for the specific language governing permissions and
...
 import { DismissButton, useOverlay } from "@react-aria/overlays";
 import { FocusScope } from "@react-aria/focus";
 import classNames from "classnames";
-import styles from "./Popover.module.css";
 import { useObjectRef } from "@react-aria/utils";
 
-export const Popover = forwardRef(
+import styles from "./Popover.module.css";
+
+interface PopoverProps {
```
robintown, line 25: I think just calling this `Props` would be fine

src/popover/Popover.tsx
```diff
@@ -14,14 +14,23 @@ See the License for the specific language governing permissions and
...
+  isOpen: boolean;
+  onClose: () => void;
+  className: string;
+  children: JSX.Element;
+  [index: string]: unknown;
```
robintown, line 30: Rather than use an index signature, could we make this interface just extend `HTMLAttributes<HTMLDivElement>`?

src/popover/PopoverMenu.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+import styles from "./PopoverMenu.module.css";
+import { Popover } from "./Popover";
+
+interface PopoverMenuTriggerProps {
+  children: JSX.Element;
+  placement: Placement;
+  className: string;
+  disableOnState: boolean;
+  [index: string]: unknown;
```
robintown, line 33: useMenuTriggerState seems to have types, so I suggest this interface extend that rather than use an index signature

src/popover/PopoverMenu.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+export const PopoverMenuTrigger = forwardRef<
```
robintown, line 35:
```suggestion
}

export const PopoverMenuTrigger = forwardRef<
```

toger5:
@robintown the you proposed are really nice.
Ready to merge?

robintown (approved):

PR #489, opened 2022-07-30
Convert `src/*.js` and `src/*.jsx` to TypeScript 

SimonBrandner:


src/Header.tsx
```diff
@@ -36,7 +55,18 @@ export function LeftNav({ children, className, hideMobile, ...rest }) {
...
+  children?: ReactNode;
+  className?: string;
+  hideMobile?: string;
+}
+
+export function RightNav({
+  children,
+  className,
+  hideMobile,
+  ...rest
```
toger5, line 68: there is no`rest` in RightNavProps it seems to be the leftover params for a `div` so maybe extend  RightNavProps from `HTMLAttributes<HTMLDivElement>`

src/ListBox.jsx
```diff
@@ -1,50 +0,0 @@
...
-      ref={ref}
-      className={classNames(styles.option, className, {
-        [styles.selected]: isSelected,
-        [styles.focused]: isFocused,
-        [styles.disables]: isDisabled,
-      })}
-    >
-      {item.rendered}
-    </li>
-  );
```
toger5, line 49: why does git not understand the rename of the file?

src/SequenceDiagramViewerPage.tsx
```diff
@@ -1,13 +1,20 @@
 import React, { useCallback, useState } from "react";
```
toger5, line 1: here we also are missing the copyright

src/Tooltip.tsx
```diff
@@ -0,0 +1,98 @@
+import React, {
```
toger5, line 1: here as well (missing copyright)

src/UserMenu.tsx
```diff
@@ -75,12 +85,11 @@ export function UserMenu({
             <UserIcon />
           )}
         </Button>
-        {() => "Profile"}
       </TooltipTrigger>
       {(props) => (
         <Menu {...props} label="User menu" onAction={onAction}>
           {items.map(({ key, icon: Icon, label }) => (
-            <Item key={key} textValue={label} className={styles.menuItem}>
+            <Item key={key} textValue={label}>
```
toger5, line 92: why is the style class not needed here (why did it get removed)?

toger5 (commented):
Thanks a lot. Look really good and ticks a lot of files from the non-ts list.
I dont really understand why we move from using the children for the tooltips
```
   const [tooltipTrigger, tooltip] = children;
 ```
 To having it part of the `TooltipTriggerProps`
 ```
 interface TooltipTriggerProps {
  children: ReactElement;
  placement?: Placement;
  delay?: number;
  tooltip: () => string;
}
```

src/UserMenu.tsx
SimonBrandner: `Item` doesn't have a `className` prop

src/ListBox.jsx
SimonBrandner: Because the evil overlord want us to suffer :D

SimonBrandner:
> Thanks a lot. Look really good and ticks a lot of files from the non-ts list. I dont really understand why we move from using the children for the tooltips
> 
> ```
>    const [tooltipTrigger, tooltip] = children;
> ```
> 
> To having it part of the `TooltipTriggerProps`
> 
> ```
> interface TooltipTriggerProps {
>  children: ReactElement;
>  placement?: Placement;
>  delay?: number;
>  tooltip: () => string;
> }
> ```

I was having trouble typing it and having it as a prop seems nicer to me somehow - why should we pass down something as a child when it is not a child?

toger5 (approved):
I see how it is nice to have the tooltip as a prop. Looks fine to me now

PR #493, opened 2022-08-01
Get volume button inline with designs

SimonBrandner:


src/video-grid/VideoTile.module.css
```diff
@@ -60,16 +59,24 @@
...
 
-.optionsButton {
-  right: 16px;
-  top: 16px;
+.infoBubbleIcon {
+  background-color: rgba(23, 25, 28, 0.85);
+}
+
+.button button:hover svg * {
+  fill: white !important; /* To override .iconButton:not(.stroke):hover svg */
```
robintown, line 67: Please use a theme variable here

src/video-grid/VideoTile.module.css
```diff
@@ -60,16 +59,24 @@
   z-index: 1;
 }
 
-.optionsButton {
-  right: 16px;
-  top: 16px;
+.infoBubbleIcon {
+  background-color: rgba(23, 25, 28, 0.85);
```
robintown, line 63: This could stand to be a theme variable as well - it looks to be a non-standard one borrowed from Element's light mode, so this would be a good opportunity to fix that by replacing it with a similar one from dark mode, making it `--background-85` or something

robintown (commented):
The designs show a gradient that appears over the top of the video tile to make the button more visible against the background, which this doesn't seem to have?

robintown:
It also looks like in the designs, the button only appears on hover

SimonBrandner:
> It also looks like in the designs, the button only appears on hover

True, thanks for pointing it out!

src/index.css
```diff
@@ -38,6 +38,7 @@ limitations under the License.
   --quinary-content: #394049;
   --system: #21262c;
   --background: #15191e;
+  --backgrouund-85: rgba(23, 25, 28, 0.85);
```
robintown, line 41:
```suggestion
  --background-85: rgba(23, 25, 28, 0.85);
```

robintown, line 41: This should be accompanied by an environment variable in .env.example and src/main.tsx

robintown (commented):
> The designs show a gradient that appears over the top of the video tile to make the button more visible against the background, which this doesn't seem to have?

Unless I'm mistaken I still don't see this?

gaelledel (approved):

robintown (approved):
:sparkles:

PR #502, opened 2022-08-07
Add full-screen mode

SimonBrandner:
Fixes https://github.com/vector-im/element-call/issues/179

<hr>

Note: We hide the screen-sharing and options button in the full-screen view because neither can be opened from it. The browser won't be able to show its screen/window/tab picking dialog over the full-screen view and our tooltips/context-menus aren't currently compatible with full-screening and would require further refactoring


gaelledel (changes requested):
Currently, the top overlay is displayed as soon as users hover anywhere on a tile. Ideally we should have the overlay appearing only when hovering on the upper section of the tile. Also might not be related to this but changing to freedom layout on screenshare is broken

gaelledel:
https://user-images.githubusercontent.com/76945183/183407517-7699d053-12d3-452f-8ab0-1716b7994cd1.mov

See what happens when I switch layout

SimonBrandner:
> Screen.Recording.2022-08-08.at.13.24.27.mov 
> See what happens when I switch layout

Turns out this is unrelated, it's on call.element.io too

robintown (commented):
Generally looks good

robintown (approved):

gaelledel (approved):

PR #512, opened 2022-08-12
Fix audio during full-screening

SimonBrandner:
Fixes https://github.com/vector-im/element-call/issues/509

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -44,10 +44,11 @@ import { UserMenuContainer } from "../UserMenuContainer";
 import { useRageshakeRequestModal } from "../settings/submit-rageshake";
 import { RageshakeRequestModal } from "./RageshakeRequestModal";
 import { useMediaHandler } from "../settings/useMediaHandler";
-import { useShowInspector } from "../settings/useSetting";
+import { useShowInspector, useSpatialAudio } from "../settings/useSetting";
 import { useModalTriggerState } from "../Modal";
 import { useAudioContext } from "../video-grid/useMediaStream";
 import { useFullscreen } from "../video-grid/useFullscreen";
+import { AudioContainer } from "../video-grid/AudioContainer";
 import { useAudioOutputDevice } from "../video-grid/useAudioOutputDevice";
 
 const canScreenshare = "getDisplayMedia" in (navigator.mediaDevices ?? {});
@@ -107,6 +108,8 @@ export function InCallView({
   const elementRef = useRef<HTMLDivElement>();
   const [layout, setLayout] = useVideoGridLayout(screenshareFeeds.length > 0);
   const { toggleFullscreen, fullscreenParticipant } = useFullscreen(elementRef);
+
+  const [spatialAudio] = useSpatialAudio();
 
   const [audioContext, audioDestination, audioRef] = useAudioContext();
   const { audioOutput } = useMediaHandler();
@@ -183,7 +186,6 @@ export function InCallView({
           key={fullscreenParticipant.id}
           item={fullscreenParticipant}
           getAvatar={renderAvatar}
-          audioOutputDevice={audioOutput}
           audioContext={audioContext}
           audioDestination={audioDestination}
           disableSpeakingIndicator={true}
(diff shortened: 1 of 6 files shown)


dbkr (commented):
Sorry, I'm really struggling to understand what the change is here. We're changing to using the web audio api... always? What's the change in useSpatialMediaStream?

Also rendering empty element feels like hooks making us do really strange things to fit the paradigm. Isn't the answer to make another hook and then just iterate over the array in a useEffect?

SimonBrandner:
> Sorry, I'm really struggling to understand what the change is here. 

Sorry, I should have added a description...

The issue we're facing is that when we full-screen a feed, all the other feeds don't get rendered and so we get no audio from them.

> We're changing to using the web audio api... always? 

Yes, we're changing to always using the web audio API. This way audio gets played even for invisible feeds. The other option would be to render a bunch of audio elements in the DOM like we do in Element Web

> What's the change in useSpatialMediaStream?

Since we route the audio through the web audio API and a separate audio element, we don't want `useMediaStream()` to play any audio, so we pass it a bunch of `undefined`s and a `true` to mute it

> Also rendering empty element feels like hooks making us do really strange things to fit the paradigm. Isn't the answer to make another hook and then just iterate over the array in a useEffect?

Yeah, hooks aren't making this easy - we could iterate over all the feeds in a hook but then whenever a feed is added/removed the hook would get re-run and so all the audio paths would be re-setup which doesn't seem like something we want. Ideally we would call a hook from inside of the the for loop inside the other hook but that is simply not possible hence the empty elements. Again we could just render a bunch of audio elments in the DOM, if you think that is the better option, we can go with that 

<hr>

Hopefully, I am making a bit more sense now... If not, sorry

dbkr:
Ah right, so:
 * Spatial audio -> special spatial audio output (web audio using panner node)
 * Not spatial audio but fullscreen -> AudioContainer using web audio
 * Not spatial audio & not fullscreen -> audio played from video element

...and the loopback to make AEC work with webaudio is enabled in all cases anyway? If that's right could it be docced somewhere?

Ideally I guess we'd amalgamate the two web audio paths into one although that could be a refactor later. Might be worth a bug for the tech debt.

On the empty elements, OK then - also might be worth a bug to track this tech debt too so we can refactor down the line: rendering audio shouldn't really have anything to do with our UI, so the fact that it's tied to React at all feels like a mis-design.

SimonBrandner:
> Ah right, so:
> 
> * Spatial audio -> special spatial audio output (web audio using panner node)
> * Not spatial audio but fullscreen -> AudioContainer using web audio
> * Not spatial audio & not fullscreen -> audio played from video element

Actually, it's 

+ Spatial audio & not fullscreen -> audio element (web audio using panner node in `useSpatialMediaStream` in `VideoTile`)
+ Other -> audio element (web audio using `AudioForParticipant`)

> ...and the loopback to make AEC work with webaudio is enabled in all cases anyway? If that's right could it be docced somewhere?

Yeah, we always use the loopback hack... We could replace the `Other` case with an array of audio elements to avoid this (aka EW approach) which should also be more readable. The main con of this approach is that we'd lose the ability to fix https://github.com/vector-im/element-call/issues/504 since that requires using the web audip API 

> Ideally I guess we'd amalgamate the two web audio paths into one although that could be a refactor later. Might be worth a bug for the tech debt.

I actually doubt we can much better than the approach taken here since if we want spatial audio we need to route the audio through the `VideoTile`s to get their position and if we are in fullscreen we on the other hand definitely do not want the handling in the `VideoTile`s as all but one is hidden. Maybe I am missing something?

> On the empty elements, OK then - also might be worth a bug to track this tech debt too so we can refactor down the line: rendering audio shouldn't really have anything to do with our UI, so the fact that it's tied to React at all feels like a mis-design.

We could write our own thingy to handle this but I suspect it would be more complex than using React + we wouldn't get to re-use the hooks

dbkr (approved):
Surely in spatial audio mode the audio isn't going through the video elements, it's going through the web audio API anyway, since that's what's doing the panning? Anyway, OK, we can go with this if it fixes the audio but I'd definitely like to find a better solution (to the empty elements too).

PR #516, opened 2022-08-12
Fix a case where someone's audio could be missing if the audio track arrived late

robintown:
The addition of an audio track wouldn't cause this hook to re-run. However, I don't think we should worry about optimizing for the case of 0 audio tracks at all, since that could only conceivably happen with screensharing feeds.


SimonBrandner (approved):
Thank you for beating me to it! :D

PR #519, opened 2022-08-13
Fix handling of streams with no audio tracks

SimonBrandner:


src/video-grid/useMediaStream.ts
```diff
@@ -213,7 +213,7 @@ export const useSpatialMediaStream = (
   const sourceRef = useRef<MediaStreamAudioSourceNode>();
 
   useEffect(() => {
-    if (spatialAudio && tileRef.current && !mute) {
+    if (spatialAudio && audioContext && tileRef.current && !mute) {
```
robintown, line 216: audioContext is intended to always be present as per the type, so I don't think this is necessary?

robintown (commented):
I think it might be nicer to make useSpatialMediaStream work with 0 audio tracks on its own, so the hook doesn't have to be re-run everytime someone unmutes (by reverting my previous change)

src/video-grid/useMediaStream.ts
SimonBrandner: I've seen an issue where it is not :/ Might be related to my SFU branch - will try to track it down and send a fix if it is relevant for regular EC

SimonBrandner:
> I think it might be nicer to make useSpatialMediaStream work with 0 audio tracks on its own, so the hook doesn't have to be re-run everytime someone unmutes (by reverting my previous change)

Good point!

robintown (approved):
Good idea separating that out into a reusable hook

PR #535, opened 2022-08-18
Fix the call type selector

robintown:
I also converted UnauthenticatedView to TypeScript for good measure, since TypeScript would've caught this problem.

Closes https://github.com/vector-im/element-call/issues/534

### DIFF
--- a/src/matrix-utils.ts
+++ b/src/matrix-utils.ts
@@ -17,7 +17,12 @@ import { Visibility, Preset } from "matrix-js-sdk/src/@types/partials";
 import { ISyncStateData, SyncState } from "matrix-js-sdk/src/sync";
 import { WidgetApi } from "matrix-widget-api";
 import { logger } from "matrix-js-sdk/src/logger";
+import {
+  GroupCallIntent,
+  GroupCallType,
+} from "matrix-js-sdk/src/webrtc/groupCall";
 
+import type { Room } from "matrix-js-sdk/src/models/room";
 import IndexedDBWorker from "./IndexedDBWorker?worker";
 import { getRoomParams } from "./room/useRoomParams";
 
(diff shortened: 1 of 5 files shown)


src/matrix-utils.ts
```diff
@@ -322,6 +328,36 @@ export async function createRoom(
...
+  const result = await createPromise;
+
+  console.log(`Creating ${ptt ? "PTT" : "video"} group call room`);
+
+  await client.createGroupCall(
+    result.room_id,
+    ptt ? GroupCallType.Voice : GroupCallType.Video,
+    ptt,
+    GroupCallIntent.Room
+  );
```
SimonBrandner, line 359: I am very confused as to how this should be here or as to how did things ever work without it?

robintown: What happened was: in https://github.com/vector-im/element-call/pull/471/ I moved group call creation to be the responsibility of `useLoadGroupCall`. I thought that the `createGroupCall` call here in `createRoom` was superfluous, so I removed it, and I also thought that `useLoadGroupCall` would still be able to choose between PTT and non-PTT correctly, since it takes a `createPtt` flag.

There were two problems with this: the `createPtt` flag doesn't necessarily indicate whether the user selected walkie-talkie mode in the dropdown, instead it just reflects whether the `?ptt` URL parameter is present. The second problem is that the code was still passing a PTT flag to `createRoom`, even though I removed that. That should've never happened - if we had finished migrating to TypeScript at the time of https://github.com/vector-im/element-call/pull/471/, the errors probably would've forced me to rethink my decisions.

SimonBrandner: Ah, I see - thank you for the explanation!

SimonBrandner (approved):

PR #564, opened 2022-09-01
Fix bug where additional group calls could be created

dbkr:
This (hopefully) fixes the remaining bug where extra group calls
could be created when entering a room.

We waited for the Room event to arrive, but didn't wait for the
group call event handler to actually process the event, so it would
have depended what order the event handlers were run in.

If this doesn't fix it, it at least adds logging so we'll have more
to go on next time.

Fixes https://github.com/vector-im/element-call/issues/563

### DIFF
--- a/src/room/useLoadGroupCall.ts
+++ b/src/room/useLoadGroupCall.ts
@@ -58,7 +58,7 @@ export const useLoadGroupCall = (
             resolve(room);
           }
         };
-        client.on(ClientEvent.Room, onRoomEvent);
+        client.on(GroupCallEventHandlerEvent.Room, onRoomEvent);
       });
 
       // race the promise with a timeout so we don't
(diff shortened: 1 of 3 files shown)


src/room/useLoadGroupCall.ts
```diff
@@ -53,7 +54,7 @@ export const useLoadGroupCall = (
       const waitPromise = new Promise<Room>((resolve) => {
         const onRoomEvent = async (room: Room) => {
           if (room.roomId === roomId) {
-            client.removeListener(ClientEvent.Room, onRoomEvent);
+            client.removeListener(GroupCallEventHandlerEvent.Room, onRoomEvent);
```
SimonBrandner, line 57: Maybe this is just GitHub on Android lying about what lines were changed, but how does us removing a different event listener fix this?

dbkr: GitHub on Android is not lying, I just failed to update the right line, and also failed to actually commit the js-sdk change anyway...

SimonBrandner: Ah, I see

SimonBrandner (approved):

PR #568, opened 2022-09-06
Log ID instead of object

dbkr:
as otherwise it recurses and logs the entire client + store


src/room/useLoadGroupCall.ts
```diff
@@ -80,7 +80,7 @@ export const useLoadGroupCall = (
       const room = await fetchOrCreateRoom();
       logger.debug(`Fetched / joined room ${roomIdOrAlias}`);
       const groupCall = client.getGroupCallForRoom(room.roomId);
-      logger.debug("Got group call", groupCall);
+      logger.debug("Got group call", groupCall.groupCallId);
```
robintown, line 83: `groupCall` can be null, which would make this explode

robintown (changes requested):

src/room/useLoadGroupCall.ts
dbkr: good point

PR #581, opened 2022-09-14
Maximise the active speaker when the window is small

robintown:
The purpose of this is to make Element Call look nice when shown in a picture-in-picture view inside Element Web:

Screenshot 2022-09-14 at 19-08-50 Element Test room

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -215,7 +215,7 @@ export function InCallView({
   // window is too small to show everyone
   const maximisedParticipant = useMemo(
     () =>
-      fullscreenParticipant ?? bounds.height <= 500
+      fullscreenParticipant ?? (bounds.height <= 500 && bounds.width <= 500)
         ? items.find((item) => item.focused) ??
           items.find((item) => item.callFeed) ??
           null
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -16,6 +16,8 @@ limitations under the License.
 
 import React, { useEffect, useCallback, useMemo, useRef } from "react";
 import { usePreventScroll } from "@react-aria/overlays";
+import useMeasure from "react-use-measure";
+import { ResizeObserver } from "@juggle/resize-observer";
 import { MatrixClient } from "matrix-js-sdk/src/client";
 import { RoomMember } from "matrix-js-sdk/src/models/room-member";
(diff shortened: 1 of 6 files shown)


src/room/InCallView.tsx
```diff
@@ -197,7 +208,20 @@ export function InCallView({
...
     return participants;
-  }, [userMediaFeeds, activeSpeaker, screenshareFeeds, layout]);
+  }, [userMediaFeeds, activeSpeaker, screenshareFeeds]);
+
+  // The maximised participant: either the participant that the user has
+  // manually put in fullscreen, or the focused (active) participant if the
+  // window is too small to show everyone
+  const maximisedParticipant = useMemo(
+    () =>
+      fullscreenParticipant ?? bounds.height <= 500
```
SimonBrandner, line 218: I think we ought to check for width too - not sure about the exact use-case but you could have the window with a low height but high width and want to see multiple tiles

src/room/InCallView.tsx
```diff
@@ -217,24 +241,27 @@ export function InCallView({
     []
   );
 
-  const renderContent = useCallback((): JSX.Element => {
+  const renderContent = (): JSX.Element => {
```
SimonBrandner, line 244: Why not `useCallback()`?

(/me sees the realization that I lack react hooks knowledge coming :sweat_smile:)

SimonBrandner (approved):
Otherwise, LGTM

src/room/InCallView.tsx
robintown: It's unnecessary since `renderContent` is called directly within this component rather than being passed as a prop, so `useCallback` won't make this any more efficient (it'll just cause React to do additional unnecessary comparisons of the dependencies)

And sorry, I think I was the one to suggest that you put `useCallback` here, I guess I wasn't paying attention to where the function was used at the time

SimonBrandner: > It's unnecessary since `renderContent` is called directly within this component rather than being passed as a prop, so `useCallback` won't make this any more efficient (it'll just cause React to do additional unnecessary comparisons of the dependencies)

Ah, I see! Btw, thank you for the occasional lesson on react-hooks :D (I should probably allocate some time to do some more in-depth research :sweat_smile: )

> And sorry, I think I was the one to suggest that you put `useCallback` here, I guess I wasn't paying attention to where the function was used at the time

No worries!

PR #595, opened 2022-09-22
Fix scroll bug in spotlight view

dbkr:
This was a confusion between indices of the tile and the tile position: the spotlight tile is the 0th TilePosition, ie. the tile with order 0, not the tile with index 0.

Also comment one method to hopefully make this slightly easier to understand.

Fixes https://github.com/vector-im/element-call/issues/499

### DIFF
--- a/src/video-grid/VideoGrid.tsx
+++ b/src/video-grid/VideoGrid.tsx
@@ -660,6 +660,8 @@ function getSubGridPositions(
   return newTilePositions;
 }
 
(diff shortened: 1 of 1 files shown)


robintown (approved):
Yay :)

robintown:
Closes https://github.com/vector-im/element-call/issues/398 as well, right?

dbkr:
ah, indeed, yes

PR #602, opened 2022-09-23
Fix bug causing mic/webcam to remain open after call

dbkr:
Fixes https://github.com/vector-im/element-call/issues/596


robintown (approved):
Confirmed that this works with the EW integration too

PR #603, opened 2022-09-26
Clear storage after logout

dbkr:


robintown (approved):
I sort of wonder if this could have been causing some of the client desync confusion that we've seen

PR #605, opened 2022-09-26
Let the maximized video feed fill the window

robintown:
instead of getting letterboxed.


dbkr:
Does this keep the right aspect ratio?

robintown:
Yes, all our video elements have `object-fit` set on them to preserve it

dbkr (approved):

PR #609, opened 2022-09-27
Use device labels rather than IDs in widget API

dbkr:
device IDs are different for each origin, so won't match up when passed in & out of widgets. Use the label instead.

For https://github.com/vector-im/element-web/issues/23331

### DIFF
--- a/src/media-utils.ts
+++ b/src/media-utils.ts
@@ -1,0 +1,23 @@
+/*
+Copyright 2022 Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


robintown (approved):
This side looks good

src/media-utils.ts
```diff
@@ -14,10 +14,55 @@ See the License for the specific language governing permissions and
...
+ * stream from the default camera which can mean, for example, that the
+ * light for the FaceTime camera turns on briefly even if you selected
+ * another camera. Once the Permissions API
+ * (https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API)
+ * is ready for primetime, this should allow us to avoid this.
+ *
+ * @return The available media devices
+ */
+export async function getDevices(): Promise<MediaDeviceInfo[]> {
+  let stream;
```
robintown, line 48: Type?

robintown (approved):

PR #618, opened 2022-10-12
Set up translation with i18next

robintown:
This sets the app up to use i18next as its translation framework, automatically detect the user's language, and allow the language to be overridden with a URL parameter such as https://call.element.io/#?lang=de-DE. I also took the opportunity to clean up our strings, and add a couple of German translations to demonstrate that translation works.

Closes https://github.com/vector-im/element-call/issues/369


src/App.tsx
```diff
@@ -43,34 +43,36 @@ export default function App({ history }: AppProps) {
...
-                </SentryRoute>
-                <SentryRoute path="*">
-                  <RoomRedirect />
-                </SentryRoute>
-              </Switch>
-            </OverlayProvider>
-          </Sentry.ErrorBoundary>
-        </InspectorContextProvider>
-      </ClientProvider>
+      <Suspense fallback={null}>
```
SimonBrandner, line 46: Why is this necessary, do we really need to use something that is afaik experimental?

SimonBrandner (approved):
LGTM otherwise

src/App.tsx
robintown: react-i18next triggers a suspense whenever it needs to fetch a translation file, so without a `Suspense` at some level in the component tree this will crash the application. There is [a way to use it without suspense](https://react.i18next.com/latest/usetranslation-hook#not-using-suspense), but it's rather cumbersome as it involves handling the loading state separately for every single component, and suspense is already stable in React 18 (which we should update to ASAP anyways…)

PR #619, opened 2022-10-13
Leave audio elements unmuted regardless of mute state

robintown:
Because to-device signaling can lag behind a user's actual audio, when you combine this with federation latency, you might end up cutting off the first few words of a user's sentence after they unmute.

I'd like to get this reviewed by @dbkr specifically before merging, to make sure this doesn't conflict with any walkie-talkie logic. (We did ultimately decide to leave feeds in walkie-talkie mode unmuted anyways, right?)

Closes https://github.com/vector-im/voip-internal/issues/63

### DIFF
--- a/src/video-grid/AudioContainer.tsx
+++ b/src/video-grid/AudioContainer.tsx
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React, { useEffect, useRef } from "react";
+import React, { FC, useEffect, useRef } from "react";
 
 import { Participant } from "../room/InCallView";
 import { useCallFeed } from "./useCallFeed";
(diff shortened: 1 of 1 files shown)


SimonBrandner (dismissed):
What about the spatial audio hook?

robintown:
Spatial audio already works like this, leaving everything except the local feed unmuted regardless of mute state

SimonBrandner (commented):
Ah, yeah, you're rignt

robintown:
@dbkr Since you're away but we need this a bit urgently, the decision was to merge this and have you review it some time after you're back

dbkr:
I'd say this looks fine - I think we were generally assuming we were playing back audio from participants, even if they were muted, including in PTT mode - we did consider muting them but I think in practice it was more important to not miss audio from people rather than ensure people were really muted when we thought they were.

PR #621, opened 2022-10-14
Hide the invite button in non-public rooms

robintown:
Closes https://github.com/vector-im/element-web/issues/23335, since the app currently doesn't have the ability to see join rule events in matryoshka mode, and so will always think a room is private in that case. It's a slightly roundabout way of solving the issue but it also ensures that private rooms on call.ems.host (they can technically exist) won't show the invite button either


src/room/useRoomState.ts
```diff
@@ -0,0 +1,51 @@
...
+export const useRoomState = <T>(room: Room, f: (state: RoomState) => T): T => {
+  const [numUpdates, setNumUpdates] = useState(0);
+  useTypedEventEmitter(
+    room,
+    RoomStateEvent.Update,
+    useCallback(() => setNumUpdates((n) => n + 1), [setNumUpdates])
+  );
+  // We want any change to the update counter to trigger an update here
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  return useMemo(() => f(room.currentState), [room, f, numUpdates]);
```
SimonBrandner, line 38: Why not call `f()` directly?

robintown: Because f might be some expensive calculation, or might return a different object on every call. In the case of useRoomAvatar and useJoinRule it's not, but I want this to be a hook we can use across the codebase, and memos as a rule never hurt

SimonBrandner (approved):

PR #627, opened 2022-10-14
Consolidate some similar strings

robintown:


src/room/RoomAuthView.tsx
```diff
@@ -98,7 +98,7 @@ export function RoomAuthView() {
         </main>
         <Body className={styles.footer}>
           <Trans>
-            {"Not registered yet? "}
+            Not registered yet?{" "}
```
SimonBrandner, line 101:
```suggestion
            Not registered yet?&nbsp;
```

Why not ^?

SimonBrandner (approved):

src/room/RoomAuthView.tsx
robintown: Hmm I mean, I don't really see a reason to make it a non-breaking space? If this and the link were on two separate lines, that would still read naturally to me, perhaps even more naturally

PR #631, opened 2022-10-17
Fix fullscreen buttons fullscreening the wrong feed

robintown:
Turns out `??` binds more tightly than `... ? ... : ...`

Closes https://github.com/vector-im/element-call/issues/624


SimonBrandner (approved):
I probably never would have realized this...

PR #641, opened 2022-10-19
Fix a soft crash when exiting fullscreen

robintown:
I can't quite remember how I triggered this, but it's apparently possible for exitFullscreen to be called while you're not in fullscreen, which throws an error. As a quick fix, this catches the errors.


robintown:
This wasn't the error

PR #648, opened 2022-10-21
Show tiles for members we're trying to connect to

dbkr:
This should help give more context on what's going wrong in splitbrain scenarios.

If users leave calls uncleanly, their tile will remain in until their member event times out, which will be an hour from when they joined the call. See https://github.com/vector-im/element-call/issues/639.

Part of https://github.com/vector-im/element-call/issues/616

### DIFF
--- a/src/room/GroupCallView.tsx
+++ b/src/room/GroupCallView.tsx
@@ -76,7 +76,6 @@ export function GroupCallView({
     toggleScreensharing,
     requestingScreenshare,
     isScreensharing,
-    localScreenshareFeed,
     screenshareFeeds,
     participants,
     unencryptedEventsFromUsers,
@@ -221,6 +220,7 @@ export function GroupCallView({
           client={client}
           roomName={groupCall.room.name}
           avatarUrl={avatarUrl}
+          participants={participants}
           microphoneMuted={microphoneMuted}
           localVideoMuted={localVideoMuted}
           toggleLocalVideoMuted={toggleLocalVideoMuted}
@@ -230,7 +230,6 @@ export function GroupCallView({
           onLeave={onLeave}
           toggleScreensharing={toggleScreensharing}
           isScreensharing={isScreensharing}
-          localScreenshareFeed={localScreenshareFeed}
           screenshareFeeds={screenshareFeeds}
           roomIdOrAlias={roomIdOrAlias}
           unencryptedEventsFromUsers={unencryptedEventsFromUsers}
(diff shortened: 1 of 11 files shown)


src/room/GroupCallView.tsx
```diff
@@ -76,7 +76,6 @@ export function GroupCallView({
     toggleScreensharing,
     requestingScreenshare,
     isScreensharing,
-    localScreenshareFeed,
```
dbkr, line 79: This was just unused

src/room/InCallView.tsx
```diff
@@ -82,14 +83,16 @@ interface Props {
...
-  localScreenshareFeed: CallFeed;
   roomIdOrAlias: string;
   unencryptedEventsFromUsers: Set<string>;
   hideHeader: boolean;
 }
 
-export interface Participant {
+// Represents something that should get a tile on the layout,
+// ie. a user's video feed or a screen share feed.
+export interface TileDescriptor {
```
dbkr, line 93: Renamed to try & resolve naming conflict as the list of room members was also called 'participants'. These aren't really participants as they also include extra tiles for the screenshares.

src/room/InCallView.tsx
```diff
@@ -185,39 +188,48 @@ export function InCallView({
...
+
+      // NB. this assumes that the same user can't join more than once from multiple
+      // devices, but the participants are just RoomMembers, so this assumption is baked
+      // into GroupCall itself.
+      tileDescriptors.push({
+        id: p.userId,
+        member: p,
+        callFeed: userMediaFeed,
+        focused: screenshareFeeds.length === 0 && p.userId === activeSpeaker,
+        isLocal: p.userId === client.getUserId(),
```
SimonBrandner, line 206:
```suggestion
        isLocal: userMediaFeed.isLocal(),
```

src/video-grid/useMediaStream.ts
```diff
@@ -35,21 +35,31 @@ declare global {
 export const useMediaStreamTrackCount = (
```
SimonBrandner, line 35: I am confused about what the changes here are doing

src/room/InCallView.tsx
dbkr: We might not have a userMediaFeed now, though

src/video-grid/useMediaStream.ts
robintown: I think the type of `stream` should be changed to reflect that then (or the other option is splitting `AudioForParticipant` into an outer and inner component to avoid adding nullability)

src/room/InCallView.tsx
SimonBrandner: Ah, right

src/video-grid/useMediaStream.ts
dbkr: Done -  I was planning on going through & doing a strict-mode sweep soon so was sort of neglecting these.

SimonBrandner (approved):
LGTM from my POV

PR #652, opened 2022-10-21
Fix screen sharing

dbkr:
* Make the embedded mode screen sharing a request-each-way rather than request-and-reply, since replies time out and so can't wait for the user.
 * Try normal screen sharing first, then fall back to using the widget API if it fails (for lack of a good way of detecting when we should be using the widget API).

Fixes https://github.com/vector-im/element-call/issues/649
Requires https://github.com/matrix-org/matrix-react-sdk/pull/9485

### DIFF
--- a/src/room/useGroupCall.ts
+++ b/src/room/useGroupCall.ts
@@ -27,10 +27,11 @@ import { MatrixCall } from "matrix-js-sdk/src/webrtc/call";
 import { CallFeed } from "matrix-js-sdk/src/webrtc/callFeed";
 import { RoomMember } from "matrix-js-sdk/src/models/room-member";
 import { useTranslation } from "react-i18next";
+import { IWidgetApiRequest } from "matrix-widget-api";
 
 import { usePageUnload } from "./usePageUnload";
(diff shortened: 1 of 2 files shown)


src/room/useGroupCall.ts
```diff
@@ -302,36 +303,84 @@ export function useGroupCall(groupCall: GroupCall): UseGroupCallReturnType {
...
-              });
+      } catch (e) {
+        // this will fail in Electron because getDisplayMedia just throws a permission
+        // error, so if we have a widget API, try requesting via that.
+        if (widget) {
+          const reply = await widget.api.transport.send(
+            ElementWidgetActions.ScreenshareRequest,
+            {}
+          );
+          if (!reply.pending) {
```
SimonBrandner, line 325: Should this be awaited?

SimonBrandner (approved):

src/room/useGroupCall.ts
dbkr: Nope - we await for the send which yields the reply, so `pending` is just a boolean flag on the reply object.

PR #658, opened 2022-10-24
Add posthog Telemetry (Anonymity Logic + call duration telemetry)

toger5:
Fixes: https://github.com/vector-im/element-call/issues/353
Fixes: https://github.com/vector-im/element-call/issues/659
Depends on: #682


src/PosthogAnalytics.ts
```diff
@@ -0,0 +1,340 @@
...
+  SignupTracker,
+  MuteCameraTracker,
+  MuteMicrophoneTracker,
+} from "./PosthogEvents";
+
+/* Posthog analytics tracking.
+ *
+ * Anonymity behaviour is as follows:
+ *
+ * - If Posthog isn't configured in `config.json`, events are not sent.
```
SimonBrandner, line 34: Don't think we have a `config.json`?

src/PosthogAnalytics.ts
```diff
@@ -0,0 +1,340 @@
...
+ * - If [Do Not Track](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack) is
+ *   enabled, events are not sent (this detection is built into posthog and turned on via the
+ *   `respect_dnt` flag being passed to `posthog.init`).
+ * - If the `feature_pseudonymous_analytics_opt_in` labs flag is `true`, track pseudonomously by maintaining
```
SimonBrandner, line 38: Do we really use this name `feature_pseudonymous_analytics_opt_in`?

SimonBrandner (commented):
Otherwise LGTM

src/PosthogAnalytics.ts
```diff
@@ -0,0 +1,340 @@
...
+}
+
+export enum RegistrationType {
+  Guest,
+  Registered,
+}
+
+interface PlatformProperties {
+  appVersion: string;
+  appPlatform: "embedded" | "jssdk";
```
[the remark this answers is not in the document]
toger5: for now the platform only represents which backend is used.
I think it should not be called platform then...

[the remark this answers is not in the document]
toger5: But now that you are saying it, it makes sense that in the future the PlatformProperties should also include Android, iOS, Web, Desktop...

src/PosthogAnalytics.ts
```diff
@@ -0,0 +1,336 @@
+/*
+Copyright 2021 The Matrix.org Foundation C.I.C.
```
dbkr, line 2: year

src/PosthogAnalytics.ts
```diff
@@ -0,0 +1,336 @@
...
+ *
+ * Anonymity behaviour is as follows:
+ *
+ * - If Posthog isn't configured in `config.json`, events are not sent.
+ * - If [Do Not Track](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack) is
+ *   enabled, events are not sent (this detection is built into posthog and turned on via the
+ *   `respect_dnt` flag being passed to `posthog.init`).
+ * - If the posthog analytics are explicitly activated by the user in the element call settings,
+ *   a randomised analytics ID is created and stored in account_data for that user (shared between devices)
+ *   so that the user can be identify in posthog.
```
dbkr, line 42:
```suggestion
 *   so that the user can be identified in posthog.
```

dbkr (approved):
otherwise lgtm - thanks for the detailed comments, they really help!

PR #667, opened 2022-10-26
Add option to allow ICE server fallback

SimonBrandner:


dbkr (commented):
Hmm - I don't think we can do this by default as it would mean hitting matrix.org on deployments that otherwise aren't associated with it.

SimonBrandner:
> Oh, I think I misread this and the default was off anyway? Either way, I guess good for the example to be consistent with the default.

Yeah and agreed :D

PR #674, opened 2022-10-27
Avoid Olm loading loop 

SimonBrandner:


dbkr (approved):
I also note that we have the Suspense here, although looks like this is only for lazy-loaded components. Not sure what components we lazy load. Anyway, this is good for now & fixes the app.

robintown:
The `Suspense` is there for `useTranslation` hooks, since they may need to fetch translation files

dbkr:
Ah, thanks. I guess we still can't use the suspense for the olm library loading?

PR #682, opened 2022-10-29
Async config file

toger5:
I could not add `VITE_PRODUCT_NAME` and `VITE_DEFAULT_HOMESERVER` for now since those are used in `vite.config.js` which to my understanding is executed before `App` or `main`.

For the sentry initialization I needed to introduce a `onAppLoaded` section in app so that the config is available for the senty initialization but we can show the loading screen (`App`-component) while the config is still downloading.


.gitignore
```diff
@@ -5,3 +5,4 @@ dist
 dist-ssr
 *.local
 .idea/
+config.json
```
SimonBrandner, line 8:
```suggestion
config.json

```

src/main.tsx
```diff
@@ -149,8 +138,22 @@ i18n
     },
   });
 
+// Initilization done after the App config has loaded:
+function onConfigLoaded() {
+  Sentry.init({
```
SimonBrandner, line 143: Is this async? Should we also await this? Could we put it into `App.tsx` since most of the init code is already there?

src/settings/submit-rageshake.ts
```diff
@@ -252,7 +253,7 @@ export function useSubmitRageshake(): {
         }
 
         await fetch(
-          (import.meta.env.VITE_RAGESHAKE_SUBMIT_URL as string) ||
+          Config.instance.config.rageshake?.submit_url ??
```
SimonBrandner, line 256: IMO, it would be better to use the defaults object somehow here

src/config/ConfigOptions.ts
```diff
@@ -0,0 +1,14 @@
...
+  sentry?: {
+    dns: string;
+    environment: string;
+  };
+  rageshake?: {
+    submit_url: string;
+  };
+}
+
+export const DEFAULT: IConfigOptions = {};
```
SimonBrandner, line 14:
```suggestion
export const DEFAULT_CONFIG: IConfigOptions = {};
```

To be more descriptive

src/App.tsx
```diff
@@ -31,33 +31,49 @@ import { usePageFocusStyle } from "./usePageFocusStyle";
...
 import { InspectorContextProvider } from "./room/GroupCallInspector";
 import { CrashView, LoadingView } from "./FullScreenView";
+import { Config } from "./config/Config";
 
 const SentryRoute = Sentry.withSentryRouting(Route);
 
 interface AppProps {
   history: History;
+  onConfigLoaded: () => void;
+}
```
SimonBrandner, line 41:
```suggestion
}

```

src/config/Config.ts
```diff
@@ -0,0 +1,62 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { DEFAULT, IConfigOptions } from "./ConfigOptions";
+
+export class Config {
+  static instance: Config;
```
SimonBrandner, line 20: Could this be a getter that throws when the internal instance is undefined, i.e. `init()` has not been called yet

src/config/Config.ts
```diff
@@ -0,0 +1,62 @@
...
+    Config.instance.initPromise = new Promise<void>((resolve) => {
+      downloadConfig("../config.json").then((config) => {
+        Config.instance.config = { ...Config.instance.config, ...config };
+        resolve();
+      });
+    });
+    return Config.instance.initPromise;
+  }
+
+  config: IConfigOptions;
```
SimonBrandner, line 35: Could we please be explicit about member visibility here?

src/config/Config.ts
```diff
@@ -0,0 +1,62 @@
...
+  initPromise: Promise<void>;
+  constructor() {
+    this.config = DEFAULT;
```
SimonBrandner, line 38: Something about this doesn't feel completely right

SimonBrandner (commented):
Looking good

src/config/Config.ts
```diff
@@ -14,29 +14,31 @@ See the License for the specific language governing permissions and
...
 
-import { DEFAULT, IConfigOptions } from "./ConfigOptions";
+import { IConfigOptions } from "./ConfigOptions";
 
 export class Config {
-  static instance: Config;
-  static init(): Promise<void> {
-    if (Config?.instance?.initPromise) {
-      return Config.instance.initPromise;
+  private static _instance: Config;
```
SimonBrandner, line 20:
```suggestion
  private static internalInstance: Config;
```

(Personal preference)

src/config/Config.ts
```diff
@@ -14,29 +14,31 @@ See the License for the specific language governing permissions and
...
+    if (Config?._instance?.initPromise) {
+      return Config._instance.initPromise;
     }
-    Config.instance = new Config();
-    Config.instance.initPromise = new Promise<void>((resolve) => {
+    Config._instance = new Config();
+    Config._instance.initPromise = new Promise<void>((resolve) => {
       downloadConfig("../config.json").then((config) => {
-        Config.instance.config = { ...Config.instance.config, ...config };
+        Config._instance.config = { ...Config._instance.config, ...config };
```
SimonBrandner, line 33:
```suggestion
        Config._instance.config = { ...Config._instance.config, ...DEFAULT_CONFIG };
```

src/useDependenciesLoaded.tsx
```diff
@@ -0,0 +1,123 @@
...
+      seti18nState(LoadState.Loaded);
+    }
+  }, [configState, i18nState, olmState, sentryState]);
+
+  return (
+    olmState === LoadState.Loaded &&
+    configState === LoadState.Loaded &&
+    sentryState === LoadState.Loaded &&
+    i18nState === LoadState.Loaded
+  );
```
SimonBrandner, line 138: This could probably be written by putting all the vars into an array and using `find(() => {})` to check if there is any value that is not loaded

src/useDependenciesLoaded.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+// We need to synchonysly track the if init on the deps is already called:
+// otherwise strict mode will trigger the effect twice with xxxState still beeing LoadState.None
+// resulting in calling init twice.
+let alreadyInitializedOlm = false;
+let alreadyInitializedConfig = false;
+let alreadyInitializedSentry = false;
+let alreadyInitializedi18n = false;
+
+export function useDependenciesLoaded(): boolean {
```
SimonBrandner, line 45: Do we take the `alreadyInitialized...` vars into account here? Perhaps this could be simplified further by using a singleton store to which the hook would only subscribe to

(when suggesting using a custom hook I did not realize we would need to use the global vars - sorry I overlooked that)

SimonBrandner: Alternatively, I would do `useState(alreadyInitializedX ? LoadState.Loaded : LoadState.None)`, I think that should work?

src/initializer.tsx
```diff
@@ -0,0 +1,149 @@
...
+  olm: LoadState = LoadState.None;
+  config: LoadState = LoadState.None;
+  sentry: LoadState = LoadState.None;
+  i18n: LoadState = LoadState.None;
+
+  allDepsAreLoaded() {
+    const notLoadedDeps = Object.values(this).filter(
+      (s) => s !== LoadState.Loaded
+    );
+    return notLoadedDeps.length === 0;
```
SimonBrandner, line 46:
```suggestion
    return !Object.values(this).some(
      (s) => s !== LoadState.Loaded
    );
```

Will `allDepsAreLoaded()` also be one of the `values`?

src/initializer.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+}
+
+export class Initializer {
+  private static _instance: Initializer;
+
+  public static init(): Promise<void> | null {
+    if (Initializer?._instance?.initPromise) {
+      return null;
```
SimonBrandner, line 55: null?

src/App.tsx
```diff
@@ -31,33 +29,29 @@ import { usePageFocusStyle } from "./usePageFocusStyle";
 import { SequenceDiagramViewerPage } from "./SequenceDiagramViewerPage";
 import { InspectorContextProvider } from "./room/GroupCallInspector";
 import { CrashView, LoadingView } from "./FullScreenView";
+import { Initializer } from "./initializer";
 
 const SentryRoute = Sentry.withSentryRouting(Route);
 
 interface AppProps {
   history: History;
 }
```
SimonBrandner, line 38:
```suggestion
}

```

src/initializer.tsx
toger5: otherwise we will attach multiple `then`s to the promise.
which would call `setLoaded(true);` multiple times. Not an issue but if someone adds functionality it might be confusing why its called multiple times? Or would that be expected?
```
useEffect(() => {
    Initializer.init()?.then(() => {
      setLoaded(true);
    });
  });
  ```

File: src/initializer.tsx (hunk 13 above)
Comment (toger5) [reply src/initializer.tsx:46]: I was also worried but that seems to be not the case with class instances.
`obj = {a:()=>{return 0}}` would definitly be included in the values.

File: src/initializer.tsx (hunk 14 above)
Comment (toger5) [reply src/initializer.tsx:55]: Looks like its fine:
image

Comment (SimonBrandner) [reply src/initializer.tsx:55]: Ah, yeah, right. Makes sense

File: src/initializer.tsx (hunk 13 above)
Comment (toger5) [reply src/initializer.tsx:46]: Looks like its fine
image

Comment (SimonBrandner) [reply src/initializer.tsx:46]: Great!

**Review by SimonBrandner — COMMENTED** [review]
LGTM now 

(requesting Dave's review just to be sure)

**Review by SimonBrandner — APPROVED** [review]

File: sample.config.json [hunk 16]
```diff
@@ -0,0 +1,13 @@
+{
+  "posthog": {
+    "api_key": "examplePosthogKey",
+    "api_host": "https://posthog.com"
+  },
+  "sentry": {
+    "environment": "main-branch",
+    "dns": "https://examplePublicKey@o0.ingest.sentry.io/0"
```
dbkr, line 8: DSN (Data Source name)

src/initializer.tsx
```diff
@@ -0,0 +1,146 @@
+/*
+Copyright 2021 New Vector Ltd
```
dbkr, line 2: matrix.org copyright, also the right year :)

dbkr (commented):
otherwise lgtm! Let's merge this once we've cut the release.

src/initializer.tsx
SimonBrandner: I thought since this is in `vector-im` we should use `New Vector LTD`?

toger5: I updated the year and matrix.org.
It seems we are using both in element call.
Should we unify that and move everything to one of the two?

toger5:
Olm got removed from the initializer since it is alrady handled during `initClient`:
https://github.com/vector-im/element-call/commit/d4caa1585b6874c0578658a19e9e3e3274d1e8a4#

PR #695, opened 2022-11-01
Fix missing tile bug

dbkr:
The 'connecting' tile change meant that we could have tiles right at the start of the call where we wouldn't have before, and in fact could have tiles for other users before we even had a tile for ourself. This threw off the logic for ordering tiles which had a special case for 1:1 calling which assumed that one of the tiles in a 1:1 call was the local user. In this case, this assumption wasn't true at the very start of the call, so the tile orders got assigned incorrectly and then persisted for the rest of the call.

Fixes https://github.com/vector-im/element-call/issues/694

### DIFF
--- a/src/room/useGroupCall.ts
+++ b/src/room/useGroupCall.ts
@@ -51,7 +51,7 @@ export interface UseGroupCallReturnType {
   requestingScreenshare: boolean;
   isScreensharing: boolean;
   screenshareFeeds: CallFeed[];
-  localDesktopCapturerSourceId: string;
+  localDesktopCapturerSourceId: string; // XXX: This looks unused?
   participants: RoomMember[];
   hasLocalParticipant: boolean;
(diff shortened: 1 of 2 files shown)


src/room/useGroupCall.ts
```diff
@@ -51,7 +51,7 @@ export interface UseGroupCallReturnType {
   requestingScreenshare: boolean;
   isScreensharing: boolean;
   screenshareFeeds: CallFeed[];
-  localDesktopCapturerSourceId: string;
+  localDesktopCapturerSourceId: string; // XXX: This looks unused?
```
SimonBrandner, line 54: True, I guess there could be a use case but it doesn't seem like something to worry about now

dbkr: Oh yes - I commented this a while back when I was looking at strict mode failures but apparently never committed it... until now!

PR #715, opened 2022-11-03
Enable fullscreen button for all tiles

notramo:
Currently, the fullscreen button in the top right corner of the video tile is only shown for screenshare.
This PR enables it for every video feed.

It is a useful feature, and it doesn't bring any considerable disadvantages.

Closes https://github.com/vector-im/element-call/issues/566

Signed-off-by: notramo <notramo@protonmail.com>

### DIFF
--- a/src/video-grid/VideoTile.tsx
+++ b/src/video-grid/VideoTile.tsx
@@ -82,16 +82,14 @@ export const VideoTile = forwardRef<HTMLDivElement, Props>(
         />
       );
 
-      if (screenshare) {
-        toolbarButtons.push(
-          <FullscreenButton
-            key="fullscreen"
-            className={styles.button}
-            fullscreen={fullscreen}
-            onPress={onFullscreen}
-          />
-        );
-      }
+      toolbarButtons.push(
+        <FullscreenButton
+          key="fullscreen"
+          className={styles.button}
+          fullscreen={fullscreen}
+          onPress={onFullscreen}
+        />
+      );
     }
 
     const caption = hasFeed ? name : t("{{name}} (Connecting...)", { name });


SimonBrandner (approved):
LGTM codewise

This is going to need a design review to determine if this is something we want, could you please add some screen-shots?

robintown:
@notramo Could you please [sign off](https://github.com/vector-im/element-web/blob/develop/CONTRIBUTING.md#sign-off) your changes?

notramo:
@robintown I updated the comment with the sign-off line.

@SimonBrandner there are the screenshots (of a camera feed). Notice the fullscreen icon in the top right corner of the card. 
| current | this PR |
|-|-|
|call-current|call|

dbkr:
Our fullscreen behaviour does feel a little odd right now and I'm not sure that being able to full-screen individual feeds is really a sensible thing, screenshare or not: normally I would expect to just full screen the whole app, separately having the option to hide other feeds perhaps, which would achieve the same thing, although this is a question for design.

notramo:
Why is it a problem? If the user doesn't like it, then they doesn't use it. But if somebody finds it useful, then they could use it.

gaelledel:
Hi there, thanks for the suggestion. Full screen is not necessary in the case of a tile because you can expand it on double click and change layout to access larger view. What might be useful however is a full screen at app level.

gaelledel (commented):
Hi there, thanks for the suggestion. Full screen is not necessary in the case of a tile because you can expand it on double click and change layout to access larger view. What might be useful however is a full screen at app level.

spongythecake:
> 
> Hi there, thanks for the suggestion. Full screen is not necessary in the case of a tile because you can expand it on double click and change layout to access larger view. What might be useful however is a full screen at app level.
> 

You said fullscreen is not necessary. Have you consulted with an accessibility auditor on this decision or papers? 

Yes, a tile can be expanded. However, you still can not make full use of a monitor. 
Every other video app has a fullscreen, and Element Call seems unique in their thinking. 

Also, if you watched Matrix Live pre-ElementCall, the hosts would fullscreen their guests. So, the point is that people closest to you in your office were using a feature that you don't consider important.

notramo:
This is the kind of feature that will encourage downstream forks instead of contributing to upstream so everyone can benefit.

BujuArena:
What, this has a real implementation but it's actually closed, intentionally, and more than 2 years ago?  Is there a pop-out feature planned where we can full-screen the pop-out, at least?  Otherwise, Element Call doesn't make sense for video calls on desktop computers, meaning it doesn't suit what seems to be its primary purpose.

notramo:
On top of that, the Picture-In-Picture button was also disabled. Why?

The Element Call team should notice that not everyone works on 3 or more huge 4K monitors... Even the latest macBook Pro has a 14 inch version, not mentioning all the other laptops (e.g. ThinkPad X series with 12.1 inch display) which are quite popular. Please support those devices too.

BujuArena:
I can't see my friends' screen shares clearly, which makes Element Call useless for pair programming and similar screen-sharing activities which work just fine in Slack, for example.

lenoregilbert:
Seconding the frustration here - A pop-out feature in lieu of this is pretty important and is a feature parity with other popular (less secure) chat apps many users are accustomed to. I found this thread while looking to see if anyone else had discussed per tile full-screen with some motivation to attempt adding the pop-out feature myself but this seems pretty discouraging!

PR #716, opened 2022-11-03
Preserve aspect ratio for fullscreen video

notramo:
Current behavior: crop fullscreen video to 16:9
This PR: preserve aspect ratio when video is fullscreen.

Note that screensharing already preserves the aspect ratio, so this is only relevant for regular streams in fullscreen (#715).

### DIFF
--- a/src/video-grid/VideoTile.module.css
+++ b/src/video-grid/VideoTile.module.css
@@ -47,6 +47,7 @@
   width: 100%;
 }
 
+.videoTile.maximised > video,
 .videoTile.screenshare > video {
   object-fit: contain;
 }


robintown:
@notramo Could you please add a screenshot if possible?

@gaelledel We'll need your input on whether we should make this change. Note that it will also affect Element Web's picture-in-picture view, by causing the active speaker's video to be letterboxed.

robintown (approved):
Code looks fine for what it does

notramo:
Screen Shot 2022-11-03 at 17 41 18

robintown:
Since the decision we've come to is to not allow usermedia tiles to be put in fullscreen, I don't believe this PR is still relevant.

PR #717, opened 2022-11-03
Add config file for CD version

dbkr:


config/element_io_develop.json
```diff
@@ -0,0 +1,13 @@
+{
+  "posthog": {
+    "api_key": "phc_MhClVy9DiV20vazSYIiedFkM5Xi3z1LPBwrdn9PYZQQ",
```
toger5, line 3: This means the keys are just fully public on gh?
Which is not a big deal since they can be read inside the a browser terminal anyways?

config/element_io_develop.json
```diff
@@ -0,0 +1,13 @@
...
+    "api_host": "https://app.posthog.com"
```
robintown, line 4: Is that the right place? I thought we were using a self-hosted Posthog

dbkr: Good point - @toger5 ?

toger5 (approved):

config/element_io_develop.json
dbkr: Yeah, this is what we do for element-web: https://github.com/vector-im/element-web/blob/develop/element.io/app/config.json#L49

config/element_io_develop.json
toger5: No not yet. We are using a seperate instance for call. And a second self hosted one would cost a lot more.

PR #721, opened 2022-11-03
Initialize i18next before mounting components

robintown:
Components such as the loading screen expect it to already be initialized.

### DIFF
--- a/src/initializer.tsx
+++ b/src/initializer.tsx
@@ -15,13 +15,8 @@ limitations under the License.
 */
 
 import { Integrations } from "@sentry/tracing";
-import i18n from "i18next";
-import { initReactI18next } from "react-i18next";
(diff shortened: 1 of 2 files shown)


toger5:
I was thinking moving things out of main into the `initializer.ts`.
Here is what I had in mind: https://github.com/vector-im/element-call/pull/723
It does not really bring a benefit except making main cleaner.

toger5 (approved):
Looks good! Thx
If this is the preferred solution compared to https://github.com/vector-im/element-call/pull/723 you can merge this.
(Sorry that I missed your PR. Otherwise I would have just approved this one here...)
Feel free to merge this

robintown:
Let's merge https://github.com/vector-im/element-call/pull/723 to keep main.tsx clean

PR #726, opened 2022-11-04
Disable posthog in widgets

toger5:
There have to be some widget issues that make the posthog class not work. Check for widget environment before enabling posthog.


src/PosthogAnalytics.ts
```diff
@@ -124,7 +124,10 @@ export class PosthogAnalytics {
         respect_dnt: true,
         advanced_disable_decide: true,
       });
-      this.enabled = true;
+      if (PosthogAnalytics.getPlatformProperties().matrixBackend === "jssdk") {
```
robintown, line 127: I'd suggest moving this to the `if (posthogConfig.project_api_key && posthogConfig.api_host)` to not initialize Posthog in widgets either

robintown (approved):

PR #727, opened 2022-11-04
Fix close button on join call modal

dbkr:
It just errored as we didn't pass onClose through to the modal


robintown (commented):
Is there a reason we haven't made `onClose` required for modals? Since the modal component does seem to expect it to be defined

dbkr:
Mmm, it's weird that it's optional since it looks like there's always a close button.

robintown (approved):

PR #728, opened 2022-11-04
Add aria-describedby associations

dbkr:


src/input/Input.tsx
```diff
@@ -122,6 +122,7 @@ export const InputField = forwardRef<
             type={type}
```
robintown, line 122: The textarea variant above probably needs aria-describedby too

src/input/Input.tsx
```diff
@@ -122,6 +122,7 @@ export const InputField = forwardRef<
...
             checked={checked}
             disabled={disabled}
+            aria-describedby={description ? id + "-desc" : undefined}
```
robintown, line 125: `id` could be undefined, so I'd suggest generating one. react-aria has a `useId` hook: https://react-spectrum.adobe.com/react-aria/useId.html

dbkr: Ah, useful - thanks.

robintown (dismissed):

robintown (approved):

PR #747, opened 2022-11-10
Make room / call creation logging more accurate

dbkr:
It said it was creating the room when actually it was creating the call

### DIFF
--- a/src/matrix-utils.ts
+++ b/src/matrix-utils.ts
@@ -277,7 +277,7 @@ export async function createRoom(
 
   const result = await createPromise;
 
-  console.log(
+  logger.log(
     `Creating ${ptt ? "PTT" : "video"} group call in ${result.room_id}`
   );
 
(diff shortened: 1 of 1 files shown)


SimonBrandner (approved):

src/matrix-utils.ts
```diff
@@ -225,6 +225,7 @@ export async function createRoom(
   name: string,
   ptt: boolean
 ): Promise<[string, string]> {
+  console.log(`Creating room for froup call`);
```
robintown, line 228:
```suggestion
  logger.log(`Creating room for group call`);
```

src/matrix-utils.ts
```diff
@@ -276,7 +277,9 @@ export async function createRoom(
 
   const result = await createPromise;
 
-  console.log(`Creating ${ptt ? "PTT" : "video"} group call room`);
+  console.log(
```
robintown, line 280:
```suggestion
  logger.log(
```

PR #749, opened 2022-11-11
Add 'm' and 'space' shortcuts for mute/unmuting during a call

erikjohnston:
I rely on using the `m` and `space` shortcuts to toggle mute in Jitsi all the time, and I *really* miss it.

I have *no* idea what I'm doing when it comes to react and javascript or whatever language I have just written, but it seems to work when I test it :shrug: 

Related issue: https://github.com/vector-im/element-call/issues/218

### DIFF
--- a/public/locales/en-GB/app.json
+++ b/public/locales/en-GB/app.json
@@ -134,6 +134,7 @@
   "Walkie-talkie call": "Walkie-talkie call",
   "Walkie-talkie call name": "Walkie-talkie call name",
   "WebRTC is not supported or is being blocked in this browser.": "WebRTC is not supported or is being blocked in this browser.",
+  "Whether to enable the keyboard shortcuts to e.g. mute/unmute the local microphone.": "Whether to enable the keyboard shortcuts to e.g. mute/unmute the local microphone.",
   "Yes, join call": "Yes, join call",
   "You can't talk at the same time": "You can't talk at the same time",
   "Your recent calls": "Your recent calls"
--- a/public/locales/en-GB/app.json
+++ b/public/locales/en-GB/app.json
@@ -42,6 +42,7 @@
   "Display name": "Display name",
   "Download debug logs": "Download debug logs",
   "Element Call Home": "Element Call Home",
+  "Enable keyboard shortcuts": "Enable keyboard shortcuts",
   "Entering room…": "Entering room…",
   "Exit full screen": "Exit full screen",
   "Fetching group call timed out.": "Fetching group call timed out.",
--- a/public/locales/en-GB/app.json
+++ b/public/locales/en-GB/app.json
@@ -134,7 +134,7 @@
   "Walkie-talkie call": "Walkie-talkie call",
   "Walkie-talkie call name": "Walkie-talkie call name",
   "WebRTC is not supported or is being blocked in this browser.": "WebRTC is not supported or is being blocked in this browser.",
-  "Whether to enable the keyboard shortcuts to e.g. mute/unmute the local microphone.": "Whether to enable the keyboard shortcuts to e.g. mute/unmute the local microphone.",
+  "Whether to enable the keyboard shortcuts, e.g. 'm' to mute/unmute the mic.": "Whether to enable the keyboard shortcuts, e.g. 'm' to mute/unmute the mic.",
   "Yes, join call": "Yes, join call",
   "You can't talk at the same time": "You can't talk at the same time",
   "Your recent calls": "Your recent calls"
--- a/public/locales/en-GB/app.json
+++ b/public/locales/en-GB/app.json
@@ -42,7 +42,7 @@
   "Display name": "Display name",
   "Download debug logs": "Download debug logs",
   "Element Call Home": "Element Call Home",
-  "Enable keyboard shortcuts": "Enable keyboard shortcuts",
+  "Single-key keyboard shortcuts": "Single-key keyboard shortcuts",
   "Entering room…": "Entering room…",
   "Exit full screen": "Exit full screen",
   "Fetching group call timed out.": "Fetching group call timed out.",
@@ -134,7 +134,7 @@
   "Walkie-talkie call": "Walkie-talkie call",
   "Walkie-talkie call name": "Walkie-talkie call name",
   "WebRTC is not supported or is being blocked in this browser.": "WebRTC is not supported or is being blocked in this browser.",
-  "Whether to enable the keyboard shortcuts, e.g. 'm' to mute/unmute the mic.": "Whether to enable the keyboard shortcuts, e.g. 'm' to mute/unmute the mic.",
+  "Whether to enable single-key keyboard shortcuts, e.g. 'm' to mute/unmute the mic.": "Whether to enable single-key keyboard shortcuts, e.g. 'm' to mute/unmute the mic.",
   "Yes, join call": "Yes, join call",
   "You can't talk at the same time": "You can't talk at the same time",
   "Your recent calls": "Your recent calls"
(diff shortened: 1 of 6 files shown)


robintown:
Single-button shortcuts like M and especially space can create accessibility issues, so I think we may want to explore alternatives, or make them opt-in

jakewb-b (commented):
I agree we should have the feature but, as Robin says, it can create accessibility issues and I think would be flagged under WCAG guidelines so I'd prefer to have it as a setting - perhaps in the new 'advanced' tab that we're putting other features in. @gaelledel could you confirm you'd be happy with this?

Also, this wouldn't be a blocked to merging, but is it possible to set it so that if you short-press on the shortcut button then it toggles mute/unmute, but if you press-and-hold on it while muted, then it unmutes you temporarily and re-mutes you when you release. We might need to experiment with what the right length of time is to differentiate the behaviours, but other apps have this function and it works nicely.

erikjohnston:
Ah, I hadn't realised that we'd need to make this configurable. I can have a quick look to see how hard it would be for me to add that.


> Also, this wouldn't be a blocked to merging, but is it possible to set it so that if you short-press on the shortcut button then it toggles mute/unmute, but if you press-and-hold on it while muted, then it unmutes you temporarily and re-mutes you when you release. We might need to experiment with what the right length of time is to differentiate the behaviours, but other apps have this function and it works nicely.

That sounds a bit like what `space` is doing in this PR (again, copied shamelessly from jitsi): holding space will unmute you until you release it. There is no delay though, so if you just tap space it'd always mute you. The `m` shortcut is a standard toggle.

jakewb-b:
> holding space will unmute you until you release it. There is no delay though, so if you just tap space it'd always mute you. The m shortcut is a standard toggle.

Ah that's perfect then - exactly what I was after. Thanks!

erikjohnston:
Have stuck an option under "Advanced", with the setting defaulting to true:

Screenshot from 2022-11-14 10-25-09

erikjohnston:
Does I need to do anything else for translations now that I've added an extra string?

jakewb-b:
Thanks. I don't know if we should default to false - even though I think most people would prefer to have the setting, if it's an accessibility issue we might be better off not making it the default. I might need to dig into the accessibility rules as I'm not certain - unless @gaelledel or @robintown know?

Also - @robintown do we have any other keyboard shortcuts currently? I'm happy having all keyboard shortcuts linked to this one setting - that's probably better than having a whole bunch of settings for different shortcuts, but in that case we need to make sure they actually _are_ all linked to it. And I wonder if we should list what the shortcuts are?

erikjohnston:
> Also - @robintown do we have any other keyboard shortcuts currently? I'm happy having all keyboard shortcuts linked to this one setting - that's probably better than having a whole bunch of settings for different shortcuts, but in that case we need to make sure they actually _are_ all linked to it. And I wonder if we should list what the shortcuts are?

The only instances in the code I can see are for PTT, FWIW

jakewb-b:
> The only instances in the code I can see are for PTT, FWIW

We should probably tie that to this if it's not already. It's *essentially* an unmute button anyway so I think it would be a bit odd if you turned the new shortcut off, and the PTT shortcut wasn't affected.

That said, I don't really want the PTT shortcut to be off-by-default as it's quite a fundamental behaviour and will be a hassle to always have to go and turn it on. So I'm inclined to stick with your current behaviour of on-by-default for now, and then we can always revisit if this does become an accessibility issue. In the meantime I'll do some digging into exactly what the rules/guidance say.

gaelledel:
We'd need to have a description of the keyboard shortcut in settings - What it is and what is does. It needs to be disabled by default.

erikjohnston:
I've naively added the PTT stuff behind the setting. Not actually sure how to test it though (or really what it is)

jakewb-b:
> It needs to be disabled by default.

@gaelledel is that definitive based on WCAG guidelines? It's not a huge deal, but it is a slight nuisance for PTT

erikjohnston:
> We'd need to have a description of the keyboard shortcut in settings - What it is and what is does.

Does this need to include the list of shortcuts? I'm a bit worried about listing all the keyboard shortcuts there as it would likely get out of date quickly.

public/locales/en-GB/app.json
```diff
@@ -133,6 +134,7 @@
   "Walkie-talkie call": "Walkie-talkie call",
   "Walkie-talkie call name": "Walkie-talkie call name",
   "WebRTC is not supported or is being blocked in this browser.": "WebRTC is not supported or is being blocked in this browser.",
+  "Whether to enable the keyboard shortcuts to e.g. mute/unmute the local microphone.": "Whether to enable the keyboard shortcuts to e.g. mute/unmute the local microphone.",
```
erikjohnston, line 137: I'm not particularly happy with this description, but put it up as a straw man for people to comment on.

src/room/useGroupCall.ts
```diff
@@ -395,6 +403,43 @@ export function useGroupCall(groupCall: GroupCall): UseGroupCallReturnType {
...
+
+    const keyUpListener = (event) => {
+      // Check if keyboard shortcuts are enabled
+      const keyboardShortcuts = getSetting("keyboard-shortcuts", true);
+      if (!keyboardShortcuts) {
+        return;
+      }
+
+      if (event.key === " ") {
+        setMicrophoneMuted(true);
```
dbkr, line 430: Note that there are also gotchas here when the windows gets unfocused while the user has the button held which means the app never gets the keyup event. The PTT code should handle this so we may want to re-use the code from there, although it's probably a thing we could do later.

public/locales/en-GB/app.json
```diff
@@ -42,6 +42,7 @@
   "Display name": "Display name",
   "Download debug logs": "Download debug logs",
   "Element Call Home": "Element Call Home",
+  "Enable keyboard shortcuts": "Enable keyboard shortcuts",
```
dbkr, line 45: "Single-key keyboard shortcuts" perhaps? (since it's specifically ones that activate without modifier keys held)

public/locales/en-GB/app.json
dbkr: We could also just explain by example here, like:

"e.g. 'm' to mute/unmute the mic"

src/room/useGroupCall.ts
```diff
@@ -395,6 +403,43 @@ export function useGroupCall(groupCall: GroupCall): UseGroupCallReturnType {
...
+      if (!keyboardShortcuts) {
+        return;
+      }
+
+      if (event.key === "m") {
+        toggleMicrophoneMuted();
+      }
+      if (event.key === " ") {
+        setMicrophoneMuted(false);
+      }
```
dbkr, line 419: We might also want to think about adding 'v' if we're aiming for Jitsi parity (ie. camera mute toggle)

src/room/useGroupCall.ts
erikjohnston: So it looks to me like PTT auto-mutes if the window loses focus. I think the equivalent here would be to mute if the window loses focus while the space bar is down? I'm not sure which is more intuitive: that it mutes or stays unmuted in that case.

Happy to make the change, though I would need some guidance on how best to track a variable for "is the spacebar currently down?"

public/locales/en-GB/app.json
dbkr: I can't speak for product but the single-key shortcuts are the ones that cause issues for a11y (especially spacebar since that's used for interacting with elements on the page on place of a click).

src/room/useGroupCall.ts
dbkr: sgtm - probably easiest to just add another state variable (`const { spacebarHeld, setSpacebarHeld } = useState(false);`) and then `setSpacebarHeld()` with the appropriate values in keydown/keyup and when the window loses focus.

public/locales/en-GB/app.json
jakewb-b: Yeah, agreed - we can make this exclusively about single-key shortcuts.

erikjohnston: Ok cool, will change the wording then.

erikjohnston: Done

src/room/useGroupCall.ts
erikjohnston: Done.

PR #752, opened 2022-11-15
Add a 'waiting for video' state to media tiles

dbkr:
This will show if the call is waiting for media to connect (in practice doesn't actually seem to happen all that often) but also show if the media connection is lost, with the js-sdk change.

Requires https://github.com/matrix-org/matrix-js-sdk/pull/2880
Fixes: https://github.com/vector-im/element-call/issues/669

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -14,7 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React, { useEffect, useCallback, useMemo, useRef } from "react";
+import React, {
+  useEffect,
+  useCallback,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 import { usePreventScroll } from "@react-aria/overlays";
 import useMeasure from "react-use-measure";
 import { ResizeObserver } from "@juggle/resize-observer";
@@ -25,6 +31,11 @@ import { CallFeed } from "matrix-js-sdk/src/webrtc/callFeed";
 import classNames from "classnames";
 import { useTranslation } from "react-i18next";
 import { JoinRule } from "matrix-js-sdk/src/@types/partials";
+import {
+  CallEvent,
+  CallState,
+  MatrixCall,
+} from "matrix-js-sdk/src/webrtc/call";
 
 import type { IWidgetApiRequest } from "matrix-widget-api";
 import styles from "./InCallView.module.css";
@@ -73,6 +84,7 @@ interface Props {
   client: MatrixClient;
   groupCall: GroupCall;
   participants: RoomMember[];
+  calls: MatrixCall[];
   roomName: string;
   avatarUrl: string;
   microphoneMuted: boolean;
@@ -90,6 +102,12 @@ interface Props {
   hideHeader: boolean;
 }
 
+export enum ConnectionState {
+  ESTABLISHING_CALL = "establishing call", // call hasn't been established yet
+  WAIT_MEDIA = "wait_media", // call is set up, waiting for ICE to connect
+  CONNECTED = "connected", // media is flowing
+}
+
 // Represents something that should get a tile on the layout,
 // ie. a user's video feed or a screen share feed.
 export interface TileDescriptor {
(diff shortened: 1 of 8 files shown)


src/room/InCallView.tsx
```diff
@@ -231,11 +292,19 @@ export function InCallView({
         focused: true,
         isLocal: screenshareFeed.isLocal(),
         presenter: false,
+        connectionState: ConnectionState.CONNECTED, // by definition since the screen shares arrived on the same connection
```
SimonBrandner, line 295: What if the connection dropped later?

src/room/InCallView.tsx
```diff
@@ -154,6 +174,46 @@ export function InCallView({
 
   const { hideScreensharing } = useUrlParams();
 
+  const makeConnectionStatesMap = useCallback(() => {
+    const newConnStates = new Map<string, ConnectionState>();
+    for (const participant of participants) {
+      const userCall = groupCall.getCallByUserId(participant.userId);
+      const feed = userMediaFeeds.find((f) => f.userId === participant.userId);
+      let connectionState = ConnectionState.ESTABLISHING_CALL;
+      if (feed && feed.isLocal()) {
```
SimonBrandner, line 183: How does the fact we have a local feed mean we are connected?

src/room/InCallView.tsx
```diff
@@ -90,6 +102,12 @@ interface Props {
   hideHeader: boolean;
 }
 
+export enum ConnectionState {
```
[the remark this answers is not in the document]
dbkr: Looks like we do? https://github.com/vector-im/element-call/blob/main/src/home/CallTypeDropdown.tsx#L32

src/room/InCallView.tsx
dbkr: Well, it's our own feed, so I think that effectively means it's always connected? We could also say connection state is N/A for a local feed, although it would make it slightly more complex.

src/room/InCallView.tsx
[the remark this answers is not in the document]
SimonBrandner: I meant all caps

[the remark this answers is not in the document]
SimonBrandner: I think we don't use that...

src/room/InCallView.tsx
dbkr: True - have passed through the connection state of the relevant user (although we don't display it for screenshares anyway).

src/room/InCallView.tsx
```diff
@@ -154,6 +174,46 @@ export function InCallView({
...
+  const makeConnectionStatesMap = useCallback(() => {
+    const newConnStates = new Map<string, ConnectionState>();
+    for (const participant of participants) {
+      const userCall = groupCall.getCallByUserId(participant.userId);
+      const feed = userMediaFeeds.find((f) => f.userId === participant.userId);
+      let connectionState = ConnectionState.ESTABLISHING_CALL;
+      if (feed && feed.isLocal()) {
+        connectionState = ConnectionState.CONNECTED;
+      } else if (userCall) {
+        if (userCall.state === CallState.Connected) {
```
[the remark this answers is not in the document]
dbkr: I don't think we really care about these for the purposes of displaying the tile: they'll all use the default of `ESTABLISHING_CALL`.

src/room/InCallView.tsx
SimonBrandner: Ah, sorry, I was thinking about this the wrong way

src/room/InCallView.tsx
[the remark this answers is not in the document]
dbkr: Oh - of course, sorry 🤦

src/video-grid/VideoTile.tsx
```diff
@@ -94,7 +95,20 @@ export const VideoTile = forwardRef<HTMLDivElement, Props>(
       }
     }
 
-    const caption = hasFeed ? name : t("{{name}} (Connecting...)", { name });
+    let caption: string;
+    switch (connectionState) {
+      case ConnectionState.EstablishingCall:
+        caption = t("{{name}} (Connecting...)", { name });
+
```
robintown, line 102:
```suggestion
```

robintown (approved):
LGTM

PR #761, opened 2022-11-21
Enable users to join calls from multiple devices

robintown:
This updates to https://github.com/matrix-org/matrix-js-sdk/pull/2902 and addresses many places where the app previously assumed that a given user can only join the call once. One notable exception is the inspector, which still works, but doesn't output very helpful diagrams if view a user who was joined on multiple devices.


dbkr (approved):
🎉 Thanks for sorting this out properly!

PR #763, opened 2022-11-22
Fix click leaking through to DOM element underneath

dbkr:
See comment, although this is quite hack - I'm torn on whether this is worth it for the bugfix. Upgrading react-aria doesn't fix it either (and also breaks everything in React strict mode).

Fixes https://github.com/vector-im/element-call/issues/762

### DIFF
--- a/src/ListBox.tsx
+++ b/src/ListBox.tsx
@@ -73,6 +73,19 @@ function Option<T>({ item, state, className }: OptionProps<T>) {
     ref
   );
 
+  // Hack: remove the onPointerUp event handler and re-wire it to
+  // onClick. Chrome Android triggers a click event after the onpointerup
+  // event which leaks through to elements underneath the z-indexed select
+  // popover. preventDefault / stopPropagation don't have any effect, even
+  // adding just a dummy onClick handler still doesn't work, but it's fine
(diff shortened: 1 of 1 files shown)


src/ListBox.tsx
```diff
@@ -73,6 +73,19 @@ function Option<T>({ item, state, className }: OptionProps<T>) {
...
+  // event which leaks through to elements underneath the z-indexed select
+  // popover. preventDefault / stopPropagation don't have any effect, even
+  // adding just a dummy onClick handler still doesn't work, but it's fine
+  // if we handle just onClick.
+  // https://github.com/vector-im/element-call/issues/762
+  const origPointerUp = optionProps.onPointerUp;
+  delete optionProps.onPointerUp;
+  optionProps.onClick = (e) => {
+    origPointerUp(e as unknown as React.PointerEvent<HTMLElement>);
+  };
```
robintown, line 87:
```suggestion
  optionProps.onClick = useCallback((e) => {
    origPointerUp(e as unknown as React.PointerEvent<HTMLElement>);
  }, [origPointerUp]);
```

robintown (approved):

PR #767, opened 2022-11-28
Posthog widget embedding

toger5:
Fixes: https://github.com/vector-im/element-call/issues/764
This loads the "analyticsID" form element web in case the user already has given consent to send analytics to posthog.
Relies on: https://github.com/matrix-org/matrix-react-sdk/pull/9637

Allows to check in what context the app gets used:
image


src/PosthogAnalytics.ts
```diff
@@ -218,29 +235,56 @@ export class PosthogAnalytics {
...
+          await this.setAnalyticsId(analyticsID, client);
         }
-        this.posthog.identify(analyticsID);
       } catch (e) {
         // The above could fail due to network requests, but not essential to starting the application,
         // so swallow it.
         logger.log("Unable to identify user for tracking" + e.toString());
       }
+      if (analyticsID) {
+        this.posthog.identify(analyticsID);
```
t3chguy, line 257: This looks like it could leak an EW analytics ID to a different posthog instance?

toger5: The plan is so that element web and element call use the same analytics id. That mirrors the behaviour of using the standalone element call client. (There the analytics id would be loaded from the account data if already available)
The url parameters from where the id is passed are element call specific. So no other widget will have access to the id.

t3chguy: Don't they have separate configs and thus could point at different posthog instances?

src/PosthogAnalytics.ts
```diff
@@ -16,9 +16,10 @@ limitations under the License.
 
 import posthog, { CaptureOptions, PostHog, Properties } from "posthog-js";
 import { logger } from "matrix-js-sdk/src/logger";
+import { MatrixClient } from "matrix-js-sdk";
```
robintown, line 19:
```suggestion
import { MatrixClient } from "matrix-js-sdk/src/client";
```

Importing from `matrix-js-sdk` directly can break things IIRC

src/PosthogAnalytics.ts
```diff
@@ -218,29 +235,56 @@ export class PosthogAnalytics {
...
         // The above could fail due to network requests, but not essential to starting the application,
         // so swallow it.
         logger.log("Unable to identify user for tracking" + e.toString());
       }
+      if (analyticsID) {
+        this.posthog.identify(analyticsID);
+      } else {
+        console.warn(
+          "No analyticsID is availble. Should not try to setup posthog"
+        );
```
robintown, line 261: This feels like it should be a `logger.info`

robintown (commented):
Code looks good overall

src/PosthogAnalytics.ts
t3chguy: As long as you're consistent with the import it is fine, `matrix-js-sdk` maps to `/src/` or `/lib/` depending on if release or develop repsectively

src/PosthogAnalytics.ts
toger5: This is now fixed by only sending a hash of `'ec' + accountAnalyticsId + client.getUserId()` to posthog.

src/PosthogAnalytics.ts
```diff
@@ -285,6 +354,11 @@ export class PosthogAnalytics {
       : Anonymity.Disabled;
     this.setAnonymity(anonymity);
     if (anonymity === Anonymity.Pseudonymous) {
+      this.setRegistrationType(
+        window.matrixclient.isGuest() || window.isPasswordlessUser
```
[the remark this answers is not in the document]
toger5: Can i get acces to `isPasswordlessUser` from the client peg? It seems like this is a element call specific property not part of the client interface?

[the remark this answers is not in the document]
dbkr: Ah, no you can't since the js-sdk doesn't know about it. You would indeed need the context, or which is impossible using a shared instance method - we have an impedance mismatch between singletons and contexts. We should probably decide on one or the other rather than mixing them, but this is probably an okay workaround for now.

dbkr (approved):

src/PosthogAnalytics.ts
```diff
@@ -124,9 +125,7 @@ export class PosthogAnalytics {
         const { analyticsID } = getUrlParams();
         // if the embedding platform (element web) already got approval to communicating with posthog
         // element call can also send events to posthog
-        if (analyticsID) {
-          setSetting("opt-in-analytics", true);
-        }
+        setSetting("opt-in-analytics", !!analyticsID);
```
dbkr, line 128: I think `Boolean(analyticsID)` is favoured (wonder if there's a lint rule for this).

dbkr (approved):

src/PosthogAnalytics.ts
toger5: Okay good to know. A lint rule would be great. `!!` is still used in a couple of occasions (I checked before using it).

PR #773, opened 2022-12-02
Autofocus authentication forms

robintown:


robintown:
Linter has pointed out that autofocus can harm accessibility, so let's actually not

PR #791, opened 2022-12-16
Make spatial audio Firefox-only

dbkr:
Hopefully explained in comment: we have a heisenbug where we sometimes lack audio from a certain participant, so this simplifies the audio path by removing the workaround required to do AEC with spatial audio on chrome.

Screenshot 2022-12-16 at 16 43 04


src/video-grid/VideoTileContainer.tsx
```diff
@@ -72,7 +72,7 @@ export function VideoTileContainer({
     audioContext,
     audioDestination,
     localVolume,
-    isLocal || maximised
+    isLocal
```
SimonBrandner, line 75: I don't think I see why this got removed?

SimonBrandner (approved):
LGTM otherwise

src/video-grid/VideoTileContainer.tsx
dbkr: Yeah, working out why it was there in the first place was also rather subtle. We routed audio differently when a video was maximised, using the AudioContainer things. This is no longer necessary since the audio elements can just stay in the DOM.

SimonBrandner: Well, I don't think we would want to play audio both from `AudioSink`s and the spatial audio thingy when full-screened? What am I missing?

dbkr: Yep, the spatial audio pathways are only active when spatial audio is turned on. Previously we used a separate audio context based audio path when in fullscreen because all the other video elements went away. Now we use the audio elements whether fullscreen or not.

SimonBrandner: Sorry, I think I am still missing something... - what about the case where you have spatial audio and are in fullscreen, how is audio routed in this case?

dbkr: Ah yes sorry, turns out I probably did have double audio here for the maximised video as it does go via the audio elements when maximised, spatial audio or no spatial audio. Fixed.

SimonBrandner (approved):

PR #798, opened 2022-12-20
Move default homeserver to config file

dbkr:


src/config/ConfigOptions.ts
```diff
@@ -19,21 +19,24 @@ export interface ConfigOptions {
   rageshake?: {
     submit_url: string;
   };
-}
 
-export interface ResolvedConfigOptions extends ConfigOptions {
```
robintown, line 24: I intended to use the `ConfigOptions` vs `ResolvedConfigOptions` distinction to keep ahead of strict mode errors. Because we always merge the default config with the downloaded config, the result of this merge had a more specific type than just `ConfigOptions`, which can help with null safety around the optional parts of the config - could we please not remove this?

src/config/ConfigOptions.ts
```diff
@@ -19,21 +19,24 @@ export interface ConfigOptions {
...
-  sentry: { DSN: "", environment: "production" },
-  rageshake: {
-    submit_url: "https://element.io/bugreports/submit",
+export const DEFAULT_CONFIG: ConfigOptions = {
+  default_server_config: {
+    ["m.homeserver"]: {
+      // These are probably poor guesses - we may want to just not work without
+      // a config file.
+      base_url: `${window.location.protocol}//${window.location.host}`,
+      server_name: window.location.host,
```
robintown, line 39: I would suggest default to `http://localhost:8008`/`localhost`, since that's what the README in https://github.com/vector-im/element-call/pull/799 claims it uses anyways

src/config/Config.ts
```diff
@@ -41,7 +37,26 @@ export class Config {
...
 
-  public config: ResolvedConfigOptions;
+  // Convenience accessors
+  public static defaultHomeserverUrl(): string | undefined {
+    const defaultServerConfig = Config.instance.config.default_server_config;
+    if (!defaultServerConfig) {
+      return undefined;
+    }
+
+    return defaultServerConfig["m.homeserver"]?.base_url;
```
robintown, line 47: The types say that `default_server_config` and `m.homeserver` are always defined - however `Config.instance.config` is the thing that actually might not be

robintown (commented):
Mostly strict mode concerns, we should turn on strict mode for real soon

src/config/ConfigOptions.ts
dbkr: Ah, I thought it was probably something like this, although it was only shallow merging so it didn't seem like it was doing anything.

src/auth/LoginPage.tsx
```diff
@@ -76,11 +76,9 @@ export const LoginPage: FC = () => {
...
 
   const homeserverHost = useMemo(() => {
-    try {
-      return new URL(homeserver).host;
-    } catch (error) {
-      return defaultHomeserverHost;
-    }
+    // XXX: This isn't really correct: the server name of an HS may not
+    // be the same as the hostname of the client API endpoint.
+    return new URL(homeserver).host;
```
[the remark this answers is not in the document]
dbkr: I guess this was written to allow changing the HS in future, but yeah, probably better to just keep it simple for now.

src/config/ConfigOptions.ts
```diff
@@ -19,21 +19,24 @@ export interface ConfigOptions {
...
+      base_url: string;
+      server_name: string;
+    };
   };
 }
 
-export const DEFAULT_CONFIG: ResolvedConfigOptions = {
-  sentry: { DSN: "", environment: "production" },
-  rageshake: {
-    submit_url: "https://element.io/bugreports/submit",
```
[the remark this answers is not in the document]
dbkr: Ah yep, thanks. I've just made rageshake throw an exception for now - I'll follow up with a PR to actually remove the option.

src/config/ConfigOptions.ts
dbkr: Actually - doesn't it work if we just make the types non-nullable in ConfigOptions itself? I think the only reason we had to have them nullable before was because we returned an empty config object if there was no config, where we should have been returning the default.

robintown: I personally think it still makes sense to keep the version of `ConfigOptions` with optional values, to type `downloadConfig` accurately which can return partial configs, and as documentation for what shape a config file is expected to have (ConfigOptions.ts is linked from the README), but none of that's critical.

src/config/ConfigOptions.ts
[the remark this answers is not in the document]
robintown: sounds good

src/config/Config.ts
```diff
@@ -39,25 +41,25 @@ export class Config {
 
   // Convenience accessors
   public static defaultHomeserverUrl(): string | undefined {
-    const defaultServerConfig = Config.instance.config.default_server_config;
+    const defaultServerConfig = Config.get().default_server_config;
     if (!defaultServerConfig) {
       return undefined;
     }
```
robintown, line 47: I still don't think this is necessary

dbkr: Oh yeah, I after all that I forgot to remove this bit

src/config/ConfigOptions.ts
dbkr: Fair enough - have added it back with a comment.

robintown (approved):

PR #799, opened 2022-12-20
Make sample config just have the HS config

dbkr:
The other options are things people usually won't need or want to set, so just keep it to the basics.

 * Rename config sample to match element-web
 * Update / simplify build instructions
 * Rename nginx config file to differentiate it from EC config


SimonBrandner (approved):

robintown (approved):
Looks good given https://github.com/vector-im/element-call/pull/798

PR #810, opened 2022-12-22
Upload source maps to sentry

dbkr:
Uses the vite sentry plugin which should also embed the release (in theory).

Also builds separately in the publish stage, as uploading source maps as part building the docker image is weird. This also means we can publish a tarball of just the built app.

Also fix the docker image again because I forgot to specify the destination file for the config file.

Secrets are added and should be ready to go.


SimonBrandner (approved):
LGTM

PR #816, opened 2023-01-03
send posthog events instantly for callEnded in embedded mode

toger5:
In embedded mode the iFrame will be killed right after the call ended. This prohibits the event from beeing sent. With the send_instantly flag we skip the posthog batching queue to send the event before the iFrame closes.


dbkr (approved):
lgtm, although the justification you give in the PR description is great and could go in the comment on line 187 too maybe?

dbkr (approved):

PR #818, opened 2023-01-03
Fix copyright headers

dbkr:
This is an Element project (in the vector-im repo) so the copyright should be for New Vector: it was incorrectly attributed to the foundation for some files (and some files were missing headers).


SimonBrandner (approved):
Spot check looks fine

PR #825, opened 2023-01-04
decrease wait after call ended to 10ms

toger5:


robintown (approved):
Oh, nice, 10ms makes this quite negligible

robintown (approved):
https://github.com/matrix-org/matrix-react-sdk/pull/9862 will make this hangup sequence work reliably

PR #833, opened 2023-01-07
Hide local volume controls for tiles with no audio

SimonBrandner:


src/video-grid/useCallFeed.ts
```diff
@@ -38,6 +39,7 @@ function getCallFeedState(callFeed: CallFeed | undefined): CallFeedState {
     videoMuted: callFeed ? callFeed.isVideoMuted() : true,
     audioMuted: callFeed ? callFeed.isAudioMuted() : true,
     localVolume: callFeed ? callFeed.getLocalVolume() : 0,
+    hasAudio: callFeed ? callFeed.stream.getAudioTracks().length >= 1 : false,
```
robintown, line 42: Might this need to react to tracks being added or removed?

SimonBrandner: The `track` event should cause a `new_stream` event to be emitted by `CallFeed` which should call `onUpdateCallFeed()` which then updates the `hasAudio` value

PR #844, opened 2023-01-12
Fix caching headers on Docker image

dbkr:
Disable any caching other than etag based for everything other than the assets which have hash-based filenames.

Fixes https://github.com/vector-im/element-call/issues/840 (ie. this will avoid people being on different versions, unless they are super unlucky and different participants actually loaded the page before & after a deploy).

### DIFF
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -2,9 +2,24 @@ server {
     listen       8080;
     server_name  localhost;
 
+    root   /app;
+
     location / {
-        root   /app;
+        # disable cache entriely by default (apart from Etag which is accurate enough)
+        add_header Cache-Control 'private no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0';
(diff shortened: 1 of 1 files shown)


SimonBrandner (approved):
Looks like a config change

config/nginx.conf
```diff
@@ -2,9 +2,24 @@ server {
     listen       8080;
```
MadLittleMods, line 2: > this will avoid people being on different versions, unless they are super unlucky and different participants actually loaded the page before & after a deploy

This seems inevitable. If different versions aren't compatible, we should have a way to warn that you joined a room with a host with a newer/older version which may cause issues. Maybe a little warning icon in the corner with the tooltip explaining the version mismatch of any participants.

Or perhaps a warning if Element Call sees signaling in the Matrix room that it doesn't understand. I don't know what kind of incompatibility there was between the versions in question.

PR #853, opened 2023-01-16
Change rageshake to save much more regularly

dbkr:
First commit does a bit of cleanup & refactoring. Second commit makes the actual change - more detail in commit messages. Also renames the logger member so it doesn't collide with the logger import.

### DIFF
--- a/src/settings/rageshake.ts
+++ b/src/settings/rageshake.ts
@@ -48,7 +48,7 @@ const MAX_LOG_SIZE = 1024 * 1024 * 5; // 5 MB
 // Shortest amount of time between flushes. We are just appending to an
 // IndexedDB table so we don't expect flushing to be that expensive, but
 // we can batch the writes a little.
-const MAX_FLUSH_INTERVAL = 2 * 1000;
+const MAX_FLUSH_INTERVAL_MS = 2 * 1000;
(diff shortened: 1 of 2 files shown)


src/settings/rageshake.ts
```diff
@@ -54,15 +37,23 @@ limitations under the License.
...
-// the frequency with which we flush to indexeddb
-const FLUSH_RATE_MS = 30 * 1000;
-
 // the length of log data we keep in indexeddb (and include in the reports)
 const MAX_LOG_SIZE = 1024 * 1024 * 5; // 5 MB
 
+// Shortest amount of time between flushes. We are just appending to an
+// IndexedDB table so we don't expect flushing to be that expensive, but
+// we can batch the writes a little.
+const MAX_FLUSH_INTERVAL = 2 * 1000;
```
SimonBrandner, line 51:
```suggestion
const MAX_FLUSH_INTERVAL = 2 * 1000; // ms
```

I assume

SimonBrandner (approved):

src/settings/rageshake.ts
dbkr: I've put it in the name of the constant instead

PR #860, opened 2023-01-19
Lowercase room alias before joining

dbkr:
As per comment. This was breaking PTT embedding because it generated a mixed case room alias.

Also add a comment to an undocumented and very strange function.

### DIFF
--- a/src/matrix-utils.ts
+++ b/src/matrix-utils.ts
@@ -212,7 +212,7 @@ export function fullAliasFromRoomName(
 }
 
 /**
- * XXX What is this trying to do? It looks like it's getting the localpart from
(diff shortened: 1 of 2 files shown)


src/matrix-utils.ts
```diff
@@ -211,6 +211,12 @@ export function fullAliasFromRoomName(
   return `#${roomAliasLocalpartFromRoomName(roomName)}:${client.getDomain()}`;
 }
 
+/**
+ * XXX What is this trying to do? It looks like it's getting the localpart from
```
SimonBrandner, line 215:
```suggestion
 * XXX: What is this trying to do? It looks like it's getting the localpart from
```

SimonBrandner (approved):

PR #879, opened 2023-01-25
Grid layout supporting larger conferences

dbkr:
Based on https://github.com/vector-im/element-call/pull/804

This merges @robintown's changes from https://github.com/robintown/element-call/tree/big-grid into https://github.com/vector-im/element-call/pull/805 and adds simulcast functionality to it


fkwp:
now redundant since it has been already merged

PR #882, opened 2023-01-26
Be more clear about homeserver requirements in README.md

dinosmm:
Updated README.md to include more detailed information about the limitations of Element Call and a recommended homeserver set up.

The reason for this update is that as a random person trying to self-host Element Call, I struggled to figure out how it works in relation to the HS it uses, and took some time to figure out it does not work unless open registration is enabled (which is generally bad practice, so I also eventually noticed the public instance at call.element.io uses a separate HS which is not federated).

I included my findings in the README so others trying to self-host have a clearer picture of the app's inner workings and particularities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,12 @@ server {
 }
 ```
 
+By default, the app expects you to have [Synapse](https://matrix-org.github.io/synapse/latest/setup/installation.html) installed locally and running on port 8008. If you wish to use another homeserver, you can add a config file as above.
+
+Element Call requires a Synapse homeserver with registration enabled without any 3pid or token requirements, if you want it to be used by unregistered users. Furthermore, it cannot be used to log in to an existing homeserver where user accounts have joined normal rooms, as it cannot handle those yet.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,6 +42,12 @@ server {
 }
 ```
 
+By default, the app expects you to have [Synapse](https://matrix-org.github.io/synapse/latest/setup/installation.html) installed locally and running on port 8008. If you wish to use another homeserver, you can add a config file as above.
+
+Element Call requires a Synapse homeserver with registration enabled without any 3pid or token requirements, if you want it to be used by unregistered users. Furthermore, it cannot be used to log in to an existing homeserver where user accounts have joined normal rooms, as it cannot handle those yet.
+
+Therefore, to use a self-hosted Synapse homeserver, this is recommended to be a new server where any user account created has not joined any normal rooms anywhere in the Matrix federated network. The HS used can be setup to disable federation, so as to prevent spam registrations (if you keep registrations open) and to ensure Element Call continues to work in case any user decides to log in to their Element Call account using the standard Element app and joins normal rooms that Element Call cannot handle.
```
SimonBrandner, line 49: I don't think we require Synapse specifically, could Synapse only be used as an example?

SimonBrandner (commented):
Thank you!

README.md
dinosmm: Sure!
Sentence 45 there is verbatim from the existing README. I have now edited it to be more clear that other homeservers can be used.

SimonBrandner (approved):
Awesome!

PR #885, opened 2023-01-27
e2e: add end-to-end test workflow

EnricoSchw:
- The tests are executed in the Docker container. This is not perfect.Because we start an HttpServer it is the most stable and easiest way.
- Depends on: https://github.com/vector-im/static-call-participant/pull/12
- Before we merge, the TODOs have to be done and removed from source!
- An image is built in the workflow and that's an overhead. I will remove that in a follow-up PR. But it must be clear which image we can use here.


EnricoSchw:
The e2e tests failing because the private repository with test framework cannot be checked out. I need a GitHub Access Token. @dbkr Do you know who can create this for me?

SimonBrandner (approved):
Looks like a GH workflow

PR #889, opened 2023-01-30
Add call feed size debug info

SimonBrandner:


src/video-grid/VideoTile.module.css
```diff
@@ -96,7 +96,7 @@ limitations under the License.
 
   display: flex;
   align-items: center;
-  justify-content: flex-end;
+  justify-content: space-between;
```
robintown, line 99: This won't affect the layout of the toolbar when it holds multiple buttons, will it?

src/video-grid/VideoTileContainer.tsx
```diff
@@ -66,6 +66,7 @@ export function VideoTileContainer({
     speaking,
     stream,
     purpose,
+    debugInfo,
```
robintown, line 69: This should be added to the props interface

SimonBrandner: Props interface? This is coming from a hook

src/video-grid/VideoTile.module.css
SimonBrandner: No, it should not since they are wrapped in a container of their own

fkwp:
Whenever, a tile is resized and requests a new simulcast layer it's shifted around on the screen. wdyt? Would it be helpful to have the debug overlay visible all the time (if enabled), such that it is more easy to follow?

SimonBrandner:
> Whenever, a tile is resized and requests a new simulcast layer it's shifted around on the screen. wdyt? Would it be helpful to have the debug overlay visible all the time (if enabled), such that it is more easy to follow?

Do I understand correctly, that you suggest **always** showing the debug info instead of showing it on hover?

fkwp:
> > Whenever, a tile is resized and requests a new simulcast layer it's shifted around on the screen. wdyt? Would it be helpful to have the debug overlay visible all the time (if enabled), such that it is more easy to follow?
> 
> Do I understand correctly, that you suggest **always** showing the debug info instead of showing it on hover?

yes (if debug is enabled). I could image that going forward we will also add other stuff to the debug-view

robintown (approved):

src/video-grid/VideoTileContainer.tsx
SimonBrandner: No worries

PR #899, opened 2023-02-03
Try to eliminate empty rows

dbkr:


robintown:
Hmm, this doesn't seem to be working for me most of the time:

Screenshot 2023-02-04 at 00-51-42 Element Call test

This also appears to have regressed tile enlargement, such that you can now get the grid into states where clicking on a tile causes it to shuffle around with other tiles in an unpredictable way, and sometimes placing tiles on top of one another. I also managed to cause a crash:

```
00:47:16.039 Uncaught TypeError: areas[row(...)] is undefined
    slotGridStyle NewVideoGrid.tsx:586
    forEachCellInArea NewVideoGrid.tsx:159
    slotGridStyle NewVideoGrid.tsx:586
    React 3
    NewVideoGrid NewVideoGrid.tsx:568
    React 13
NewVideoGrid.tsx:586:17
```

dbkr:
Hmm - fun. Given how much of a hack this ended up being, I think the best bet is to close it & start again — we can always take inspiration from it.

PR #909, opened 2023-02-13
Extend developer view with media debug information

EnricoSchw:
This PR contains a video call interceptor. With this debug view we can display the Track IDs and Stream IDs depending on the users of a call. This makes it easier to determine changes in the median track flow.
We generally extend our Debug View and other PRs will follow. For example In the next PR we will assign some  RTP transceiver properties to the displayed tracks.


EnricoSchw:
The Linter fails because an update of `matrix-js-sdk` was done before the PR.
I will fix the lint issue in the `matrix-js-sdk`. 

:-)

public/locales/de/app.json
```diff
@@ -83,6 +83,8 @@
   "Sending…": "Senden …",
   "Settings": "Einstellungen",
   "Share screen": "Bildschirm teilen",
+  "Show call feed debug info": "Call-Feed-Inspector anzeigen",
+  "Show VoIP debug info": "VoIP-Inspektor anzeigen",
```
dbkr, line 87: These files should only be edited by Weblate otherwise it causes conflicts

src/inspectors/VoIPInspector.tsx
```diff
@@ -0,0 +1,132 @@
...
+import styles from "./VoIPInspector.module.css";
+
+interface VoIPViewerProps {
+  client: MatrixClient;
+  groupCall: GroupCall;
+  userMediaFeeds: CallFeed[];
+  screenshareFeeds: CallFeed[];
+}
+
+export function VoIPViewer({
```
dbkr, line 31: Maybe 'Media' would be a better name for this tab? The whole app is arguably VoIP.

src/inspectors/VoIPInspector.tsx
```diff
@@ -0,0 +1,132 @@
...
+  client,
+  groupCall,
+  userMediaFeeds,
+  screenshareFeeds,
+}: VoIPViewerProps) {
+  return (
+    <div className={styles.scrollContainer}>
+      <div className={styles.voIPInspectorViewer}>
+        <Table name="Media Feeds" feeds={userMediaFeeds} />
```
dbkr, line 40: Don't forget to add translation wrappers around these

src/inspectors/VoIPInspector.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+function Table({ name, feeds }: TableProp): JSX.Element {
+  // Catch case if feeds is empty
+  if (feeds.length === 0) {
+    return (
+      <div className={styles.section}>
+        <p className={styles.sectionTitle}>{name}</p>
+
+        <div className={styles.centerMessage}>
+          <p>No Feeds..</p>
```
dbkr, line 62: Translation wrapper again (also should be the ellipsis character (…).

src/inspectors/VoIPInspector.tsx
```diff
@@ -0,0 +1,132 @@
...
+      </header>
+      {feeds.map((feed, i) => {
+        const user = feed.isLocal()
+          ? "local"
+          : feed.getMember() !== null
+          ? feed.getMember()?.name
+          : feed.userId;
+        return (
+          <TableRow
+            key={i}
```
dbkr, line 86: key should be something persistent to the thing that the row represents, so probably the feed ID. Similarly for the index prop passed into the Component for the same thing.

src/room/GroupCallInspector.tsx
```diff
@@ -460,12 +466,19 @@ export function GroupCallInspector({
       className={styles.inspector}
     >
       <div className={styles.toolbar}>
-        <button onClick={() => setCurrentTab("sequence-diagrams")}>
-          Sequence Diagrams
-        </button>
-        <button onClick={() => setCurrentTab("inspector")}>Inspector</button>
+        {showCallFeedDebugInfo === true && (
```
dbkr, line 469: This is what I was talking about in standup, btw — call feed debug info isn't related to the sequence diagrams. I don't think there's any reason to have options to show/hide tabs in the debug inspector. I would just show all of these tabs unconditionally.

dbkr:
Oh, also don't forget to give the PR a changelog-friendly name.

public/locales/de/app.json
EnricoSchw: Oh ok.  I will remove this lines

src/inspectors/VoIPInspector.tsx
EnricoSchw: Renamed it to Media

dbkr (approved):

PR #911, opened 2023-02-14
New grid system

robintown:
This adds a completely rewritten grid component that lays out tiles in a big interactive grid intended for calls of more than 12 people. Eventually, once it reaches feature parity, it will replace the current grid system. For now, it can be enabled for testing with a check box found in developer settings.

This is the first step towards https://github.com/vector-im/element-call/issues/838.


src/useReactiveState.ts
```diff
@@ -0,0 +1,46 @@
+import {
+  DependencyList,
+  Dispatch,
+  SetStateAction,
+  useCallback,
+  useRef,
+  useState,
+} from "react";
+
+export const useReactiveState = <T>(
```
robintown, line 10: Needs documentation

src/useReactiveState.ts
```diff
@@ -0,0 +1,46 @@
+import {
```
robintown, line 1: This and a couple other files are missing copyright headers

SimonBrandner (approved):
Looks good!!!

PR #912, opened 2023-02-14
Typescript update to 4.9.0

EnricoSchw:
This PR fixes the lint issue we have because the `matrix-jas-sdk` uses a higher typescript version.

Unfortunately I had to set `mermaid` to a pre release version. The typescript bug they have is only fixed in the develop branch: https://github.com/mermaid-js/mermaid/commit/9f2d29b68b00d9719add89b4f23938ff2db45077
https://github.com/mermaid-js/mermaid/pull/3914

We using `mermaid` in Inspector  for Sequenz Diagrams. I think, there is possible to use a pre version.


SimonBrandner (approved):
Looks like a TS update

PR #924, opened 2023-02-21
Allow Element Call to be started without audio / video interface

EnricoSchw:
Allow Element Call to be started without audio / video interface for the sake of a screen-sharing only session#902

Is blocked by: https://github.com/matrix-org/matrix-js-sdk/pull/3162

### DIFF
--- a/src/room/GroupCallView.tsx
+++ b/src/room/GroupCallView.tsx
@@ -81,6 +81,7 @@ export function GroupCallView({
     screenshareFeeds,
     participants,
     unencryptedEventsFromUsers,
+    initCallWithoutVideoAndAudio,
   } = useGroupCall(groupCall);
 
   const { t } = useTranslation();
@@ -300,6 +301,7 @@ export function GroupCallView({
         roomIdOrAlias={roomIdOrAlias}
         isEmbedded={isEmbedded}
         hideHeader={hideHeader}
+        isOnlyScreenShare={initCallWithoutVideoAndAudio}
       />
     );
   }
--- a/src/room/GroupCallView.tsx
+++ b/src/room/GroupCallView.tsx
@@ -263,6 +263,7 @@ export function GroupCallView({
           roomIdOrAlias={roomIdOrAlias}
           unencryptedEventsFromUsers={unencryptedEventsFromUsers}
           hideHeader={hideHeader}
+          isOnlyScreenshare={initCallWithoutVideoAndAudio}
         />
       );
     }
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -91,6 +91,7 @@ interface Props {
   roomIdOrAlias: string;
   unencryptedEventsFromUsers: Set<string>;
   hideHeader: boolean;
+  isOnlyScreenshare: boolean;
 }
 
 export function InCallView({
@@ -113,6 +114,7 @@ export function InCallView({
   roomIdOrAlias,
   unencryptedEventsFromUsers,
   hideHeader,
+  isOnlyScreenshare,
 }: Props) {
   const { t } = useTranslation();
   usePreventScroll();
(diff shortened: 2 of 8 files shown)


src/room/InCallView.tsx
```diff
@@ -372,26 +374,27 @@ export function InCallView({
...
-        <HangupButton onPress={onLeave} />
-      </div>
-    );
   } else {
-    footer = (
-      <div className={styles.footer}>
-        <MicButton muted={microphoneMuted} onPress={toggleMicrophoneMuted} />
+    const buttons: JSX.Element[] = [];
+
+    if (!isOnlyScreenshare) {
```
SimonBrandner, line 380: Didn't we still want to show the mic and video button but for them just to be disabled?

EnricoSchw: The LocalFeed has no tracks. Unmuting and muting would results in an error. I'm catching this error in the SDK, but we decided to disable the buttons anyway. Actually, you would have to check whether an audio or video device is available. If this device is available we can display the buttons. But this is just a work around feature and I don't want ed change too much.

dbkr: The ideal solution from my PoV would be to display the buttons and clicking them causes it to try open the corresponding device again (I think this matches what other platforms do). This seems inconsistent since we don't hide the video button if you lack a webcam. We can ask a designer though.

src/room/GroupCallView.tsx
```diff
@@ -262,6 +263,7 @@ export function GroupCallView({
           roomIdOrAlias={roomIdOrAlias}
           unencryptedEventsFromUsers={unencryptedEventsFromUsers}
           hideHeader={hideHeader}
+          isOnlyScreenshare={initCallWithoutVideoAndAudio}
```
dbkr, line 266: I don't think these are quite the same thing? `initCallWithoutVideoAndAudio` controls whether a call without no audio or video is permitted, but `isOnlyScreenshare` implies we definitely don't have audio or video.

src/room/InCallView.tsx
EnricoSchw: I tried to find a solution thats fits with the current behavior and software design. Currently this feature not change or touch old source code ore behaviour. 

I totally agree, this would be the best solution. But for this solution I need more time.

src/room/GroupCallView.tsx
EnricoSchw: `initCallWithoutVideoAndAudio` is a matrix js sdk property describe the behavior of the sdk
`isOnlyScreenshare` is a element call property describe the feature we implementing.

@dbkr What is your suggestion how we can rename one of the property?

src/room/InCallView.tsx
SimonBrandner: Makes sense, my suggestion here would be to open an issue to fix this in the future, perhaps when we end up doing the refactoring

src/room/GroupCallView.tsx
dbkr: I think both of these might be the same problem: the `isOnlyScreenshare` and the fact that the buttons are hidden entirely implies that EC won't try to open audio & video devices at all, but the js-sdk will still try to open devices with the flag enabled, the difference is that it will continue if it fails (as I understand it).

I think we probably should choose either:
 * Allowing calls to proceed when they have no audio or video devices, continue to show the buttons (they could just do nothing if no corresponding feed for now rather than retrying `getUserMedia()`): this way the `isOnlyScreenshare` flag is no longer needed since we don't need to change the UI.
 * Have a special mode where calls are *only* screenshares and never try to open audio or video devices and we hide the mute buttons entirely. I feel like this would need a nontrivial amount of UI design though, since you would want to make it a different thing in the UI.

I have a fairly strong preference for the former.

EnricoSchw: @dbkr I will implement this like your first suggestion. But I need this property either, because.. 
In the case the property is true and no track is in the feed i disable the buttons in a call. Even i have to hide the status message `User is connecting...` This view status is covert by the feeds and is not disappearing if the feed has no tracks.

EnricoSchw: I will rename the property in `allowCallWithoutVideoAndAudio`

EnricoSchw: @dbkr: I reduced the `matrix-js-sdk` dependencies to a minimum. Except for one point, the EC works independently of the feature, now.

dbkr (approved):

SimonBrandner (approved):

README.md
```diff
@@ -30,6 +30,20 @@ cp config/config.sample.json public/config.json
...
+```json
+{
+  ...
+
+  "features": {
+    "feature_group_calls_without_video_and_audio": true
+  }
+}
+```
+
```
SimonBrandner, line 46: Could this be moved to the end of the `Host it yourself` section? Also it would be good to have this under a `Features` section or similar

README.md
```diff
@@ -62,6 +48,22 @@ Element Call requires a homeserver with registration enabled without any 3pid or
 
 Therefore, to use a self-hosted homeserver, this is recommended to be a new server where any user account created has not joined any normal rooms anywhere in the Matrix federated network. The homeserver used can be setup to disable federation, so as to prevent spam registrations (if you keep registrations open) and to ensure Element Call continues to work in case any user decides to log in to their Element Call account using the standard Element app and joins normal rooms that Element Call cannot handle.
 
+### Features
+
+#### Allow to join group calls without camera and audio device
+
+You can allow to join a group call without video and audio by setup this feature in your `config.json`:
```
SimonBrandner, line 55:
```suggestion
#### Allow joining group calls without a camera and a microphone

You can allow joining a group call without video and audio enabling this feature in your `config.json`:
```

SimonBrandner (approved):
Tiny nitpick

PR #932, opened 2023-02-28
Behave sensibly if a full room alias is entered

dbkr:
Check explicitly to see if the room name that's enetered into the box looks like a room alias and if so, do the sensible thing.

Fixes https://github.com/vector-im/element-call/issues/852

### DIFF
--- a/src/matrix-utils.ts
+++ b/src/matrix-utils.ts
@@ -212,6 +212,31 @@ export function fullAliasFromRoomName(
 }
 
 /**
+ * Applies some basic sanitisation to an room name that the user
+ * has given us
+ * @param input The room name from the user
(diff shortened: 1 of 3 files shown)


src/matrix-utils.ts
```diff
@@ -211,6 +211,31 @@ export function fullAliasFromRoomName(
   return `#${roomAliasLocalpartFromRoomName(roomName)}:${client.getDomain()}`;
 }
 
+/**
+ * Applies some basic sanitisation to an room name that the user
```
SimonBrandner, line 215:
```suggestion
 * Applies some basic sanitisation to a room name that the user
```

SimonBrandner (approved):
Looks sane

PR #935, opened 2023-03-01
Add e2eEnabled parameter to Widget client

alariej:
Allow to-device signaling in matryoshka mode to be unencrypted, like in the standalone client. This is useful especially for testing and also for clients which do not yet support encryption, like Quadrix. The default value for the optional `e2eEnabled` URL parameter is "true", so this should not have any impact on existing implementations using the matrix-widget-api, like Element Web.

I have tested this commit in a test build of Quadrix, at: https://github.com/alariej/quadrix/blob/7f134a3646932e3d4d92538ade1e6a63cc94fd5c/src/modules/ElementCall/index.tsx#L314

### DIFF
--- a/src/widget.ts
+++ b/src/widget.ts
@@ -101,7 +101,7 @@ export const widget: WidgetHelpers | null = (() => {
       // We need to do this now rather than later because it has capabilities to
       // request, and is responsible for starting the transport (should it be?)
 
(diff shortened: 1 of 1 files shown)


robintown (approved):
Seems sensible to me, thanks

PR #943, opened 2023-03-03
Add proposal for mid based signalling

dbkr:


doc/mid-based-signalling.md
```diff
@@ -0,0 +1,252 @@
...
+        "call_id": "35657a5b793ce",
+        "conf_id": "bbe53499f82e3",
+        "invitee": "@bob:example.org",
+        "lifetime": 60000,
+        "party_id": "123456",
+        "description": {
+            "type": "offer",
+            "sdp": "[...]",
+        }
+    },
```
SimonBrandner, line 37: Perhaps this should also be split up into several pieces?

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,252 @@
...
+    "m.call.capabilities": {
+        "m.call.transferee": false,
+        "m.call.dtmf": false,
+    },
+    "m.tracks.describe": {
```
SimonBrandner, line 42: `m.tracks`? Shouldn't it be `m.call`?

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,252 @@
...
+        "1": { // transceiver mid 1
+            "media_uuid": "aaaa-aaaa-aaaaaa-aaaa-aaaa",
+            "media_group_uuid": "1234-1234-123456-1234-1234", // rather than 'track group ID' to match media UUID?
```
SimonBrandner, line 45: Makes sense this way :heavy_check_mark:

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,252 @@
...
+us an idea of whether it could work or not.**
+
+m.call.invite:
+```
+{
+    "m.negotiate": {
+        "version": 1,
+        "call_id": "35657a5b793ce",
+        "conf_id": "bbe53499f82e3",
+        "invitee": "@bob:example.org",
```
SimonBrandner, line 30: What's the point of this field? Isn't it clear who the recipient is if this is to-device

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,252 @@
...
+                },
+            },
+        }
+    },
+}
+```
+
+XXX: How flat vs deep do we want the structure to be here? I've done it quite deep here,
+organised by user ID / device ID / media group UUID, but they could also just be a flat
+list of tracks. It would be more duplication but maybe less effort to read.
```
SimonBrandner, line 149: It would also be useful to think about how would we actually implement this and how it fits in with the current implementation

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,252 @@
...
+            "height": 576,
+        },
+        "bbbb-bbbb-bbbbbb-bbbb-bbbb": {},
+    },
+},
+```
+
+This has also been rearrnaged a little to make the media UUIDs the keys and remove the
+unsubscribe section which is unnecessary if we always send the complete set of tracks we
+want to receive (we unsubscribe by just removing the media UUID from the dict).
```
SimonBrandner, line 184: I am pretty sure we actually wanted to avoid this to keep the messages short. Perhaps we could use `null` to say we want to unsubscribe. We could use a similar technique for unpublishing

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,252 @@
...
+
+This also now contains a sequence number, so the focus can reply with a an ack:
```
SimonBrandner, line 186: Do all events now have a sequence number? What does the number mean?

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,252 @@
...
+This outlines how a switch to mid + media UUID based signalling would work in more detail.
+
+**Note: I've made this mixin about describing the tracks that are being sent on the transceivers
+and the keys are the mids rather than the media UUIDs. Hoping that sketching this out will give
+us an idea of whether it could work or not.**
+
+m.call.invite:
+```
+{
+    "m.negotiate": {
```
SimonBrandner, line 26: Should this be `m.call. negotiate to avoid collisions with other types of negotiations (whatever they may be)?

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,252 @@
...
+        "description": {
+            "type": "answer",
+            "sdp": "[...]",
+        }
+    },
+    "m.call.capabilities": {
+        "m.call.transferee": false,
+        "m.call.dtmf": false,
+    },
+    "m.tracks.advertise": {
```
SimonBrandner, line 129: `m.call`?

SimonBrandner (commented):
This really does seem to look great apart of some nits. The only thing I think we should think deeper about is the depth :D

doc/mid-based-signalling.md
fkwp: Add feedback to track subscriptions with seqnums so every response can be matched to a request

doc/mid-based-signalling.md
dbkr: Yeah, it's really just consistency with 1:1 calls, although if we're essentially deprecating them then maybe we shouldn't worry about it.

doc/mid-based-signalling.md
dbkr: As in split the common call stuff out from the actual negotiation? Interesting idea.

doc/mid-based-signalling.md
dbkr: Yes, maybe I am barking up the wrong tree with this and it should just be the deltas each time, especially if we have sequence numbers to ensure we don't lose events.

doc/mid-based-signalling.md
dbkr: Yep. I've added a bit more detail in the doc.

doc/mid-based-signalling.md
SimonBrandner: Well `m.call.candidates` would still include some of this, so it would make sense to have that in a common mixin

doc/mid-based-signalling.md
dbkr: I've expanded on this a bit in 457578a, was this the sort of thing you were thinking about?

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+m.call.subscribe
+```
+"m.call.subscribe": {
+    "seq": 1,
+    "media_uuids": {
+        "aaaa-aaaa-aaaaaa-aaaa-aaaa": {
+            "width": 1024,
+            "height": 576,
+        },
+        "bbbb-bbbb-bbbbbb-bbbb-bbbb": {},
```
SimonBrandner, line 199: For pull-based media in the future, this is going to need to include the full path (e.g. `userId/deviceId/mediaGroupUUID/mediaUUID`)

dbkr: Wait, why would this be different?

SimonBrandner: So that the focus knows from where to request the track in case it can't read state

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
+# MID Based Signalling
```
SimonBrandner, line 1: This might be nice opportunity to fix conf vs call id somehow

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+        "description": {
+            "type": "answer",
+            "sdp": "[...]",
+        }
+    },
+    "m.call.capabilities": {
+        "m.call.transferee": false,
+        "m.call.dtmf": false,
+    },
+    "m.call.describe": {
```
SimonBrandner, line 91:
```suggestion
    "m.call.describe_media": {
```

Perhaps to be clearer?

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+        "description": {
+            "type": "answer",
+            "sdp": "[...]",
+        }
+    },
+    "m.call.capabilities": {
+        "m.call.transferee": false,
+        "m.call.dtmf": false,
+    },
+    "m.call.advertise": {
```
SimonBrandner, line 129:
```suggestion
    "m.call.advertise_media": {
```

Perhaps this to be clearer

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+    "selected_party_id": "678910",
+}
+```
+
+Once the client receives this, it decides what tracks it wants to receive and then sends
+a subscribe message over the data channel:
+
+m.call.subscribe
+```
+"m.call.subscribe": {
```
SimonBrandner, line 192:
```suggestion
"m.call.subscribe_to_media": {
```

Perhaps this for clarity

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+                    "purpose": "m.usermedia",
+                    "kind": "video",
+                }, {
+                    "media_uuid": "bbbb-bbbb-bbbbbb-bbbb-bbbb":
+                    "purpose": "m.usermedia",
+                    "kind": "audio",
+                },
+            },
+        }
+    },
```
SimonBrandner, line 143: From the js-sdk class structure POV, it would make much more sense for this to be structured the same way metadata is atm (`user_id` and `device_id` being props of media group uuid) since the individual `CallFeed`s (media groups) have the `userId`s and `deviceId`s as props, so the mapping makes much more sense that way

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+            "media_group_uuid": "1234-1234-123456-1234-1234",
+            "purpose": "m.usermedia",
+            "kind": "audio",
+        },
+    },
+}
+```
+
+A focus, however, will not send any tracks by default and therefore does not include an
+`m.call.describe` bblock. Instead, it includes an `m.track.advertise` block advertising
```
EnricoSchw, line 109: type: bblock -> block

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+        "description": {
+            "type": "offer",
+            "sdp": "[...]",
+        }
+    },
+    "m.call.capabilities": {
+        "m.call.transferee": false,
+        "m.call.dtmf": false,
+    },
+    "m.call.describe": {
```
SimonBrandner, line 42: The question we should ask ourselves is if we'd ever want to have metadata for media groups too which would then mean it would make sense to have this structured in a way similar to the current metadata structure

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+    "errcode": "M_UNKNOWN",
+    "error": "Internal server error",
+}
+```
+
+This may give some indication as to why some tracks were not available (should it have errors per
+media UUID, perhaps?)
+
+If the focus needs to renegotiate to send the tracks, it does so, describing the media UUIDs it intends to send on the
+transceivers once the negotiation is complete:
```
EnricoSchw, line 236: In this case, thats the transceiver id's (mid's) from the Focus side. Right?

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+XXX: How flat vs deep do we want the structure to be here? I've done it quite deep here,
+organised by user ID / device ID / media group UUID, but they could also just be a flat
+list of tracks. It would be more duplication but maybe less effort to read.
+
+The expected behaviour here would be for foci to essentially maintain a structure with all
+tracks being pushed to it. This structure would probably have call IDs as a top level index,
+then look very similar to the structure of the `m.call.advertise` event. It could either keep
+a reference to the transceiver it was receiving media on in the structure itself alongside
+the media UUID, or maintain a separate map of media UUID to transceiver / peer connection such
+that the first structure could be marshalled to JSON and sent to clients as-is.
```
EnricoSchw, line 156: I think the SFU do not need to save and maintain the map of media UUID to transceiver / peer connection. The mid's are only needed to identify the tracks inside the SDP. If the connection is negotiated, they are no longer required.

doc/mid-based-signalling.md
dbkr: Yes, definitely. afaik its just the MSC that is confused somewhere, but note to check.

doc/mid-based-signalling.md
dbkr: Wouldn't we do this in the advertise message though? That's where the bulk of the metadata should be.

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+(invite, answer, negotiate).
+
+The `m.call.describe` block is a 'mixin' block in externsible events terms and describes the media being sent on each
+transceiver by the sending user. It *always* refers only to the media being sent by the device that sends the event.
+
+The same is sent either to a focus or a peer client.
+
+A focus will gather the various streams that it's receiving by the `conf_id` from the `m.negotiate` section. In future
+when stream are advertised via the room state, this could be unnecessary and foci need not have any knowledge of what
+group calls are happening (assuming we don't need them to enforce viewship based on this).
```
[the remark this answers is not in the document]
dbkr: Correct. This is assuming that they would have access to that data through some mechanism.

doc/mid-based-signalling.md
dbkr: So basically remove the user ID & device ID levels and move them to attributes of the media group ID? It would be towards the flatter end of possible structures and maybe a nice compromise between depth & duplication. I'm not sure we should necessarily be basing the choices on what the js-sdk does directly, but if the js-sdk has a god reason for doing it that way then that's certainly valid.

doc/mid-based-signalling.md
dbkr: Well, it need to know what media UUID corresponds to what it's sending in each case: whether it does that by mapping them to mid or track ID or whatever is up to the implementation - is that what you meant?

doc/mid-based-signalling.md
dbkr: Ah right, yes, although in this case the address would also need to include the SFU that the user was publishing to? I wonder if this should look more like:

```
media: [
  {
    "media_uuid": "bbbb-bbbb-bbbbbb-bbbb-bbbb",
    "user_id": "@alice:example.org",
    "device_id": "aaaaaa",
    "sfu_user_id": "@sfu:example.org",
    "sfu_device_id": "sfusfusfu",
  }
]
```

doc/mid-based-signalling.md
EnricoSchw: Yes, exactly, that is what I meant. I only just wanted to pointed out that's MID's quite unstable and only needed to init this process. And we should not based on it in such `whatever implementation`.

doc/mid-based-signalling.md
dbkr: They should be the same, no? Since the media lines must be the same on each side.

doc/mid-based-signalling.md
SimonBrandner: Well we might (or not) have some fast-changing metadata on the media group level, so it would be useful for that case

doc/mid-based-signalling.md
SimonBrandner: Yeah, that's what I mean. I'd say the js-sdk doesn't have a **god** reason for this since I was the one who wrote and I haven't gone mad enough to consider myself a diety (yet) :D But it does have a **good** reason - I think it makes sense in terms of media groups being what you display in the UI (but I wrote it so I am biased)

(sorry, I sometimes can't help myself :sweat_smile: )

doc/mid-based-signalling.md
EnricoSchw: Right, the  media lines are bidirectional and correlate to the remote side. 

What i mean is, whoever initiates the offer determines what is sent over the mid-line. When Focus only receives tracks via a connection, the Focus will never have a reason for start a media renegotiation, because on their side the media will never change. 

In case the Focus start a renegotiation and creates an offer, then the focus sending media as well. In this case the Focus defines, what is sending over the media lines. But this means the client will receive a track from another peer. 

If the client not sending media to the Focus the Focus starts with  0 media line. If the client already sending to the Focus the Focus could use as well the 0 media line. But we decide in this case the Focus will use a new Transceiver. 

Long story short: The media lines alone define not what will sending over it. I know we want use pre defined Transceiver and reserve the 0-media lines and 1-media lines for Client sending medias. However, I have still my doubts that this is a good idea.

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+SDP to work around this).
+
+We also propose switch the structure of MSC3401 call events to be compatible with extensible
+events. Legacy calls would have to remain the same for backwards compatibility. It is probably
+impractical to have any call events support both formats because this would mean the SDP, which
+can be quite large,would have to be duplicated in the JSON. Therefore, there's a significant
+advantage to taking to opportunity to migrate to an extensible event format now.
+
+We use the same event types for call invites and answers, so it is important that these events
+are only send via to device message directly to client known to support them. They must not be
```
daniel-abramov, line 14: (minor typo fixes unless my English knowledge is outdated 🙂)

```suggestion
are only sent via to-device messages directly to clients known to support them. They must not be
```

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+            "kind": "audio",
+        },
+    },
+}
+```
+
+Note that the SDP content is now in a section called `m.negotiate`. This is a mixin block common to all negotiation events
+(invite, answer, negotiate).
+
+The `m.call.describe` block is a 'mixin' block in externsible events terms and describes the media being sent on each
```
daniel-abramov, line 62:
```suggestion
The `m.call.describe` block is a 'mixin' block in extensible events terms and describes the media being sent on each
```

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+media UUID it finds to an `m.call.subscribe` message.
+
+The most complex part is that when the `m.call.describe` message arrives back from the focus,
+it will have to search through the data from the `m.call.advertise` message to map the tracks
+it is now receving to the right user IDs (XXX: it could build them into a map to make this lookup
+efficient, although if we make the advertise message indexed by media UUID then it already has
+a map indexed by the correct thing...)
+
+The `select_answer` is also tweaked to be more extensible-event like although is essentially
+the same:
```
daniel-abramov, line 174: Maybe this is a dumb question, but do we need a `select_answer` still? It seems like after exchanging the SDP offer and SDP answer, we could establish a connection and start streaming.

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+        "aaaa-aaaa-aaaaaa-aaaa-aaaa": {
+            "width": 1024,
+            "height": 576,
+        },
+        "bbbb-bbbb-bbbbbb-bbbb-bbbb": {},
+    },
+},
+```
+
+This has also been rearrnaged a little to make the media UUIDs the keys and remove the
```
daniel-abramov, line 204:
```suggestion
This has also been rearranged a little to make the media UUIDs the keys and remove the
```

doc/mid-based-signalling.md
```diff
@@ -0,0 +1,276 @@
...
+unsubscribe section which is unnecessary if we always send the complete set of tracks we
+want to receive (we unsubscribe by just removing the media UUID from the dict).
+
+This also now contains a sequence number. This is a monotonically increasing integer, starting
+at 0 and scoped to the lifetime of the peer connection. The focus will send a reply containing
+this sequence number to acknowledge that it has processed the message. This can be a positive ack:
```
daniel-abramov, line 210: Nice 🚀 

Btw, do we want to use a sequence number or rather some sort of a randomly generated transaction ID?

I know that the advantage of a sequence number is that it still can be used as a transaction ID and we probably need it since To-Device messages are not ordered. The only problem that I can think of though is that what if one party receives a message with a sequence number `10000` after a message with a sequence number `5`? What's the behavior? Should we drop the `10000` message or should we buffer it and process it once the previous message arrives (this opens up a possible attack vector of sending 1, 2, 3, 10000000 and causing the receiving client to deplete their own buffer)?

Also, what to do when the SFU gets restarted and does not know the last sent `seq`? - I suppose this is solved by "scoped to the lifetime of the peer connection", but I wonder if we need to have `seq` for all events (negotiate events could also fail).

doc/mid-based-signalling.md
dbkr: Understood – hopefully c06d3d4 clarifies this a bit.

doc/mid-based-signalling.md
dbkr: Ah yes, I see. My gut feeling is that the media groups would probably be purely a way to associate two tracks together as a pair (or group) and deliberately not add any further data, but perhaps.

doc/mid-based-signalling.md
dbkr: I did wonder whether to remove it for group calls, but I think I'm inclined to keep it for consistency. The connection doesn't need to be blocked on the select_answer though, we can still start streaming after just the offer & answer.

doc/mid-based-signalling.md
dbkr: I made it `m.call.subscribe_media` as I think it's fine as an abbreviation for consistency.

doc/mid-based-signalling.md
dbkr: True, seqnums allow for a little more flexibility but we wouldn't need the re-ordering capabilities if we're using a channel that guarantees ordering which the WebRTC DC will give us. A randomly generated ID could avoid clients thinking they have to do reordering and they wouldn't have to worry about masking them monotonically increasing.

doc/mid-based-signalling.md
SimonBrandner: True, that does make sense, but perhaps, not sure

doc/mid-based-signalling.md
dbkr: I'm not sure I understand. From my PoV, the three different scenarios work like this:

If Alice (the client) isn't sending any media to the focus, the focus starts sending media to her and starts a renegotiation to add a transceiver, which becomes mid 0. It includes an `m.call.describe_media` saying that the media on mid 0 is Bob's media.

If Alice is already sending media, the focus renegotiates but decides to add a new transceiver. This becomes mid 2, so its `m.call.describe_media` informs the client that mid 2 contains Bob's media. There is no entry for mid 0 or 1 because the focus isn't sending anything on those transceivers, it's only receiving.

If Alice is already sending media and the focus decides to re-use one of those transceivers, its `m.call.describe_media` informs the client that Bob's media will arrive on mid 0. Alice will re-send her own `m.call.describe_media` in her answer re-iterating that her own media is being sent on mid 0 since she's still sending it in the other direction. The focus's `m.call.describe_media` describes what's flowing on the transceivers in the focus -> client direction and the client's `m.call.describe_media` describes what's going client -> focus.

doc/mid-based-signalling.md
dbkr: So something like c62296c?

toger5:
@dbkr Since full mesh is further away into the future now I wonder if we can close this?

dbkr:
Yeah, seems so.

PR #961, opened 2023-03-15
OpenTelemetry

dbkr:
Sends basic telemetry data to OpenTelemetry, and also adds a PostHog exporter so the same data can go to PostHog too.

The Posthog exporter is less complete and still contains console.log statements. I feel like we may as well include it anyway, although we could also remove it for now and add it later.

Nothing will be enabled without a collector URL in the config. Also requires the user to opt in to analytics, although this states that the analytics are anonymous which OpenTelemetry data currently is not. We need to make a separate opt-in for non-anonymous data or something before this can actually be enabled anywhere outside of a dev setup.


config/otel_dev/docker-compose.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - "13133:13133" # health_check extension
+      - "4317:4317" # OTLP gRPC receiver
+      - "4318:4318" # OTLP HTTP receiver
+      - "55670:55679" # zpages extension
+    depends_on:
+      - jaeger-all-in-one
+  nginx:
+    image: nginxinc/nginx-unprivileged:latest
+    volumes:
+      - ./nginx_otel.conf:/etc/nginx/conf.d/default.conf:ro
```
EnricoSchw, line 27: Did you forget to add the configuration file? `./nginx_otel.conf`

dbkr: Oops, yes I did. Fixed.

src/otel/OTelGroupCallMembership.ts
```diff
@@ -0,0 +1,177 @@
...
+    } else if (event.type === "sendEvent") {
+      this.callMembershipSpan.addEvent(
+        `matrix.sendToRoomEvent_${event.eventType}`,
+        flattenVoipEvent(event)
+      );
+    }
+  }
+
+  public onToggleMicrophoneMuted(newValue: boolean) {
+    this.callMembershipSpan.addEvent("matrix.toggleMicMuted", {
```
robintown, line 149: This errors if you try to mute yourself for the lobby screen - I think `callMembershipSpan` should be nullable

EnricoSchw (approved):

PR #971, opened 2023-03-22
End-to-end audio observability

robintown:
This reports via OpenTelemetry when particular participants are speaking, as an easy way to observe the delivery of audio in calls.

Closes https://github.com/vector-im/element-call/issues/959


src/otel/OTelGroupCallMembership.ts
```diff
@@ -110,10 +112,11 @@ export class OTelGroupCallMembership {
   }
 
   public onLeaveCall() {
-    this.callMembershipSpan?.addEvent("matrix.leaveCall");
-
-    // and end the main span to indicate we've left
-    if (this.callMembershipSpan) this.callMembershipSpan.end();
+    this.callMembershipSpan!.addEvent("matrix.leaveCall");
```
MadLittleMods, line 115: `this.callMembershipSpan` could easily be `undefined` if `onLeaveCall` is called twice which seems easy to accidentally do as well. It would be good to trace these extra calls as well.

Perhaps some before-join and after-leave spans are necessary to add the events to.

MadLittleMods: From the [recent update](https://github.com/vector-im/element-call/pull/971/commits/928f1c1d6face57e152b6e959be0d236a934d2f5) to address this, this doesn't feel like enough protection,

https://github.com/vector-im/element-call/blob/928f1c1d6face57e152b6e959be0d236a934d2f5/src/otel/OTelGroupCallMembership.ts#L129-L132

We would still be blind in the tracing if this kind of thing happened.

public/config.json
```diff
@@ -0,0 +1,19 @@
+{
```
dbkr, line 1: Presumably this needs to go too?

robintown: Right…

dbkr (approved):

src/otel/OTelGroupCallMembership.ts
robintown: Sorry for missing this, but I disagree about OpenTelemetry traces being the place to log an API misuse of this kind. Our primary use case for telemetry, rageshakes, already include logs, and it seems a bit roundabout to use OpenTelemetry to detect bugs in the OpenTelemetry code itself. I think it's reasonable to trust that the one place in the code that this is called from gets the calls right.

MadLittleMods: > to detect bugs in the OpenTelemetry code itself. 

This would be an app bug not OpenTelemetry. Am I missing something?

> I think it's reasonable to trust that the one place in the code that this is called from gets the calls right.

This seems like a bad assumption to me. The app could misbehave in an unlimited number of ways. This is the code that is called in order to start and end traces around joining/leaving.

If I were debugging some problem in Element Call, chances are I just check the trace because it gives a really nice overview and I can dive in at any granularity necessary. I would want the trace to be accurate to what actually happened instead of just dropping calls on the floor with a log I could easily miss in the sea of logs.

robintown: Oh, I understand what you are worried about a little better now, I thought you were just referring to ensuring that the OTelGroupCallMembership API is used correctly. I can add an event to the group call span when a double-join is attempted, which I'll do in another PR now, but for double-leaves, I honestly don't feel that adding new spans to encompass the app's entire lifecycle just to have a place to stick such events is worth it. That would seem to completely duplicate the purpose of logs—unless you have something more bespoke in mind, like using spans to model the user's entire journey through the app's pages?

robintown: (still, when adding instrumentation, our time is limited, and we're not trying to literally cover every possible failure mode, just the ones that have actually been elusive in practice)

MadLittleMods: We should probably make sure what's instrumented is accurate at least.

These kind of logic bugs are going to very arduous to spot when we're debugging an actual situation. Otherwise, someone is first going to have to track down this problem again to get the information they need, then track down the actual bug.

PR #981, opened 2023-04-03
Move call events to the call span

dbkr:


src/otel/OTelGroupCallMembership.ts
```diff
@@ -205,13 +205,15 @@ export class OTelGroupCallMembership {
     const eventType = event.eventType as string;
     if (!eventType.startsWith("m.call")) return;
 
+    const callTrackingInfo = this.callsByCallId.get(call.callId);
+
     if (event.type === "toDevice") {
-      this.callMembershipSpan?.addEvent(
+      callTrackingInfo.span.addEvent(
```
EnricoSchw, line 211: In line [236](https://github.com/vector-im/element-call/blob/277081ee2a82cac68d24a6e87e0dab66e35bf2c7/src/otel/OTelGroupCallMembership.ts#L236) we check call is existing.  Please make sure we don't need this here. 


Otherwise I approve the PR

EnricoSchw (approved):

src/otel/OTelGroupCallMembership.ts
dbkr: Good spot, it should probably be here too.

PR #983, opened 2023-04-04
Track call rejoins

robintown:
Call rejoins will be one of the KPIs we track in PostHog to measure call quality. I've also reverted the previous behavior which logged all OpenTelemetry spans to PostHog, since we should only be sending small, anonymized bits of data there.


dbkr (approved):
Nice, thanks.

PR #984, opened 2023-04-04
Add OpenTelemetry events for PeerConnection state changes / errors

dbkr:
Creates a new class to represent individual calls and adds the listeners
there.

Requires https://github.com/matrix-org/matrix-js-sdk/pull/3251
Based on https://github.com/vector-im/element-call/pull/974


src/otel/OTelCall.ts
```diff
@@ -0,0 +1,103 @@
...
+    } else {
+      this.call.once(
+        CallEvent.PeerConnectionCreated,
+        this.addCallPeerConnListeners
+      );
+    }
+  }
+
+  public dispose() {
+    this.call.peerConn.removeEventListener(
```
EnricoSchw, line 44: We don't delete the other listeners? Is only a question because the connection object is dead after closing anyway.

EnricoSchw (approved):
Very practical and efficient. :-) Only one question.

src/otel/OTelCall.ts
dbkr: Oops, thanks.

PR #985, opened 2023-04-05
Fix exception when loading PostHog

dbkr:
PostHog was expecting the matrix client object to be initialised at the point it ran its setup, which wasn't the case. Check to see if it's there on login and add an onLoginStatusChanged hook that to re-check.

Also make a few methods private that didn't need to be public.

Also fix a few instances where the OpenTelemetry group call tried to report metrics using a tracer which didn't exist anymore, if the user disabled analytics and then joined the same call again.

### DIFF
--- a/src/otel/OTelGroupCallMembership.ts
+++ b/src/otel/OTelGroupCallMembership.ts
@@ -124,6 +124,8 @@ export class OTelGroupCallMembership {
   }
 
   public onJoinCall() {
+    if (!ElementCallOpenTelemetry.instance) return;
+
     // Create the main span that tracks the time we intend to be in the call
     this.callMembershipSpan =
       ElementCallOpenTelemetry.instance.tracer.startSpan(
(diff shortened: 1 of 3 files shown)


src/otel/OTelGroupCallMembership.ts
```diff
@@ -174,7 +176,7 @@ export class OTelGroupCallMembership {
     for (const [userId, userCalls] of calls.entries()) {
       for (const [deviceId, call] of userCalls.entries()) {
         if (!this.callsByCallId.has(call.callId)) {
-          const span = ElementCallOpenTelemetry.instance.tracer.startSpan(
+          const span = ElementCallOpenTelemetry.instance?.tracer.startSpan(
```
EnricoSchw, line 179: We should add ` if (!ElementCallOpenTelemetry.instance) return;` here as well. Because `span` could be `undefined`. This isn't quite as obvious as `ElementCallOpenTelemetry.instance` doesn't have `undefined` in the signature.

EnricoSchw (approved):
one comment. otherwise approved

src/otel/OTelGroupCallMembership.ts
dbkr: Good point. I've skipped the block rather than the whole function so we still remove calls from the active calls array.

PR #988, opened 2023-04-05
Allow different OpenTelemetry collectors to be enabled/disabled

dbkr:
Always enable OpenTelemetry, but conditionally enable the OTLP exporter, as per comment.

Fixes https://github.com/vector-im/element-call/issues/987

### DIFF
--- a/src/otel/otel.ts
+++ b/src/otel/otel.ts
@@ -40,7 +40,20 @@ export class ElementCallOpenTelemetry {
   private _otlpExporter: OTLPTraceExporter;
 
   static globalInit(): void {
-    recheckOTelEnabledStatus();
+    // we always enable opentelemetry in general. We only enable the OTLP
+    // collector if a URL is defined (and in future if another setting is defined)
+    // The posthog exporteer is always enabled, posthog reporting is enabled or disabled
+    // within the posthog code.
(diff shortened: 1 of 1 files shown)


src/otel/otel.ts
```diff
@@ -38,23 +37,17 @@ export class ElementCallOpenTelemetry {
   private _provider: WebTracerProvider;
   private _tracer: Tracer;
   private _anonymity: Anonymity;
+  private _otlpExporter: OTLPTraceExporter;
 
   static globalInit(): void {
-    settingsBus.on("opt-in-analytics", recheckOTelEnabledStatus);
-    recheckOTelEnabledStatus(getSetting("opt-in-analytics", false));
+    recheckOTelEnabledStatus();
```
robintown, line 43: Since this is now always enabled, and the config can't change, could we get rid of the whole recheck-loop thing and move the initialization to inside `get instance`?

dbkr: Yeah, may as well I guess - I assume we'll want a config setting at some point, but it can come back later. I've kept the globalInit since I think it's nice that it only reads from the config once the config is ready. wdyt?

robintown: Oh, sure, that's fine

robintown (approved):

PR #989, opened 2023-04-05
Don't require an OTLP collector to enable OpenTelemetry

robintown:
Since we're now exporting OpenTelemetry traces to PostHog and soon rageshakes, an OTLP collector shouldn't be a requirement for recording telemetry.


robintown:
Duplicate of https://github.com/vector-im/element-call/pull/988

PR #990, opened 2023-04-05
Log double-join attempts on the group call span

robintown:
This is to hopefully address https://github.com/vector-im/element-call/pull/971#discussion_r1145660452 a little better.

### DIFF
--- a/src/otel/OTelGroupCallMembership.ts
+++ b/src/otel/OTelGroupCallMembership.ts
@@ -128,6 +128,7 @@ export class OTelGroupCallMembership {
     if (!ElementCallOpenTelemetry.instance) return;
     if (this.callMembershipSpan !== undefined) {
       logger.warn("Call membership span is already started");
+      this.callMembershipSpan.addEvent("matrix.doubleJoin");
       return;
     }
 


src/otel/OTelGroupCallMembership.ts
```diff
@@ -128,6 +128,7 @@ export class OTelGroupCallMembership {
     if (!ElementCallOpenTelemetry.instance) return;
     if (this.callMembershipSpan !== undefined) {
       logger.warn("Call membership span is already started");
+      this.callMembershipSpan.addEvent("matrix.doubleJoin");
```
MadLittleMods, line 131: Potential better name?

```suggestion
      this.callMembershipSpan.addEvent("matrix.duplicateJoin");
```

MadLittleMods, line 131: We should also catch `duplicateLeave` situations in `onLeaveCall(...)`

dbkr (approved):

src/otel/OTelGroupCallMembership.ts
MadLittleMods, line 131: Perhaps it would be better to track this sort of thing by instrumenting `GroupCall.enter()` and `GroupCall.leave()` directly

fkwp:
we can close this due to our change to LiveKit

PR #991, opened 2023-04-06
Add tiny release notes script

dbkr:
As part of documenting how we release element call (although the script is not a critical stage of the release process by any means)


scripts/reformat-release-notes.py
```diff
@@ -0,0 +1,15 @@
...
+# project (ie. we don't really need to mention the author of every PR)
+#
+# eg. in: * OpenTelemetry by @dbkr in https://github.com/vector-im/element-call/pull/961
+#    out: * OpenTelemetry (https://github.com/vector-im/element-call/pull/961)
+
+import sys
+import re
+
+for line in sys.stdin:
+    matches = re.match(r'^\* (.*) by (.*) in (.*)$', line.strip())
```
robintown, line 14:
```suggestion
    matches = re.match(r'^\* (.*) by (\S+) in (\S+)$', line.strip())
```

So it doesn't match wrong on a PR title like "Improve Element Call by making it faster"

robintown (approved):

PR #997, opened 2023-04-12
Fix a crash when adding call events to telemetry

robintown:
Since `typeof null` is `'object'`, the `flattenVoipEventRecursive` function was mistakenly casting nulls to `Record<string, unknown>` in its `typeof v === "object"` case, causing `Object.entries` to explode.

Closes https://github.com/vector-im/element-call/issues/993

### DIFF
--- a/src/otel/OTelGroupCallMembership.ts
+++ b/src/otel/OTelGroupCallMembership.ts
@@ -74,7 +74,7 @@ function flattenVoipEventRecursive(
     );
 
   for (const [k, v] of Object.entries(obj)) {
-    if (["string", "number", "boolean"].includes(typeof v)) {
+    if (["string", "number", "boolean"].includes(typeof v) || v === null) {
       flatObject[prefix + k] = v;
     } else if (typeof v === "object") {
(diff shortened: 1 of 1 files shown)


src/otel/OTelGroupCallMembership.ts
```diff
@@ -74,7 +74,7 @@ function flattenVoipEventRecursive(
     );
 
   for (const [k, v] of Object.entries(obj)) {
-    if (["string", "number", "boolean"].includes(typeof v)) {
+    if (["string", "number", "boolean"].includes(typeof v) || v === null) {
       flatObject[prefix + k] = v;
     } else if (typeof v === "object") {
       flattenVoipEventRecursive(
```
EnricoSchw, line 80: FYI: We have a new [`flattenVoipEventRecursive`](https://github.com/vector-im/element-call/blob/a2b3e098b6c650f05e6db495d0ebabe17a2567d3/src/otel/ObjectFlattener.ts#L64)   which is an extended version of this here. The new one can also handle Maps.

EnricoSchw (approved):

PR #1001, opened 2023-04-14
Fix broken posthog requests in embedded mode

EnricoSchw:
In embedded mode the posthog otel export event is not triggert. This is a fix. I think it has something to do with the iframe in which the EC is embedded. In the first attempt I make a variable global (this is a hack and I'm still looking for a better solution)


EnricoSchw:
Fixed by https://github.com/vector-im/element-call/pull/1002 ! Can be closed.

PR #1019, opened 2023-04-20
Make webrtc stats configurable

EnricoSchw:
Setup webrtc stats collection interval

This PR based on https://github.com/matrix-org/matrix-js-sdk/pull/3307


src/room/useLoadGroupCall.ts
```diff
@@ -112,12 +117,17 @@ export const useLoadGroupCall = (
...
+          .createGroupCall(
+            room.roomId,
+            createPtt ? GroupCallType.Voice : GroupCallType.Video,
+            createPtt,
+            GroupCallIntent.Room
+          )
+          .then((groupCall) => {
+            groupCall.statsCollectIntervalTime = STATS_COLLECT_INTERVAL_TIME_MS;
+            return groupCall;
+          });
```
dbkr, line 130: I would probably just do `const newGroupCall = await client.createGroupCall(...); newGroupCall. statsCollectIntervalTime = STATS_COLLECT_INTERVAL_TIME_MS; return newGroupCall;` at this point, as I found my self trying to work out whether the `await` was still necessary above.

dbkr (approved):

PR #1025, opened 2023-04-24
Remove the console exporter

dbkr:
I'm not sure how this got left in, presumably it was not intended.

### DIFF
--- a/src/room/useGroupCall.ts
+++ b/src/room/useGroupCall.ts
@@ -133,7 +133,6 @@ function getParticipants(
             : ConnectionState.WaitMedia
           : ConnectionState.EstablishingCall;
       }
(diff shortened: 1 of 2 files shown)


src/room/useGroupCall.ts
```diff
@@ -133,6 +133,7 @@ function getParticipants(
             : ConnectionState.WaitMedia
           : ConnectionState.EstablishingCall;
       }
+      console.log(`${member.userId} conn state is ${connectionState}`);
```
robintown, line 136: This should go through `logger`

robintown (approved):

src/room/useGroupCall.ts
dbkr: Oops, actually didn't mean to commit that at all.

Issue #978, opened 2023-03-31
"JavaScript heap out of memory" in the build process


MomentQYC:
### Steps to reproduce

1. run `yarn` and `yarn build`
2. The error occurs and the process is interrupted.

#### What did you expect?
Successfully build
#### What happened instead?
```shell
<--- Last few GCs --->

[1995771:0x66ed170]    96045 ms: Scavenge (reduce) 2042.8 (2081.6) -> 2042.6 (2082.3) MB, 14.8 / 0.0 ms  (average mu = 0.475, current mu = 0.384) allocation failure; 
[1995771:0x66ed170]    98419 ms: Mark-sweep (reduce) 2043.6 (2082.3) -> 2043.2 (2082.8) MB, 2372.4 / 0.0 ms  (average mu = 0.270, current mu = 0.019) allocation failure; scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: 0xb7b3f0 node::Abort() [node]
 2: 0xa8c89a  [node]
 3: 0xd631b0 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xd63557 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xf40c55  [node]
 6: 0xf5313d v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [node]
 7: 0xf2d83e v8::internal::HeapAllocator::AllocateRawWithLightRetrySlowPath(int, v8::internal::AllocationType, v8::internal::AllocationOrigin, v8::internal::AllocationAlignment) [node]
 8: 0xf2ec07 v8::internal::HeapAllocator::AllocateRawWithRetryOrFailSlowPath(int, v8::internal::AllocationType, v8::internal::AllocationOrigin, v8::internal::AllocationAlignment) [node]
 9: 0xf0fdda v8::internal::Factory::NewFillerObject(int, v8::internal::AllocationAlignment, v8::internal::AllocationType, v8::internal::AllocationOrigin) [node]
10: 0x12d36cf v8::internal::Runtime_AllocateInYoungGeneration(int, unsigned long*, v8::internal::Isolate*) [node]
11: 0x1700579  [node]
Aborted
```
Node.js v18.14.2
[...]

PR #1026, opened 2023-04-25
Add data ID tags for trafficlight adapter to find

michaelkaye:
This PR adds an initial set of data-testid tags to enable certain elements on the page to be found via playwright's lookup mechanisms. 

It also includes a quick change for the issue #978, which i was otherwise running into each time I built. I should get rid of that commit and maybe put it in another branch somewhere, but would appreciate guidance on what to do with it properly - this change seems to JustWork for me, so i don't know why we wouldn't want to include it.


michaelkaye:
I should also mention there will be further sets of changes incoming; but this is a short set that will help for now.

SimonBrandner (approved):
LGTM

package.json
```diff
@@ -3,7 +3,7 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "build": "vite build",
+    "build": "NODE_OPTIONS=--max-old-space-size=16384 vite build",
```
dbkr, line 6: We should remove this from the github action if we're putting it in here

michaelkaye: Removed these in 9c8692d

PR #1037, opened 2023-05-05
Settings improvements

SimonBrandner:
Fixes https://github.com/vector-im/element-call/issues/945

<hr>

Unclear things: 

- [x] Beta flag/tab/whatever - how does it behave, what does it mean, is it useful at the moment?
- [x] We probably need the user ID in the profile tab
- [x] It probably makes sense to keep the button to change the display name to make it clear when the change is done
- [x] We now have a separate developer settings tab - do we want to keep it?


robintown:
Here's my understanding of the current situation:

- Beta flag/tab/whatever - how does it behave, what does it mean, is it useful at the moment?
I believe this tab is something that Gaëlle is proposing for once EC exits beta - a tab to connect users with any beta-quality experiments we're running, such as SFU calls. For now it's not relevant
- We probably need the user ID in the profile tab
I was about to say I disagree since there's no need to share your contact information in EC, but then I remembered that you need to know your user ID in order to log in, so yes, we should keep it where it is for now.
- It probably makes sense to keep the button to change the display name to make it clear when the change is done
A button like this certainly made sense for when the profile interface was a modal, because modals are expected to have a primary action button, but I worry that the button would feel out of place in a larger settings interface / risk people ignoring it when going to change their profile. I suggest we make the form automatically save when the profile tab unmounts.
- We now have a separate developer settings tab - do we want to keep it?
Yep, that tab was added because we wanted to avoid polluting the new settings interface with developer options for ordinary users.

src/settings/ProfileSettingsTab.tsx
```diff
@@ -78,64 +73,51 @@ export function ProfileModal({ client, ...rest }: Props) {
...
+          avatarUrl={avatarUrl}
+          displayName={displayName}
+          onRemoveAvatar={onRemoveAvatar}
+        />
+      </FieldRow>
+      <FieldRow>
+        <InputField
+          id="userId"
+          name="userId"
+          label={t("User ID")}
```
robintown, line 92:
```suggestion
          label={t("Username")}
```

This is what the MXID is called on the login and registration pages, so we should be consistent with it

src/settings/FeedbackSettingsTab.tsx
```diff
@@ -0,0 +1,93 @@
...
+          />
+        </FieldRow>
+        {error && (
+          <FieldRow>
+            <ErrorMessage error={error} />
+          </FieldRow>
+        )}
+        <FieldRow>
+          <Button type="submit" disabled={sending}>
+            {sending ? t("Submitting feedback…") : t("Submit feedback")}
```
robintown, line 87:
```suggestion
            {sending ? t("Submitting…") : t("Submit")}
```

To match the designs

src/settings/FeedbackSettingsTab.tsx
```diff
@@ -0,0 +1,93 @@
...
+interface Props {
+  roomId?: string;
+}
+
+export function FeedbackSettingsTab({ roomId }: Props) {
+  const { t } = useTranslation();
+  const { submitRageshake, sending, error } = useSubmitRageshake();
+  const sendRageshakeRequest = useRageshakeRequest();
+
+  const onSubmitFeedback = useCallback(
```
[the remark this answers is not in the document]
SimonBrandner: Great suggestion!

robintown:
If someone else from the team could look over the last changes I've made, that'd be great :)

robintown:
Screenshots:

Screenshot 2023-05-22 at 15-43-06 Element Call test
Screenshot 2023-05-22 at 15-42-50 Element Call test

Issue #1072, opened 2023-05-18
Unable to mute video when no audio device present on system


michaelkaye:
### Steps to reproduce

1. Register new user
2. Create new room
3. See video preview works
4. Click video mute button
5. State of video mute button does not alter; video remains unmuted.

Repeatable replication available via trafficlight test (test not complete so not committed yet)

#### What did you expect?
Even if no audio device present, to be able to mute video device.

#### What happened instead?
Video device could not be muted.

Logs say:
```
* Trafficlight asked to execute action "set_mute", data = {"video_mute":true}:
	Action for "set_mute" found in action-map  ✔
Video state currently unmuted, clicking.
PP0: GroupCall 1684422238299arroN3r6MjWal2Bf setLocalVideoMuted() (stream=7d3686d0-d200-4058-b977-8d9afeb1af5f, muted=true)
PP0: GroupCall 1684422238299arroN3r6MjWal2Bf setLocalVideoMuted() no device or permission to receive local stream, muted=true
```
And the icon state doesn't change; which fails the test.

### Operating system

Trafficlight/Docker/Linux

### Browser information

Chromium 114.0.5735.35 i believe; installed in playwright environment

### URL for webapp

element-call main

### Will you send logs?

No

michaelkaye:
https://github.com/vector-im/element-call/assets/1917473/f8b899b5-3666-4d84-80c3-171a124eadd7

Attempt to click the video mute button is at 0:11 - and the remaining 30s of the test video is waiting for the button to change.

michaelkaye:
IMHO, at minimum; it should be clear to the end user that despite hitting the video mute their video has failed to be muted; ideally it should also mute the video ^^

michaelkaye:
The exact same test works fine when run outside of docker; the difference which makes me think it's audio related being that inside docker there is only this device:
```
[
  {
    deviceId: '8d1ed80dcf9218e731dbbfb508c327d0e424d1b3da022c8ea6262e9f873a10b0',
    kind: 'videoinput',
    label: 'X_12',
    groupId: '923de623a74867356996efa2c1a9b844f56a55c7e24310295f1f08aacbd66c33'
  }
]
```

Outside docker there are these devices:
```
[
  {
    deviceId: 'default',
    kind: 'audioinput',
    label: 'Default',
    groupId: 'b1639839d99626b67b1a6384ba940d936ddd4a1d3d07e7c415da0ca4454b3a99'
  },
  {
[...]

PR #1073, opened 2023-05-18
Identify SVGs representing video muted and video unmuted

michaelkaye:
The underlying tests don't currently function due to #1072 but this PR at least allows us to identify the state and things to click - we can just avoid running the actual tests for now.


michaelkaye:
This isn't needed, superseded by https://github.com/vector-im/element-call/pull/1363

(at least me not merging this explains why the data-testids ended up missing when i came back to use them :D

PR #1080, opened 2023-05-25
move webrtc etc. events from groupCall to matrix.call span

EnricoSchw:
Move the stats as span under call span

Bildschirmfoto 2023-05-25 um 15 10 40

This PR depends on: https://github.com/matrix-org/matrix-js-sdk/pull/3413


robintown (approved):
Looks good to go!

PR #1084, opened 2023-06-05
Add quality survey at the end of the call

toger5:
fixes: https://github.com/vector-im/element-call/issues/236

Shows a quality survey at the end of the call:

https://github.com/vector-im/element-call/assets/16718859/4cfa2e08-a737-46e1-b6fa-40145be5766d

 - only show survey if analytics are enabled
 - Show survey before the "create account" dialog
 - Data is posted anonomysly on posthog as an event
 - pressing submit is showing a submit button for 1s and a thanks for your feedback label for 1 s. (those delays are added artificially to give feedback that the submission worked)


public/locales/en-GB/app.json
```diff
@@ -109,6 +111,7 @@
   "Stop sharing screen": "Stop sharing screen",
   "Submit": "Submit",
   "Submit feedback": "Submit feedback",
+  "Submitting...": "Submitting...",
   "Submitting…": "Submitting…",
```
robintown, line 115: These strings are redundant, we should prefer the second one

src/input/Input.tsx
```diff
@@ -130,15 +130,17 @@ export const InputField = forwardRef<
...
-
-        <label htmlFor={id}>
-          {type === "checkbox" && (
-            <div className={styles.checkbox}>
-              <CheckIcon />
-            </div>
-          )}
-          {label}
-        </label>
+        {/* Dont create label element if not set -> prohibits broken animation on focus*/}
```
robintown, line 133: For screen reader accessibility and usability in general, it's important that we always label fields. Even though the designs technically don't show a label on the feedback field (only a placeholder), I suggest we use 'Your feedback' as the label. (It could easily have been an accidental omission in the designs)

src/input/StarRatingInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+            type="button"
+            key={index}
+            className={styles.star}
+            onClick={() => {
+              setRating(index);
+              onChange(index);
+            }}
+            onMouseEnter={() => setHover(index)}
+            onMouseLeave={() => setHover(rating)}
+          >
```
robintown, line 56: I think this is going to be confusing for screen reader users, because the buttons are without any `aria-label` and disconnected from the number input, which has no effect when changed. A good way to solve this might be to model the stars as a group of radio buttons.

src/room/CallEndedView.module.css
```diff
@@ -17,20 +17,31 @@ limitations under the License.
...
+  white-space: pre;
 }
 
 .callEndedContent {
   text-align: center;
-  max-width: 360px;
+  max-width: 450px;
+}
+.callEndedContent p {
+  font-size: 18px;
```
robintown, line 28: Let's use one of these font size variables: https://github.com/vector-im/element-call/blob/15350b6989fee1b2e458994400daec8c4f4e6c82/src/index.css#L33

src/room/CallEndedView.tsx
```diff
@@ -14,19 +14,127 @@ See the License for the specific language governing permissions and
...
+      );
+
+      setSubmitting(true);
+
+      setTimeout(() => {
+        setSubmitDone(true);
+
+        setTimeout(() => {
+          if (isPasswordlessUser) {
+            // setting this renders the callEndedView with the invatation to create an account
```
robintown, line 68:
```suggestion
            // setting this renders the callEndedView with the invitation to create an account
```

src/room/CallEndedView.tsx
```diff
@@ -14,19 +14,127 @@ See the License for the specific language governing permissions and
...
+            id="feedbackText"
+            name="feedbackText"
+            placeholder={t("Your feedback")}
+            type="textarea"
+            required
+          />
+        </FieldRow>{" "}
+        <FieldRow>
+          {submitDone ? (
+            <p>Thanks for your feedback!</p>
```
robintown, line 122: This should be translated

src/settings/FeedbackSettingsTab.tsx
```diff
@@ -68,9 +69,10 @@ export function FeedbackSettingsTab({ roomId }: Props) {
       <form onSubmit={onSubmitFeedback}>
         <FieldRow>
           <InputField
+            className={feedbackStyles.feedback}
             id="description"
             name="description"
-            label={t("Your feedback")}
+            placeholder={t("Your feedback")}
```
robintown, line 75: Same thing as before, we should always have labels

robintown (commented):
Looking good, here are my suggestions

src/input/StarRatingInput.tsx
toger5: I changed it to radio buttons and tried it with a screen reader. It seemed fine (was reading out as: "one star, selected radio button one of five ...)

src/input/Input.tsx
toger5: this makes this option to not add a label very bad. reverted it and changed to look to diverge from the designs and showing the label in the feedback text field

src/input/StarRatingInput.tsx
```diff
@@ -34,34 +35,46 @@ export function StarRatingInput({
...
+                onChange(index);
+              }}
+              required
+            />
+            <label
+              className={styles.hideElement}
+              key={"lbl" + String(index)}
+              htmlFor={String(index)}
+            >
+              {index + " " + t("star")}
```
robintown:
```suggestion
              {t("{{count}} stars", { count: index })}
```

PR #1092, opened 2023-06-07
LiveKit Experiment

daniel-abramov:
See our internal documents for more details.

**The PR is created for the sole purpose of creating an automatic deployment!**


OmElDonia:
I am testing the latest deployment and unlike both the stable and dev deployments, it says "Oops, something's gone wrong." for some testers.

After further investigations, it seems like not having a video source is causing this.

```
TypeError: Cannot read properties of undefined (reading 'muted')
    at Tet (GroupCallView.tsx:192:1)
    at DY (react-dom.production.min.js:167:137)
    at nj (react-dom.production.min.js:197:258)
    at She (react-dom.production.min.js:292:88)
    at _he (react-dom.production.min.js:280:389)
    at Cxe (react-dom.production.min.js:280:320)
    at dB (react-dom.production.min.js:280:180)
    at hj (react-dom.production.min.js:271:88)
    at mhe (react-dom.production.min.js:268:429)
    at j (scheduler.production.min.js:13:203)
```

Adding a video source fixes it.

daniel-abramov:
> I am testing the latest deployment and unlike both the stable and dev deployments, it says "Oops, something's gone wrong." for some testers.
> 
> After further investigations, it seems like not having a video source is causing this.
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'muted')
>     at Tet (GroupCallView.tsx:192:1)
>     at DY (react-dom.production.min.js:167:137)
>     at nj (react-dom.production.min.js:197:258)
>     at She (react-dom.production.min.js:292:88)
>     at _he (react-dom.production.min.js:280:389)
>     at Cxe (react-dom.production.min.js:280:320)
>     at dB (react-dom.production.min.js:280:180)
>     at hj (react-dom.production.min.js:271:88)
>     at mhe (react-dom.production.min.js:268:429)
>     at j (scheduler.production.min.js:13:203)
> ```
> 
> Adding a video source fixes it.

Ah, right the `video` source is `undefined` if there are no video devies and our (non-strict) lints did not capture it.

It's this part in a `GroupCallView.tsx`:

```
        userChoices={{
          videoMuted: lkState.localMedia.video.muted,
          audioMuted: lkState.localMedia.audio.muted,
        }}
```

As a quick fix, you can replace it with:
```
        userChoices={{
          videoMuted: lkState?.localMedia.video?.muted ?? true,
          audioMuted: lkState?.localMedia.audio?.muted ?? true,
        }}
```

We probably need to submit a patch for that (cc @toger5).

src/room/InCallView.tsx
```diff
@@ -165,6 +165,13 @@ export function InCallView({
     options
   );
 
+  // TODO: move the room creation into the useRoom hook and out of the useLiveKit hook.
+  // This would than allow to not have those 4 lines
+  livekitRoom.options.audioCaptureDefaults.deviceId =
+    mediaDevices.state.get("audioinput").selectedId;
+  livekitRoom.options.videoCaptureDefaults.deviceId =
+    mediaDevices.state.get("videoinput").selectedId;
```
daniel-abramov, line 173: NB: The options could be passed directly to the `useLiveKitRoom` below.

src/livekit/useLiveKit.ts
```diff
@@ -102,27 +125,32 @@ export function useLiveKit(): LiveKitState | undefined {
   return state;
 }
 
+// if a room is passed this only affects the device selection inside a call. Without room it changes what we see in the lobby
 function useMediaDevicesState(room: Room): MediaDevicesState {
+  let connectedRoom: Room;
+  if (room.state !== ConnectionState.Disconnected) {
+    connectedRoom = room;
+  }
```
daniel-abramov, line 133: What if we move it to the `useLiveKitState()` function above and accept a `room?: Room`? I think it will be much more flexible and elegant since we'll then be able to use the function from multiple places (also useful for the further refactoring).

src/livekit/useLiveKit.ts
```diff
@@ -19,6 +25,10 @@ type LiveKitState = {
   room: Room;
 };
 
+function emptyToUndef(str) {
+  return str === "" ? undefined : str;
+}
+
```
daniel-abramov, line 31: Oh, when I meant moving things to a function, I meant that thing that selects an active device based on the active device setting and the default device setting, not the conversion to undefined 🙂

src/livekit/useLiveKit.ts
```diff
@@ -32,21 +42,35 @@ export function useLiveKit(): LiveKitState | undefined {
   // Create a React state to store the available devices and the selected device for each kind.
   const mediaDevices = useMediaDevicesState(room);
 
-  // Create local video track.
+  const [settingsDefaultDevices] = useDefaultDevices();
```
daniel-abramov, line 45: A left over? We seem to call this function from the `useMediaDevicesState()`.

src/room/InCallView.tsx
toger5: yes I 100% agree. This should be refactored as mentioned in the TODO comment above

src/livekit/useLiveKit.ts
toger5: I was considering this. But since we have `activeAudioDevice`, `settingsDefaultDevices.audioinput` `audioDevices[0]?.deviceId` there are 3 parameters that need to be different for each function call. So it is not really possible to make it very generic. I would end up to be a function that looks like so:
```
function (a,b,c){
a === "" ? (b === "" ? c : b) : a
)
```
which i found less readable than `??` chaingin. WIth `??` chaing it is at least obvious what priority the different variables have.
wdyt?

src/livekit/useLiveKit.ts
toger5: How could i get the settings from the `useMediaDevicesState()`
I am calling it again to get the settings val in theis scope. since it is a hook that should be no perf disadvantag?

src/video-grid/VideoGrid.tsx
```diff
@@ -692,38 +782,53 @@ interface DragTileData {
...
+// ie. a user's video feed or a screen share feed.
+export interface TileDescriptor<T> {
+  id: string;
+  focused: boolean;
+  isPresenter: boolean;
+  isSpeaker: boolean;
+  hasVideo: boolean;
+  local: boolean;
+  largeBaseSize: boolean;
+  placeNear?: string;
```
daniel-abramov, line 816: First of all, thanks for the changes that bring new behavior to the grid layout, that was faster than expected 🙂 I think since we don't have to rush anymore, this seems like a good opportunity to discuss things that we could improve!

---

This change in the `TileDescriptor` (and the corresponding changes in the code that depends on it) seems to introduce a lot of complexity.  Originally (after the first refactoring that I did by generalizing the descriptor and removing unnecessary parts), we only had 2 boolean properties to describe a single tile: `focused` and `local` that contained 4 bits of information. Although `local` felt out of place (i.e. the layout/grid was expected to know what the "locality of participant" is and it was mostly used to control the layout in special cases of 1:1 calls, the complexity was still low (knowing that there are 4 potential states for the tile [although it was actually only 3] was bearable). Now we have 6 flags (**64 bits of information!**) along with the optional `placeNear`. So alone the flags allow a user to create **64 different combinations** of a tile description. This seems like much more state possibilities than we actually want to have 😉

The second problem is the encapsulation of logic. I think that in order to create a simple and reliable video grid layout logic, we could think of a grid layout as a separate generic component that exists in a separate library and does not know much about the type of content that is being rendered inside it, it simply accepts some input (the array of description of the tiles) and arranges them. Currently, we also have generic fields (`focused`) mixed with conferencing-specific fields that explicitly convey their meaning by their names (`hasVideo`, `isSpeaker`, etc). Moreover, the values of the flags are set in an unobvious way where the values overlap with each other in terms of meaning, e.g.:

```typescript
        const userMediaTile = {
          id,
          focused: false,
          isPresenter: sfuParticipant.isScreenShareEnabled,
          isSpeaker:
            (sfuParticipant.isSpeaking || hadSpokedInTime) &&
            !sfuParticipant.isLocal,
          hasVideo: sfuParticipant.isCameraEnabled,
          local: sfuParticipant.isLocal,
          largeBaseSize: false,
          data: {
            member,
            sfuParticipant,
            content: TileContent.UserMedia,
          },
        };
```

Note that `local` (`isLocal`) is set explicitly (i.e. the video grid knows when the field is local), but at the same time `isLocal` appears in the `isSpeaker` calculation which is counterintuitive - if the video grid knows that the tile is local, why does it expect that for calculating of certain other fields (`isSpeaker` in this case), the caller must include `local` in calculation? (it defeats the purpose of passing `local` to the video grid in the first place). Also note that `focused` is always set to `false` which also changes the meaning of what `focused` used to mean. It seems like it's easy to create a combination of flags that do not correspond to any set of the actual GUI state of a grid (i.e. to create a description that is "not supposed to be used").

I've shortly checked how these values are used and it seems like they are used to create some sort of a ranking system. Also, the meaning of some flags seems to change depending on the layout being used.

**One of the ways** to overcome the aforementioned issues is to implement the ranking explicitly - to pass a single integer value `priority` (or `rank`, or `importance` if you will) instead of a set of flags. This would allow the caller to specify the importance of a tile and to change the implementation of who is important and who isn't without ever changing the implementation logic of the video grid. This also allows generalizing the video grid algorithm to work for a more generic case. It also separates two concerns well. Also, it will be much more natural for the caller to describe the importance of the tiles as this will be done *explicitly* (instead of expecting that the video grid _knows_ that e.g. screen sharing tiles are more important than the rest of the tiles).

Then, calculating the importance of a tile may be as compact and elegant as something like this:

```rust
let importances = participants.map(|p| {
    let mut importance = 0;
    for criteria in &[p.isSpeaking, p.isScreensharing, p.hasCameraEnabled] {
        if criteria {
            importance += 1;
        }
    }
    return importance
});
```

If we discover that there are actually special cases that change the rendering so much that it does not fit the concept well, the better choice may be to create different layouts for them as different layouts may require different algorithms and hence different input parameters.

daniel-abramov:
We can close this one in favor of [our new development branch](https://github.com/vector-im/element-call/tree/livekit) as we don't really plan to merge it in `feature_sfu`.

PR #1093, opened 2023-06-07
Don't show the quality survey if the app is a widget

robintown:
As explained by the comment, we don't yet have designs that account for this combo.

### DIFF
--- a/src/room/GroupCallView.tsx
+++ b/src/room/GroupCallView.tsx
@@ -272,7 +272,16 @@ export function GroupCallView({
       );
     }
   } else if (left) {
-    if (isPasswordlessUser || PosthogAnalytics.instance.isEnabled()) {
+    // The call ended view is shown for two reasons: prompting guests to create
+    // an account, and prompting users that have opted into analytics to provide
+    // feedback. We don't show a feedback prompt to widget users however (at
+    // least for now), because we don't yet have designs that would allow widget
+    // users to dismiss the feedback prompt and close the call window without
+    // submitting anything.
+    if (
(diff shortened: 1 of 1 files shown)


src/room/GroupCallView.tsx
```diff
@@ -272,7 +272,16 @@ export function GroupCallView({
...
-    if (isPasswordlessUser || PosthogAnalytics.instance.isEnabled()) {
+    // The call ended view is shown for two reasons: prompting guests to create
+    // an account, and prompting users that have opted into analytics to provide
+    // feedback. We don't show a feedback prompt to widget users however (at
+    // least for now), because we don't yet have designs that would allow widget
+    // users to dismiss the feedback prompt and close the call window without
+    // submitting anything.
+    if (
+      isPasswordlessUser ||
+      (PosthogAnalytics.instance.isEnabled() && !isEmbedded)
```
toger5, line 283: Is there ever the case that we have a passwordless user in embedded mode?
Or do we just never want to show the call ended view in embedded mode. In this case you can change the condition to just never be true if embedded?
```suggestion
      (isPasswordlessUser || PosthogAnalytics.instance.isEnabled()) && !isEmbedded
```

toger5 (approved):
Makes sense.
I dont know if you want to change the condition but either way I can approve

EnricoSchw (approved):

src/room/GroupCallView.tsx
robintown: There can be no passwordless users in embedded mode, so the condition is equivalent to what you've suggested

PR #1097, opened 2023-06-09
Don't require textual feedback

robintown:
We want to encourage scoring as much as possible for the purpose of our KPIs, even if it means we don't always get detailed textual feedback.


SimonBrandner (approved):

toger5:
I was thinking about this as well. Its annoying to not be able to only send stars

PR #1099, opened 2023-06-09
Merge main into livekit-experiment

robintown:


src/video-grid/VideoTileContainer.tsx
```diff
@@ -14,48 +14,82 @@ See the License for the specific language governing permissions and
...
+  x?: SpringValue<number>;
+  y?: SpringValue<number>;
+  width?: SpringValue<number>;
+  height?: SpringValue<number>;
+  onDragRef?: RefObject<
+    (
+      tileId: string,
+      state: Parameters<Handler<"drag", EventTypes["drag"]>>[0]
+    ) => void
+  >;
```
daniel-abramov, line 58: Oh wow, that's many additional parameters.

Is there a way we could separate them from the rest of the props so that the child components don't need to receive and handle them?

My concern is that apart from the [simple] rendering logic of the in a `VieoTile`, we pass many video-grid specific parameters that are essentially only passed down and/or being used from one place.

It looks like this logic could be separated, so that when someone works with their own React component that is used as a child, they don't really need to know the implementation details of the parent component (video grid) in order to make it work properly (this was also the case in the version of the old video grid, which I was concerned about, but it was not that critical as it contained 1-2 related parameters).

So essentially, my expectation is that ideally we want the following properties:
- it should be easy to add a React component as a child element to the grid layout without the knowledge of how the video grid layout works and without the need to handle the passed values.
- the changes in the video grid logic (parent) must not require changing child components

Otherwise, we run into a problem the video grid and any child element are strongly coupled together and can't be used separately, also increases the overall complexity of the child as it means that anyone who develops child components cannot properly implement them without knowing about the parent component much.

Does it make any sense? :)

daniel-abramov: I agree that in rare cases, for the optimization's sake, the child may need to get _something_ from a parent and e.g. call a function. While such cases are rare, they do indeed occur. However, in such cases, we still may simplify it down to a single parameter (closure) that each child gets and calls e.g. at the beginning, something like `useAnimations()`.

I don't know if it helps, but I think [LiveKit created some nice and easy-to-use abstractions](https://github.com/livekit/components-js/blob/4a461513c17d421a1e61bc845c7c1ef96b1cc4b4/packages/react/src/components/layout/GridLayout.tsx#L25) for it. While their layout is certainly simpler than ours, it would be certainly possible to minimize the list of "dependencies" of the child 🙁

robintown: See https://github.com/vector-im/element-call/pull/1104 - it ought to improve this situation

daniel-abramov: Oh, nice, thanks for addressing it promptly 🙂 I did not expect we would make this improvement right now, but it's really nice to see that the first improvements started to appear 👍

PR #1101, opened 2023-06-12
Connection lost banner

toger5:
The connectivity is determined by the sync state of the client. A sync error with the state: `"ConnectionError"` will result in a banner stating "Connectivity to the server has been lost." (The same wording that is used in element web)
 
Screenshot 2023-06-12 at 10 16 32
Screenshot 2023-06-12 at 10 16 44


src/initializer.tsx
```diff
@@ -45,6 +45,12 @@ class DependencyLoadStates {
 
 export class Initializer {
   private static internalInstance: Initializer;
+  private isInitialized = false;
+
+  public static isInitialized(): boolean {
+    return Boolean(Initializer.internalInstance?.isInitialized);
```
robintown, line 51: This cast seems redundant?

src/ClientContext.tsx
```diff
@@ -183,17 +205,23 @@ export const ClientProvider: FC<Props> = ({ children }) => {
         }
       }
     };
-
+    let clientWithListener;
```
robintown, line 208: This should have a type

src/ClientContext.tsx
```diff
@@ -107,8 +119,18 @@ export const ClientProvider: FC<Props> = ({ children }) => {
...
   });
 
+  const onSync = (state: SyncState, _old: SyncState, data: ISyncStateData) => {
+    setState((currentState) => {
+      return {
+        ...currentState,
+        disconnected: isDisconnected(state, data),
+      };
+    });
+  };
```
robintown, line 132: I'm slightly concerned that this could cause the entire component tree to re-render on every sync, because it sets the state to a new object every time (and React [doesn't perform a shallow comparison on state updates in functional components, but rather an `Object.is` comparison](https://react.dev/reference/react/useState#setstate-caveats)). Something like this would avoid that issue:

```suggestion
  const onSync = (state: SyncState, _old: SyncState, data: ISyncStateData) => {
    setState((currentState) => {
      const disconnected = isDisconnected(state, data)
      return disconnected === currentState.disconnected
        ? currentState
        : { ...currentState, disconnected };
    });
  };
```

robintown (commented):
Just a couple suggestions - also, the text in the banner appears to be off center, do you have any idea why that is?

src/ClientContext.tsx
toger5: Thanks!

robintown (approved):

PR #1104, opened 2023-06-12
Decouple video grid from video tile components

robintown:
This is an attempt to address the feedback in https://github.com/vector-im/element-call/pull/1099#discussion_r1226863404 that the video grid and video tile components have become too tightly coupled. After this change, the only requirements that the video grid makes of its child components are:

- They accept ref, style, and item props
- They attach the ref and styles to a react-spring animated element

Note: I removed the video grid Storybook file, because I'm not aware of anyone using Storybook for development of Element Call beyond Robert, and it would take some effort to fix to work with these changes.

Incidentally, this fixes https://github.com/vector-im/element-call/issues/1081 and https://github.com/vector-im/element-web/issues/24601.

Closes https://github.com/vector-im/element-call/issues/1081
Closes https://github.com/vector-im/element-web/issues/24601


dbkr (approved):

daniel-abramov (approved):
Thanks Robin! The new list of "dependencies" of a child from a parent is undoubtedly more manageable! 👍

PR #1106, opened 2023-06-13
Display SFU participants who did not publish an `m.call.member` event

daniel-abramov:
Previously it was only possible to show the participants who respected MSC3401 in the sense that all displayed participants were Matrix participants. This is good, however, it does not allow to show the **actual** participants who might have entered the room without publishing the `m.call.member` event, such as test participants created by the load testing tool.

This PR removes this limitation.

In the future, we'll rely upon `m.call.member` event being published in a room and to ensure that no one can spy on participants, E2EE will be used, so only the participants who published the `m.call.member` event will be able to decrypt the media.

**It seems like sometimes there are problems with rendering (unnecessary re-renders and jumping tiles whenever someone joins), but apart from that it works more or less.**

PS: This PR contains **merge commits** (and in fact 2 of them since I ~~rebased~~ "synchronized" with a base branch twice). This certainly looks ugly and has recently been discussed internally, but since it's the way other PRs are created in the `element-call`, I've kept it the same way here without doing the rebase.


src/video-grid/VideoTileContainer.tsx
```diff
@@ -50,35 +50,64 @@ interface Props {
...
 }
 
-export const VideoTileContainer: FC<Props> = memo(
-  ({ item, targetWidth, targetHeight, getAvatar, onDragRef, ...rest }) => {
-    const { rawDisplayName } = useRoomMemberName(item.member);
-    const tileRef = useRef<HTMLElement | null>(null);
+export const VideoTileContainer: React.FC<Props> = React.memo(
+  ({ item, id, targetWidth, targetHeight, onDragRef, ...rest }) => {
+    // Handle display name changes.
+    const [displayName, setDisplayName] = React.useState<string>("[👻]");
```
SimonBrandner, line 64: Perhaps we could export this into a const for better clarity

SimonBrandner (approved):
Looks sane

PR #1108, opened 2023-06-14
Build EC room use method

EnricoSchw:


EnricoSchw:
I will clean up the log statements and add common logger

PR #1115, opened 2023-06-14
Revert `useRoom()`

daniel-abramov:
Sorry @EnricoSchw , I had to revert the change as it added some undesired side-effect (see below). I’ve noticed that the `useLiveKitRoom()` has been replaced with a custom `useRoom()` and I did not quite get what issue it was meant to solve.

`useRoom()` is a copy-paste from LiveKit’s `useLiveKitRoom()` minus the initialization/forwarding of the HTML props and handling of a `simulateParticipants` parameter. However it still uses the same input type for the props which makes it even more confusing because it means that these parameters are ignored (in fact [you passed `simulateParticipants: 10`](https://github.com/vector-im/element-call/pull/1092/commits/20c9c092586983e206a37576a23346e9dc012aea#diff-20eaa30bd23102b20e1a663910bafe1f71ce4db711fbcb08d21019e29fe63457R169) and it did no effect because it was not handled).

At the same time, it introduces a regression by activating local participant devices one-by-one without awaiting the returned promise object (setXXX functions return a `Promise` and must be awaited upon). This means that the passed error callback will not be called in a `catch` block should the promise throw an exception.

If the idea was to prevent warnings in the console, all that must have been done is to memoize the params (as mentioned in our discussion room), so that they are not re-created on each render (which you successfully achieved by moving the creation of callback objects to the outside of a component). What I can’t understand is what was the purpose of the `useRoom()`.

So far a tiny change (to not re-create callbacks on each render) is more than enough to fix the issue without introducing additional undesired effects. Let me know if I missed something.


daniel-abramov:
> Hi @daniel-abramov . Don't revert that. I know what you are referring to. Just add your five lines for `simulateParticipants` to `useRoom`, that should be enough then. This is the beginning of a refactoring of the room state. I'll wait until @toger5 is done with device selection and @dbkr with the mute camera bug.

Could you please elaborate on that, @EnricoSchw?

Camera and device selection bugs are coupled with the room state.

And the `useRoom()` is no different from `useLiveKitRoom()`. I think we really really need to be careful with such changes as it introduces unnecessary accidental complexity.

EnricoSchw:
You can understand the premises, just not the conclusion i did. And its because about `copy-paste`?

daniel-abramov:
> You can understand the premises, just not the conclusion i did. 

I understand your points regarding the use of `room` and agree with them, but the problem is that they are orthogonal to the changes that have been done (I would understand if you refactored `useLiveKitState()` to not use the `Room` etc and that would be understandable, in fact, I've done it in a branch also since I also don't like the way it's currently used).

My problem is that originally (from what I understood) you tried to eliminate the warning of being connected to the room. As it was shortly exchanged in chat, it would be enough to memoize the arguments to fix it (a tiny change, about 5 LOC). This would be a perfect little commit that does the job.

But instead of this, there was a different commit that just replaces the LiveKit's hook (`useLiveKitRoom()`) with the almost identical copy-paste of the body of this function. I don't understand why it was necessary, or to be more precise, why it was merged in the first place. I'm just trying to understand your thought process, i.e. why do you think that it was necessary to have this change? What's the purpose? One line of code (`useLiveKitRoom()`) was replaced with over 100 LOC of a code copied from the `useLiveKitRoom()` body (while also introducing a bug).

I sort of could understand if you merged it by mistake, that's why I created this PR and explained the reasons for that. But you disagree with the revert, i.e. you think that the change was necessary. I'm just trying to understand why you think so.

toger5:
superseeded: https://github.com/vector-im/element-call/pull/1120

PR #1120, opened 2023-06-16
LiveKit Device Usage Refactoring

daniel-abramov:
@EnricoSchw, this also superseded https://github.com/vector-im/element-call/pull/1115

(I still don't understand why and how it would be needed)

I've also refactored not using the `Room` component in the `GroupCallView`, so you might be interested in it.


src/room/VideoPreview.tsx
```diff
@@ -75,21 +60,76 @@ export function VideoPreview({
...
+
+      return {
+        selectedId: deviceId,
+        enabled,
+      };
+    };
+
+    onUserChoicesChanged({
+      video: createChoices(videoEnabled, activeVideoId),
+      audio: createChoices(audioEnabled, activeAudioId),
```
toger5, line 95: This has some latency to it right?
Choosing in the settings updates `mediaDevices` -> `mediaDevices.videoIn.selectedId` (and `mediaDevices.videoIn.selectedId` ) replace the track inside the usePreviewDevice hook. ->
Only when the new track is done initializing `activeVideoId` changes. -> the settings get updated.
not sure that is an issue.
But i think there is no other way to make the transfer of the permission request choice work...
Also it is unlikely some can start the call quick enough before the track is initialzied

daniel-abramov: If I got you right, do you mean the case where someone selects a media and then clicks on the button before choices are getting initialized? - The `InCallView` is only created when the choices are not `undefined`. I.e. it's only on the render which has the user choices set the `InCallView` will be set.

I can imagine that perhaps there is a case where the user does not accept the dialog and instead presses the join button, in which case the old choices will be used. But this seems to be an expected behavior (i.e. if you have not selected anything, then the previous selection will be used).

PR #1123, opened 2023-06-16
Hide connection quality indicators behind a developer setting

robintown:
Calls are an environment with high cognitive load, so it's important that we keep extra UI elements like these to a minimum and stick to what's been explicitly designed. I assume that this was here as a developer feature to diagnose reliability of the back end components, which is perfectly fine, so I've kept it behind a developer setting rather than fully removing it.

### DIFF
--- a/public/locales/en-GB/app.json
+++ b/public/locales/en-GB/app.json
@@ -86,6 +86,7 @@
   "Settings": "Settings",
   "Share screen": "Share screen",
   "Show call inspector": "Show call inspector",
+  "Show connection stats": "Show connection stats",
   "Sign in": "Sign in",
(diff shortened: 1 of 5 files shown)


fkwp:
would it be an idea to present that indicator in case of bad network connectivity?

robintown:
Potentially, but my hypothesis is that even then, these indicators are still of limited value, because users already intuitively understand that jittery video and audio concealment indicate network issues, and there's not a lot of manual action that users can take to resolve such issues, anyways. There are also cases where some users get a poor connection quality rating, but the call is still qualitatively running quite well.

Another unrelated issue is that the connection quality component from LiveKit lacks an accessible textual description.

PR #1134, opened 2023-06-23
Force to allow calls without video and audio in embedded mode

EnricoSchw:
This is related to: https://github.com/vector-im/voip-internal/issues/98 and need to be discuss!


robintown (approved):
Looks fine!

PR #1135, opened 2023-06-23
Fix device selection never completing in widget mode

robintown:
By reinstating some of the code from the full mesh branch that was there to handle this


SimonBrandner (approved):
Thanks!

PR #1144, opened 2023-06-26
Fix branch names in CI

dbkr:


.github/workflows/netlify-livekit.yaml
```diff
@@ -64,12 +64,11 @@ jobs:
         with:
           publish-dir: dist
           deploy-message: "Deploy from GitHub Actions"
-          production-branch: livekit-experiment
+          production-branch: livekit
           production-deploy: true
           # These don't work because we're in workflow_run
           enable-pull-request-comment: false
           enable-commit-comment: false
-          github-deployment-environment: livekit-experiment
```
dbkr, line 72: It doesn't look like there's anything in these, I think it's unnecessary - going to try without it.

EnricoSchw (approved):

PR #1166, opened 2023-06-28
Support for getting SFU config using OIDC

### LINKED ISSUE #1149: Pass OpenID token to JWT service
Use OIDC endpoint to verify homeserver/user exitsence (but not whether the user is in a given call) \[Note on OIDC endpoint. This one is handled by Synapse rather than mandating an OIDC auth provider\] Example given for embedded mode, but is similar with non-embedded.

* EC asks client for OIDC token ([https://github.com/matrix-org/matrix-widget-api/blob/master/src/interfaces/WidgetApiAction.ts\#L37](https://github.com/matrix-org/matrix-widget-api/blob/master/src/interfaces/WidgetApiAction.ts#L37))
* Client asks HS for OIDC token ([https://spec.matrix.org/v1.7/client-server-api/\#post\_matrixclientv3useruseridopenidrequest\_token](https://spec.matrix.org/v1.7/client-server-api/#post_matrixclientv3useruseridopenidrequest_token))
* HS returns token to client
* Client returns token to EC
* EC sends token to LK JWT service with JWT request
* JWT service contact the user's native homeserver with OIDC token ([https://spec.matrix.org/v1.7/server-server-api/\#openid](https://spec.matrix.org/v1.7/server-server-api/#openid))
* HS returns user info,JWT service verifies that the user really exists on that HS
* JWT service retruns JWT

Note the JWT Service also returns the SFU web socket URL

dbkr:
* Change `jwt_service_url` to `livekit_service_url`
 * Make it a POST so we can send the openID token sensibly
 * Get an OIDC token & pass it with the request
 * Read the SFU URL from there too

I made some changes to the auth server here before realising this wasn't it's actual home. I'm leaving the basic support here for now so the dev setup works, but I intend to remove this as soon as the backend service builds docker images we can just use from the dev script.

We'll need to update configs & the JWT service before merging this.

Fixes https://github.com/vector-im/element-call/issues/1149


src/room/GroupCallView.tsx
```diff
@@ -222,9 +222,10 @@ export function GroupCallView({
     return <ErrorView error={error} />;
   } else if (state === GroupCallState.Entered && userChoices) {
     return (
-      <ActiveCall
-        groupCall={groupCall}
+      <OpenIDLoader
```
toger5, line 233: This makes it hard to read because it is not immediately obvious that the active call is inside this wrapper.

src/livekit/OpenIDLoader.tsx
```diff
@@ -0,0 +1,55 @@
...
+        logger.error("Failed to fetch SFU config: ", e);
+        setError(new Error("Failed to fetch SFU config"));
+      }
+    })();
+  }, [client, roomName]);
+
+  if (error) {
+    return <ErrorView error={error} />;
+  } else if (sfuConfig) {
+    return <ActiveCall client={client} {...rest} sfuConfig={sfuConfig} />;
```
toger5: is there a reason for having `rest` in the middle compared to the end?

toger5 (approved):
Makes sense. I really like the whole approach we are taking here.

src/room/GroupCallView.tsx
dbkr: Yeah, possibly I could add the call as a child & have it set a param or put it in the context

dbkr: OK yeah, this is cleaner.

src/livekit/openIDSFU.ts
```diff
@@ -0,0 +1,58 @@
...
+  roomName: string
+): Promise<SFUConfig> {
+  const openIdToken = await client.getOpenIdToken();
+  logger.debug("Got openID token", openIdToken);
+
+  const livekitCfg = Config.get().livekit;
+
+  if (!livekitCfg?.livekit_service_url) {
+    throw new Error("No livekit service URL defined");
+  }
```
daniel-abramov: We probably want to move this part outside of the `getSFUConfigWithOpenID`, making the URL a mandatory parameter of a function. This way we ensure that the function can only be called with a "valid input parameters".

So far we've been using the `Config` as a global variable in different places. I think it would be easier if we read the config once (when the app is initialized) and then just pass the right values to the functions. I think for the sake of this PR it would be sufficient to just move this logic to the higher level component that calls the `getSFUConfigWithOpenID()` (similar to how it was before).

Rationale:
1. Reading/accessing the config file does not seem to belong to the responsibilities of the `getSFUConfigWithOpenID()` function. From my perspective, explicitly passing the configuration for such nice and clean 'library-looking' functions that have a single responsibility would be a better choice 🙂 
2. I imagine that if LiveKit is not configured, the branches of code that are responsible for the LiveKit initialization won't be called at all.

What do you think?

src/livekit/openIDSFU.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const res = await fetch(livekitCfg.livekit_service_url + "/sfu/get", {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({
+      room: roomName,
+      openid_token: openIdToken,
+      remove_me_user_id: client.getUserId(), // the service will get this from the openid request
```
daniel-abramov: It seems like we're not using this value atm. Should we pass it?

src/livekit/openIDSFU.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { Config } from "../config/Config";
+
+export interface SFUConfig {
+  url: string;
+  jwt: string;
+}
+
+export async function getSFUConfigWithOpenID(
+  client: MatrixClient,
```
daniel-abramov: It feels like adding `MatrixClient` is a quite large dependency for this function that we probably don't want to have as it adds too much power/responsibility for this function to handle 🙂 From what I can see, we only require 2 fields from the `client`. It would be safer if we could pass these fields directly.

Perhaps instead of `client`, we could pass a single small structure, something like `clientIdentity`, or, in case we only need a device ID, just a device ID would do.

src/livekit/openIDSFU.ts
```diff
@@ -0,0 +1,58 @@
...
+    body: JSON.stringify({
+      room: roomName,
+      openid_token: openIdToken,
+      remove_me_user_id: client.getUserId(), // the service will get this from the openid request
+      device_id: client.getDeviceId(),
+    }),
+  });
+  if (res.status / 100 !== 2) {
+    throw new Error("SFO Config fetch failed with status code " + res.status);
+  }
```
daniel-abramov: Perhaps we could describe it a bit more explicitly, i.e. check for the "good" status and return the config to the user and fail otherwise, `status / 100 != 2` looks like too many magic numbers 🙂 (it's understandable that we do it based on the numeric numbers for different statuses, but I think explicitly written expectations are better and easier to refactor in the future should we use something different than HTTP).

What do you think?

src/livekit/openIDSFU.ts
```diff
@@ -0,0 +1,58 @@
...
+  const sfuConfig = await res.json();
+
+  return sfuConfig;
```
daniel-abramov: Nit: perhaps would be a bit more compact with `return await res.json();`, but it's a matter of preference, so not a big deal (feel free to resolve and ignore if you disagree 🙂).

daniel-abramov (commented):
Thanks @dbkr again for making the PR!

I've a left a couple of comments.

daniel-abramov:
@dbkr, it seems like we may need to remove the SFU part from this PR as it's already part of the https://github.com/vector-im/lk-jwt-service/pull/1

src/room/GroupCallView.tsx
toger5: thanks.

toger5 (approved):

src/livekit/openIDSFU.ts
dbkr: Done

src/livekit/openIDSFU.ts
dbkr: This is fairly standard in an HTTP client as HTTP defines anything in the 200-300 range as successful - it doesn't matter what the actual code is if you only care whether it worked or not. Checking for just one status would mean the client would break if the server changed to return a different 2xx code, which is perfectly valid. Would status >= 200 && status < 300 be any clearer? Or just a comment?

daniel-abramov: Right, I understand that [a range from 200-299 is a successful response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses), however, what I meant is that it's getting uncommon nowadays to rely on numeric client codes *in a high-level application code* (for the sake of the readability, being explicit / not needing to memorize the codes, examples in libraries: [one](https://www.npmjs.com/package/http-status-codes), [two](https://docs.rs/http/latest/http/status/struct.StatusCode.html#method.is_success)).

In this particular case, I think it would be more reliable (and arguably more readable) if we instead used the [`ok` property of a `Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok).

Also,

```typescript
if (res.status / 100 !== 2) {
    throw new Error("SFO Config fetch failed with status code " + res.status);
  }
```

will throw an exception in cases that we don't really expect since if the `status` is e.g. `202`, then `202 / 100` for a `number` will not be `2`, which will throw an exception.

src/livekit/openIDSFU.ts
dbkr: Yeah, seems fair enough, I was mostly just copying the way it was done before. Done.

src/livekit/openIDSFU.ts
dbkr: Oh, sure, we can use response.ok instead - done.

src/livekit/openIDSFU.ts
dbkr: Done

src/livekit/openIDSFU.ts
daniel-abramov: Thanks! I think that we actually probably don't even need to use `Pick<>`, because it mentions `MatrixClient` (as if we're saying that we need only parts of the `MatrixClient`, whereas ideally the component does not really even need to know about the `MatrixClient`), perhaps a better choice would be to define an `interface` that has 2 required properties. This way neither `OpenIDLoader`, nor `getSFUConfigWithOpenID()` require to mention and import a `MatrixClient`: strictly speaking we don't really want to have a subset of a `MatrixClient`, we just need some type `T`, such that satisfies a given `interface` (2 fields), that way the code remains more flexible and also less "external dependencies" 🙂 What do you think?

src/livekit/openIDSFU.ts
dbkr: I thought I replied to this but apparently not. Either way, yes, seems reasonable - done.

src/livekit/openIDSFU.ts
dbkr: I was reluctant to start duplicating properties of matrixClient, but yeah, you could also do it that way, I don't really mind particularly.

src/livekit/OpenIDLoader.tsx
```diff
@@ -0,0 +1,80 @@
...
+  } else if (sfuConfig) {
+    return (
+      <SFUConfigContext.Provider value={sfuConfig}>
+        {children}
+      </SFUConfigContext.Provider>
+    );
+  } else {
+    return <LoadingView />;
+  }
+}
```
daniel-abramov: Cool! A small hint: it seems like we have 3 distinct states here: loading that then resolves into **either** loaded **or** failed (it's not possible to have both `error` and `sfuConfig` set). Also, the current state won't really compile with strict lints since the actual type of each state is `SFUConfig | undefined` and `Error | undefined` respectively.

It seems to be a good idea to model the state explicitly (stronger types, no possibility to have mutually exclusive states) like this:

```suggestion
  const [state, setState] = useState<
    SFUConfigLoading | SFUConfigLoaded | SFUConfigFailed
  >({ kind: "loading" });

  useEffect(() => {
    (async () => {
      try {
        const result = await getSFUConfigWithOpenID(
          client,
          livekitServiceURL,
          roomName
        );
        setState({ kind: "loaded", sfuConfig: result });
      } catch (e) {
        logger.error("Failed to fetch SFU config: ", e);
        setState({ kind: "failed", error: e });
      }
    })();
  }, [client, livekitServiceURL, roomName]);

  switch (state.kind) {
    case "loading":
      return <LoadingView />;
    case "failed":
      return <ErrorView error={state.error} />;
    case "loaded":
      return (
        <SFUConfigContext.Provider value={state.sfuConfig}>
          {children}
        </SFUConfigContext.Provider>
      );
  }
}

type SFUConfigLoading = {
  kind: "loading";
};

type SFUConfigLoaded = {
  kind: "loaded";
  sfuConfig: SFUConfig;
};

type SFUConfigFailed = {
  kind: "failed";
  error: Error;
};
```

It prevents all kinds of problems (especially if the code gets larger, and has a more complicated state) and is also a bit more elegant 🙂

dbkr: oh, so have them as a union type: yeah, that works.

daniel-abramov (approved):
Generally, it looks ok, thanks for addressing the feedback! A quick summary for transparency to not lose track of open threads:

1. [As mentioned in one of the comments](https://github.com/vector-im/element-call/pull/1166#issuecomment-1618071294), we probably should remove the `server.go` from the repository. It's ok if we do it in a separate PR! However, we need to at least remove the changes that have been done in this PR so that the diff for the part that we don't use (`server.go`) is not affected (it has been superseded by the PR in the JWT service repository). 
2. I'm not really sure if we want/need to have an `OpenIDLoader` component that creates a React context that could be used by the children. The context is [generally discouraged](https://react.dev/learn/passing-data-deeply-with-context#before-you-use-context) and is only useful in rather rare scenarios when a certain value is required throughout a large component tree (such as a configuration that many components want to use or something similar). In our case, we only need the `jwt + sfuUrl` variables in a single place, so it seems like an overkill to use a context there given that we don't really intend to use these from other components in the future. To me, it feels like what we want to have instead **is a hook**, something like `useSfuJwtConfig()` that would return the aforementioned union type representing the current state config. This would allow us to only create an `ActiveCall` when the config is resolved and properly set. Currently, since we use a context, [this line](https://github.com/vector-im/element-call/pull/1166/files#diff-20eaa30bd23102b20e1a663910bafe1f71ce4db711fbcb08d21019e29fe63457R99) may return `undefined` which we don't even check in our code (the line that uses the config would fail if strict lints were enabled). And we can't work around it by checking the `sfuConfig` and calling the `useLiveKit()` conditionally as we can't/shouldn't call hooks conditionally (React design/peculiarity), which means that in the current code, the fix would need to make `useLiveKit()` to accept an optional `sfuConfig?` and return early if it's not passed. While it's not bad on its own, we could have avoided this issue by structuring the architecture a bit differently.
3. Also I think we should [follow the principle of the least privilege](https://github.com/vector-im/element-call/pull/1166#discussion_r1251130375) when defining components/hooks and their inputs and outputs, so they look "as pure as possible" and expose as little as possible information to each component, i.e. we must not import or use big global types like `MatrixClient` (or other similar things) from the components/functions that don't really need to know much about them.

@dbkr, I think in order to merge this PR, it would be ok to just solve (1) for now and squash the commits (I don't want to bother/overwhelm too much with my reviews 🙂), so that then someone else can work on (2) and (3) in the upcoming PRs (when time is available, etc) if we agree that they make sense.

dbkr:
>  1. [As mentioned in one of the comments](https://github.com/vector-im/element-call/pull/1166#issuecomment-1618071294), we probably should remove the `server.go` from the repository...

I left the changes in so that the dev setup still worked, but I now have the code ready to have the docker compose just pull down the built image from the container registry rather than having to copy the code into here. Just let me know what you'd rather do.

> 2. I'm not really sure if we want/need to have an `OpenIDLoader` component that creates a React context that could be used by the children...

This was proving to be problematic for reasons I can't remember specifically but due to the fact that hooks can't be rendered conditionally. It's possible we might be able to it this way now though. Generally this kind of feedback is better to provide earlier rather than later.

> 3. Also I think we should [follow the principle of the least privilege](https://github.com/vector-im/element-call/pull/1166#discussion_r1251130375) when defining components/hooks and their inputs and outputs...

Could you reply to https://github.com/vector-im/element-call/pull/1166#discussion_r1251287787 please? If you have a strong preference, that's fine.

daniel-abramov:
> I left the changes in so that the dev setup still worked, but I now have the code ready to have the docker compose just pull down the built image from the container registry rather than having to copy the code into here. Just let me know what you'd rather do.

Ideally, I think we may need just a single `docker-compose` with the jwt service image pointing to the GitHub registry (I believe that's what you are using locally at the moment if I got you right). For the sake of this PR (to not increase its scope), I think it would be fine to just revert the changes that changed the `server` part (given that we don't really want to use it, we want to use the service from the JWT token repository). I'm not really sure why the code of the backend was copy-pasted into the frontend repository, but judging from Git history, Enrico tried to make a quick way to just test things locally from a single folder or something.

> This was proving to be problematic for reasons I can't remember specifically but due to the fact that hooks can't be rendered conditionally. It's possible we might be able to it this way now though.

We can't call hooks conditionally, but we can create components conditionally. I.e. regardless of the context, we would need to somehow handle the `undefined` value, oftentimes it either means that the other hook that needs this value receives an optional parameter or that the component that calls the hook that needs the value is rendered conditionally, such as:

```typescript
const sfuConfig = useJwtSfuConfig();
switch (sfuConfig.kind) {
    case "loaded":
        return (<ActiveCallView sfuConfig={sfuConfig});
...
}
```

Currently, it's almost the same except that the `sfuConfig` is passed via the context which means that the `sfuConfig` may be `undefined` (current code does not handle this case, but the strict lint would complain about it). But nothing that would block merging this PR (I can submit a small improvement PR later if that's fine).

> Could you reply to https://github.com/vector-im/element-call/pull/1166#discussion_r1251287787 please? If you have a strong preference, that's fine.

👍

src/livekit/openIDSFU.ts
daniel-abramov: So my main concern about passing the `MatrixClient` around is that it's rather a large and powerful object, most components don't really need to know about what `MatrixClient` is, so instead of passing a large object, the components should have simple interfaces/types/structures that they expect as an input. The code would then become less coupled - the `getSFUConfigWithOpenID()` would work fine even without a matrix client, i.e. it would work with any type `T` that can provide a device ID and an open ID token.

Such things don't look particularly bad at first sight, but I decided to mention it since it's easier to explain with simpler examples. It gets much more complicated when components take many parameters, each of which relies on some large data types even if they intend to use 2-3 properties of it, but because of the type it also forces their parent components and callers to pass the large type around. As mentioned I don't want to block the merge of this PR, so I've already approved it with the current changes, but thought that it may make sense to clarify some points for the future and upcoming PRs 🙂

PR #1167, opened 2023-06-28
Remove avatar from header

### LINKED ISSUE #565: Avatars look broken on call.element.io

### LINKED ISSUE #1152: Remove room avatars
Design feedback - can we remove the room avatar within a call. There is no easy way to set it, and it adds clutter to the layout.

SimonBrandner:
Fixes https://github.com/vector-im/element-call/issues/565
Fixes https://github.com/vector-im/element-call/issues/1152

### DIFF
--- a/src/Header.tsx
+++ b/src/Header.tsx
@@ -24,9 +24,7 @@ import styles from "./Header.module.css";
 import { useModalTriggerState } from "./Modal";
 import { Button } from "./button";
 import { ReactComponent as Logo } from "./icons/Logo.svg";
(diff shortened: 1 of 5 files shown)


robintown (commented):
This seems fine, but we could also clean up a little more by removing the room avatar field from `MatrixInfo`

SimonBrandner:
> This seems fine, but we could also clean up a little more by removing the room avatar field from `MatrixInfo`

Good point!

robintown (approved):

PR #1173, opened 2023-06-30
Enable stricter lints

### LINKED ISSUE #1132: Enable strict lints and checks
1. We must enable as maximum "compile time" checks as possible to catch potential problems early (not on runtime).
2. We also must forbid merging when there are failed links or formatting issues.
3. Ideally we want to have a pre-push hook that checks the lints, so that the author of the PR knows about issues in advance (before pushing the branch or creating a PR). This would avoid not particularly useful "fix formatting", "fix lints" and other commits.

daniel-abramov:
This is an attempt to fix https://github.com/vector-im/element-call/issues/1132

### Rationale

The main idea is that after this PR we'll have a decent set of TypeScript strictness and we'll forbid the `@ts-ignore`, so that we know that our code is sound [type-wise].

### Important Notes

1. I haven't enabled the strictest lints possible for the reasons discussed in a team meeting (i.e. too many errors and incompatibilities).
2. I deliberately used `@ts-ignore` in certain parts where the fix of the issue looked untrivial or where I'm not familiar with fixing it (feel free to submit a follow-up or push a change to the branch if you know a quick and easy way to fix it). This is primarily seen in `VideoGrid`, `NewVideoGrid`, `rageshake.ts`, `GroupCallInspector`. I also did not want to touch the video grid logic much since @robintown works on a video grid.
3. In many cases I had to use `!` to assert the type in cases where it looked like we were expecting the actual type.
4. I have refactored the `ClientContext` and tried to keep the compatibility layer for the components that expected a different "plain" representation of a client state.
5. I [intentionally] did not fix some "logical errors" (weirdness and unnecessary complications of code, i.e. accidental complexity) in some parts as that would result in a massive PR with possibly more regressions, this PR is the first step to enable clients and make sure that we can start building on top of a stricter type system.
6. ~~I have not yet performed any smoke tests, so things may fail, especially if some assumptions (e.g. those mentioned in (3)) were wrong.~~


src/Avatar.tsx
```diff
@@ -99,10 +99,10 @@ export const Avatar: React.FC<Props> = ({
     [size]
   );
 
-  const resolvedSrc = useMemo(
-    () => resolveAvatarSrc(client, src, sizePx),
-    [client, src, sizePx]
-  );
+  const resolvedSrc = useMemo(() => {
+    if (!client || !src || !sizePx) return undefined;
+    return resolveAvatarSrc(client, src, sizePx);
```
robintown: `resolveAvatarSrc` also checks for whether the client is undefined, so you can remove that redundant check above now

src/UserMenuContainer.tsx
```diff
@@ -49,7 +48,9 @@ export function UserMenuContainer({ preventNavigation = false }: Props) {
           modalState.open();
           break;
         case "logout":
-          logout();
+          if (logout) {
+            logout();
```
robintown: For future reference you can write this more concisely as `logout?.()`

src/home/useGroupCallRooms.ts
```diff
@@ -79,25 +79,30 @@ function sortRooms(client: MatrixClient, rooms: Room[]): Room[] {
...
 
         return {
           roomId: room.getCanonicalAlias() || room.roomId,
           roomName: room.name,
-          avatarUrl: room.getMxcAvatarUrl(),
+          avatarUrl: room.getMxcAvatarUrl()!,
           room,
           groupCall,
-          participants: [...groupCall.participants],
+          participants: [...groupCall!.participants.keys()],
```
robintown, line 102: Oh nice, I never guessed that we actually implemented showing the participants like this, but then broke the feature by messing up the types

src/profile/useProfile.ts
```diff
@@ -38,41 +38,40 @@ type ProfileSaveCallback = ({
   removeAvatar: boolean;
 }) => Promise<void>;
 
-export function useProfile(client: MatrixClient) {
-  const [{ loading, displayName, avatarUrl, error, success }, setState] =
+export function useProfile(client?: MatrixClient) {
```
robintown:
```suggestion
export function useProfile(client: MatrixClient | undefined) {
```

Let's do this so the caller can't completely forget to pass something for the `client` argument

src/room/GridLayoutMenu.tsx
```diff
@@ -46,7 +46,12 @@ export function GridLayoutMenu({ layout, setLayout }: Props) {
         </Button>
       </TooltipTrigger>
       {(props: JSX.IntrinsicAttributes) => (
-        <Menu {...props} label={t("Grid layout menu")} onAction={setLayout}>
+        <Menu
+          {...props}
+          label={t("Grid layout menu")}
+          onAction={(key) => setLayout(key.toString() as Layout)}
+          onClose={() => {}}
```
robintown: Could we follow the established pattern here of wrapping all callbacks in `useCallback` as a general rule?

src/room/GroupCallLoader.tsx
```diff
@@ -60,5 +60,5 @@ export function GroupCallLoader({
     return <ErrorView error={error} />;
   }
 
-  return <>{children(groupCall)}</>;
+  return groupCall ? <>{children(groupCall)}</> : <></>;
```
robintown: I believe it's safe to use an assertion `groupCall!` here since both `loading` and `error` are checked above

src/room/GroupCallView.tsx
```diff
@@ -141,14 +141,14 @@ export function GroupCallView({
         PosthogAnalytics.instance.eventCallStarted.track(groupCall.groupCallId);
 
         await Promise.all([
-          widget.api.setAlwaysOnScreen(true),
-          widget.api.transport.reply(ev.detail, {}),
+          widget?.api.setAlwaysOnScreen(true),
+          widget?.api.transport.reply(ev.detail, {}),
```
robintown:
```suggestion
          widget!.api.setAlwaysOnScreen(true),
          widget!.api.transport.reply(ev.detail, {}),
```

due to the `if (widget)` above

src/room/GroupCallView.tsx
```diff
@@ -141,14 +141,14 @@ export function GroupCallView({
...
         ]);
       };
 
       widget.lazyActions.on(ElementWidgetActions.JoinCall, onJoin);
       return () => {
-        widget.lazyActions.off(ElementWidgetActions.JoinCall, onJoin);
+        widget?.lazyActions.off(ElementWidgetActions.JoinCall, onJoin);
```
robintown:
```suggestion
        widget!.lazyActions.off(ElementWidgetActions.JoinCall, onJoin);
```

src/room/GroupCallView.tsx
```diff
@@ -207,12 +207,12 @@ export function GroupCallView({
     if (widget && state === GroupCallState.Entered) {
       const onHangup = async (ev: CustomEvent<IWidgetApiRequest>) => {
         leave();
-        await widget.api.transport.reply(ev.detail, {});
-        widget.api.setAlwaysOnScreen(false);
+        await widget?.api.transport.reply(ev.detail, {});
+        widget?.api.setAlwaysOnScreen(false);
```
robintown:
```suggestion
        await widget!.api.transport.reply(ev.detail, {});
        widget!.api.setAlwaysOnScreen(false);
```

src/room/GroupCallView.tsx
```diff
@@ -207,12 +207,12 @@ export function GroupCallView({
...
       };
       widget.lazyActions.once(ElementWidgetActions.HangupCall, onHangup);
       return () => {
-        widget.lazyActions.off(ElementWidgetActions.HangupCall, onHangup);
+        widget?.lazyActions.off(ElementWidgetActions.HangupCall, onHangup);
```
robintown:
```suggestion
        widget!.lazyActions.off(ElementWidgetActions.HangupCall, onHangup);
```

src/room/InCallView.tsx
```diff
@@ -200,11 +203,11 @@ export function InCallView({
     if (widget) {
       const onTileLayout = async (ev: CustomEvent<IWidgetApiRequest>) => {
         setLayout("freedom");
-        await widget.api.transport.reply(ev.detail, {});
+        await widget?.api.transport.reply(ev.detail, {});
```
```suggestion
        await widget!.api.transport.reply(ev.detail, {});
```

src/room/InCallView.tsx
```diff
@@ -200,11 +203,11 @@ export function InCallView({
...
       };
       const onSpotlightLayout = async (ev: CustomEvent<IWidgetApiRequest>) => {
         setLayout("spotlight");
-        await widget.api.transport.reply(ev.detail, {});
+        await widget?.api.transport.reply(ev.detail, {});
```
```suggestion
        await widget!.api.transport.reply(ev.detail, {});
```

src/room/InCallView.tsx
```diff
@@ -214,8 +217,8 @@ export function InCallView({
       );
 
       return () => {
-        widget.lazyActions.off(ElementWidgetActions.TileLayout, onTileLayout);
-        widget.lazyActions.off(
+        widget?.lazyActions.off(ElementWidgetActions.TileLayout, onTileLayout);
+        widget?.lazyActions.off(
```
```suggestion
        widget!.lazyActions.off(ElementWidgetActions.TileLayout, onTileLayout);
        widget!.lazyActions.off(
```

src/home/CallTypeDropdown.tsx
```diff
@@ -52,7 +52,15 @@ export const CallTypeDropdown: FC<Props> = ({ callType, setCallType }) => {
         </Headline>
       </Button>
       {(props: JSX.IntrinsicAttributes) => (
-        <Menu {...props} label={t("Call type menu")} onAction={setCallType}>
+        <Menu
+          {...props}
+          label={t("Call type menu")}
+          onAction={(key) => {
```
robintown: this could also use `useCallback`

src/button/LinkButton.tsx
```diff
@@ -46,7 +46,7 @@ export function LinkButton({
     <Link
       className={classNames(
         variantToClassName[variant || "secondary"],
-        sizeToClassName[size],
+        size ? sizeToClassName.lg : [],
```
robintown: As long as we're using `sizeToClassName` to prepare for the possibility of there being more than one button size, we should probably not hardcode `lg` here

src/UserMenuContainer.tsx
```diff
@@ -59,19 +60,22 @@ export function UserMenuContainer({ preventNavigation = false }: Props) {
...
-        isAuthenticated={isAuthenticated}
-        isPasswordlessUser={isPasswordlessUser}
-        avatarUrl={avatarUrl}
-        onAction={onAction}
-        displayName={
-          displayName || (userName ? userName.replace("@", "") : undefined)
-        }
-      />
-      {modalState.isOpen && (
+      {avatarUrl && (
```
robintown: Hiding the menu if the user doesn't have an avatar set seems wrong

src/UserMenu.tsx
```diff
@@ -119,7 +119,7 @@ export function UserMenu({
           )}
         </Button>
       </TooltipTrigger>
-      {(props) => (
+      {(props: any) => (
```
robintown: It's probably fine to leave this for a future PR to avoid making this PR too big, but we should consider changing `PopoverMenuTrigger` to take this function as a normal prop rather than as a child, so it's easier to write types for

src/ClientContext.tsx
```diff
@@ -299,61 +249,141 @@ export const ClientProvider: FC<Props> = ({ children }) => {
...
-      isPasswordlessUser,
-      client,
-      changePassword,
-      logout,
-      userName,
-      setClient,
-    ]
-  );
+  const [state, setState] = useState<ClientState | undefined>(undefined);
+  React.useEffect(() => {
```
robintown: This combination of `useState` + `useEffect` could be replaced with a single `useMemo` hook to achieve the same thing with fewer renders

robintown (commented):
Thanks a lot, this will improve the developer experience of working on Element Call significantly! I'm not super happy about all the `@ts-ignore` comments we have now, but I guess working iteratively to improve our types is a good strategy.

daniel-abramov:
> Thanks a lot, this will improve the developer experience of working on Element Call significantly! I'm not super happy about all the `@ts-ignore` comments we have now, but I guess working iteratively to improve our types is a good strategy.

**Fully agree.** AFAIK one of the strategies that some teams use to approach a component that has too many 'violations' of the strictness, is to suppress all of the warnings/errors with a `@ts-ignore` at the first stage and introduce a rule "no new `@ts-ignore` anymore", so that at least the new code benefits from a stricter set of rules, while at the same time consider removing the existing `@ts-ignore` when working on features/further improvements.

I decided to go a step further and fix those that looked easy to set, venturing into `ClientContext` refactoring for a better way to fix them. Still, it looked like it takes too much time, so I thought it would be better to suppress some of them that I can't fix quickly or don't have enough context to fix.

src/Avatar.tsx
daniel-abramov: Ah, but the type on the `resovleAvatarSrc` was not optional (though it used optional checks inside the function 🙂), so the type error was triggered and I did not check how the function looks. Now, after looking, I realized that the function could be inlined since everything is checked. Thanks for pointing this out! Fixed.

src/UserMenuContainer.tsx
daniel-abramov: Thank you! My bad, was probably too lazy to check how it must be written. Fixed.

src/profile/useProfile.ts
daniel-abramov: That makes sense, agree! Fixed.

src/room/GridLayoutMenu.tsx
daniel-abramov: Good catch, fixed!

OOC: Do you by chance know if there is a lint for that or something similar so that such things could be caught in advance? I'm always forgetting about these sorts of things (not a frontend/React developer 🙂).

src/home/CallTypeDropdown.tsx
daniel-abramov: Agreed, done!

src/button/LinkButton.tsx
daniel-abramov: Agree, fixed.

src/UserMenuContainer.tsx
daniel-abramov: Good point, fixed!

src/UserMenu.tsx
daniel-abramov: I'm not an expert, but sounds sensible 🙂 I'm definitely not a fan of `any` and weaker types.

src/ClientContext.tsx
daniel-abramov: My poor React skills, fixed 🙂 It looks much better with `useMemo()`, thanks!

daniel-abramov:
**UPD.** Rebased against `livekit` to resolve conflicts.

src/room/GroupCallLoader.tsx
daniel-abramov: Ah, so the fields in the structure were "either/or`? - Ok, now I see, checked the source code of the hook. Thanks for pointing this out!

After checking the internals of a function, I could not stand an implicit (not properly modeled) state, so I decided to refactor the hook to make it more explicit and prevent impossible states. The whole thing [looks much more elegant now IMHO](https://github.com/vector-im/element-call/pull/1173/commits/03ba4ac45ae061c672fab7062f99efa9d7aeedae#diff-25657a65748b36f55c850afaa4a56aab67db55013fe52e3eb3ecda1a1fe2cb56R48) and also strongly typed. Hope you like it too 🙂

src/room/GroupCallView.tsx
daniel-abramov: Just to double-check: we guarantee that the `widget` cannot be modified otherwise, right?

The reason I'm asking is that at this place the `widget` is used from the callback that is defined inside the `if`, not from the `if` block by itself, hence the warning from a transpiler, i.e. by the point the callback is called, the widget may become `null` again. The `on()` action that subscribes to the event does not trigger this warning since it's called directly inside an `if` (i.e. guaranteed to be non-null).

src/room/GroupCallView.tsx
daniel-abramov: (pending on https://github.com/vector-im/element-call/pull/1173#discussion_r1251176236)

src/room/GroupCallView.tsx
daniel-abramov: (pending on https://github.com/vector-im/element-call/pull/1173#discussion_r1251176236)

src/room/GroupCallView.tsx
daniel-abramov: (pending on https://github.com/vector-im/element-call/pull/1173#discussion_r1251176236)

src/room/InCallView.tsx
daniel-abramov: (pending on https://github.com/vector-im/element-call/pull/1173#discussion_r1251176236)

src/room/InCallView.tsx
daniel-abramov: (pending on https://github.com/vector-im/element-call/pull/1173#discussion_r1251176236)

src/room/InCallView.tsx
daniel-abramov: (pending on https://github.com/vector-im/element-call/pull/1173#discussion_r1251176236)

src/room/GridLayoutMenu.tsx
robintown: I don't think so, sadly… but you're very right that it's easy to forget these things. Usually it's not a big deal to leave things unmemoized, but Meta is also experimenting with an auto-memoizing compiler that could improve the developer experience here.

src/room/GroupCallLoader.tsx
robintown: Fantastic :+1:

src/room/GroupCallView.tsx
robintown: Correct, `widget` can't be modified because it's a `const`.

I think the reason that TypeScript doesn't cast it to a non-null value here is because you're supposed to be able to replace a declaration like `export const widget` with `export let widget` without that being a breaking change in your module's API, and so the type checker has to pretend that `widget` could change to `null` by the time the function is invoked.

daniel-abramov: I see, thanks for explaining! Fixed.

daniel-abramov:
**UPD.** So it seems like now all the problems are solved, so the only thing that is left that keeps us from merging (apart from a smoke test that I'll do) is the errors that I mentioned in the description of the issue ([they also fail in the pipeline](https://github.com/vector-im/element-call/actions/runs/5455178543/jobs/9926266319?pr=1173)) - namely the errors that originate from the `matrix-js-sdk`.

The `exclude` pattern in `tsconfig` does not seem to do anything in this case as `matrix-js-sdk` is directly imported and so it looks like it can't be disabled via `exclude` (we have `node_modules` in the `exclude`). The `skipLibCheck: false` option does not seem to do anything particularly useful in this case either.

package.json
```diff
@@ -9,8 +9,9 @@
     "build-storybook": "build-storybook",
     "prettier:check": "prettier -c .",
     "prettier:format": "prettier -w .",
-    "lint": "yarn lint:types && yarn lint:js",
-    "lint:js": "eslint --max-warnings 0 src",
+    "lint": "yarn lint:types && yarn lint:eslint",
+    "lint:eslint": "eslint --max-warnings 0 src",
```
dbkr, line 13: Why the rename from lint:js to lint:eslint?

daniel-abramov: Oh, I think I did it while comparing `eslint` with `tsc` and thinking that perhaps calling it `lint:eslint` would be slightly more explicit since it only calls `eslint` (`element-web`, `matrix-js-sdk` contain a `lint:js` command that consists of running multiple subcommands, both calling `eslint` though).

But I don't have any strong preference here, so I could revert change if you prefer 🙂

robintown (approved):
Looks good once the errors are fixed.

For the first one, I see that matrix-js-sdk has the file https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/%40types/another-json.d.ts which is supposed to suppress that error, so I think you can fix this by including that file in our TypeScript config like matrix-react-sdk does: https://github.com/matrix-org/matrix-react-sdk/blob/2a7780052e60f23a8a7a90d549f27dea471e5ccd/tsconfig.json#L22

Not sure about the second error though.

package.json
robintown: FWIW I like the more explicit name

dbkr: Fair enough, seems reasonable, as long as we're consistent between projects.

dbkr:
https://github.com/matrix-org/matrix-rust-sdk/blob/a0f40606d4e379b395b524b3c86d2abc261bf6a8/bindings/matrix-sdk-crypto-js/CHANGELOG.md?plain=1#L34 suggests they've just been added recently, so maybe we need to update js-sdk to pull in a newer copy of the crypto lib.

daniel-abramov:
> Looks good once the errors are fixed.
> 
> For the first one, I see that matrix-js-sdk has the file https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/%40types/another-json.d.ts which is supposed to suppress that error, so I think you can fix this by including that file in our TypeScript config like matrix-react-sdk does: https://github.com/matrix-org/matrix-react-sdk/blob/2a7780052e60f23a8a7a90d549f27dea471e5ccd/tsconfig.json#L22
> 
> Not sure about the second error though.

That did the trick, awesome, thanks!

daniel-abramov:
> https://github.com/matrix-org/matrix-rust-sdk/blob/a0f40606d4e379b395b524b3c86d2abc261bf6a8/bindings/matrix-sdk-crypto-js/CHANGELOG.md?plain=1#L34 suggests they've just been added recently, so maybe we need to update js-sdk to pull in a newer copy of the crypto lib.

You're right! I ran `yarn lint` in the `matrix-js-sdk` and it exhibited the same errors, but once updating the rust crypto SDK in the `package.json`, `yarn lint` succeeds!

Sadly, I can't push to the `livekit` branch, it seems like I lack permissions. Could someone please apply the patch or grant me permissions to push to the branches? 🙂

<details>
The patch is as simple as:

```diff
Author: Daniel Abramov <daniel.abramov@element.io>
Date:   Thu Jul 6 11:15:43 2023 +0100

    Fix linting errors caused by the old crypto SDK

    See https://github.com/vector-im/element-call/pull/1173#issuecomment-1621306979
    for more details!

diff --git a/package.json b/package.json
index 8ffe173ea..470fc65f1 100644
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     ],
     "dependencies": {
         "@babel/runtime": "^7.12.5",
-        "@matrix-org/matrix-sdk-crypto-js": "^0.1.0-alpha.10",
+        "@matrix-org/matrix-sdk-crypto-js": "^0.1.0",
         "another-json": "^0.2.0",
         "bs58": "^5.0.0",
         "content-type": "^1.0.4",
diff --git a/yarn.lock b/yarn.lock
index dd15a07b0..1d0622de4 100644
--- a/yarn.lock
+++ b/yarn.lock
@@ -1426,10 +1426,10 @@
   dependencies:
     lodash "^4.17.21"

-"@matrix-org/matrix-sdk-crypto-js@^0.1.0-alpha.10":
-  version "0.1.0-alpha.10"
-  resolved "https://registry.yarnpkg.com/@matrix-org/matrix-sdk-crypto-js/-/matrix-sdk-crypto-js-0.1.0-alpha.10.tgz#b6a6395cffd3197ae2e0a88f4eeae8b315571fd2"
-  integrity sha512-8V2NKuzGOFzEZeZVgF2is7gmuopdRbMZ064tzPDE0vN34iX6s3O8A4oxIT7SA3qtymwm3t1yEvTnT+0gfbmh4g==
+"@matrix-org/matrix-sdk-crypto-js@^0.1.0":
+  version "0.1.0"
+  resolved "https://registry.yarnpkg.com/@matrix-org/matrix-sdk-crypto-js/-/matrix-sdk-crypto-js-0.1.0.tgz#766580036d4df12120ded223e13b5640e77db136"
+  integrity sha512-ra/bcFdleC1iRNms2I96UXA0NvQYWpMsHrV5EfJRS7qV1PtnQNvgsvMfjMbkx8QT2ErEmIhsvB5fPCpfp8BSuw==

 "@matrix-org/olm@https://gitlab.matrix.org/api/v4/projects/27/packages/npm/@matrix-org/olm/-/@matrix-org/olm-3.2.14.tgz":
   version "3.2.14"
```

</details>

dbkr:
Can you make a PR to the livekit branch as normal?

daniel-abramov:
> Can you make a PR to the livekit branch as normal?

I tried, but it did not work out. Apparently, I don't have any push permissions.

daniel-abramov:
After [merging `develop` into `livekit` in the Matrix JS SDK](https://github.com/matrix-org/matrix-js-sdk/pull/3573) the original errors disappeared, but now it has a new portion of errors that is similar to the `another-json` errors were resolved following a suggestion from @robintown by including the file from the Matrix JS SDK that contained the `declare module` part.

Now we have the same error but for `uuid`. Should we add it in a similar way to the Matrix JS SDK? (I'm not really sure why it happens at the first place)

cc @toger5

dbkr:
I've submitted https://github.com/matrix-org/matrix-js-sdk/pull/3593 to make these types packages regular dependencies in the js-sdk which I think is the right solution here.

PR #1174, opened 2023-06-30
Fix big grid crashing due to missing React import

robintown:
by fixing the cause rather than the symptom: this upgrades the code to use the [new, recommended JSX transform mode of React 17+](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html), which no longer requires you to import React manually just to write JSX.


SimonBrandner (approved):
Looks sane

PR #1175, opened 2023-07-03
Fix new grid layout

SimonBrandner:


dbkr (approved):

src/room/InCallView.tsx
```diff
@@ -245,7 +249,8 @@ export function InCallView({
     [fullscreenItem, noControls, items]
   );
 
-  const Grid = NewVideoGrid;
+  const Grid =
+    items.length > 12 && layout === "freedom" ? NewVideoGrid : VideoGrid;
```
toger5, line 253: Do we ever show Grid if we are not in freeform layout? Is it really required to check for `layout === "freeform"`

toger5 (approved):

PR #1177, opened 2023-07-03
Maintain friendly URL in browser URL bar

### LINKED ISSUE #1153: Maintain friendly URL in browser URL bar
Design feedback - can we keep the call.element.io/foo URL displaying in the URL bar rather than switch to the full room name? It avoids confusion, especially when people send each other links.

SimonBrandner:
Fixes https://github.com/vector-im/element-call/issues/1153


src/App.tsx
```diff
@@ -68,14 +67,11 @@ export default function App({ history }: AppProps) {
                     <SentryRoute exact path="/register">
                       <RegisterPage />
                     </SentryRoute>
-                    <SentryRoute path="/room/:roomId?">
```
robintown, line 74: Removing this route will break widget mode, as Element Web relies on setting the path to `/room` to get to the room page. Unless you have any better ideas, I think the best way to handle this would be to keep it here as `<SentryRoute path="/room">`? for when you only have a room ID

src/matrix-utils.ts
```diff
@@ -346,15 +346,11 @@ export async function createRoom(
...
-      return `${window.location.protocol}//${window.location.host}/room/${roomIdOrAlias}`;
-    } else {
-      return `${window.location.protocol}//${window.location.host}/${localPart}`;
-    }
+    return `${window.location.protocol}//${window.location.host}/${
+      roomIdOrAlias.substring(1).split(":")[0]
+    }`;
   } else {
-    return `${window.location.protocol}//${window.location.host}/room/#?roomId=${roomIdOrAlias}`;
+    return `${window.location.protocol}//${window.location.host}?roomId=${roomIdOrAlias}`;
```
robintown: Here is another place we should keep the `/room` route

src/home/CallList.tsx
```diff
@@ -74,7 +74,10 @@ function CallTile({
 }: CallTileProps) {
   return (
     <div className={styles.callTile}>
-      <Link to={`/room/${roomId}`} className={styles.callTileLink}>
+      <Link
+        to={`/${roomId.substring(1).split(":")[0]}`}
```
robintown: The fact that the variable is called `roomId` is worrying, because this approach only works for room aliases. Would you mind also renaming it to something more appropriate?

src/UrlParams.ts
```diff
@@ -93,14 +95,38 @@ interface UrlParams {
...
+    }
+    // Add "#", if not present
+    if (!roomAlias.includes("#")) {
+      roomAlias = `#${roomAlias}`;
+    }
+    // Add server part, if missing
+    if (!roomAlias.includes(":")) {
+      roomAlias = `${roomAlias}:${Config.defaultServerName()}`;
+    }
+  }
```
robintown: Sorry to ask you to redo this :sweat_smile: but I would much rather limit this function to handling parameters passed in the query string / fragment, and leave the parsing of the path to React Router and the RoomPage component, if possible. For example if we use a route like `"/:roomName"` instead of `"*"`, I think we can call React Router's `useParams` hook in the page component to get the name out of the URL.

src/App.tsx
robintown: (and also still keep a `/room/:roomId` route with a `Redirect` to the simplified URL, for backwards compatibility)

SimonBrandner: I don't think that is necessary? `getUrlParams()` should already handle `/room/...` URLs

src/matrix-utils.ts
SimonBrandner: The `getUrlParams()` function should handle the `roomId?` without `/room/...`

src/App.tsx
robintown: Oh, I see, the `*` route would cause `/room` to be routed to the room page, even though it looks like it needs to be a special case. I didn't catch that at first.

SimonBrandner: No worries :)

src/matrix-utils.ts
robintown: In a sense, yes, but if you follow this URL, our React Router setup will send you to the homepage rather than the room page

robintown: (at least I think so. certainly possible that I've got this wrong)

src/UrlParams.ts
robintown: I think we may have miscommunicated, because what I meant by this is that I'd like `getUrlParams` to continue to receive only the query string and fragment as parameters (i.e. only the non-routable parts of the URL), and then let the logic in `App` be the single source of truth for how we parse URL paths and route them in the app.

I believe this would be worthwhile since it would let us implement things like redirects from `/room/:roomAlias` to `/:roomName` more easily by directly applying the tools that React Router gives us, and avoids spreading the routing code across multiple places.

That's the idea at least, though I recognize that the room ID sometimes taking the form of `#?roomId=...` does complicate this… Do you have any thoughts on this?

src/matrix-utils.ts
SimonBrandner: Ah, right, yeah, that makes sense!

src/UrlParams.ts
SimonBrandner: I see, another problem might be that I am not really sure how the redirecting would behave in embedded mode - I am scared you might get unnecessary loading screens then which is why I put everything in here to avoid the redirects

To be frank I am not 100% sure I see the problem in what you're describing - I might need to think through more closely, but I see the paragraph above as a bigger issue

robintown (approved):
Hey, on second thought I don't want to leave this blocked on the above comment. I may be dealing with the routing code soon anyways to upgrade to React Router v6, which would be a good opportunity for me to show how I was imagining it could work, in a less confusing way than my explanations here.

SimonBrandner:
> Hey, on second thought I don't want to leave this blocked on the above comment. I may be dealing with the routing code soon anyways to upgrade to React Router v6, which would be a good opportunity for me to show how I was imagining it could work, in a less confusing way than my explanations here.

Hey, sounds good - looking forward to seeing the solution you have in mind :)

PR #1181, opened 2023-07-04
Use unique room names for LiveKit rooms

daniel-abramov:
Fixes https://github.com/vector-im/element-call/issues/1165.

Note that this PR is based on `lint-fixes` and merges into `lint-fixes`. I suggest we merge `lint-fixes` first, then merging this one into a default branch (`livekit`) will be trivial. The reason I made it based on `lint-fixes` is that I wanted to benefit from a stricter type system + because some of the parts that these PR changes were also updated in the `lint-fixes`. Doing it on top of `livekit` (default branch) would lead to another round of rebases in `lint-fixes`.

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -68,7 +68,6 @@ import { ElementWidgetActions, widget } from "../widget";
 import { GridLayoutMenu } from "./GridLayoutMenu";
 import { GroupCallInspector } from "./GroupCallInspector";
 import styles from "./InCallView.module.css";
-import { MatrixInfo } from "./VideoPreview";
 import { useJoinRule } from "./useJoinRule";
 import { ParticipantInfo } from "./useGroupCall";
 import { ItemData, TileContent, VideoTile } from "../video-grid/VideoTile";
@@ -117,7 +116,6 @@ export interface InCallViewProps {
   onLeave: () => void;
   unencryptedEventsFromUsers: Set<string>;
   hideHeader: boolean;
-  matrixInfo: MatrixInfo;
   otelGroupCallMembership?: OTelGroupCallMembership;
 }
 
@@ -129,7 +127,6 @@ export function InCallView({
   onLeave,
   unencryptedEventsFromUsers,
   hideHeader,
-  matrixInfo,
   otelGroupCallMembership,
 }: InCallViewProps) {
   const { t } = useTranslation();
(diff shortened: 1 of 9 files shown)


SimonBrandner (approved):
Looks sane

src/room/InCallView.tsx
```diff
@@ -98,7 +98,7 @@ export function ActiveCall(props: ActiveCallProps) {
   const livekitRoom = useLiveKit(props.userChoices, {
     sfuUrl: Config.get().livekit!.server_url,
     jwtUrl: `${Config.get().livekit!.jwt_service_url}/token`,
-    roomName: props.matrixInfo.roomName,
+    roomName: `{props.matrixInfo.roomName}({props.matrixInfo.roomId})`,
```
robintown: I think the problem that gave rise to https://github.com/vector-im/element-call/issues/1165 isn't that the LiveKit room name wasn't unique - it was that it wasn't stable. Including the Matrix room name in this string means that the result still won't be stable (it could either change over time, or vary between different participants when in a room without an explicit name), so I would instead suggest a combination of room ID + call ID.

robintown: Note that without `$` before the braces, this is actually the literal string `"{props.matrixInfo.roomName}({props.matrixInfo.roomId})"`

robintown (changes requested):

src/room/InCallView.tsx
SimonBrandner: Yeah, this make sense

SimonBrandner: Oh, I was totally blind when reviewing this...

daniel-abramov: My bad, thanks, fixed!

daniel-abramov: Oh, you're right, thanks! Fixed. Just to make sure that I did not mess up things: `groupCall.groupCallId` is the same as the `conf_id` (conference ID), right? (not the `callID` that was different for each participant who participated in a conference - the problem that we observed when testing `waterfall` several months ago)

robintown: Yep, that's right

src/room/InCallView.tsx
```diff
@@ -99,17 +98,19 @@ export function ActiveCall(props: ActiveCallProps) {
   const livekitRoom = useLiveKit(props.userChoices, {
     sfuUrl: Config.get().livekit!.server_url,
     jwtUrl: `${Config.get().livekit!.jwt_service_url}/token`,
-    roomName: props.matrixInfo.roomName,
+    roomName: `${props.matrixInfo.roomName}(${props.groupCall.groupCallId})`,
```
robintown: That's still the room name, rather than the room ID :D

daniel-abramov: Sorry, I messed things up. I've just rebased the PR against `lint-fixes` to contain the newest changes that were rebased against `livekit`. Now it should finally be fine 🙂

robintown (approved):

PR #1191, opened 2023-07-06
Improve the double click detection

robintown:
So that it doesn't cause unnecessary renders, and interprets a series of three clicks as a double-click followed by a single click, rather than two overlapping double-clicks. (That behavior felt odd to me during testing of NewVideoGrid, which is why I picked up this small change.)


SimonBrandner (approved):
Looks sane (though I don't mind the triple clicks :sweat_smile:)

PR #1192, opened 2023-07-06
Make the expand and collapse interactions inverses of one another

### LINKED ISSUE #1156: Consider positioning of tiles when expanded and reduced to minimise movement
Tiles can move around in confusing ways if I expand one and then shrink it down again. I know it might be difficult to achieve, but as I user I would expect that if I expand a tile and then contract it again, the layout will be exactly as it was before I expanded it. 
How are we currently deciding on tile positioning when expanding and contracting? Is there any way we can improve the model to make it feel more logical and consistent?

robintown:
For the most part, at least. If the edge cases where they differ still feel weird, I can iterate on this further.

[Screencast from 2023-07-05 16-40-36.webm](https://github.com/vector-im/element-call/assets/48614497/ff7b9538-f773-4b85-a3be-b91f5c19d999)

The diff is unfortunately a bit impenetrable, because I had to change both the `fillGaps` and `cycleTileSize` core algorithms used by the big grid layout. But: the main change of significance is the addition of a function `vacateArea`, which clears out an area within the grid in a specific way that mirrors the motion performed by `fillGaps`.

Closes https://github.com/vector-im/element-call/issues/1156

### DIFF
--- a/src/array-utils.ts
+++ b/src/array-utils.ts
@@ -22,10 +22,10 @@ limitations under the License.
 // Array.prototype.findLastIndex
 export function findLastIndex<T>(
   array: T[],
-  predicate: (item: T) => boolean
+  predicate: (item: T, index: number) => boolean
 ): number | null {
   for (let i = array.length - 1; i >= 0; i--) {
-    if (predicate(array[i])) return i;
+    if (predicate(array[i], i)) return i;
   }
 
   return null;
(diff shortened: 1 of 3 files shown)


dbkr (approved):
Wow, this ended up being quite the algorithm. Thanks for the detailed comments, it definitely helps.

toger5:
This is super cool! Amazing it worked out to be done today

PR #1193, opened 2023-07-06
Add E2EE banner

### LINKED ISSUE #1180: e2ee Disclaimer
### Your use case

Prior to releasing the livekit version to call.element.io, we require a prominent disclaimer. 

Ideally, I would like a suitable icon - e.g. an open or greyed-out padlock icon, with the text 'Not E2EE'. If possible this would then have a tooltip stating: `Element Call is temporarily not E2EE while we test scalability`.

@rufuskahler any chance you could put something together on this?

_No response_

### Additional context

_No response_

SimonBrandner:
Fixes https://github.com/vector-im/element-call/issues/1180

<hr>

image
image
image


src/E2EEBanner.module.css
```diff
@@ -0,0 +1,22 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+.e2eeBanner {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  gap: 16px;
```
toger5: the font size in the [figma](https://www.figma.com/file/AjvSw4lI9iGO5aTe4WAl4r/Element-Call?type=design&node-id=3842-74107&mode=design&t=yo0QxoFmpsS4st3E-0) desings is 12px. This should be the same than the font size of the "By clicking "Go", you agree to our [Terms and conditions](https://call.ems.host/_matrix/consent?v=1.0)" Label.

toger5 (approved):
Looks great.
Thanks

PR #1200, opened 2023-07-06
Combined permission request with newer livekit sdk version

toger5:
requires: https://github.com/livekit/client-sdk-js/pull/773

This superseeds: https://github.com/vector-im/element-call/pull/1172


SimonBrandner (approved):
Looks sane though I have to admit it's a little hard keeping track of all things

package.json
```diff
@@ -130,5 +130,8 @@
       "text",
       "cobertura"
     ]
+  },
+  "resolutions": {
+    "@types/react": "^18.2.12"
   }
```
toger5: This would fix the lint issues that were present in the build before this change. Is that acceptable?

SimonBrandner: Yeah, I think so, does anyone else from @vector-im/element-call-reviewers have opinions?

toger5: it was enought to update the yarn lock file

dbkr (approved):
Whoo, thanks!

src/room/VideoPreview.tsx
```diff
@@ -149,25 +172,21 @@ export function VideoPreview({ matrixInfo, onUserChoicesChanged }: Props) {
...
-            />
-          )}
+          <MicButton
+            muted={!audioEnabled}
+            onPress={() => setAudioEnabled(!audioEnabled)}
+          />
+          <VideoButton
+            muted={!videoEnabled}
+            onPress={() => setVideoEnabled(!videoEnabled)}
+          />
```
daniel-abramov, line 182: Just a quick question: what if we only have a video track or audio track?

toger5: You are right. I did not like the jumpy behaviour of the buttons (popping in and out after the track is available.) That is why i disabled it.
This is addressed in: https://github.com/vector-im/element-call/pull/1218

Issue #1165, opened 2023-06-28
Use room_id as room identifier for LiveKit since the roomDisplayName is not unique


fkwp:

SimonBrandner:
Fixed in https://github.com/vector-im/element-call/pull/1181

PR #1207, opened 2023-07-07
Release preparations

### LINKED ISSUE #1189: Fix simulcast screen sharing
Screen-sharing does not work in Chrome.

daniel-abramov:
Minimal set of fixes for the #1165 and 30 fps screen-sharing.

Fixes screen-sharing in Chrome, but disables simulcast (tbd #1189).

### DIFF
--- a/src/livekit/options.ts
+++ b/src/livekit/options.ts
@@ -16,10 +16,6 @@ const defaultLiveKitPublishOptions: TrackPublishDefaults = {
   simulcast: true,
   videoSimulcastLayers: [VideoPresets.h180, VideoPresets.h216] as VideoPreset[],
   screenShareEncoding: ScreenSharePresets.h1080fps30.encoding,
-  screenShareSimulcastLayers: [
-    new VideoPreset(1920, 1080, 1_500_000, 5, "medium"),
-    ScreenSharePresets.h1080fps15,
-  ] as VideoPreset[],
   stopMicTrackOnMute: false,
   videoCodec: "vp8",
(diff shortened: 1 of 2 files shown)


SimonBrandner (approved):

src/livekit/options.ts
```diff
@@ -17,8 +17,9 @@ const defaultLiveKitPublishOptions: TrackPublishDefaults = {
   videoSimulcastLayers: [VideoPresets.h180, VideoPresets.h216] as VideoPreset[],
   screenShareEncoding: ScreenSharePresets.h1080fps30.encoding,
   screenShareSimulcastLayers: [
-    new VideoPreset(1920, 1080, 1_500_000, 5, "medium"),
+    new VideoPreset(1920, 1080, 500_000, 5, "medium"),
```
toger5: is the screensharing fixed if we remote the 5fps layer entirely?

daniel-abramov: Sadly it only gets fixed if we remove all simulcast layers. Apparently, temporal layers don't work in Chrome or something like this.

But I think I get your point :) I.e. it's better to have screen sharing working in both browsers rather than having one browser with screen sharing support and another without.

I've removed them and only left the 30fps preset for the screen-sharing encoding. Now it should work in both browsers in 30 fps mode (but without a simulcast).

toger5: where is the change with the removed screen sharing layers?

daniel-abramov: I've forgotten to push, done :)

PR #1209, opened 2023-07-07
Fallback to config LiveKit URL if there is no URL coming from the js-sdk

SimonBrandner:


toger5:
That was quick (give me a sec to review)

src/room/GroupCallView.tsx
```diff
@@ -219,7 +220,9 @@ export function GroupCallView({
     undefined
   );
 
-  const livekitServiceURL = groupCall.foci[0]?.livekitServiceUrl;
+  const livekitServiceURL =
+    groupCall.foci[0]?.livekitServiceUrl ??
+    Config.get().livekit.livekit_service_url;
```
toger5, line 225: I guess I am assuming correctly, that a failed network request or an unset value will result in foci[0] to be undefined?

toger5 (approved):
Cool we have this in

src/room/GroupCallView.tsx
SimonBrandner: That's actually the following step. This only falls back when the value is undefined (i.e. the `m.call` event doesn't have the field). After we get the URL here we pass it over to the `OpenIDLoader` which actually does the request

toger5 (approved):

PR #1212, opened 2023-07-08
Update README with livekit config

xmj:
In line with the livekit updates to GroupCallView etc, modify README.md to contain a working configuration sample.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,19 +90,19 @@ yarn dev
 
 Add in you `.env` in root dir with:
 
-```yaml
+```shell
 # Develop backend settings:
 LIVEKIT_KEY="devkey"
 LIVEKIT_SECRET="secret"
 ```
 
-Add SFU parameter in your local config `./public/config.yml`:
+Add SFU parameter in your local config `./public/config.json`:
 
-```yaml
-"livekit": {
-    "server_url": "ws://localhost:7880",
-    "jwt_service_url": "http:/localhost:8881"
-  },
+```json
+,
+  "livekit": {
+      "livekit_service_url": "http://localhost:7881"
+    }
 ```
 
 Run backend components:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -90,19 +90,19 @@ yarn dev
...
-```yaml
-"livekit": {
-    "server_url": "ws://localhost:7880",
-    "jwt_service_url": "http:/localhost:8881"
-  },
+```json
+,
+  "livekit": {
+      "livekit_service_url": "http://localhost:7881"
+    }
```
SimonBrandner, line 105:
```suggestion
,
{
  "livekit": {
    "livekit_service_url": "http://localhost:7881"
  }
}
```

SimonBrandner (approved):
Thank you!!

README.md
xmj: This should probably be `http://localhost:7880" but I'm still having an issue with @dbkr's OpenID stuffs in the SFU code here:
https://github.com/vector-im/element-call/blob/livekit/src/livekit/openIDSFU.ts#L39

so I haven't been able to test that yet.

Which livekit-version do I need to get it running?

SimonBrandner: I think it was you with whom I chatted earlier? Are things clearer now?

toger5:
@xmj are you happy in how the readme now looks?
I think this PR still has the change to the sample config which is nice.
Do you think you can update it to only have those changes?

xmj:
> @xmj are you happy in how the readme now looks? I think this PR still has the change to the sample config which is nice. Do you think you can update it to only have those changes?

Hi

My usecase for Element-call went away over time - we did some hackathon @ hack4glarus.ch to get an instance running, that's where these changes came from.

happy for someone else to carry the torch :-D

toger5:
Closed in favor of: https://github.com/element-hq/element-call/pull/2483

PR #1216, opened 2023-07-10
Don't show profile settings in embedded mode

### LINKED ISSUE #1179: Element Call embedded mode: Clean up Settings Menu
In embedded mode some of the Element Call settings options do not make sense:
- Profil

SimonBrandner:
Fixes https://github.com/vector-im/element-call/issues/1179

### DIFF
--- a/src/settings/SettingsModal.tsx
+++ b/src/settings/SettingsModal.tsx
@@ -46,6 +46,7 @@ import {
   MediaDevices,
   MediaDevicesState,
 } from "../livekit/useMediaDevicesSwitcher";
+import { useUrlParams } from "../UrlParams";
 
 interface Props {
   mediaDevicesSwitcher?: MediaDevicesState;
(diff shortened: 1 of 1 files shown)


src/settings/SettingsModal.tsx
```diff
@@ -59,6 +60,8 @@ interface Props {
 export const SettingsModal = (props: Props) => {
   const { t } = useTranslation();
 
+  const { isEmbedded } = useUrlParams();
```
toger5, line 63: I remember to also have seen some `!widget` logit for detecting embedded mode? Is this the proper way. Is this also consistend with all other code?

toger5 (approved):
Looks good (I added one question)

src/settings/SettingsModal.tsx
SimonBrandner: Hmm, so as far as I can tell both are used and I am not sure if their usage differs

SimonBrandner: For posterity: we decided to go with `isEmbedded` as it makes more sense for UI

PR #1218, opened 2023-07-10
Disable mute button if media is not yet available. 

toger5:
This also fixes the jumpy mute unmute animation which can be observed at the very start of the lobby view.
Screenshot 2023-07-10 at 11 41 48


src/button/Button.module.css
```diff
@@ -61,6 +61,17 @@ limitations under the License.
...
 }
 
+.button:disabled,
+.toolbarButton:disabled,
+.toolbarButtonSecondary:disabled,
+.iconButton:disabled,
+.iconCopyButton:disabled,
+.secondary:disabled,
+.secondaryHangup:disabled,
+.copyButton:disabled {
```
SimonBrandner: Do we need to list these all? Shouldn't they all have class of `.button`?

SimonBrandner (approved):
Looks good otherwise

src/button/Button.module.css
toger5: good question. this is copied from the focus field above.

toger5: also works with just button.
The question is, if we really want all buttons to behave like this...

toger5: I think for now to just have this style for the toolbarButton might be the best option to avoid style regressions.

SimonBrandner (approved):

PR #1219, opened 2023-07-10
Adjust simulcast options for video

### LINKED ISSUE #1214: Add higher quality/resolution video feed layer
### Your use case

Curently the video is limited to 480p even if the network would allow for higher resolutions.
This is a good resolution for one person. For a camere setup which shows a whole group of ppl it becomes apperant, that one cannot see the detials of individual ppl.

_No response_

### Additional context

It would be great to activate the 720p or 1080p resolution layer.

SimonBrandner:
Fixes https://github.com/vector-im/element-call/issues/1214

### DIFF
--- a/src/livekit/options.ts
+++ b/src/livekit/options.ts
@@ -14,12 +14,12 @@ const defaultLiveKitPublishOptions: TrackPublishDefaults = {
   red: true,
   forceStereo: false,
   simulcast: true,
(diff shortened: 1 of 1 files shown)


toger5 (approved):
Thanks a lot. I think this is a very reasonable set of resolutions.

PR #1225, opened 2023-07-10
Add the livekit URL in matroyshka mode

### LINKED ISSUE #1201: Fix embedded mode
- Don't revert anything
- Use the livekit url from our config as a default, so if the state event doesn't have one, we'll use the one from our config.
- EW will continue creating calls with no livekit url
- Ideally we can still get embedded mode working today so that we don't ship it broken on call.element.io

dbkr:
This is a bit of a hack, as commented. The way to make it less of a hack is there too.

Fixes embedded mode, although users will get a permission prompt without https://github.com/matrix-org/matrix-react-sdk/pull/11209.

Fixes https://github.com/vector-im/element-call/issues/1201

### DIFF
--- a/src/widget.ts
+++ b/src/widget.ts
@@ -23,6 +23,7 @@ import type { MatrixClient } from "matrix-js-sdk/src/client";
 import type { IWidgetApiRequest } from "matrix-widget-api";
 import { LazyEventEmitter } from "./LazyEventEmitter";
 import { getUrlParams } from "./UrlParams";
+import { Config } from "./config/Config";
 
 // Subset of the actions in matrix-react-sdk
 export enum ElementWidgetActions {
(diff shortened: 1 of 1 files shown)


SimonBrandner (approved):

src/widget.ts
```diff
@@ -156,9 +157,33 @@ export const widget: WidgetHelpers | null = (() => {
...
+          // element-web to use waitForIFrameLoad=false. Once that change has rolled out,
+          // we can just start the client after we've fetched the config.
+          foci: [],
         }
       );
-      const clientPromise = client.startClient().then(() => client);
+
+      const clientPromise = new Promise<MatrixClient>((resolve) => {
+        (async () => {
+          await Config.init();
```
toger5, line 175: this seems to be redundant with what the initializer is doing? do we not have the initializer in embedded mode?

dbkr: Ah, sorry, added a comment

toger5: This migh result in duplicated downloads of the json right?

robintown: `Config.init` has a few lines that look like this:

```typescript
    if (Config.internalInstance?.initPromise) {
      return Config.internalInstance.initPromise;
    }
```

so that doesn't appear to be the case. If there's already a request promise in flight, it returns that same promise.

robintown (approved):

PR #1226, opened 2023-07-10
Fix E2EE disclaimer

SimonBrandner:
As per https://github.com/vector-im/element-call/issues/1180#issuecomment-1629023419


toger5 (approved):
thx

PR #1238, opened 2023-07-11
Remove unecassary state for lobby device initialization

toger5:
@daniel-abramov

### DIFF
--- a/src/room/VideoPreview.tsx
+++ b/src/room/VideoPreview.tsx
@@ -66,11 +66,6 @@ export function VideoPreview({ matrixInfo, onUserChoicesChanged }: Props) {
   const [videoEnabled, setVideoEnabled] = useState<boolean>(true);
   const [audioEnabled, setAudioEnabled] = useState<boolean>(true);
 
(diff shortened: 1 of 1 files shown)


daniel-abramov (approved):
Cool!

Do we need `videoAvailableAndEnabled`? It seems like it's not being used anymore if we rely on `videoEnabled`? (same for the audio counterpart) We only use it in an `onPress` callback, but it the `muted` does not use it anymore.

toger5:
@daniel-abramov and for the actual value we pass the the userChoices.
It makes things more readbale imo.

daniel-abramov:
> @daniel-abramov and for the actual value we pass the the userChoices. It makes things more readbale imo.

But should not we refer to the same value in `muted` and `onPress`?

toger5:
@daniel-abramov you are right. Since the button is disabled we never get into the situation that the video can be enabled/disabeld or changed in any way if the tracks are not available... (no track -> no button interaction)
So this was making it unecassary save (and a bit confusing)

SimonBrandner (approved):

PR #1243, opened 2023-07-12
Add support for accessing the web server from other devices

SimonBrandner:
`@vitejs/plugin-basic-ssl` enables SSL when running the Vite which means the `--host` flag (`yarn dev --host`) can be used to then access the app from other devices on the network which is useful for testing on other devices (e.g. mobile phones)


toger5 (approved):
This works and is really great

PR #1247, opened 2023-07-12
Explain how to start https devserver in the readme README.md

toger5:


robintown (commented):
Hmm, does this actually work? Last time I tried it, I wasn't able to use Element Call from a mobile device this way, because an HTTP dev server on a host other than localhost doesn't count as a [secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts).

robintown (approved):
Oh, now I've seen https://github.com/vector-im/element-call/pull/1243

toger5:
Yes this is mostly documentiong https://github.com/vector-im/element-call/pull/1243
So that we dont forget how to do that the next time I want to.

dbkr:
`prettier -w README.md` should make this happy: it probably just wants it wrapped to some number of columns.

PR #1266, opened 2023-07-14
Put ghost participants behind a developer flag

robintown:
Because of the slight delay between a participant leaving the LiveKit room and them leaving the Matrix call, they can appear as a ghost for a second while leaving, even during normal usage. Since ghosts are really just a developer feature used for load testing, I suggest we put this behind a developer flag to avoid 👻 spooking people.

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -58,6 +58,7 @@ import {
 import {
   useShowInspector,
   useShowConnectionStats,
+  useShowGhosts,
 } from "../settings/useSetting";
 import { useModalTriggerState } from "../Modal";
 import { PosthogAnalytics } from "../analytics/PosthogAnalytics";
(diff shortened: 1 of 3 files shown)


robintown:
This is blocked behind two things: getting end-to-end encryption, so that we no longer need to show ghosts as a defense against eavesdropping, and fixing matrix-js-sdk to retry sending the state event when it runs into rate limits, so that this change wouldn't introduce split-brains.

toger5:
@robintown is this worth keeping around or will this be done differently after the UX refactor?
If so I would propose closing this? (and maybe creating an issue instead?)

robintown:
Yeah okay, this has lots of conflicts and we're still a ways out from being able to undraft it. It's waiting on individual sender keys shared via to-device messages to become a thing, basically.

PR #1273, opened 2023-07-18
Set video enabled even if tracks are not yet initialized

toger5:
This could fix "muted on join issues" but could introduce issues where the buttons show unmuted even if no device is available.

### DIFF
--- a/src/room/VideoPreview.tsx
+++ b/src/room/VideoPreview.tsx
@@ -106,11 +106,11 @@ export function VideoPreview({ matrixInfo, onUserChoicesChanged }: Props) {
     onUserChoicesChanged({
       video: {
         selectedId: videoIn.selectedId,
-        enabled: videoEnabled && !!videoTrack,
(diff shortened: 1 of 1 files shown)


dbkr (approved):
This does seem to make more sense and it's not obvious to me why we'd need to factor in whether we had a track or not.

toger5:
> This does seem to make more sense and it's not obvious to me why we'd need to factor in whether we had a track or not.

Agreed. The idea was that if there is no track, there is no device, hence the user will not share video or audio. The idea was that it could be the least confusing if it also would be shown as muted to the user to not make them question if they actually have no device plugged in.

toger5:
> This does seem to make more sense and it's not obvious to me why we'd need to factor in whether we had a track or not.

Agreed. The idea was that if there is no track, there is no device, hence the user will not share video or audio. The idea was that it could be the least confusing if it also would be shown as muted to the user to not make them question if they actually have no device plugged in.

PR #1285, opened 2023-07-20
Add disconnected screen for when livekit disconnects from the call

### LINKED ISSUE #1268: Failure mode on reconnecting to a room(?)
### Steps to reproduce

There is some failure mode, possibly when reconnecting to a room after being disconnected, where the livekit SFU thinks the client has connected fine and is publishing a stream, but the client just says 'peer connection failed'. It seems like the same keeps happening even after a browser restart. The user in question had to change browsers to rejoin the call. Must be something getting left in a state where it just doesn't work at all on an unclean termination of the client.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

Yes

dbkr:
eg. when it gives up retrying to connect

Also refactors the rageshake button into a dedicated component that we can just put wherever we want.

Not the prettiest screen:
Jul-20-2023 17-54-26

Fixes https://github.com/vector-im/element-call/issues/1268


src/room/InCallView.tsx
```diff
@@ -188,6 +189,29 @@ export function InCallView({
...
+    onLeave();
+  }, [onLeave]);
+
+  useEffect(() => {
+    livekitRoom.on(RoomEvent.Disconnected, onDisconnected);
+
+    return () => {
+      livekitRoom.off(RoomEvent.Disconnected, onDisconnected);
+    };
+  }, [onDisconnected, livekitRoom]);
```
robintown: I think `useEventEmitter` would work here

src/settings/RageshakeButton.module.css
```diff
@@ -0,0 +1,22 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+.rageshakeControl {
+  height: 50px;
+  width: 176px;
```
robintown: I worry that hardcoding the dimensions like this is not a good idea for i18n

robintown (commented):
Generally looks good, I'd be curious to know if this screen has been tested on mobile though, given the other ticket about the feedback screen

src/settings/RageshakeButton.module.css
dbkr: Yeah, fair. I've just made it text which means the button changes size causing some jumping, but avoids the hardcoding.

robintown (approved):

PR #1287, opened 2023-07-21
Stop devices from automatically sleeping

### LINKED ISSUE #1087: Phone goes to sleep when in calls
### Steps to reproduce

As far as I can tell/remember, this happened in both:
* Android: Google Chrome (stable) : `114.0.5735.57`
* Android: Firefox Nightly (nightly): `116.0a1`

In both cases, the phone fell asleep after the standard screen time-out time.

I think these could help:
* https://developer.mozilla.org/en-US/docs/Web/API/Screen_Wake_Lock_API

Keywords: screen lock, screensaver, idle inhibit, screen goes to sleep, etc.

#### What did you expect?
* The screen to stay awake, as if I were watching a video in the browser?

#### What happened instead?

* The screen kept going to sleep. And this is not Element Call's fault, but Android's handling of rotation is truly awful in a number of cases, and one such example is that the orientation of the screen resets when the screen locks, further adding to the frustration felt here.

#### Extra

This form asks for logs, I specifically looked for a place to get/collect/send feedback in the app and couldn't find it. Very easily could be that I'm tired but maybe it could be more obvious?

### Operating system

Android

### Browser information

(see above)

### URL for webapp

call.element.io

### Will you send logs?

No

robintown:
A TypeScript upgrade was necessary to bring in the types for the wake lock API.

Closes https://github.com/vector-im/element-call/issues/1087


src/useWakeLock.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * React hook that inhibits the device from automatically going to sleep.
+ */
+export const useWakeLock = () => {
+  useEffect(() => {
+    if ("wakeLock" in navigator) {
+      let mounted = true;
+      let lock: WakeLockSentinel | null = null;
+
+      const onVisiblityChange = async () => {
```
dbkr, line 31: I guess the assumption here is that the device will continue to works for audio whilst asleep, so if it's not visible then we don't need to keep it awake? Should we be releasing the lock when the visibility changes to invisible then?

dbkr (commented):
Also looks like something has got confused with oidc-client-ts being dropped as a dep for matrix-js-sdk but still being needed, although it seems to still be there so not sure why yarn thinks it isn't.

src/useWakeLock.ts
robintown: Actually, the deal is that the browser will automatically release the lock whenever the window becomes invisible, so when the user comes back we have to reacquire it. I'll put that in a comment.

dbkr: Ah, I see - thanks.

dbkr (approved):

PR #1293, opened 2023-07-21
Connection Lost Banner

dbkr:
https://github.com/vector-im/element-call/pull/1292/files to the livekit branch, but with the logic more or less rewritten because the client state stuff had all been refactored in the livekit branch.


src/DisconnectedBanner.tsx
```diff
@@ -0,0 +1,53 @@
...
+  children,
+  className,
+  ...rest
+}: DisconnectedBannerProps) {
+  const { t } = useTranslation();
+  const clientState = useClientState();
+  let shouldShowBanner = false;
+
+  if (clientState?.state === "valid") {
+    const validClientState = clientState as ValidClientState;
```
robintown, line 39: I think TypeScript is normally smart enough to perform this cast automatically

robintown (approved):

src/DisconnectedBanner.tsx
dbkr: I would have thought this, but VSCode seemed to be unhappy without it.

toger5:
Thanks for getting this back in.

PR #1296, opened 2023-07-22
Update README.md

SnoutBug:
Update readme to reflect the host that uses the current latest version of the livekit branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ For prior version of the Element Call that relied solely on full-mesh logic, che
 
 ![A demo of Element Call with six people](demo.jpg)
 
-To try it out, visit our hosted version at [call.element.io](https://call.element.io). You can also find the latest development version continuously deployed to [element-call.netlify.app](https://element-call.netlify.app).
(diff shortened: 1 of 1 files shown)


robintown (approved):
Thanks, we forgot about this link!

dbkr:
Uh, this isn't true yet, call.element.dev isn't really "live" yet (it doesn't auto update). What's your source here?

SnoutBug:
@dbkr the other link does not host the latest version of the livekit branch, but the full-mesh branch, while call.element.dev does host the current latest version of the livekit branch
I stumbled upon this because I was confused as to why the hosted instance provided by element-call.netlify.app seemed so outdated when compared to call.element.io and call.element.dev

dbkr:
What other link?

Current situation:
 * element-call.netlify.app: latest full-mesh
 * element-call-livekit.netlify.app: latest livekit
 * call.element.io: latest livekit release
 * call.element.dev: outdated deployment of the livekit branch (not yet auto-updating)

I don't believe we've ever talked about call.element.dev (because it's not ready yet) so I'm slightly confused why you would have tried it at all. If it's been advertised somewhere, I'd like to clear up the confusion.

PR #1297, opened 2023-07-24
Automatically mute the user, if there already are more than 8 participants in the call

### LINKED ISSUE #1202: Automatically mute users when joining a call where participant number is greater than 8
### Your use case

When users join an in-progress large call, we should automatically set them to muted to avoid disruption and background noise on large calls. 

The behaviour should be that when they hit the lobby screen, they are pre-set to muted, and ideally there is some explanatory text such as 'this call is already busy so we have automatically muted you'. However they can still choose to un-mute themselves at this point, all we are doing is setting the default state.

_No response_

### Additional context

_No response_

SimonBrandner:
Fixes https://github.com/vector-im/element-call/issues/1202

### DIFF
--- a/src/room/GroupCallView.tsx
+++ b/src/room/GroupCallView.tsx
@@ -279,6 +279,7 @@ export function GroupCallView({
           setUserChoices(choices);
           enter();
         }}
+        muteAudio={participants.size > 8}
         isEmbedded={isEmbedded}
         hideHeader={hideHeader}
       />
--- a/src/room/VideoPreview.tsx
+++ b/src/room/VideoPreview.tsx
@@ -40,10 +40,15 @@ export type MatrixInfo = {
 
 interface Props {
   matrixInfo: MatrixInfo;
+  muteAudio: boolean;
   onUserChoicesChanged: (choices: UserChoices) => void;
 }
 
-export function VideoPreview({ matrixInfo, onUserChoicesChanged }: Props) {
+export function VideoPreview({
+  matrixInfo,
+  muteAudio,
+  onUserChoicesChanged,
+}: Props) {
   const { client } = useClient();
   const [previewRef, previewBounds] = useMeasure({ polyfill: ResizeObserver });
 
@@ -64,7 +69,13 @@ export function VideoPreview({ matrixInfo, onUserChoicesChanged }: Props) {
 
   // Create local media tracks.
   const [videoEnabled, setVideoEnabled] = useState<boolean>(true);
-  const [audioEnabled, setAudioEnabled] = useState<boolean>(true);
+  const [audioEnabled, setAudioEnabled] = useState<boolean>(!muteAudio);
+
+  useEffect(() => {
+    if (muteAudio) {
+      setAudioEnabled(false);
+    }
+  }, [muteAudio]);
 
   // The settings are updated as soon as the device changes. We wrap the settings value in a ref to store their initial value.
   // Not changing the device options prohibits the usePreviewTracks hook to recreate the tracks.
--- a/src/livekit/useLiveKit.ts
+++ b/src/livekit/useLiveKit.ts
@@ -11,7 +11,7 @@ export type UserChoices = {
 };
 
 export type DeviceChoices = {
-  selectedId: string;
+  selectedId?: string;
   enabled: boolean;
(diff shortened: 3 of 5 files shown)


robintown (commented):
I think the desired behavior that we discussed was to reflect this automatic mute state in the lobby, so it's more apparent to the user what their mute state in the call will be. Is that easily done?

SimonBrandner:
> I think the desired behavior that we discussed was to reflect this automatic mute state in the lobby, so it's more apparent to the user what their mute state in the call will be. Is that easily done?

Oh, that should be doable though I wonder if all users won't turn it on again? I guess not

robintown (approved):

src/room/VideoPreview.tsx
```diff
@@ -64,7 +69,13 @@ export function VideoPreview({ matrixInfo, onUserChoicesChanged }: Props) {
...
   // Create local media tracks.
   const [videoEnabled, setVideoEnabled] = useState<boolean>(true);
-  const [audioEnabled, setAudioEnabled] = useState<boolean>(true);
+  const [audioEnabled, setAudioEnabled] = useState<boolean>(!muteAudio);
+
+  useEffect(() => {
+    if (muteAudio) {
+      setAudioEnabled(false);
+    }
+  }, [muteAudio]);
```
toger5, line 78: Are you sure we need this? As soon as `audioEnabled` is set to false the user settings should contain the correct value (passed over: `onUserChoicesChanged`)

toger5: Sorry for the late review...

src/room/GroupCallView.tsx
```diff
@@ -279,6 +279,7 @@ export function GroupCallView({
           setUserChoices(choices);
           enter();
         }}
+        muteAudio={participants.size > 8}
```
daniel-abramov, line 282: It looks like each time the participant count changes (and the participant count is more than 8), the user will get muted, is it an expected behavior? - I.e. if you have 7 participants in a room, and you're in a lobby, unmuted, then suddenly the UI gets muted (because the 8th participant joined the call), at that point you may unmute yourself and then get muted again (9th participant joined).

I believe the check for participant size is something that we probably want to do only once upon joining.

src/room/VideoPreview.tsx
SimonBrandner: I am not sure I follow here...

src/room/GroupCallView.tsx
SimonBrandner: Yeah, I see what you mean though I wonder what exactly is the desired behaviour here... R previously said:

> I think the desired behavior that we discussed was to reflect this automatic mute state in the lobby, so it's more apparent to the user what their mute state in the call will be. Is that easily done?

src/room/VideoPreview.tsx
toger5: I think i now understand better what this is for. It guarantees, that we update the AudioEnabled flag if the muteAudio prop changes. (if more than 8 participants have joined)
Imo we should only use the participant count when opening the lobby to not have a unexpected mute state change that could confuse the user. In other words: if the participant count exceeds 8 while the user is in the lobby should not change the button mute state.

SimonBrandner: Yes, I guess that would make sense, yes - /me gets it done

src/room/GroupCallView.tsx
daniel-abramov: Yeah, it makes sense, but what I meant to say is that we probably don't want to change the mute state each time when the amount of participants changes as it means that if you unmute the video, it will get muted automatically and each time you unmute it, it will get muted again once someone in the room joins (which won't be obvious for the lobby user).

I remember that some other solution does it like this: when you join the room, your devices are muted (and there is a tooltip telling you that), but if you unmute yourself, you don't get muted once someone new joins. I.e. I think the amount of participants and setting a mute state should probably be done once, not each time someone enters/leaves a call. Does it make any sense? :)

SimonBrandner: Yeah, I think so. Does [this](https://github.com/vector-im/element-call/pull/1306) make sense?

PR #1306, opened 2023-07-25
Don't change mute state while in the lobby

SimonBrandner:
Adjust https://github.com/vector-im/element-call/pull/1297

CC @toger5 and @daniel-abramov

### DIFF
--- a/src/room/GroupCallView.tsx
+++ b/src/room/GroupCallView.tsx
@@ -37,6 +37,12 @@ import { findDeviceByName } from "../media-utils";
 import { OpenIDLoader } from "../livekit/OpenIDLoader";
 import { ActiveCall } from "./InCallView";
 
+/**
+ * If there already is this many participants in the call, we automatically mute
+ * the user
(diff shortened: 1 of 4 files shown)


src/room/GroupCallView.tsx
```diff
@@ -293,7 +295,7 @@ export function GroupCallView({
           setUserChoices(choices);
           enter();
         }}
-        muteAudio={participants.size > 8}
+        initWithMutedAudio={participants.size > 0}
```
toger5: This should be > 8 ?
Or did the logic change?

toger5 (commented):
I think there should be an 8. Maybe we also want this to become a constant somewhere?

src/room/GroupCallView.tsx
SimonBrandner: Oops

toger5 (approved):
I was just about proposing to make it a const. Great you made it a const anyways

PR #1308, opened 2023-07-25
Add screen-sharing local volume control

### LINKED ISSUE #1233: Local volume control of screen-share tiles
Branched of https://github.com/vector-im/element-call/issues/1231
Depends on https://github.com/livekit/client-sdk-js/issues/442

SimonBrandner:
Fixes https://github.com/vector-im/element-call/issues/1233

This should work in theory but doesn't work in practice


dbkr (approved):
lgtm, assuming the comment is stale and it does actually work in practice, as I think I saw you mention.

PR #1314, opened 2023-07-28
Switch to new group call if the call in a room changes

dbkr:
This PR comes with a lot of caveats currently:
 * It removes the leaveCall() from when the useGroupCall() hook is torn down. I can't see another way of maintaining the correct joined state (because switching group calls will cause the effects to re-run) but this does mean that if the user somehow navigates within the app without using the hangup button, they won't leave the call.
 * The state of the call will be reset on switch, including the state of the screenshare, what devices are being used and even the mute states.

Unsure if it's better to try & fix these up by moving state around or whether to refactor more in js-sdk so the group call can be switched more transparently.

Requires: https://github.com/matrix-org/matrix-js-sdk/pull/3636

### DIFF
--- a/src/room/useGroupCall.ts
+++ b/src/room/useGroupCall.ts
@@ -94,6 +94,8 @@ interface State {
   requestingScreenshare: boolean;
   participants: Map<RoomMember, Map<string, ParticipantInfo>>;
   hasLocalParticipant: boolean;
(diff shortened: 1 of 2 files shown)


dbkr:
Superseded by https://github.com/vector-im/element-call/pull/1348

PR #1321, opened 2023-08-02
Fix multiple issues with device settings

### LINKED ISSUE #1316: Camera/microphones no longer have a name
### Steps to reproduce

1. Visit any call URL
2. Go to settings
3. Look at the microphone / camera options

#### What did you expect?

Devices names

#### What happened instead?

Numbered devices

image

### Operating system

Ubuntu 22.04.2 LTS

### Browser information

Firefox 116.0

### URL for webapp

https://call.element.io/

### Will you send logs?

Yes

### LINKED ISSUE #1317: Device choice should be remembered

Element Call should remember the previous choice of camera and microphone.

Save time to select the correct device every time from the settings modal.

-

-

### Additional context

-

robintown:
To track media devices, we were previously relying on a combination of LiveKit's useMediaDeviceSelect hook, and an object called UserChoices. Device settings should be accessible from outside a call, but the latter hook should only be used with a room or set of preview tracks, so it couldn't be raised to the app's top level. I also felt that the UserChoices was hard to follow due to lack of clear ownership of the object.

To bring clarity to media device handling and allow device settings to be shown outside a call, I refactored these things into a single MediaDevicesContext which is instantiated at the top level of the app. Then, I had to manually sync LiveKit's device state with whatever is present in the context. This refactoring ended up fixing a couple other bugs with device handling along the way:

Closes https://github.com/vector-im/element-call/issues/1316
Closes https://github.com/vector-im/element-call/issues/1317
https://github.com/vector-im/element-call/issues/1289 related


toger5:
I am getting a: `Uncaught (in promise) SyntaxError: JSON.parse: unexpected non-whitespace character after JSON data at line 1 column 4 of the JSON data` error. Is that my local setup? Do I need to reset the cache?

dbkr:
Ah, this is our pipeline not working for forked PRs again. Maybe we should just accept the tradeoff and always fetch the config from the netlify branch rather than trying to find it from the source branch.

robintown:
Replaced by https://github.com/vector-im/element-call/pull/1322 which ought to have a working build

PR #1322, opened 2023-08-03
Fix multiple issues with device settings

### LINKED ISSUE #1316: Camera/microphones no longer have a name

### LINKED ISSUE #1317: Device choice should be remembered

robintown:
To track media devices, we were previously relying on a combination of LiveKit's useMediaDeviceSelect hook, and an object called UserChoices. Device settings should be accessible from outside a call, but the latter hook should only be used with a room or set of preview tracks, so it couldn't be raised to the app's top level. I also felt that the UserChoices code was hard to follow due to lack of clear ownership of the object.

To bring clarity to media device handling and allow device settings to be shown outside a call, I refactored these things into a single MediaDevicesContext which is instantiated at the top level of the app. Then, I had to manually sync LiveKit's device state with whatever is present in the context. This refactoring ended up fixing a couple other bugs with device handling along the way:

Closes https://github.com/vector-im/element-call/issues/1316
Closes https://github.com/vector-im/element-call/issues/1317
https://github.com/vector-im/element-call/issues/1289 related


dbkr (approved):
Wow, this ended up being quite the refactor: thanks for that, looks good I think!

toger5:
Really nice you gave this topic another look. 
For the global device settings menu: This of course implies, that we will have permission prompts when going into the settings right?

This also brings back the issue with the multiple device requests on android mobile. But this seems to be a regressions anyways. Have you ever tried removing the change that was done after it was all working (the change with the hardware mute in the lobby if the mute button is pressed so the indicator led turns off?)

robintown:
@toger5 Yes, permission prompts in settings are expected (and sadly I couldn't find a way to make them combined, but I think this is acceptable as a start)

You mean, on Chrome on Android, there are multiple device requests in the lobby? That's interesting to know. I did remove the muting change that you describe, because I noticed that it would be enough to let usePreviewTracks re-create the tracks from scratch each time the video configuration changes, which meant I could reduce the code complexity there. So I have a hard time seeing what would cause this…

toger5:
Oh I was beeing unprecise the issue only occurs on firefox mobile android.

PR #1325, opened 2023-08-04
Put E2EE password into the URL and store it in local storage

SimonBrandner:


src/UrlParams.ts
```diff
@@ -103,8 +109,17 @@ export const getUrlParams = (
   hash = window.location.hash
 ): UrlParams => {
   let roomAlias: string | null = null;
+  let password: string | null = null;
+
+  const passwordIndex = hash.indexOf(PASSWORD_STRING);
```
dbkr: Using a question mark here is probably technically fine but seems a bit confusing given the overlap with the query param. Maybe just parse it as a regular query string, so separate fields with '&'?

src/home/RegisteredView.tsx
```diff
@@ -70,11 +73,14 @@ export function RegisteredView({ client, isPasswordlessUser }: Props) {
...
-        const [roomAlias] = await createRoom(client, roomName, ptt);
+        const [roomAlias, roomId] = await createRoom(client, roomName, ptt);
 
-        if (roomAlias) {
-          history.push(`/${roomAlias.substring(1).split(":")[0]}`);
-        }
+        setLocalStorageItem(
+          getRoomSharedKeyLocalStorageKey(roomId),
+          randomString(32)
+        );
```
dbkr, line 81: If we're storing one item in localstorage per room, I think we need to handle running out of space unfortunately: I'm not sure how though, I guess we would have to start forgetting old rooms? Also, what happens in firefox private mode?

dbkr (commented):
The UX here feels confusing to me: specifically having the URLs be room-specific but not be able to c+p the URL bar to have someone else join, although this is just me guessing without any data to back it up. I think we'll have to have some kind of proper design work here before we launch this.

src/UrlParams.ts
SimonBrandner: Good point

src/home/RegisteredView.tsx
SimonBrandner: From what I found online, there is a 5 MB - 10 MB limit though I didn't find anything about an item count limit?

SimonBrandner: > Also, what happens in firefox private mode?

The non-nightly versions of FF still don't have insertable streams, so it doesn't matter much atm...

src/matrix-utils.ts
```diff
@@ -343,13 +343,18 @@ export async function createRoom(
 }
 
 // Returns a URL to that will load Element Call with the given room
-export function getRoomUrl(roomIdOrAlias: string): string {
+export function getRoomUrl(
+  roomIdOrAlias: string,
+  password: string = ""
+): string {
+  password = password === "" ? "" : "#?" + PASSWORD_STRING + password;
```
dbkr, line 350: Seems like we've still ended up with a '?' in here. If that's intentional, okay, but my suggestion was just to have "#password=foo" and then just parsing the hash with `URLSearchParams ` rather than writing our own parsing.

src/home/RegisteredView.tsx
dbkr: But it's still going to fail after you joined a certain number of e2ee rooms, whether the limit is in in count or bytes?

Do we at least know what the failure mode is without localstorage?

SimonBrandner:
Closing in favour of https://github.com/vector-im/element-call/pull/1344/ as it contains more fixes and generally makes more sense to merge in one piece

PR #1334, opened 2023-08-09
Addressing Deprecating save-state and set-output commands

fkwp:


.github/workflows/publish.yaml
```diff
@@ -69,7 +69,7 @@ jobs:
             type=sha,format=short,event=branch
             type=semver,pattern={{version}}
             type=raw,value=latest-ci,enable={{is_default_branch}}
-            type=raw,value=latest-ci_{{ steps.current_date.outputs.unix_time }},enable={{is_default_branch}}
+            type=raw,value=latest-ci_{{steps.current.time.outputs.unix_time}},enable={{is_default_branch}}
```
robintown: Did you mean `current-time` here, perhaps?

fkwp: thx yes :-)

robintown (approved):

PR #1335, opened 2023-08-09
fix typo -> {{}} needs to be prefixed by $

fkwp:


.github/workflows/publish.yaml
```diff
@@ -69,7 +69,7 @@ jobs:
             type=sha,format=short,event=branch
             type=semver,pattern={{version}}
             type=raw,value=latest-ci,enable={{is_default_branch}}
-            type=raw,value=latest-ci_{{steps.current-time.outputs.unix_time}},enable={{is_default_branch}}
+            type=raw,value=latest-ci_${{steps.current-time.outputs.unix_time}},enable={{is_default_branch}}
```
robintown, line 72: Does that fix need to apply to all the other uses of `{{}}` in these lines, too?

fkwp: tbh. I don't have an idea... I saw lines with and without $

fkwp: https://docs.github.com/en/actions/learn-github-actions/expressions

robintown: Okay, well let's try this and see what it does

robintown (approved):

PR #1339, opened 2023-08-09
Disable full mesh in widget mode

robintown:
We disabled full mesh on the client that's used in SPA mode, but forgot to do the same for the widget mode client.


SimonBrandner (approved):

toger5:
What a great find!

PR #1344, opened 2023-08-11
Complete SPA E2EE work

SimonBrandner:


src/e2ee/sharedKeyManagement.ts
```diff
@@ -0,0 +1,81 @@
...
+export const useManageRoomSharedKey = (roomId: string): string | null => {
+  const { password } = useUrlParams();
+  const [e2eeSharedKey, setE2EESharedKey] = useInternalRoomSharedKey(roomId);
+
+  useEffect(() => {
+    if (!password) return;
+    if (password === "") return;
+    if (password === e2eeSharedKey) return;
+
+    setE2EESharedKey?.(password);
```
dbkr: Why does this need an optional?

dbkr (approved):
I think this looks good now, modulo same concerns about the UX around e2ee call URLs, but we can revisit that with proper design work.

PR #1348, opened 2023-08-18
Switch to the MatrixRTCSession layer in js-sdk

### LINKED ISSUE #1301: Update call when a new MatrixRTC related state changes
### Steps to reproduce

There is always a possibility that two users trying to create a call close enough in time will race and create two calls. When this happens, we need to resolve the glare and pick one of them. Currently, users already on the call will stay using the old call event and users who join subsequently will end up in the new call, creating a splitbrain.

We already have an algorithm for picking which call in a room we use, we just need to re-run it when a new call event arrives, leaving the current call and joining a new one if we get a different call event.

[Evolution of MatrixRTC related state events](https://docs.google.com/document/d/1nTQNVNf0bjEcvMtEhs_4lVgWPGJxMd42bqQeoZQF2lU/edit)

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

dbkr:
Requires https://github.com/matrix-org/matrix-js-sdk/pull/3663
Fixes https://github.com/vector-im/element-call/issues/1301


src/livekit/openIDSFU.ts
```diff
@@ -14,82 +14,78 @@ See the License for the specific language governing permissions and
...
 export interface SFUConfig {
   url: string;
   jwt: string;
 }
 
+export function sfuConfigEquals(a?: SFUConfig, b?: SFUConfig): boolean {
+  if (a === undefined && b === undefined) return true;
+  if (a === undefined || b === undefined) return false;
+
+  return a.jwt === b.jwt && a.url === b.url;
```
SimonBrandner, line 34:
```suggestion
  return a?.jwt === b?.jwt && a?.url === b?.url;
```

Why not ^?

src/livekit/openIDSFU.ts
```diff
@@ -14,82 +14,78 @@ See the License for the specific language governing permissions and
...
+  const activeFocus = useActiveFocus(rtcSession);
+
+  useEffect(() => {
+    (async () => {
+      const sfuConfig = activeFocus
+        ? await getSFUConfigWithOpenID(client, activeFocus)
+        : undefined;
+      setSFUConfig(sfuConfig);
+    })();
+  }, [client, activeFocus]);
```
SimonBrandner, line 58:
```suggestion
  useEffect(async () => {
    const sfuConfig = activeFocus
      ? await getSFUConfigWithOpenID(client, activeFocus)
      : undefined;
    setSFUConfig(sfuConfig);
  }, [client, activeFocus]);
```

Wouldn't this be enough?

src/room/InCallView.tsx
```diff
@@ -442,53 +424,63 @@ export function InCallView({
...
 }
 
+function findMatrixMember(
+  room: MatrixRoom,
+  id: string
+): RoomMember | undefined {
+  if (!id) return undefined;
+
+  const parts = id.split(":");
+  if (parts.length < 2) {
```
SimonBrandner: Should this be `3`, if the ID looks like `@simonb:element.io:FDU546DFE`?

SimonBrandner (commented):
This is looking really good, afaict

Requesting team review again as I feel like this should get more reviewers

src/livekit/openIDSFU.ts
dbkr: I find that much harder to understand, personally, and less obvious that it's doing the thing it intends to do. I'm fine with it, but I think it could use a comment, whereas I find the previous code a bit more self-documenting.

src/livekit/openIDSFU.ts
dbkr: React won't let you have an async function as an effect callback.

src/room/InCallView.tsx
dbkr: I don't think so? There should be at least two parts: the user ID and the device ID. If there are fewer than that, it's an error.

src/livekit/openIDSFU.ts
SimonBrandner: I think that for me it's quicker to read my version as "are they equal" but I don't feel strongly about it. FWIW this would also work:

```suggestion
  if (a.url !== b.url) return false;
  if (a.jwt !== b.jwt) return false

  return true;
```

src/room/InCallView.tsx
SimonBrandner: Well `@simonb:element.io.split(":") === 2`, doesn't it?

dbkr: Oh, I see what you mean. Well, this was not assuming anything in particular about the format of the user IDs other than that they might contain colons, so we just take the last part off the end. Really we ought to pick a separator that doesn't appear in either, escaping/encoding the parts if necessary.

src/livekit/openIDSFU.ts
dbkr: Wouldn't that also need conditionals to avoid dereferencing `undefined`?

src/room/InCallView.tsx
SimonBrandner: I agree, we should use something different and that the code that joins the two user ID parts works but this check doesn't...

src/livekit/openIDSFU.ts
SimonBrandner: I am not sure what you mean?

dbkr: The first thing it does is read `a.url` but `a` could be undefined in which case that will throw an exception.

src/room/InCallView.tsx
dbkr: Fair enough, I guess we specifically know it's a user ID anyway so we should treat it as such.

src/livekit/openIDSFU.ts
SimonBrandner: Ahm rug
```suggestion
  if (a?.url !== b?.url) return false;
  if (a?.jwt !== b?.jwt) return false

  return true;
```

Ah, right!

dbkr: OOB we've decided this doesn't really matter - resolving

SimonBrandner (approved):

PR #1350, opened 2023-08-22
E2EE for embeded mode

SimonBrandner:
Based on https://github.com/vector-im/element-call/pull/1348
Requires https://github.com/matrix-org/matrix-js-sdk/pull/3667


src/livekit/useLiveKit.ts
```diff
@@ -53,25 +60,38 @@ interface UseLivekitResult {
...
-
-    (e2eeOptions.keyProvider as ExternalE2EEKeyProvider).setKey(
-      e2eeConfig?.sharedKey
-    );
-  }, [e2eeOptions, e2eeConfig?.sharedKey]);
+    if (!e2eeOptions) return;
+    if (!e2eeConfig) return;
+
+    if (e2eeConfig.mode === E2EEMode.PerParticipantKey) {
+      (e2eeOptions.keyProvider as MatrixKeyProvider).setRTCSession(rtcSession);
```
dbkr, line 88: Should `setRTCSession` be called with `undefined` somewhere to remove the listener?

src/e2ee/matrixKeyProvider.ts
```diff
@@ -0,0 +1,63 @@
...
+    this.rtcSession.on(
+      MatrixRTCSessionEvent.EncryptionKeyChanged,
+      this.onEncryptionKeyChanged
+    );
+
+    for (const [
+      participant,
+      encryptionKey,
+    ] of this.rtcSession.encryptionKeys.entries()) {
+      this.onEncryptionKeyChanged(
```
dbkr: Some comments here wouldn't go amiss: I'm not immediately seeing why the RTCSession changing would obviously imply a key change.

src/livekit/useLiveKit.ts
SimonBrandner: I assume that if the class gets destroyed, so do the listeners, so there is no need to call it with `undefined`?

src/e2ee/matrixKeyProvider.ts
SimonBrandner: Not necessarily but the new session could be aware of keys of which the old session wasn't. Also, if we end up linking the encryption key events to individual calls as you suggested in the other PR (which makes sense), then I assume the keys would change with the session (that is if I understand correctly what the class represents - i.e. there would be multiple instances of this class per room if there were two call ids being used in that room)

dbkr:
```suggestion
      // The new session could be aware of keys of which the old session wasn't, so emit a key changed event.
      this.onEncryptionKeyChanged(
```

src/livekit/useLiveKit.ts
dbkr: Well, not explicitly: there's no destructors run here (it's javascript) so the emitter will continue to retain a reference to the listener, we'd just be relying on them being GCed, so it depends whether anything else holds on to the thing doing the listening (the key provider in this case I think). Assuming everything gets torn down then everything should be detached from the GC root, but we shouldn't be relying on it given the choice IMO.

SimonBrandner: I am not really sure I see a case where this would be necessary -> I don't think I see where/when we should destroy it manually...

dbkr:
I've created https://github.com/vector-im/element-call/pull/1772 to re-apply the changes here (or most of them) as this is super conflicty now.

PR #1354, opened 2023-08-28
Adopt the Compound color system

robintown:
As a first step towards adopting the Compound design system and the new Element Call designs, this pulls in Compound's color tokens and applies them to all existing components. I've tried to choose tokens based on the semantics of where they're used, but in some cases, where the new and old design systems differ in semantics, it was necessary to choose tokens based on their resulting color. These hacks can be removed as we implement more of the new designs.

There were a set of environment variables that we used for custom themes, but Compound has way too many design tokens for that approach to still be a good idea, so I decided to replace them all with a single environment variable that just lets you write arbitrary custom CSS.

Before|After
-|-
Screenshot 2023-08-28 at 17-18-12 Element Call Home|Screenshot 2023-08-28 at 17-21-12 Element Call Home
Screenshot 2023-08-28 at 17-16-59 Element Call|Screenshot 2023-08-28 at 17-16-20 Element Call


SimonBrandner (approved):
Thanks!

PR #1358, opened 2023-08-30
Enable E2EE by default

SimonBrandner:


dbkr:
Won't this prevent us from releasing unless we want to also enable e2ee by default on prod?

dbkr (approved):
OOB: we intend to have e2ee by default in the next release, so this is a logical step!

PR #1374, opened 2023-08-31
Implement new designs for in-call footer buttons

robintown:
This basically amounts to swapping out some CSS, the icons, and the Tooltip component.

Screenshot 2023-08-30 at 21-57-33 Element Call

Requires https://github.com/vector-im/element-call/pull/1377


SimonBrandner (approved):
Looks sane codewise though CI seems unhappy

dbkr:
I think this is just because it needed a new compound with the new icons. Renovate has provided https://github.com/vector-im/element-call/pull/1377

PR #1382, opened 2023-08-31
Make wildcard (*) a possible parentUrl

toger5:
This is needed for the way android, ios, and tauri implement EC.
-> They are not running in an IFrame hence they need to send the request to themselves. Alternativly we could provide the same url that we use to setup the IFrame

### DIFF
--- a/src/widget.ts
+++ b/src/widget.ts
@@ -75,7 +75,9 @@ export const widget: WidgetHelpers | null = (() => {
     const parentUrl = query.get("parentUrl");
 
     if (widgetId && parentUrl) {
-      const parentOrigin = new URL(parentUrl).origin;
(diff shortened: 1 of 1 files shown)


toger5:
I am alsmost thinking, make parentUrl the same as the app url migh be the better solution. I will close htis once I made a choice. Feel free to comment in case you have an opinion

SimonBrandner (approved):

SimonBrandner:
Looks good codewise not sure what the nicest approach is though

toger5:
Instead of adding a wildcard we aill jsut provdide the same url as we use for the webview itself
```
parentUrl = https://call.element.io
viebview url: https://call.element.io/room?embed=&... all the other props ...
```

PR #1387, opened 2023-08-31
Switch to `Avatar` component from Compound

SimonBrandner:


src/video-grid/VideoTile.tsx
```diff
@@ -145,6 +145,8 @@ export const VideoTile = forwardRef<HTMLDivElement, Props>(
     // Firefox doesn't respect the disablePictureInPicture attribute
     // https://bugzilla.mozilla.org/show_bug.cgi?id=1611831
 
+    console.log(`LOG VideoTIle mxcSrc=${member?.getMxcAvatarUrl()}`);
```
robintown: Seems to be left over from debugging?

robintown (approved):
Thanks for doing this! It looks sensible to me

robintown (commented):
Not a blocker, but still worth noting: this seems to have regressed the appearance of avatars during grid resizes. They're meant to stay circular and scale smoothly with the containing tile.

[Screencast from 2023-09-08 14-51-51.webm](https://github.com/vector-im/element-call/assets/48614497/1582eed9-bda0-4969-96cd-ef02299d6240)

PR #1410, opened 2023-09-04
Smooth Focus Switching

dbkr:
For a few seconds after a focusn switch, keep old tiles from the previous
focus so that it doesn't look like everyone leaves & comes back.

Based on https://github.com/vector-im/element-call/pull/1348
Requires https://github.com/livekit/components-js/pull/620


src/room/InCallView.tsx
```diff
@@ -555,5 +564,40 @@ function useParticipantTiles(
...
+      setTimeout(() => {
+        setIsSwitchingFocus(false);
+      }, POST_FOCUS_PARTICIPANT_UPDATE_DELAY_MS);
+    }
+  }, [connState, setIsSwitchingFocus, isSwitchingFocus]);
+
+  if (
+    connState === ECAddonConnectionState.ECSwitchingFocus ||
+    isSwitchingFocus
+  ) {
```
toger5, line 589: I am assuming `connState` is only set to the `ECSwitchingFocus` case for a short time and `isSwitchingFocus` is used to extend the "show prev. tiles" phase for a longer time period?
If so, I am wondering if we ever hit the case where `connState === ECAddonConnectionState.ECSwitchingFocus` is `true` but `isSwitchingFocus` is false?

toger5: what happens, if the timeout `POST_FOCUS_PARTICIPANT_UPDATE_DELAY_MS` has passed but there are still missing reconnects?

src/room/InCallView.tsx
```diff
@@ -555,5 +564,40 @@ function useParticipantTiles(
...
+        newItems.push(prevTile);
+      }
+    }
+
+    return newItems;
+  }
+
+  previousTiles.current = items;
+
   return items;
```
toger5: Nit (and maybe already in personal preference territory)
Can we make it more obvious, that we only return the items if the previous condition is not met?

```suggestion
  } else {
   previousTiles.current = items;
   return items;
}
```

toger5 (commented):
Looks good and makes sense.
Overall I am wondering if the logic could be swapped and made more solid by relying on the member state events more directly.
Currently I understand it like this:
 - on ECSwtichingFocus we store the prev participants
 - than we show all the users which are connected + all the prev participants for this (while prioritizing the connected ones and use the prev ones as a fallback) for this `POST_FOCUS_PARTICIPANT_UPDATE_DELAY_MS` amount of time -> then we only show the connected participants
 
So if ppl take even longer this would still cause a "flicker" also I dont know what time we would choose for `POST_FOCUS_PARTICIPANT_UPDATE_DELAY_MS`. So I am wondering if we cannot just always run the logic without any time constraint attached like this:
 - independent of the connection state compute a list of items based on the membership events (go through all members and find according tiles in this order):
     1. check id in actual tile items -> use it
     2. check id in prev items -> use it
     3. as a last resort add a dummy tile/ connecting tile for all membership events that are not part of one of the arrays.
     
To sum up. I dont see the immediate necassity for the timer. It seems not a perfect fit for the issue? Since there is no clear way to figure out what the timeout should be it seems not obvious why we would use one in the first place. Does this make sense or am I misunderstanding or missing parts of it?

src/room/InCallView.tsx
dbkr: Added a comment in a0226156 which hopefully answers both questions!

toger5 (approved):

PR #1420, opened 2023-09-06
Add apple site association

dbkr:
So when you open EC in a browser, it will display a banner at the top asking you if you want to open element x. We'll still be able to use EC in the browser.


stefanceriu (approved):
Thaank you! 👏

toger5 (approved):

Issue #1345, opened 2023-08-14
Unmuted when UI shows that I am muted


germain-gg:
### Steps to reproduce

1. Join a call
2. Press the "spacebar" for a split second

#### What did you expect?

To the muted as soon as the spacebar is released

#### What happened instead?

People DM'ed me saying that I was unmuted, even though the UI was showing that i was muted.

#### Extra context

Toggling the mute state actually made me go on mute and solved the issue.

I was wearing AirPods, and when i pressed the spacebar, it actually toggled the "mic mode" of the airpods

This can't be reproduced all the time but i added as much context as i could.

### Will you send logs?

No

daniellekirkwood:
On the 'other side' I saw @germain-gg's icon as 'unmuted' and his background noise moved his tile to towards the top of the screen. Interesting that I saw as unmuted but Germain didn't have that state reflected in their own view.

dbkr:
https://github.com/vector-im/element-call/pull/1421 might fix this although I could only repro the opposite case, ie. the mute button claimed I was unmuted when I was muted (the mute icon in my video feed still updated).

PR #1421, opened 2023-09-06
Fix the mute icon being incorrect when using PTT

dbkr:
We were manipulating the participant's mute state directly for some reason, just for setting the mute state directly, which bypased the mutestates hook.

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -190,8 +190,7 @@ export function InCallView({
     containerRef1,
     toggleMicrophone,
     toggleCamera,
-    (muted) =>
-      muteStates?.audio?.setEnabled && muteStates.audio.setEnabled(!muted)
+    (muted) => muteStates.audio.setEnabled?.(!muted)
   );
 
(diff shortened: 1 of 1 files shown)


src/room/InCallView.tsx
```diff
@@ -190,7 +190,8 @@ export function InCallView({
     containerRef1,
     toggleMicrophone,
     toggleCamera,
-    async (muted) => await localParticipant.setMicrophoneEnabled(!muted)
+    (muted) =>
+      muteStates?.audio?.setEnabled && muteStates.audio.setEnabled(!muted)
```
```suggestion
    (muted) => muteStates.audio.setEnabled?.(!muted)
```

robintown (approved):
Would this possibly close https://github.com/vector-im/element-call/issues/1345?

dbkr:
> Would this possibly close #1345?

Hmm, I can reproduce the *opposite* behaviour (ie. the button shows I'm unmuted when I'm muted) but not that way around. It would definitely mess up the state though, so possibly.

PR #1425, opened 2023-09-08
Always publish microphone track when joining

### LINKED ISSUE #1412: Keep the mic device open during a call
We currently close both the video and audio devices when the user mutes, opening them again when the user unmutes. This is great for video because is means the webcam light turns off giving the user confidence that nobody can see them, but for audio, mobile browsers tend to kill the page unless we keep a mic device open to convince it that we're on a call (experimentally, iOS currently works with the mic muted but Android unloads the page fairly quickly). Additionally, users expect to be able to speak as soon as they hit the unmute button, but opening an audio device can take time so we could lose the start of what someone is saying, especially if they're using a bluetooth mic which takes longer to open the device.

This will mean that we'll keep any mic indicators on even when muted, but this seems fairly standard behaviour (Jitsi does it) and on mobile is actually somewhat reassuring to know that the call is still there in the background.

tl;dr: Continue releasing video devices on video mute but keep audio devices open even when muted.

dbkr:
Use `publishTrack()` to explicitly publish the audio track when we connect, otherwise if we start muted, it will only be published once we unmute (see also comment).

Based on: https://github.com/vector-im/element-call/pull/1410
Fixes: https://github.com/vector-im/element-call/issues/1412

### DIFF
--- a/src/livekit/useECConnectionState.ts
+++ b/src/livekit/useECConnectionState.ts
@@ -14,7 +14,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { ConnectionState, Room, RoomEvent } from "livekit-client";
+import {
+  AudioCaptureOptions,
+  ConnectionState,
+  Room,
+  RoomEvent,
+} from "livekit-client";
 import { useCallback, useEffect, useRef, useState } from "react";
 import { logger } from "matrix-js-sdk/src/logger";
 
@@ -42,7 +47,28 @@ function sfuConfigValid(sfuConfig?: SFUConfig): boolean {
   return Boolean(sfuConfig?.url) && Boolean(sfuConfig?.jwt);
 }
 
+async function doConnect(
+  livekitRoom: Room,
+  sfuConfig: SFUConfig,
+  audioEnabled: boolean,
+  audioOptions: AudioCaptureOptions
+): Promise<void> {
+  await livekitRoom!.connect(sfuConfig!.url, sfuConfig!.jwt);
+  const audioTracks = await livekitRoom!.localParticipant.createTracks({
+    audio: audioOptions,
+  });
+  if (audioTracks.length < 1) {
+    logger.info("Tried to pre-create local audio track but got no tracks");
+    return;
+  }
+  if (!audioEnabled) await audioTracks[0].mute();
+
+  await livekitRoom?.localParticipant.publishTrack(audioTracks[0]);
+}
+
 export function useECConnectionState(
+  initialAudioOptions: AudioCaptureOptions,
+  initialAudioEnabled: boolean,
   livekitRoom?: Room,
   sfuConfig?: SFUConfig
 ): ECConnectionState {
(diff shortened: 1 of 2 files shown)


src/livekit/useECConnectionState.ts
```diff
@@ -89,12 +115,33 @@ export function useECConnectionState(
...
+      !sfuConfigValid(currentSFUConfig.current) &&
+      sfuConfigValid(sfuConfig)
+    ) {
+      // if we're transitioning from an invalid config to a valid one (ie. connecting)
+      // then do an initial connection, including publishing the microphone track:
+      // livekit (by default) keeps the mic track open when you mute, but if you start muted,
+      // doesn't publish it until you unmute. We want to publish it from the start so we're
+      // always capturing audio: it helps keep bluetooth headsets in the right mode and
+      // mobile browsers to know we're doing a call.
+      doConnect(
```
robintown, line 135: Should there be an `await` here?

src/room/InCallView.tsx
```diff
@@ -442,53 +427,66 @@ export function InCallView({
...
+  const parts = id.split(":");
+  if (parts.length < 2) {
+    logger.warn(
+      "Livekit participants ID doesn't look like a userId:deviceId combination"
+    );
+    return undefined;
+  }
+
+  parts.pop();
+  const userId = parts.join(":");
```
toger5: why do we reconstrcut the `userId` from the parts if we have `id` already?

src/room/InCallView.tsx
```diff
@@ -504,7 +502,14 @@ function useParticipantTiles(
             : false;
 
         const id = sfuParticipant.identity;
-        const member = matrixParticipants.get(id);
+        const member = findMatrixMember(matrixRoom, id);
+        // We always start with a local participant wit the empty string as their ID before we're
```
toger5:
```suggestion
        // We always start with a local participant with the empty string as their ID before we're
```

src/room/InCallView.tsx
```diff
@@ -504,7 +502,14 @@ function useParticipantTiles(
...
+        // connected, this is fine and we'll be in "all ghosts" mode.
+        if (id !== "" && member === undefined) {
+          logger.warn(
+            `Ruh, roh! No matrix member found for SFU participant '${id}': creating g-g-g-ghost!`
+          );
+        }
```
toger5: Is this the behaviour we want? I feel like we could also treat any sfu participant who is not part of the room state as nonexistent. Basically treating it like this: RoomState > sfu participants. If someone wants to be part of a rtc session the most important step is to get oneself into the room state, so that the matrix authentication/identity system is always the source of truth.

toger5:
@dbkr can we make this a PR ontop of the dbkr/smooth_focus_switch?
Otherwise its hard to review?
(But looking at smooth_focus_switch It seems that I should first review that and only review this one once it is merged)

toger5 (commented):
I will keep these comments but noticed that it makes more sense to start with `dbkr/smooth_focus_switch`.

src/livekit/useECConnectionState.ts
dbkr: It wouldn't cause anything to be awaited anyway since it's within a useEffect callback which can't be async. The one above needs to be in an async wrapper to wait for the previous call to finsh before running doConnect().

src/room/InCallView.tsx
dbkr: This was more another PR, but is because it was the easiest way of removing the last component which js doesn't have a built in way of doing (I haver switched the base now to see how that goes).

src/room/InCallView.tsx
dbkr: Potentially, but it's not what we've been doing up until now: this just continues doing what we were doing but logging it explicitly.

robintown (approved):
All good from my perspective

PR #1426, opened 2023-09-08
Implement new in-call header and footer designs

robintown:
:warning: This PR is stacked with https://github.com/vector-im/element-call/pull/1387, so please only review the later commits. Also, it's blocked until a new Compound Web version releases, due to some dependencies there.

Desktop:

Screenshot 2023-09-08 at 15-13-16 Element Call

Mobile:

IMG_20230908_151918

All headers and footers now have a max width of 1180px, per the designs:

Screenshot 2023-09-08 at 14-50-28 Element Call


joelselvaraj:
The `Share` button in the top right could be a bit confusing for some. In meetings, its very common to say "Share your screen/May I share my screen". The term is more linked with screen sharing. I think, it could be called `Invite` as in invite someone to the meeting. I think this should be considered by the design team. If a non-technical person sees this screen and i ask them to share their screen, i m pretty sure some would hit the share button on the top than the icon in the bottom with no text.... Just something to consider. 

P.S. you all do a wonderful job. thank you.

src/room/LobbyView.tsx
```diff
@@ -16,32 +16,39 @@ limitations under the License.
...
 
 interface Props {
+  client: MatrixClient;
   matrixInfo: MatrixInfo;
   muteStates: MuteStates;
   onEnter: () => void;
   isEmbedded: boolean;
   hideHeader: boolean;
+  participatingMembers: RoomMember[];
+  onShareClick: (() => void) | null;
```
SimonBrandner, line 40:
```suggestion
  onShareClick?: () => void;
```

Why not ^?

src/room/GroupCallView.tsx
```diff
@@ -90,16 +89,39 @@ export function GroupCallView({
...
+    roomAvatar,
+    roomEncrypted,
+    client,
+  ]);
+
+  const participatingMembers = useMemo(() => {
+    const members: RoomMember[] = [];
+    for (const [member, deviceMap] of participants.entries()) {
+      // Repeat the member for as many devices as they're using
+      for (let i = 0; i < deviceMap.size; i++) members.push(member);
```
SimonBrandner: Is this according to the designs? It feels a bit weird to me

SimonBrandner (approved):
LGTM

src/room/GroupCallView.tsx
robintown: The designs don't say anything about what should count as a participant, so this was just my doing. But good point, now that you mention it, counting multiple devices as a single participant sounds less surprising to me. I would note that also Element Web is currently doing it this way (repeating the member if they're on multiple devices), so we should change EW to be consistent eventually.

src/room/LobbyView.tsx
robintown: Because making the prop optional means that TypeScript won't warn you if you forget to provide it, and I think that failing to provide this callback should count as a mistake (it means that we've forgotten to implement the "share" button in some screen)

PR #1434, opened 2023-09-13
Add background gradient

SimonBrandner:


toger5:
Screenshot 2023-09-13 at 21 14 00
I am wondering if we could somehow make this look less awkward?

Is there an easy way to use an svg as the background of a label by mapping/texturing it with the coordinates of another element?
Maybe rounding the corners of the label a little bit could also look much better already

src/index.css
```diff
@@ -49,11 +49,17 @@ limitations under the License.
...
+  --background-gradient: url("graphics/backgroundGradient.svg");
 }
 
 .cpd-theme-dark {
   --stopgap-color-on-solid-accent: var(--cpd-color-text-primary);
   --stopgap-background-85: rgba(16, 19, 23, 0.85);
+
+  background-size: calc(max(100%, 100vw)) calc(max(100%, 100vh));
+  background-repeat: no-repeat;
+  background-position: bottom right;
```
robintown: After playing around with these parameters a bit, I think the following gets a little bit closer to how the graphic is sized and placed in Figma:

```suggestion
  background-size: calc(max(1440px, 100vw)) calc(max(800px, 100vh));
  background-repeat: no-repeat;
  background-position: center;
```

src/App.tsx
```diff
@@ -35,6 +40,26 @@ import { MediaDevicesProvider } from "./livekit/MediaDevicesContext";
...
+interface BackgroundProviderProps {
+  children: JSX.Element;
+}
+
+const BackgroundProvider = ({ children }: BackgroundProviderProps) => {
+  const { pathname } = useLocation();
+
+  useEffect(() => {
+    let backgroundImage = "";
+    if (!["/login", "/register"].includes(pathname)) {
```
robintown, line 51: Non-blocking: I think it would be more elegant and React-y to apply the background by putting a CSS class on the specific page-level components that are supposed to have it, rather than inspecting the URL like this. This would also allow the background CSS to be isolated to a single CSS module file, which I think would be desirable.

robintown (commented):
Looks nice! In the interest of time, please take my comments as non-blocking: for instance, a general comment is that the background shouldn't show up in embedded mode

toger5:
> Screenshot 2023-09-13 at 21 14 00 I am wondering if we could somehow make this look less awkward?
> 
> Is there an easy way to use an svg as the background of a label by mapping/texturing it with the coordinates of another element? Maybe rounding the corners of the label a little bit could also look much better already

This also can be treated as non blocking imo.

toger5 (approved):

PR #1445, opened 2023-09-17
Use native system fonts on Android and iOS

robintown:
This behavior is called for in the new designs, but Compound Web doesn't (yet) implement it by default.

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -158,7 +158,9 @@ body,
   flex-direction: column;
 }
 
-/* On Android and iOS, prefer native system fonts */
+/* On Android and iOS, prefer native system fonts. The global.css file of
+Compound Web is where these variables ultimately get consumed to set the page's
+font-family. */
 body[data-platform="android"] {
   --cpd-font-family-sans: "Roboto", "Noto", "Inter", sans-serif;
 }
--- a/src/index.css
+++ b/src/index.css
@@ -25,12 +25,6 @@ limitations under the License.
 @import "@vector-im/compound-web/dist/style.css";
 
 :root {
-  --font-family: "Inter", -apple-system, BlinkMacSystemFont, "Segoe UI",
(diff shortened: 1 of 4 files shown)


src/room/usePageUnload.ts
```diff
@@ -53,7 +39,7 @@ export function usePageUnload(callback: () => void) {
     }
 
     // iOS doesn't fire beforeunload event, so leave the call when you hide the page.
```
toger5, line 41:
```suggestion
    // iOS doesn't fire before unload event, so leave the call when you hide the page.
```
(not even part of your pr but worth changing?) 
beforeUnload also would work.

toger5 (approved):
Thanks, looks really good.

src/index.css
```diff
@@ -149,7 +143,6 @@ body {
   color: var(--cpd-color-text-primary);
   color-scheme: dark;
   margin: 0;
-  font-family: var(--font-family);
```
toger5, line 152: Can we just remove the font-family from the body css? 
where is `--cpd-font-family-sans` used to set the computed font family?

toger5 (dismissed):

src/room/usePageUnload.ts
dbkr: This is how its spelt when adding the listener, so I'd argue this is how it should be referred to: https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event

src/index.css
robintown: Yes, we can remove this because font-family is set in the global.css file of compound-web to be whatever --cpd-font-family-sans resolves to

dbkr (approved):

src/index.css
toger5: okay wasnt awar of this. Thanks for clarifying.

PR #1446, opened 2023-09-17
Implement the new modal designs

robintown:
:warning: **Based on https://github.com/vector-im/element-call/pull/1445**, please only review the new commits

They aren't yet used anywhere, but this will let us move on to implementing specific modal interactions from the new designs.

I made the design decision of making this new Modal component always be controlled by an explicit `open` state, which was inspired by some work I did with Jetpack Compose recently, where I saw that this makes state management and the behavior of components so much more obvious.

Here is what they will look like, when used (see https://github.com/vector-im/element-call/pull/1447):

Screen Shot 2023-09-17 at 14 43 42
Screenshot from 2023-09-17 14-45-15

### DIFF
--- a/src/NewModal.module.css
+++ b/src/NewModal.module.css
@@ -1,0 +1,217 @@
+/*
+Copyright 2022 - 2023 New Vector Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+.overlay {
+  position: fixed;
+  z-index: 100;
+  inset: 0;
+  background: rgba(3, 12, 27, 0.528);
+}
+
+@keyframes fade-in {
+  from {
+    opacity: 0;
+  }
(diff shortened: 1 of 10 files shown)


src/NewModal.module.css
```diff
@@ -0,0 +1,217 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+.overlay {
+  position: fixed;
+  z-index: 100;
+  inset: 0;
+  background: rgba(3, 12, 27, 0.528);
```
toger5, line 21: Dont we have any matching color in compound `--cpd-color...`?

src/NewModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+export interface ModalProps extends AriaDialogProps {
+  title: string;
+  children: ReactNode;
+  className?: string;
+  /**
+   * The controlled open state of the modal.
+   */
+  // An option to leave the open state uncontrolled is intentionally not
+  // provided, since modals are always opened due to external triggers, and it
+  // is the author's belief that controlled components lead to more obvious code
```
toger5:
```suggestion
  // is the author's belief that controlled components lead to more obvious code.
```

toger5 (approved):
Makes lots of sense.
Thanks for making this a very cohevise chunk of logic.
(Added a missing `.` but this is non blocking.)

src/NewModal.module.css
robintown: Not that I could find, and even if there were, I don't think it would be possible to use a color variable since this value is supposed to be exactly the same in light and dark mode

PR #1447, opened 2023-09-17
Start using the new modal component

robintown:
:warning: **Based on https://github.com/vector-im/element-call/pull/1446**, please only review the new commits

This attempts to converge all our modals on the new modal component while changing their designs as little as possible. This should reduce the bundle size a bit and make the app generally feel like it's converging on the new designs, even though individual modals still remain to be revamped.

Screen Shot 2023-09-17 at 14 43 42
Screenshot from 2023-09-17 14-45-15


src/livekit/MediaDevicesContext.tsx
```diff
@@ -200,8 +200,10 @@ export const useMediaDevices = () => useContext(MediaDevicesContext);
...
+export const useMediaDeviceNames = (context: MediaDevices, enabled = true) =>
   useEffect(() => {
-    context.startUsingDeviceNames();
-    return context.stopUsingDeviceNames;
-  }, [context]);
+    if (enabled) {
+      context.startUsingDeviceNames();
+      return context.stopUsingDeviceNames;
+    }
+  }, [context, enabled]);
```
toger5, line 209: This is very similar to what I did for fixing the duplicate stream issue. Hence, there will be conflicts. But should be easy to fix.

toger5 (approved):
Looking through this I definilty agree on the choice of making the open a prop.

PR #1448, opened 2023-09-17
Add a prompt to launch Element X on mobile

### LINKED ISSUE #1429: ElementX embedding: Handle other domains than call.element.io: SPA to prompt mobile users to use EX (via `element://` or stay in the web browser)

robintown:
:warning: **Based on https://github.com/vector-im/element-call/pull/1446**, please only review the new commits

This shows a bottom sheet on mobile asking the user whether they want to open the call in Element X, as soon as the page is loaded.

Screenshot_20230917-174217_Fennec

Closes https://github.com/vector-im/element-call/issues/1429


src/room/AppSelectionModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+        ? window.location.href
+        : getRoomUrl(roomId, roomSharedKey ?? undefined)
+    );
+    // Edit the URL so that it opens in embedded mode
+    url.hash = editFragmentQuery(url.hash, (params) => {
+      params.set("isEmbedded", "");
+      return params;
+    });
+
+    const result = new URL("element://call");
```
sandhose, line 62: Not sure where that URI was decided, but this URL has `call` as hostname, which doesn't mean anything. URI without hostnames should only have a single slash, e.g. `element:/call`.
Moreover, to avoid clashes with any other custom scheme handler, we should use a vendor-specific scheme, with a reverse DNS notation, e.g. `io.element:/call`

src/room/AppSelectionModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const appUrl = useMemo(() => {
+    // If the room ID is not known, fall back to the URL of the current page
+    const url = new URL(
+      roomId === null
+        ? window.location.href
+        : getRoomUrl(roomId, roomSharedKey ?? undefined)
+    );
+    // Edit the URL so that it opens in embedded mode
+    url.hash = editFragmentQuery(url.hash, (params) => {
+      params.set("isEmbedded", "");
```
toger5, line 58: why do we need isEmbedded for the non-widget version in the webview?

toger5 (approved):
(Ignoring all the code from other PR's)
The new bits look good, there is just a question about the isEmbedded param.

src/room/AppSelectionModal.tsx
robintown: Good point, I don't think it's possible to change this immediately since Element X needs to be updated too, but I will let the teams involved know

src/room/AppSelectionModal.tsx
robintown: Good question, it's for two reasons:
- So that the app will limit the user to only visiting this specific room (because from my understanding, SPA support within Element X is meant to be more of a view-only thing)
- So that when the SPA is opened within Element X, it knows to *not* show the prompt.

I'll leave these things in a comment.

PR #1457, opened 2023-09-18
Skip `useMediaDeviceNames` in SettingsModal

toger5:


toger5:
this indeed fixes our echo cancellation.
(Bot of course also breaks device selection with proper device names)

toger5:
#### This is not actually fixing it
This seems to be a firefox issue that is caused when selecting a device for the output. (Also tracked in the firefox bugzilla)

The simplest workouround could be removing the output device selector for firefox.
+ we should investigate why showing the picker/SettingsModal is updating the output device from (risky assumption) "" -> "actual_device_id123456"

see also: https://bugzilla.mozilla.org/buglist.cgi?quicksearch=echo

PR #1459, opened 2023-09-18
Check for existing track before creating one.

toger5:
In `doConnect` a new track is created and published. If we are unmuted this leads to two tracks. In this PR we check if there is already a published track and only create the track if there is none.

### DIFF
--- a/src/livekit/useECConnectionState.ts
+++ b/src/livekit/useECConnectionState.ts
@@ -17,8 +17,10 @@ limitations under the License.
 import {
   AudioCaptureOptions,
   ConnectionState,
+  LocalTrackPublication,
   Room,
   RoomEvent,
+  Track,
 } from "livekit-client";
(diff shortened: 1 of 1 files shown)


src/livekit/useECConnectionState.ts
```diff
@@ -54,16 +56,24 @@ async function doConnect(
   audioOptions: AudioCaptureOptions
 ): Promise<void> {
   await livekitRoom!.connect(sfuConfig!.url, sfuConfig!.jwt);
-  const audioTracks = await livekitRoom!.localParticipant.createTracks({
-    audio: audioOptions,
+  const hasMicrofonTrack = Array.from(
```
dbkr: `hasMicrophoneTrack` :)

dbkr (approved):

toger5:
livekitRoom seems to be undefined. Do not merge yet.

PR #1481, opened 2023-09-18
Fix app selection prompt appearing even in the app

robintown:
Turns out, I just got the name of this URL parameter wrong

### DIFF
--- a/src/room/AppSelectionModal.tsx
+++ b/src/room/AppSelectionModal.tsx
@@ -55,7 +55,7 @@ export const AppSelectionModal: FC<Props> = ({ roomId }) => {
     // room in question, and it prevents this app selection prompt from being
     // shown a second time.
     url.hash = editFragmentQuery(url.hash, (params) => {
-      params.set("isEmbedded", "");
+      params.set("embed", "");
       return params;
     });
 


dbkr (approved):
ah, good catch

toger5 (changes requested):
I am wondering if this collides with E2EE.
https://github.com/vector-im/element-call/blob/5fa98f0afabf463c08ee08abc8538d224443638a/src/e2ee/sharedKeyManagement.ts#L77-L81
IIRC this sets up per sender e2ee in case we are in embedded mode. Which is not what we want for this kind of embedding.
I think we should change that in the `useIsRoomE2EE` fn.
I still feel like we should make a decision what the params mean:
 - embed: Should only affect user flows and styling BUT should not touch any actual behaviour changes.

Maybe as part of this PR, the docs here:
https://github.com/vector-im/element-call/blob/5fa98f0afabf463c08ee08abc8538d224443638a/src/UrlParams.ts#L28-L32
can be updated to explain what exactly embedded is doing/should be used for.

(I think even for the release we need to already patch the condition in the linked code to use `if(widget)` or the widget_id url param.
maybe:
```
    isUsingWidgetApi: hasParam("widget_id"),
```
also makes it more obvious that embed != widget mode?

PR #1482, opened 2023-09-18
Remove auto-join behavior from embedded mode

robintown:
This was a hack that we did back when we were working on PTT, to make the joining process for PTT more seamless, but it doesn't make much sense to auto-join normal calls without giving the user a chance to turn off / adjust their media. If we want this behavior back eventually, I think it would be better serviced by a separate URL parameter.


dbkr (approved):
Makes sense

PR #1485, opened 2023-09-18
Implement new lobby design

robintown:
Screen Shot 2023-09-18 at 15 49 10
Screen Shot 2023-09-18 at 15 49 03
Screen Shot 2023-09-18 at 15 48 38


toger5:
Why are the designs for mobile landscape/portrait mode so different?
I would have expected sth like this:
landscape

dbkr (approved):
Looks good I think, thank you!

robintown:
@toger5 Interesting question, I think an issue with that kind of layout is that the 'join call' button is not in an central enough location to be as obvious as it should? But I would defer to Nad or Aaron.

toger5:
Just played around and I really do like the portrait design! it looks great! Very fitting for mobile.

toger5:
> @toger5 Interesting question, I think an issue with that kind of layout is that the 'join call' button is not in an central enough location to be as obvious as it should? But I would defer to Nad or Aaron.

This is why I am so convinced by the: call buttons on the right or left side:
landscape2
It just makes sense to me since eveything else steals all the vertical space and makes it all super tight.

(But you are totally right. not a discussion for here but for design. Sorry for the tangent.)

PR #1486, opened 2023-09-18
Fix home page overflowing by a few pixels

robintown:
Because the height of our header component changed at some point, the hard-coded height values in the CSS were off by a few px and caused the page to overflow slightly.


dbkr (approved):
Ah, I wonder if this is how Daniel saw scrollbars the other day

PR #1493, opened 2023-09-18
Make touchscreen detection more robust

robintown:
Empirically, Chrome on Android can end up not matching (hover: none), but still matching (pointer: coarse). 😕


dbkr (approved):
Btw, for these things I would generally just put that explanatory comment in the code as a comment rather than in the PR description. I think it's useful context to have.

PR #1497, opened 2023-09-19
Untangle the semantics of isEmbedded

robintown:
This deletes the `isEmbedded` flag from UrlParams, replacing it with an alternative set of flags that I think is more sensible and well-defined. Thanks to Timo for the suggestion in https://github.com/vector-im/element-call/pull/1481#pullrequestreview-1632049736

### DIFF
--- a/src/UrlParams.ts
+++ b/src/UrlParams.ts
@@ -146,10 +146,6 @@ class ParamParser {
   // Normally, URL params should be encoded in the fragment so as to avoid
   // leaking them to the server. However, we also check the normal query
   // string for backwards compatibility with versions that only used that.
-  hasParam(name: string): boolean {
-    return this.fragmentParams.has(name) || this.queryParams.has(name);
-  }
-
   getParam(name: string): string | null {
     return this.fragmentParams.get(name) ?? this.queryParams.get(name);
   }
@@ -159,6 +155,11 @@ class ParamParser {
       ...this.fragmentParams.getAll(name),
       ...this.queryParams.getAll(name),
     ];
+  }
+
+  getFlagParam(name: string, defaultValue = false): boolean {
+    const param = this.getParam(name);
+    return param === null ? defaultValue : param !== "false";
   }
 }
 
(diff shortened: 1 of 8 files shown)


dbkr (approved):
Thanks, agree that naming the flags after the behaviour is a lot simpler and clearer.

src/UrlParams.ts
```diff
@@ -101,6 +110,10 @@ interface UrlParams {
   password: string | null;
 }
 
+// This is here as a stopgap, but what would be far nicer is a function that
+// takes a UrlParams and returns a query string. That would enable us to
+// consolidate all the data about URL parameters and their meanings to this one
+// file.
```
toger5, line 116: 100%

src/room/AppSelectionModal.tsx
```diff
@@ -50,12 +50,11 @@ export const AppSelectionModal: FC<Props> = ({ roomId }) => {
...
-    // Edit the URL so that it opens in embedded mode. We do this for two
-    // reasons: It causes the mobile app to limit the user to only visiting the
-    // room in question, and it prevents this app selection prompt from being
-    // shown a second time.
+    // Edit the URL to prevent the app selection prompt from appearing a second
+    // time within the app, and to keep the user confined to the current room
     url.hash = editFragmentQuery(url.hash, (params) => {
-      params.set("embed", "");
+      params.set("appPrompt", "false");
+      params.set("confineToRoom", "");
```
toger5: What is the advantage of making `"addPrompt"` a boolean while `"confinedToRoom"`, `"hideHeader"` ... just toggle by either beeing part of the url or now (this is what I mean: `...=prevVla&hideHeader=&nextprop...`)

toger5 (approved):
I really like these changes. Also the wording and seperation makes a lot of sense to me.

I will approve but I am wondering about the "true"/"false" choice. If there is a good reason behind it feel free to merge immediately.

src/room/AppSelectionModal.tsx
robintown: Good point, they can and should be handled the same way

toger5: This comment might be a little bit late. But I am wondering if we should enforce the default to be false and phrase things accordingly. So that the flags always are flags and not sometimes need to be `=flase` if the default is true and `=` if the defualt is false.

basically rename it to `hidePrompt`/`skipPrompt` is what I am thinking specifically.

robintown: That's a really good question - on the one hand, code becomes clearer if you always name boolean flags in a positive manner (so that `true` corresponds to the 'on' state), but on the other hand, having a clear default behavior for optional flags can be nice. I'm not sure what the best balance of these two concerns is, so let's bring it up at the team meeting.

Issue #1438, opened 2023-09-14
Audio for some members sometimes comes out of a different output device


dbkr:
### Steps to reproduce

Sometimes, audio for one participant will come out of a different audio device to the one that's selected and used for everyone else. A possible trigger may be someone joining later in the call? Also may be Firefox specific?

### Will you send logs?

No

toger5:
There is hope that this is also caused by the same issue that causes https://github.com/vector-im/element-call/issues/1437

PR #1510, opened 2023-09-19
Firefox audio output issues fix

### LINKED ISSUE #1437: No AEC on Firefox?
### Steps to reproduce

We have observed that some users are causing echo because the browser does not appear to be doing AEC. The common denominator seems to be Firefox: switching to Chrome fixes the problem.

Is it possible AEC doesn't work on Firefox with e2ee enabled?

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

toger5:
In firefox it causes issue to do a manaul audio output selection. Thos happen when the settings modal is opend.
As a user this can be observed as: Pressing on settings => all other participants can hear an echo.
This pr tries to disable the minimum so that the issue does not occur anymore.

Fixes: #1437 And hopefully: #1438

### DIFF
--- a/src/livekit/MediaDevicesContext.tsx
+++ b/src/livekit/MediaDevicesContext.tsx
@@ -125,7 +125,7 @@ interface Props {
 export const MediaDevicesProvider: FC<Props> = ({ children }) => {
   // Counts the number of callers currently using device names
   const [numCallersUsingNames, setNumCallersUsingNames] = useState(0);
-  const usingNames = numCallersUsingNames > 0 && !isFireFox();
+  const usingNames = numCallersUsingNames > 0;
 
   // Use output device names for output devices on all platforms except FF.
   const useOutputNames = usingNames && !isFireFox();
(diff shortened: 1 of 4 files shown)


toger5:
tested on android chrome + FF macos.

src/livekit/MediaDevicesContext.tsx
```diff
@@ -120,10 +123,17 @@ interface Props {
 }
 
 export const MediaDevicesProvider: FC<Props> = ({ children }) => {
-  // Counts the number of callers currently using device names
+  // Counts the number of callers currently using device names.
   const [numCallersUsingNames, setNumCallersUsingNames] = useState(0);
   const usingNames = numCallersUsingNames > 0;
 
+  // Use output device names for output devices on all platforms except FF.
```
dbkr, line 130: I'm not sure this comment is really telling me anything that isn't apparent from the next line: the main question would be 'why'?

dbkr (approved):
otherwise lgtm!

robintown (approved):

PR #1528, opened 2023-09-19
Human readable URLs

dbkr:
This tries to implement the new scheme for element call URLs, including:
 * All rooms have an alias
 * Legacy unencrypted rooms keep their old alias
 * New encerypted rooms get an autogenerated alias so they have a string that can always be used to join them.
 * The room name is hyphenated and included in the URL

However this is buggy as it stands. Firstly, I am gettign initialisation weirdness where it keeps thinking I'm unregistered, then remembering all my rooms again when I enter a display name... *realises* oh, this is because it's looking for config.json in the current directory, ie. /c/.

Another bug is that it doesn't update the URL bar with the canonical URL for the room, which it probably ought to.

The main blocker, however, is that it turns out we use the absence of a canonical alias to tell whether a room is supposed to be an e2ee call room, so this would need that mechanism replaced.

### DIFF
--- a/src/home/useGroupCallRooms.ts
+++ b/src/home/useGroupCallRooms.ts
@@ -15,19 +15,16 @@ limitations under the License.
 */
 
 import { MatrixClient } from "matrix-js-sdk/src/client";
-import { GroupCall } from "matrix-js-sdk/src/webrtc/groupCall";
 import { Room } from "matrix-js-sdk/src/models/room";
-import { RoomMember } from "matrix-js-sdk/src/models/room-member";
 import { GroupCallEventHandlerEvent } from "matrix-js-sdk/src/webrtc/groupCallEventHandler";
 import { useState, useEffect } from "react";
 
 export interface GroupCallRoom {
-  roomAlias: string;
+  roomAlias?: string;
   roomName: string;
   avatarUrl: string;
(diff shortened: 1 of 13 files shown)


src/home/useGroupCallRooms.ts
```diff
@@ -87,23 +84,22 @@ export function useGroupCallRooms(client: MatrixClient): GroupCallRoom[] {
         return;
       }
 
+      // XXX: This should use client.matrixRTC.getActiveRoomSessions() not the old
+      // group call stuff!
```
dbkr, line 88: Mmm, not quite Dave, because that would only return sessions with people participating in them, which is not what we want, is it? What exactly *do* we want here? Just all the rooms, probably.

toger5:
@dbkr What do you think is the best way for this PR?
I am not sure what bits of it are already fixed now. Do you think it can be closed or is it doable to rebase it?
Otherwise an issue tracking the things that are still broken probably would be the best way?

toger5:
Is there still momentum for this. Do we still see this as sth we want to fix/improve?
If not I would advocate closing it.

dbkr:
Yeah, I think we're past this now - it's still here if we do want to come back to it but it's not getting merged as-is.

PR #1529, opened 2023-09-19
Use the room ID as the react key

dbkr:
room alias doesn't work now some rooms don't have aliases.


robintown (commented):
That looks like the name, rather than the ID

dbkr:
🤦 Thank you: fixed.

robintown (approved):

PR #1530, opened 2023-09-19
Don't auto create calls from visiting the URL

### LINKED ISSUE #1419: Two people going to call.foo/foobar puts them in two different calls
### Steps to reproduce

On PR 1348 with the new SPA e2e stuff:

1. Alice goes to call.url/foobar
2. Bob navigates to the same URL

#### What did you expect?
Probably either them to end up in the same unencrypted call or for them to just get an error telling them that we've turned e2ee on and they need to update their URLs (although the latter would break everyone's calendar links).

#### What happened instead?
They both end up in different calls!

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

dbkr:
This didn't work with e2e calls and just ended up with everyone who went to the URL creating their own room because it didn't add the alias to any of them.

This has it show a very simple 404-esque screen instead. If the call already exists, it will show it as before, so existing URLs will continue to work.

Screenshot 2023-09-19 at 17 56 46

Closes https://github.com/vector-im/element-call/issues/1419


src/room/GroupCallLoader.tsx
```diff
@@ -38,6 +41,9 @@ export function GroupCallLoader({
   const { t } = useTranslation();
   const groupCallState = useLoadGroupCall(client, roomIdOrAlias, viaServers);
 
+  const history = useHistory();
+  const onHomeClick = useCallback(() => history.push("/"), [history]);
```
robintown: I think if you include an `event.preventDefault()` in this handler, that will keep the page from refreshing unnecessarily

src/room/GroupCallLoader.tsx
```diff
@@ -48,6 +54,24 @@ export function GroupCallLoader({
...
       return <>{children(groupCallState.rtcSession)}</>;
     case "failed":
-      return <ErrorView error={groupCallState.error} />;
+      if ((groupCallState.error as MatrixError).errcode === "M_NOT_FOUND") {
+        return (
+          <FullScreenView>
+            <h1>{t("Call not found")}</h1>
+            <p>
+              {t(
+                "Element Calls are now end-to-end encrypted and need to be explicitly created. This helps make sure everyone's using the same encryption key."
```
robintown: Small copy suggestion: the brand name doesn't get translated, so I think we should stick with "calls", and a further hint as to where you go to create the call wouldn't hurt

```suggestion
                "Calls are now end-to-end encrypted and need to be created from the home page. This helps make sure everyone's using the same encryption key."
```

src/room/GroupCallLoader.tsx
```diff
@@ -48,6 +54,24 @@ export function GroupCallLoader({
     case "loaded":
       return <>{children(groupCallState.rtcSession)}</>;
     case "failed":
-      return <ErrorView error={groupCallState.error} />;
+      if ((groupCallState.error as MatrixError).errcode === "M_NOT_FOUND") {
+        return (
+          <FullScreenView>
+            <h1>{t("Call not found")}</h1>
+            <p>
```
robintown: Nit: new textual elements should use the compound-web Heading and Text components

src/room/GroupCallLoader.tsx
dbkr: Ah, of course - thanks.

robintown (approved):

PR #1533, opened 2023-09-19
Include the room name in the generated URL

### LINKED ISSUE #1359: Add human readable parts to SPA URLs

dbkr:
This isn't quite the URL scheme we were aiming for, for reasons specified on https://github.com/vector-im/element-call/pull/1528, but now they look like: 

https://localhost:3000/room/#/some-new-call?roomId=!hgUlSYoxISJnxhcecA:call.ems.host&password=J2A7zeVrmkvI8sGfBotZny7mnBgwsq7a

Fixes https://github.com/vector-im/element-call/issues/1359


src/home/CallList.tsx
```diff
@@ -17,11 +17,12 @@ limitations under the License.
 import { Link } from "react-router-dom";
 import { MatrixClient } from "matrix-js-sdk/src/client";
 import { RoomMember } from "matrix-js-sdk/src/models/room-member";
+import { Room } from "matrix-js-sdk";
```
robintown: I think there have been issues in the past with importing the top-level `"matrix-js-sdk"` import from multiple places? Not sure if those still exist

robintown (approved):
Code looks good, and as for the prefix: I honestly like keeping `/room/` as it is. It's more human-readable than a single-letter abbreviation, and helps tell the story of how this is all built on Matrix primitives.

PR #1550, opened 2023-09-20
fix url by providin a last &

### LINKED ISSUE #1548: From app, it asks if I want to join from browser or app
### Steps to reproduce

1. Open call.elemen.dev in your browser
2. Start a room
3. Join the room via Element X
4. Leave the room (still from ElX)
5. Reach the home screen (still from ElX), see the video below, it starts at this step
6. Join a recent room (still from ElX)
7. It asks if I want to join from the browser or app, but I'm inside ElX.

https://github.com/vector-im/element-call/assets/946104/10a55b3e-df72-4028-971f-a8e7ecf5a802

#### What did you expect?

Not having the browser-or-app prompt.

#### What happened instead?

I do have it.

### Operating system

iOS

### Browser information

Firefox and ElementX

### URL for webapp

call.element.dev

### Will you send logs?

No

toger5:
everything after the last & will be stripped away while loading the page (i think in the same step where we remove the PWD. THat most likely also would be the better fix. (not deleting it in the first place) but for the release that should do.
 - hence we loose the last param (usually confined to room...) 
- going home kills the all the params which we need to fix!

Fixes: #1548

### DIFF
--- a/src/room/AppSelectionModal.tsx
+++ b/src/room/AppSelectionModal.tsx
@@ -63,7 +63,7 @@ export const AppSelectionModal: FC<Props> = ({ roomId }) => {
     });
 
     const result = new URL("element://call");
-    result.searchParams.set("url", url.toString());
+    result.searchParams.set("url", url.toString() + "&");
     return result.toString();
   }, [roomId, roomSharedKey]);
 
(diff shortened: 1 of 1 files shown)


dbkr (approved):
Huh, okay - good that it's commented :)

PR #1554, opened 2023-09-20
New url schema

toger5:
Is that already enough?

Goes hadn in hand with: https://github.com/vector-im/element-x-ios/pull/1752

### DIFF
--- a/src/room/AppSelectionModal.tsx
+++ b/src/room/AppSelectionModal.tsx
@@ -62,7 +62,7 @@ export const AppSelectionModal: FC<Props> = ({ roomId }) => {
       return params;
     });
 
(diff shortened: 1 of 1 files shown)


src/room/AppSelectionModal.tsx
```diff
@@ -62,7 +62,7 @@ export const AppSelectionModal: FC<Props> = ({ roomId }) => {
       return params;
     });
 
-    const result = new URL("element://call");
+    const result = new URL("io.element.call:/call");
```
robintown: In the new schema, the path is supposed to be empty

```suggestion
    const result = new URL("io.element.call:/");
```

robintown (approved):

PR #1556, opened 2023-09-20
test without pwd removal 

toger5:
This pr is just to have an easy way to try in on the phone. Dont worry about merging it.


robintown:
I guess this PR has served its purpose

PR #1557, opened 2023-09-20
Fix race where app would be opened with no e2ee key

dbkr:
The key hadn't been extracted from the URL at the point the modal was mounted, so it just didn't get the key.

### DIFF
--- a/src/e2ee/sharedKeyManagement.ts
+++ b/src/e2ee/sharedKeyManagement.ts
@@ -35,7 +35,7 @@ const useInternalRoomSharedKey = (
   return [e2eeEnabled ? roomSharedKey : null, setRoomSharedKey];
 };
 
-const useKeyFromUrl = (roomId: string) => {
+const useKeyFromUrl = (roomId: string): string | null => {
   const urlParams = useUrlParams();
   const [e2eeSharedKey, setE2EESharedKey] = useInternalRoomSharedKey(roomId);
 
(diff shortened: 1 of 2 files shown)


src/e2ee/sharedKeyManagement.ts
```diff
@@ -35,36 +35,52 @@ export const useInternalRoomSharedKey = (
...
 
-  useEffect(() => {
-    if (!password) return;
-    if (password === "") return;
-    if (password === e2eeSharedKey) return;
+  const urlPassword = useKeyFromUrl(roomId);
 
-    setE2EESharedKey(password);
-  }, [password, e2eeSharedKey, setE2EESharedKey]);
+  const [e2eeSharedKey] = useInternalRoomSharedKey(roomId);
```
toger5, line 68: why do we use useInternalRoomSharedKey here. Shouldnt useRoomSharedKey always be saver?

src/e2ee/sharedKeyManagement.ts
```diff
@@ -35,36 +35,52 @@ export const useInternalRoomSharedKey = (
...
+
 export const useRoomSharedKey = (roomId: string): string | null => {
-  return useInternalRoomSharedKey(roomId)[0];
+  // make sure we've extracted the key from the URL first
+  // (and we still need to take the value it returns because
+  // the effect won't run in time for it to save to localstorage in
+  // time for us to read it out again).
+  const passwordFormUrl = useKeyFromUrl(roomId);
+
+  return useInternalRoomSharedKey(roomId)[0] ?? passwordFormUrl;
```
toger5, line 60: this is where the magic happens right?

src/room/AppSelectionModal.tsx
```diff
@@ -43,6 +44,13 @@ export const AppSelectionModal: FC<Props> = ({ roomId }) => {
   );
 
   const roomSharedKey = useRoomSharedKey(roomId ?? "");
+  const roomIsEncrypted = useIsRoomE2EE(roomId ?? "");
+  if (roomIsEncrypted && roomSharedKey === undefined) {
+    logger.error(
+      "Generating app redirect URL for encrypted room but don't have key available!"
+    );
+  }
```
toger5, line 52: this is really good so we are less blind going forward!

toger5 (approved):
Amazing catch.

src/e2ee/sharedKeyManagement.ts
dbkr: Well, it's being used for another thing too (comparing the password in the url to the saved one) so this would change the logic here too.

src/e2ee/sharedKeyManagement.ts
dbkr: Yep

robintown (approved):

PR #1559, opened 2023-09-20
Fix double audio tracks

robintown:


robintown:
https://github.com/vector-im/element-call/pull/1560

PR #1568, opened 2023-09-22
Remove triage automation

kittykat:
The triage board isn't being used, so no reason to have automation (it's also old style and should be modernised if there's interest in using it again)


SimonBrandner (approved):
LGTM

PR #1570, opened 2023-09-22
Refactor useIsRoomE2EE

dbkr:
Make it take a room object rather than a room ID to avoid it depending on a side effect, ie. if the room object input changes, the hook will be re-run but if we can't get the room from the room ID for whatever reason, we'd be stuck.

Also add logging on why we decided a room was e2ee.

### DIFF
--- a/src/e2ee/sharedKeyManagement.ts
+++ b/src/e2ee/sharedKeyManagement.ts
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 import { useEffect, useMemo } from "react";
-import { Room } from "matrix-js-sdk";
+import { Room } from "matrix-js-sdk/src/models/room";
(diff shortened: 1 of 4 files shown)


src/e2ee/sharedKeyManagement.ts
```diff
@@ -15,10 +15,11 @@ limitations under the License.
 */
 
 import { useEffect, useMemo } from "react";
+import { Room } from "matrix-js-sdk";
```
robintown: This is another top-level import (honestly, maybe it's not a problem?)

robintown (approved):

src/e2ee/sharedKeyManagement.ts
dbkr: mmm, better to be consistent probably

PR #1571, opened 2023-09-22
Revert "Refactor useIsRoomE2EE"

dbkr:
Reverts vector-im/element-call#1570

It appears to make the problem occur every time rather than fixing it :/


dbkr:
merging since it's a revert

PR #1576, opened 2023-09-24
Add url param token for call bridging

toger5:


toger5:
Close this as it would be too much effort to keep it up to date. The approach had significant security considerations anyways. I think we are very close with the guest access to provide a proper fallback solution in the near future.

PR #1589, opened 2023-09-25
Swap the microphone and video buttons

robintown:
Per feedback and discussion with design, we should keep them in the order they were previously.

### DIFF
--- a/src/room/LobbyView.tsx
+++ b/src/room/LobbyView.tsx
@@ -129,15 +129,15 @@ export const LobbyView: FC<Props> = ({
         <div className={inCallStyles.footer}>
           {recentsButtonInFooter && recentsButton}
           <div className={inCallStyles.buttons}>
-            <VideoButton
-              muted={!muteStates.video.enabled}
-              onPress={onVideoPress}
-              disabled={muteStates.video.setEnabled === null}
-            />
             <MicButton
               muted={!muteStates.audio.enabled}
               onPress={onAudioPress}
               disabled={muteStates.audio.setEnabled === null}
+            />
+            <VideoButton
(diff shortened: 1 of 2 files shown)


src/room/LobbyView.tsx
```diff
@@ -129,16 +129,16 @@ export const LobbyView: FC<Props> = ({
...
             <MicButton
               muted={!muteStates.audio.enabled}
               onPress={onAudioPress}
               disabled={muteStates.audio.setEnabled === null}
             />
+            <VideoButton
+              muted={!muteStates.video.enabled}
+              onPress={onVideoPress}
+              disabled={muteStates.video.setEnabled === null}
+            />
```
toger5, line 141: why do we not neet the `key=1 | 2` here?

toger5 (approved):
Looks good. I am wondering why there are no keys required in the lobby however. (or why there are keys needed in the incall view)

src/room/LobbyView.tsx
robintown: React only requires keys when you use an array of ReactElements as a child, because then it can't statically know which part of the code creates which child. But in this case, since we don't need to dynamically create a buttons array, React knows that the order of these elements is stable.

toger5: The question was mostly on what makes the incallview different.
If it is in a different container incall vs lobby there is different behaviour for the buttons in the two views. What makes it necassary to have different behaviour in those two view?

toger5: The question was mostly on what makes the incallview different.
If it is in a different container incall vs lobby there is different behaviour for the buttons in the two views. What makes it necassary to have different behaviour in those two view?

PR #1592, opened 2023-09-25
Use the loglevel library's extensions

dbkr:
...instead of monkey patching the console log objects. We use a logging framework everywhere now (this fixes the times when we didn't...) so there's not really a reason to do this the hacky way anymore.

This means that log lines now appear to come from whatever else is intercepting the logger (eg. sentry) rather than rageshake.ts.

Opinions on this welcome on whether it's better or not.

### DIFF
--- a/src/settings/rageshake.ts
+++ b/src/settings/rageshake.ts
@@ -577,7 +577,7 @@ type LogExtensionFunc = (
 type LogLevelString = keyof typeof LogLevel;
 
 /**
- * This method borrowed from livekit (who also ise loglevel and in turn essentially
+ * This method borrowed from livekit (who also use loglevel and in turn essentially
  * took loglevel's example honouring log levels). Adds a loglevel logging extension
  * in the recommended way.
  */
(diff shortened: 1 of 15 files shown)


src/settings/rageshake.ts
```diff
@@ -591,3 +560,42 @@ const getCircularReplacer = (): StringifyReplacer => {
...
+}
+
+type LogExtensionFunc = (
+  level: LogLevel,
+  ...rest: (Error | DOMException | object | string)[]
+) => void;
+type LogLevelString = keyof typeof LogLevel;
+
+/**
+ * This method borrowed from livekit (who also ise loglevel and in turn essentially
```
robintown:
```suggestion
 * This method borrowed from livekit (who also use loglevel and in turn essentially
```

robintown (approved):
This solution definitely feels cleaner to me, and I especially appreciate the lint rule being in place to keep us from accidentally using `console` :+1:

PR #1611, opened 2023-09-26
Re-enable livekit rageshake logging & with depth limit

dbkr:
Puts livekit logs back in the rageshake logs and adds a recursion limit to the object serialiser in rageshake.

### DIFF
--- a/src/settings/rageshake.ts
+++ b/src/settings/rageshake.ts
@@ -566,7 +566,7 @@ const getCircularReplacer = function (): StringifyReplacer {
         depth = depthMap.get(value) ?? 0;
       }
 
-      // 'this' is supposed to be the object the value was foudn in, according to
+      // 'this' is supposed to be the object the value was found in, according to
       // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify
       // but that doesn't seem to be the case. Instead, we do a pre-pass on the children here to
       // remember what depth we saw them at.
--- a/src/settings/rageshake.ts
+++ b/src/settings/rageshake.ts
@@ -551,10 +551,10 @@ type StringifyReplacer = (
 
 // From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value#circular_references
 // Injects `<$ cycle-trimmed $>` wherever it cuts a cyclical object relationship
-const getCircularReplacer = function (): StringifyReplacer {
+const getCircularReplacer = (): StringifyReplacer => {
   const seen = new WeakSet();
   const depthMap = new WeakMap<object, number>();
-  return (key: string, value: unknown): unknown => {
+  return function (this: unknown, key: string, value: unknown): unknown {
     if (typeof value === "object" && value !== null) {
       if (seen.has(value)) {
         return "<$ cycle-trimmed $>";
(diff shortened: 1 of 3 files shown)


src/settings/rageshake.ts
```diff
@@ -548,14 +551,30 @@ type StringifyReplacer = (
...
         return "<$ cycle-trimmed $>";
       }
       seen.add(value);
+
+      let depth = 0;
+      if (key) {
+        depth = depthMap.get(value) ?? 0;
+      }
+
+      // 'this' is supposed to be the object the value was foudn in, according to
```
robintown:
```suggestion
      // 'this' is supposed to be the object the value was found in, according to
```

src/settings/rageshake.ts
```diff
@@ -548,14 +551,30 @@ type StringifyReplacer = (
...
+      // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify
+      // but that doesn't seem to be the case. Instead, we do a pre-pass on the children here to
+      // remember what depth we saw them at.
```
robintown: I think the reason `this` isn't referring to the right thing may be because we're returning an arrow function here, and arrow functions bind `this` to a specific value at their time of creation. Does swapping it out for a `function (key, value) {}` work?

dbkr: Argh, I spent some time wondering why this still didn't work after swapping the arrow function for a regular one. Of course I had just changed the wrong function. 🤦 Thanks.

src/settings/rageshake.ts
```diff
@@ -550,12 +553,21 @@ type StringifyReplacer = (
...
     if (typeof value === "object" && value !== null) {
       if (seen.has(value)) {
         return "<$ cycle-trimmed $>";
       }
       seen.add(value);
+
+      let depth = 0;
+      if (this) {
+        depth = depthMap.get(this) ?? 0;
+      }
```
toger5, line 567:
```suggestion
      let depth = depthMap.get(this) ?? 0;
```
since the keys need to be object values this should be enough? `get(undefined)` is always undefined I think.

toger5 (approved):
Looks good but I am not familiar with the circular replacer. if `getCircularReplacer` is only called once and than the return fn is used recursivly this looks good.
Added code style proposals. Dont know if they make thanks less clear. (So approving as is but for me the styling changes do make it more readable)

src/settings/rageshake.ts
dbkr: > get(undefined) is always undefined I think.

```
> const x = new Map();
> x.set(undefined, "hello");
Map(1) {undefined => 'hello'}
> x.get(undefined);
'hello'
```

So... nope. Arguably since we only enter this block if value is an object, we should never add `undefined` as a key, but I would have to reason through all of that before understanding the code. To me, I think it would be a lot more readable with the explicit `if` statement since it follows the logic: if the object has a parent, look up its parent's depth in the map. If it has no parent, it's depth is zero.

toger5: With the WeekMap we are using it is not possible however?
```
> const depthMap = new WeakMap();
> undefined
> depthMap.set(undefined, 10)
> Uncaught TypeError: WeakMap key must be an object, got undefined
    <anonymous> debugger eval code:1
 ```
 But I see your point that this is clearer! Thanks for clarifying (looking into it again)

PR #1670, opened 2023-10-02
Show the device that's actually in use

### LINKED ISSUE #1646: Device Bug: Default device shown as used when re-added
### Steps to reproduce

1. Set element call as using the default device as input (let's call it mic1)
2. Start a call
3. Unplug the default device
4. EC switches to using whatever the new default device is (mic2)
5. Plug the device back in
6. EC is still using mic2, but in settings says it's using, "Default device - mic1"

Reproed on Chrome

#### What did you expect?
It should probably change back to actually using mic1 again?

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

dbkr:
Feeds the current track (if any), and therefore device it's using into the device selection dropdown and implement a some somewhat complex logic to observe when the 'default' device has changed to a different actual devbice between when we opened it and now, and display the device we're actually using as selected instead.

The comments should explain more, including the caveat that we can't do the same detective work if the device is one of several devices on the same bit of hardware. This is because is uses the same groupId comparison technique used by Livekit in `normalizeDeviceId` to do a smiliar thing.

Fixes https://github.com/vector-im/element-call/issues/1646

### DIFF
--- a/src/settings/SettingsModal.tsx
+++ b/src/settings/SettingsModal.tsx
@@ -18,6 +18,9 @@ import { ChangeEvent, Key, useCallback, useState } from "react";
 import { Item } from "@react-stately/collections";
 import { Trans, useTranslation } from "react-i18next";
 import { MatrixClient } from "matrix-js-sdk";
+import { useMaybeRoomContext } from "@livekit/components-react";
+import { LocalTrack } from "livekit-client";
+import { logger } from "matrix-js-sdk/src/logger";
 
 import { Modal } from "../Modal";
 import styles from "./SettingsModal.module.css";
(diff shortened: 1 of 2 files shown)


src/settings/SettingsModal.tsx
```diff
@@ -66,26 +69,79 @@ export const SettingsModal = (props: Props) => {
...
+        devices.selectedId === "default")
+    ) {
+      // we work out what the actual device is based on groupId, but this only works if
+      // there is only one such device with the same group ID, which there won't be if
+      // we're using hardware with multiple sub-devices (eg. a multitrack soundcard)
+      const usedGroupId =
+        trackUsedByRoom?.mediaStreamTrack.getSettings().groupId;
+      const devicesWithMatchingGroupId = devices.available.filter(
+        (d) => d.groupId === usedGroupId && d.groupId !== "default"
+      );
```
toger5: Similar to the last pr. Could the label also work?

src/settings/SettingsModal.tsx
```diff
@@ -66,26 +69,79 @@ export const SettingsModal = (props: Props) => {
...
+        !devices.selectedId ||
+        devices.selectedId === "default")
+    ) {
+      // we work out what the actual device is based on groupId, but this only works if
+      // there is only one such device with the same group ID, which there won't be if
+      // we're using hardware with multiple sub-devices (eg. a multitrack soundcard)
+      const usedGroupId =
+        trackUsedByRoom?.mediaStreamTrack.getSettings().groupId;
+      const devicesWithMatchingGroupId = devices.available.filter(
+        (d) => d.groupId === usedGroupId && d.groupId !== "default"
```
toger5: can the groupId ever be "default" ? Or doe we want to check for deviceId !== "default"

src/settings/SettingsModal.tsx
```diff
@@ -66,26 +69,79 @@ export const SettingsModal = (props: Props) => {
...
+  ) => {
     if (devices.available.length == 0) return null;
 
+    const trackUsedByRoom = lkSource ? getTrackInUse(lkSource) : undefined;
+
+    let selectedKey = devices.selectedId;
+    // We may present a different device as the currently selected one if we have an active track
+    // from the default device, because the default device may have changed since we acquired the
+    // track, in which case we want to display the one we're actually using rather than what the
+    // default is now.
```
toger5:
```suggestion
    // We may present a different device as the currently selected one if we have an active track
    // from the default device, because the default device of the OS may have changed since we acquired the
    // track, but EC did not update the track to match the new default, in which case we want to display the one we're actually using rather than what the
    // default is now.
```

toger5 (commented):
I am wondering if the other PR does obsolete this?
In the other PR the track is always updated in case the default device changes. So the track in use should always be matching the device and hence have the same/correct groupId.

I am wondering if this PR in combination with the other one would lead to the following behaviour:
 - default OS device is changed (actual-device-id:`a` -> `b`)
 - device id in the selected variable is still `default`
 - the group id of chromes default device changed to the groupId of `b` (also the label got updated to conatining the description of `b`)
 - track is restarted (becasue we do the group id check an see that `default` is a new device) (other PR) in this case its `b` now.
 - We reach the logic introduced here,
 - we pass the conditions (`device.selectedId = "default"`)
 - we find ` (d) => d.groupId === usedGroupId && d.deviceId !== "default"` (but there will be two devices with the correct groupId because the default device and the same device with deviceId = "someActualUUIDasTheID1234")
 - the found device will be the one where groupId == usedGroupId and deviceId will be the NOT "default" so the selector does not select the default device anymore
 - chainging the device in the os again (`a` -> `b`) works as normal but I am not sure EC still updates the device since now it does not check anymore if the group Id of the defualt device has changed.
 
So I am wondering if this would break EC following the system setting on chrome after the first system setting change.

toger5:
Thinking about this topic does livekit already check if the order of the devices changes on firefox?
> A [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that is fulfilled with an array of [MediaDeviceInfo](https://developer.mozilla.org/en-US/docs/Web/API/MediaDeviceInfo) objects. Each object in the array describes one of the available media input and output devices. The order is significant — the default capture devices will be listed first.

I am wondering if the ux would be much less broken with the following approach:
 - we have a checkbox: os default settings
 - if it is checked we only show what device we get as the first device of enumerateDevices and we react to order changes and update the media device accordnigly. (on chrome we always use id="default" and update using the groupId or label trick)
 - if it is unchecked a dropdown selector will appear to choose the device. We then NEVER use "default" as the device id and never react on os cahnges and instead only react to changes in the dropdown.

src/settings/SettingsModal.tsx
dbkr: This might involve comparing here which is tricky because chrome prepends 'Default: ' so we'd have to work around that and however it i18ns on the user's system.

src/settings/SettingsModal.tsx
dbkr: No, groupid is always the groupid of the underlying device, fortunately.

dbkr:
> I am wondering if the other PR does obsolete this?
Yes - hopefully :) Although the plan basically was to make sure the dropdown always shows the truth first, then implement the device switching, so even if the device switching doesn't work for whatever reason, it still shows the truth.

> I am wondering if this PR in combination with the other one would lead to the following behaviour [...]

It will switch back to device a when you change the default device in the OS a -> b -> a (I tested). The group ID of the default device should change just the same in either case.

> I am wondering if the ux would be much less broken with the following approach: [...]

Possibly! Do you think we should pause this and wait for product / design input?

src/settings/SettingsModal.tsx
toger5: Then I am quiet certain that you wanted to use deviceId in the comparison at the end of line 117 right?

toger5:
> It will switch back to device a when you change the default device in the OS a -> b -> a (I tested). The group ID of the default device should change just the same in either case.

Then I have a thinko, just to confirm, you tested with both PR's being merged at the same.

toger5:
> Yes - hopefully :) Although the plan basically was to make sure the dropdown always shows the truth first, then implement the device switching, so even if the device switching doesn't work for whatever reason, it still shows the truth.


Considering this,
> Possibly! Do you think we should pause this and wait for product / design input?

I think pushing this to the next release and asking design is a good idea.

src/settings/SettingsModal.tsx
dbkr: Ah, I see what you mean. I've changed that and also added a few more checks that we only do this in the right conditions.

hughns:
Making draft as this is stale.

toger5:
@robintown Does this fix sth that is still broken in v0.13.0?
I am not sure what we do with default devices as of now?

BillCarsonFr:
Close this PR as stale, let's create a new one if https://github.com/element-hq/element-call/issues/1646 get priorized

PR #1687, opened 2023-10-03
Switch capture devices if the default device changes

dbkr:
This is a bit of a hack, but is the only way I can see that we can update to using the new default device when the OS-level default changes. Hopefully the comments explain everything.

### DIFF
--- a/src/livekit/useLiveKit.ts
+++ b/src/livekit/useLiveKit.ts
@@ -164,6 +164,10 @@ export function useLiveKit(
         // we need to to see if the default device has changed to a different device
         // by comparing the group ID of the device we're using against the group ID
         // of what the default device is *now*.
+        // This is special-cased for only audio inputs because we need to dig around
+        // in the LocalParticipant object for the track object and there's not a nice
+        // way to do that generically. There is usually no OS-level default video capture
+        // device anyway, and audio outputs work differently.
         if (
           id === "default" &&
           kind === "audioinput" &&
--- a/src/livekit/useLiveKit.ts
+++ b/src/livekit/useLiveKit.ts
@@ -20,6 +20,7 @@ import {
   ExternalE2EEKeyProvider,
   Room,
   RoomOptions,
+  Track,
 } from "livekit-client";
 import { useLiveKitRoom } from "@livekit/components-react";
 import { useEffect, useMemo, useRef, useState } from "react";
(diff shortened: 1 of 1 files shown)


src/livekit/useLiveKit.ts
```diff
@@ -158,12 +159,54 @@ export function useLiveKit(
...
+          if (
+            defaultDevice &&
+            activeMicTrack &&
+            // only restart if the stream is still running: LiveKit will detect
+            // when a track stops & restart appropriately, so this is not our job.
+            // Plus, we need to avoid restarting again if the track is already in
+            // the process of being restarted.
+            activeMicTrack.mediaStreamTrack.readyState !== "ended" &&
+            defaultDevice.groupId !==
+              activeMicTrack.mediaStreamTrack.getSettings().groupId
```
toger5, line 192: I am wondering if checking for the label might be an option as well. It seems odd since the lable can be changed but the group id fails in the specifci case where the default device is switched to another device in the same group. (pc with two mics? does this exist?)

toger5: Using the label would summarize the logic to: switcht the device if it is different and also restart the track if the device has changed its label. Which is super similar to the group id but restarts the track unnecassarly if it is possible to rename the default device but does not miss a restart if the device has the same group id.

toger5 (approved):
This is a nice workaround to the tricky behaviour chrome has.
I left a comment wondering if the label could work to check if chrome chagned the default device.
How does all this work with audio outputs?

src/livekit/useLiveKit.ts
dbkr: Yeah, I think a sound card with multiple inputs would have the same group ID, not that I have one to test. Checking for when the label changes could work but I think it would require us to remember what the label was when we opened the device (which will presumably be racy because it could change between us enumerating devices and calling getUserMedia) and I think would need changes to Livekit.

toger5: dont you think: `activeMicTrack.mediaStreamTrack.getSettings().label` would be the label from what it was before updating the track. I would be very surpriced if they update the label, keep the groupId and also still use the old device?
The defaultDevice on the other hand is getting fetched again -> should be the new label.
I imagine the comparison to look like this: `"Default( macBook Pro mic )" !== "Default( otherMic )"`.

PR #1703, opened 2023-10-05
Keep the password in the URL

dbkr:
We changed our minds: people do copy the URL from the bar and give that to people and expect it to work: it doesn't make sense to prioritise shorter URLs over this. There's no security advantage unless we think there's a risk someone might steal your key by taking a photo of your monitor over your shoulder and decrypting the calls they can't already hear by standing behind you.


toger5 (approved):
Cool. Looks good. Thanks

PR #1704, opened 2023-10-05
Refactor room creation code a little

dbkr:
We c+ped the code to create room passwords between two places, but we already had a createRoom utility function that knew about e2ee.

### DIFF
--- a/src/matrix-utils.ts
+++ b/src/matrix-utils.ts
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 New Vector Ltd
+Copyright 2022-2023 New Vector Ltd
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


src/matrix-utils.ts
```diff
@@ -269,11 +272,17 @@ export function isLocalRoomId(roomId: string, client: MatrixClient): boolean {
   return parts[1] === client.getDomain();
 }
 
+interface CreateRoomResult {
```
toger5, line 275: In other instances its called the other way around: CreateRoomResult vs roomCreateResult. Is this intentional?

toger5 (approved):
This is way simpler to what it was before.

src/matrix-utils.ts
dbkr: Mmm, this probably could be more consistent - fixed

PR #1707, opened 2023-10-06
Switch devices if the label doesn't match the default

dbkr:


toger5:
@dbkr does it still make sense to keep this open?
What issue was it solving? Can we merge it (whats the reason for it still being a draft PR?)

toger5:
@dbkr I hope it is okay to ping you again on this topic.
If we cannot remember what the problem that this was solving I am advocating closing this and linking it into a (to create) media device EPIC issue as reference.

dbkr:
I remember this was some kind of awful hack, but empirically it's not needed so let's close it.

PR #1709, opened 2023-10-06
Decode passwords as buffers if we can

dbkr:
As comment hopefully explains. NB. this will make old & new versions of EC incompatible for e2ee calls since they'll be using different keys.

### DIFF
--- a/src/useE2eeConfig.ts
+++ b/src/useE2eeConfig.ts
@@ -32,7 +32,10 @@ export const useE2eeConfig = (
     // as such. Otherwise, we assume it's an old-style password and use it as a string.
     try {
       const itFunc = function* () {
-        const decoded = atob(passwordString);
+        const decoded = atob(
+          // built-in atob doesn't support base64url, so convert
+          passwordString.replace("-", "+").replace("_", "/")
+        );
         for (let i = 0; i < decoded.length; ++i) {
           yield decoded.charCodeAt(i);
         }
--- a/src/useE2eeConfig.ts
+++ b/src/useE2eeConfig.ts
@@ -1,0 +1,53 @@
+/*
+Copyright 2023 New Vector Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { useMemo } from "react";
+import { logger } from "matrix-js-sdk/src/logger";
+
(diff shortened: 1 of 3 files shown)


dkasak (approved):
Took a quick look and giving a quick ✅ from our side since it looks sane to me.

robintown (commented):
At face value the code looks fine, but I'm not sure why we would want to introduce extra complexity like this for presumably no user-facing benefit. Is there additional context I'm missing?

dbkr:
Yeah, this is exactly what made me not sure about whether this was worthwhile. The only user facing difference would be possibly maybe in security, since passing the string is *probably* still fine, but it's sort of hard to say as it's a lot harder to reason about cryptographically. I'm somewhat on the fence myself.

robintown:
Right, I mean I'm not a security expert and maybe I'm making a bad assumption, but I'd think that as long as the string has sufficient entropy, it's fine? Since LiveKit is hopefully running it through a KDF internally

robintown:
Here's where they derive the key, for reference: https://github.com/livekit/client-sdk-js/blob/01c4d025e1fca6e3b096673e65d39a760dbc8997/src/e2ee/KeyProvider.ts#L86

dkasak:
LiveKit uses PBKDF2 when a string is passed and HKDF when a buffer is passed. Both KDFs are secure but serve different purposes. PBKDF2 is computationally expensive by design, used generally for low-entropy keys like passwords, while HKDF is more efficient and specifically designed for high-entropy keys. So in this instance, HKDF is more clearly the right tool for the job. From a security standpoint, there's not a significant difference between the two KDFs in our case.

That said, HKDF is cryptographically "cleaner" for high-entropy keys because it avoids unnecessary computational overhead. Generating a buffer random bytes is also more straightforward and cryptographically sound than trying to craft a high-entropy alphanumeric string. It also leads to somewhat shorter URLs, which is also a (marginal) user-facing benefit.

In fact, if the buffer is undecodable, I would've gone for an error rather than treating it as an old-style string for clarity and a further reduction in complexity.

But if we still feel the complexity isn't worth it, we can close this out.

dbkr:
We can move to it being an error eventually but we need to maintain compatibility for old URLs.

src/useE2eeConfig.ts
```diff
@@ -0,0 +1,56 @@
...
+import { E2EEConfig } from "./livekit/useLiveKit";
+
+export const useE2eeConfig = (
+  passwordString: string | null
+): E2EEConfig | undefined => {
+  return useMemo(() => {
+    if (!passwordString) return undefined;
+
+    // Since crypto.getRandomValues() gives UInt8Arrays, we get a 128 bit buffer and end up
+    // base64 encoding it to get a text string. It's a bit silly to then pass this string to
```
robintown, line 29: A couple words explaining what makes it silly would be welcome (since that was the part I was skeptical about)

robintown (approved):
Alright, well, this makes sense to me now, so it's up to you @dbkr whether you feel like merging it

dbkr:
TBH I'm not sure this is worth the extra compatibility woes. Maybe we just settle for using buffers the more sensible way in per-user key mode and eventually migrate to that, and it's fine.

PR #1717, opened 2023-10-09
Use base64url encoding for the password param

dbkr:
As base64 is fairly obviously not sensible for URLs and we were not URL encoding it so we were ending up with spaces in the URL.

Also base 64 encode the password in case, as per comment.

### DIFF
--- a/src/matrix-utils.ts
+++ b/src/matrix-utils.ts
@@ -77,6 +77,11 @@ function secureRandomString(entropyBytes: number): string {
   const key = new Uint8Array(entropyBytes);
   crypto.getRandomValues(key);
   // encode to base64url as this value goes into URLs
+  // base64url is just base64 with thw two non-alphanum characters swapped out for
+  // ones that can be put in a URL without encoding. Browser JS has a native impl
+  // for base64 encoding but only a string (there isn't one that takes a UInt8Array
+  // yet) so just use the built-in one and convert, replace the chars and strip the
+  // padding from the end (otherwise we'd need to pull in another dependency).
   return btoa(
     key.reduce((acc, current) => acc + String.fromCharCode(current), "")
   )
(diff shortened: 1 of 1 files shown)


src/matrix-utils.ts
```diff
@@ -395,9 +399,13 @@ export function getRelativeRoomUrl(
...
   password?: string
 ): string {
+  // The password shouldn't need URL encoding here (we generate URL-safe ones) but encode
+  // it in case it came from another client that generated a non url-safe one
   return `/room/#${
     roomName ? "/" + roomAliasLocalpartFromRoomName(roomName) : ""
-  }?roomId=${roomId}${password ? "&" + PASSWORD_STRING + password : ""}`;
+  }?roomId=${roomId}${
+    password ? "&" + PASSWORD_STRING + encodeURIComponent(password) : ""
+  }`;
```
toger5: can we have logging/warnings here. This could be a footgun otherwise. Having it in the log we might spot that sth odd is going on with a password from another client.

toger5: (`if (encodeURIComponent(password) != password)`)

src/matrix-utils.ts
```diff
@@ -76,9 +76,13 @@ function waitForSync(client: MatrixClient) {
...
   const key = new Uint8Array(entropyBytes);
   crypto.getRandomValues(key);
+  // encode to base64url as this value goes into URLs
   return btoa(
     key.reduce((acc, current) => acc + String.fromCharCode(current), "")
-  ).replace(/=*$/, "");
+  )
+    .replace("+", "-")
+    .replace("/", "_")
+    .replace(/=*$/, "");
```
toger5, line 90: Is there a way to have a more generic function that guarantees that all url incompatible params get replaced?
Or do we know with certainty that those are all the possible once?

toger5: (synced with Dave: seems like there isnt something really great.)

toger5 (approved):
Thanks for the additions!

PR #1720, opened 2023-10-09
Always store room passwords with the right room ID

### LINKED ISSUE #1708: Sometimes get garbled audio on unencrypted call
### Steps to reproduce

As if one or more of the other parties are using encryption, or we are trying to decrypt and unencrypted stream. The audio is garbled mess in the same way as joining an e2ee call with the wrong key & video is colourful squares in the same way.

We believe we have repro steps now:

goto lobby view of https://call.element.io/foo via the recent calls list which then translates the lobby view to https://call.element.io/room/#/foo?roomId=!JRZQgnylhPiTdVbRPS:call.ems.host
copy and paste a URL with password into the URL bar and enter (e.g. https://call.element.io/room/#/bar?roomId=!oAYRbjCVjXxtkTOKjf:call.ems.host&password=a9YvrwP0YLQ87YgGFq8BDg)
then I get room-shared-key-!JRZQgnylhPiTdVbRPS:[call.ems.host](http://call.ems.host/) with value a9YvrwP0YLQ87YgGFq8BDg

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

dbkr:
Take the room ID from the URL rather than just assuming it's still the one that was in URL params before: if only the hash changes, the app won't reload.

Fixes https://github.com/vector-im/element-call/issues/1708


src/e2ee/sharedKeyManagement.ts
```diff
@@ -17,35 +17,37 @@ limitations under the License.
...
     if (!urlParams.password) return;
     if (urlParams.password === "") return;
     if (urlParams.password === e2eeSharedKey) return;
+    if (!urlParams.roomId) return;
 
-    setE2EESharedKey(urlParams.password);
-  }, [urlParams, e2eeSharedKey, setE2EESharedKey]);
+    setLocalStorageItem(
+      getRoomSharedKeyLocalStorageKey(urlParams.roomId),
+      urlParams.password
```
toger5:
```suggestion
      // We set the Item by only using data from the url. This way we
      // make sure, we always have matching pairs in the LocalStorage,
      // as they occur in the call links.
      getRoomSharedKeyLocalStorageKey(urlParams.roomId),
      urlParams.password
```

src/e2ee/sharedKeyManagement.ts
```diff
@@ -17,35 +17,37 @@ limitations under the License.
...
 const useKeyFromUrl = (roomId: string): string | null => {
   const urlParams = useUrlParams();
-  const [e2eeSharedKey, setE2EESharedKey] = useInternalRoomSharedKey(roomId);
+  const e2eeSharedKey = useInternalRoomSharedKey(roomId);
 
   useEffect(() => {
     if (!urlParams.password) return;
     if (urlParams.password === "") return;
     if (urlParams.password === e2eeSharedKey) return;
+    if (!urlParams.roomId) return;
```
toger5, line 47: why do we not need to check:
```suggestion
    if (!urlParams.roomId) return;
    if (urlParams.roomId != roomId) return;
```
(or why do we not use the url roomId for `const e2eeSharedKey = useInternalRoomSharedKey(roomId);`)

toger5 (approved):
Looks good.
Adding the code comment does not hurt to draw extra attention on why we are not using the roomId from the fn props.
The other comment is most likely me not exactly understanding why we have the roomId prop in the fn.

src/e2ee/sharedKeyManagement.ts
dbkr: Mmm, I'm looking at this now and passing the room ID in as well is sort of confusing and it's not 100% clear what exactly it should be doing with it. I figured we would probably want to store the password in any case, as long as we store it with the correct room ID: otherwise if there's a race then we'd risk not picking up the password from the URL at all.

dbkr: OK, refactored a bit - I think it should make a bit more sense now.

src/e2ee/sharedKeyManagement.ts
```diff
@@ -17,47 +17,62 @@ limitations under the License.
...
+ * @returns A tuple of the roomId and password from the URL if the URL has both,
+ *          otherwise [undefined, undefined]
+ */
+const useKeyFromUrl = (): [string, string] | [undefined, undefined] => {
   const urlParams = useUrlParams();
-  const [e2eeSharedKey, setE2EESharedKey] = useInternalRoomSharedKey(roomId);
 
   useEffect(() => {
     if (!urlParams.password) return;
     if (urlParams.password === "") return;
```
toger5: I think those can be grouped
```suggestion
    if (!urlParams.password || !urlParams.roomId) return;
```
Also  `if (urlParams.password === "") return;` can be removed i thinks since `!"" == true`, except if you like the additional explcity?

src/e2ee/sharedKeyManagement.ts
```diff
@@ -17,47 +17,62 @@ limitations under the License.
...
   // time for us to read it out again).
-  const passwordFormUrl = useKeyFromUrl(roomId);
+  const [urlRoomId, passwordFormUrl] = useKeyFromUrl();
 
-  return useInternalRoomSharedKey(roomId)[0] ?? passwordFormUrl;
+  const storedPassword = useInternalRoomSharedKey(roomId);
+
+  if (storedPassword) return storedPassword;
+  if (urlRoomId === roomId) return passwordFormUrl;
+  return undefined;
```
toger5, line 74: This makes sense immediately when looking at it I think.

src/e2ee/sharedKeyManagement.ts
```diff
@@ -17,47 +17,62 @@ limitations under the License.
...
+ * Extracts the room password from the URL if one is present, saving it in localstorage
+ * and returning it in a tuple with the corresponding room ID from the URL.
+ * @returns A tuple of the roomId and password from the URL if the URL has both,
+ *          otherwise [undefined, undefined]
+ */
+const useKeyFromUrl = (): [string, string] | [undefined, undefined] => {
   const urlParams = useUrlParams();
-  const [e2eeSharedKey, setE2EESharedKey] = useInternalRoomSharedKey(roomId);
 
   useEffect(() => {
```
toger5, line 45: why exactly do we still need the effect?
What would happen if we do:
```js
const useKeyFromUrl = (): [string, string] | [undefined, undefined] => {
  if (urlParams.roomId && urlParams.password) {
   setLocalStorageItem(
      getRoomSharedKeyLocalStorageKey(urlParams.roomId),
      urlParams.password
    );
    return [urlParams.roomId, urlParams.password];
  }
   return [undefined, undefined];
}

src/e2ee/sharedKeyManagement.ts
dbkr: Mmm, not sure why these were both there.

src/e2ee/sharedKeyManagement.ts
dbkr: It would grab the key out of the url on every re-render: writing to localsotrage is fairly fast, but probably not fast enough I'd want to block renders on it, and obviously we don't need to do it every time.

dbkr:
Assuming this is fine - merging

PR #1721, opened 2023-10-09
Fix mute button not being in sync with actual video/audio feed.

toger5:
This happens if we toggle the button while waiting for updating the stream. It is prohibited by checking if the stream state is in sync after the update is done.

### DIFF
--- a/src/livekit/useLiveKit.ts
+++ b/src/livekit/useLiveKit.ts
@@ -101,6 +101,11 @@ export function useLiveKit(
   // block audio from being enabled until the connection is finished.
   const [blockAudio, setBlockAudio] = useState(true);
 
+  // Store if audio/video are currently updating. If to prohibit unnecessary calls
+  // to setMicrophoneEnabled/setCameraEnabled
+  const [audioMuteUpdating, setAudioMuteUpdating] = useState(false);
+  const [videoMuteUpdating, setVideoMuteUpdating] = useState(false);
+
   // We have to create the room manually here due to a bug inside
   // @livekit/components-react. JSON.stringify() is used in deps of a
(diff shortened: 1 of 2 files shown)


dbkr (approved):

src/livekit/useLiveKit.ts
```diff
@@ -101,6 +101,11 @@ export function useLiveKit(
   // block audio from being enabled until the connection is finished.
   const [blockAudio, setBlockAudio] = useState(true);
 
+  // Store if audio/video are currently updating. If to prohibit unnecessary calls
+  // to setMicrophoneEnabled/setCameraEnabled
+  const [audioMuteUpdating, setAudioMuteUpdating] = useState(false);
+  const [videoMuteUpdating, setVideoMuteUpdating] = useState(false);
```
dbkr: I suspect you probably just want `useRef` for this, otherwise setting them is going to cause a re-render.

src/livekit/useLiveKit.ts
```diff
@@ -137,22 +142,47 @@ export function useLiveKit(
...
+
+      const syncMuteStateAudio = async () => {
+        if (
+          participant.isMicrophoneEnabled !== muteStates.audio.enabled &&
+          !audioMuteUpdating
+        ) {
+          setAudioMuteUpdating(true);
+          await participant
+            .setMicrophoneEnabled(muteStates.audio.enabled)
+            .catch((e) =>
```
dbkr: probably want this to be a try block too, to avoid mixing promises / await syntax

dbkr (approved):
Right yes, the joy of hooks and capturing closures.

PR #1724, opened 2023-10-10
Re-add creation of room by visiting the URL

dbkr:
Creates an unencrypted room if you land on an old-style simple URL forn a room that doesn't exist.

WIP: needs a message that displays to warn users they've created a legacy, unencrypted room. Should this message appear for only the creator or everyone who uses the room? Should it just appear for all encrypted rooms?


toger5:
Closing this since old style room likes are mostly gone nowadays.

PR #1730, opened 2023-10-11
Add skipLobby to params

### LINKED ISSUE #1734: Add skip lobby url flag

toger5:
Previously preload automatically was skipping the lobby. In our effort to make the url params more specific instead of hiding a bunch of changes behind one parameter that is named based on usage (~`embed`~ -> `confineToRoom`... ~`preload`~ -> `preload` + `skipLobby`).

Before preload was also skipping the lobby. Now its two different things so we can also skip the lobby without using preload (most likely for the mobile 1:1 call experience).
fixes: https://github.com/vector-im/element-call/issues/1734


src/room/GroupCallView.tsx
```diff
@@ -123,78 +125,78 @@ export function GroupCallView({
...
+      // In preload mode, wait for a join action before entering
+      const onJoin = async (ev: CustomEvent<IWidgetApiRequest>) => {
+        defaultDeviceSetup(ev.detail.data as unknown as JoinCallData);
         enterRTCSession(rtcSession);
-
-        PosthogAnalytics.instance.eventCallEnded.cacheStartCall(new Date());
-        // we only have room sessions right now, so call ID is the emprty string - we use the room ID
-        PosthogAnalytics.instance.eventCallStarted.track(
-          rtcSession.room.roomId
-        );
```
robintown: Why delete this code? I don't see anything that has replaced it

robintown (commented):
One question. Also, we should make sure to create a companion PR for Element Web adding this URL parameter, otherwise this will break EC integration on web

src/room/GroupCallView.tsx
toger5: ah this should have been its seperte PR. Since this was ported from an older branch I forgot about this. This was done twice:
```js
export function enterRTCSession(rtcSession: MatrixRTCSession) {
  PosthogAnalytics.instance.eventCallEnded.cacheStartCall(new Date());
  PosthogAnalytics.instance.eventCallStarted.track(rtcSession.room.roomId);

  // This must be called before we start trying to join the call, as we need to
  // have started tracking by the time calls start getting created.
  //groupCallOTelMembership?.onJoinCall();

  // right now we assume everything is a room-scoped call
  const livekitAlias = rtcSession.room.roomId;

  rtcSession.joinRoomSession([makeFocus(livekitAlias)]);
}
```

robintown (approved):
Looks good, and I've created the Element Web companion PR so this can be unblocked

PR #1731, opened 2023-10-11
Document url properties

### LINKED ISSUE #1732: Create documenation for elemetn call urls

toger5:
As discussed we want to have documenation of the url parameters inside this repo.
Fixes: https://github.com/vector-im/element-call/issues/1732

### DIFF
--- a/docs/url-params.md
+++ b/docs/url-params.md
@@ -72,7 +72,7 @@ roomId: string | null;
 Whether the app should keep the user confined to the current call/room.
 
 ```
-confineToRoom: boolean;
+confineToRoom: boolean; (default: false)
 ```
 
 **appPrompt**
@@ -80,7 +80,7 @@ Whether upon entering a room, the user should be prompted to launch the
 native mobile app. (Affects only Android and iOS.)
 
 ```
-appPrompt: boolean;
+appPrompt: boolean; (default: true)
 ```
 
 **preload**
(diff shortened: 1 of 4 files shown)


docs/url-params.md
```diff
@@ -0,0 +1,172 @@
...
+```
+lang: string | null;
+```
+
+**fonts**
+The fonts which the interface should use, if not empty.
+
+```
+fonts: string[];
+```
```
robintown: Note that this parameter's value is defined by a `?font=font-one&font=font-two...` in the URL - the name is different, not sure how you want to handle this in the docs

docs/url-params.md
```diff
@@ -0,0 +1,172 @@
...
+confineToRoom: boolean;
+```
+
+**appPrompt**
+Whether upon entering a room, the user should be prompted to launch the
+native mobile app. (Affects only Android and iOS.)
+
+```
+appPrompt: boolean;
+```
```
robintown: A lot of these params (for example this one) have default values that would be good to document

robintown (commented):
Thanks for doing this! I really like that you've documented the general URL format in detail

docs/url-params.md
toger5: oh...
I will have a look. Very important to document it then ;)

robintown (approved):

PR #1756, opened 2023-10-13
Re-enable screen sharing on Safari

### LINKED ISSUE #1762: Enable Screensharing on Safari (Desktop?)

dbkr:
Appears to work fine now, and no reason to think it shouldn't on Livekit.

We still force no animations on Safari which uses isSafari. There's no comment to say why, so it's a mystery. I've removed the comment about disabling screensharing which is no longer true.

Fixes https://github.com/vector-im/element-call/issues/1762


robintown (approved):

fkwp:
https://github.com/vector-im/element-call/issues/1762

PR #1759, opened 2023-10-13
Fix using a non-default audio device

dbkr:
We were passing the output option when we wanted the input, so the mic track pre-creation would just always use the system default.


toger5 (approved):
good find!!!

PR #1772, opened 2023-10-16
Support key-per-participant e2ee

### LINKED ISSUE #1278: E2EE first cut: Generate media key, send an encrypted room event on call join
### Your use case

This bug covers:
 * [x] Generating some random bytes to use for an E2EE media key
 * [x] Implementing a custom E2EEKeyProvider for LiveKit to provide per-participant keys
 * [x] Sending our key to the room as a room event when we join the call
 * [x] Reading media keys from room events when we join the call and adding them to our key provider
 * [x] Reading new events as new members join the call and adding them to our key provider

Things we removed from this bug:
 * Checking whether LiveKit will tell us when we're getting encrypted media or whether we need to signal it OOB: I suspect the latter because if the browser can't do E2EE then it won't b able to inspect the packets to discover that they're scrambled, in which case we should add a field to the call event (I think?) to announce the fact it's intended to be e2ee.
 * Reading the field of the state event and enabling / disabling e2ee accordingly
   * We used the absence of a canonical alias as the indicator of whether a room is e2ee or not instead of doing these

This gets a very basic first version of end to end encryption with automatic key exchange

_No response_

### Additional context

_No response_

dbkr:
Replaces https://github.com/vector-im/element-call/pull/1350
Requires https://github.com/matrix-org/matrix-js-sdk/pull/3667
Fixes https://github.com/vector-im/element-call/issues/1278


src/widget.ts
```diff
@@ -77,6 +77,8 @@ export const widget = ((): WidgetHelpers | null => {
       logger.info("Widget API is available");
       const api = new WidgetApi(widgetId, parentOrigin);
       api.requestCapability(MatrixCapabilities.AlwaysOnScreen);
+      api.requestCapabilityToSendEvent(EventType.CallEncryptionKeysPrefix);
+      api.requestCapabilityToReceiveEvent(EventType.CallEncryptionKeysPrefix);
```
robintown, line 81: Is there a matching PR yet for matrix-react-sdk to auto-approve these capabilities?

src/livekit/useLiveKit.ts
```diff
@@ -39,9 +40,16 @@ import {
   ECConnectionState,
   useECConnectionState,
 } from "./useECConnectionState";
+import { MatrixKeyProvider } from "../e2ee/matrixKeyProvider";
+
+export enum E2EEMode {
```
robintown: I'm confused why both this and `E2eeType` need to exist. Can we consolidate them?

robintown (approved):
Generally looks good to go

src/widget.ts
dbkr: Yep, https://github.com/matrix-org/matrix-react-sdk/pull/11783 (and another previously from Simon adding them)

src/livekit/useLiveKit.ts
dbkr: Hah, probably because I added one forgetting the other existed 😂

PR #1775, opened 2023-10-17
Add limit condition to sync loop and add more logging what is actually going on.

toger5:
This should prhibit the issue of the locked client. (the loop)

And also gives us more logging what is actually happening.
+ some refactoring


dbkr (commented):
Looks good but I think your prettier config needs updating or something.

robintown (approved):

PR #1808, opened 2023-10-20
Add hideControls url param for screen recordings.

toger5:
`showControls=false` 
Screenshot 2023-10-21 at 01 05 41

`hideHeader=true`
Screenshot 2023-10-21 at 01 05 34

reference without flags
Screenshot 2023-10-21 at 01 05 24

### DIFF
--- a/src/UrlParams.ts
+++ b/src/UrlParams.ts
@@ -65,7 +65,7 @@ interface UrlParams {
   /**
    * Whether the controls should be shown. For screen recording no controls can be desired.
    */
-  hideControls:boolean;
+  showControls: boolean;
   /**
    * Whether to hide the screen-sharing button.
    */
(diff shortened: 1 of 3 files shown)


dbkr (commented):
Shouldn't they be `showHeader=false` and `showControls=false` if we're sticking to all positive flags?

toger5:
> Shouldn't they be `showHeader=false` and `showControls=false` if we're sticking to all positive flags?

you are right changing these to positive values could also be part of this PR.
I think I will just introduce this here as a positive flag for now (showControls).
changign hideHeader requires work in multiple repos. So I think it will be easier if it becomes a seperate PR.

dbkr (approved):

dreamflasher:
Why is `hideHeader` removing the logo and the view toggle? And why is `hideControls` leaving the logo?
My expectation would be two flags: `hideHeader` and `hideFooter` and they do exactly this, not more, not less.
Oh and apparently the two flags currently can't be combined?

PR #1840, opened 2023-10-30
Fix not hanging up when using the close button (EXA EXI)

toger5:
This make sure that on Ios and Adroid the we always send the hangup event when the fromWidget Hangup event is sent.

(independent if it is triggered by a toWidget hangup event or pressing the in call hangup button)

### DIFF
--- a/src/room/GroupCallView.tsx
+++ b/src/room/GroupCallView.tsx
@@ -201,22 +201,6 @@ export const GroupCallView: FC<Props> = ({
   const [leaveError, setLeaveError] = useState<Error | undefined>(undefined);
   const history = useHistory();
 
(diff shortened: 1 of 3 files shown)


dbkr (approved):
Looks good. A suggestion - could the `widgetPostHangupProcedure` just live in `leaveRTCSession? The intention of this was to collect all the common stuff we need to do when leaving and move it out of GroupCallView.

toger5:
> Could the `widgetPostHangupProcedure` just live in `leaveRTCSession? The intention of this was to collect all the common stuff we need to do when leaving and move it out of GroupCallView.

This is a good idea. I did not notice that this specifically is a wrapper for the matrix-js-sdk which should contain such kind of specific things.

dbkr (approved):

PR #1842, opened 2023-10-30
Add f-twelve web console

toger5:
This is a very quick way to get access to the log on ios and andoird without needing to rageshake and check it on the rageshake.

Could come in handy as well to be able to execute arbitrary commands in the webconsole on andoid/ios.

## Screenshots

Screenshot 2023-10-30 at 15 05 14
Screenshot 2023-10-30 at 15 05 19

Screenshot 2023-10-30 at 15 05 24

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "buffer": "^6.0.3",
     "classnames": "^2.3.1",
     "events": "^3.3.0",
+    "f-twelve": "^2.3.4",
     "i18next": "^23.0.0",
     "i18next-browser-languagedetector": "^7.0.0",
     "i18next-http-backend": "^2.0.0",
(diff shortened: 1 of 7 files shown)


dbkr (approved):
If it's useful then I guess this is okay. Main concerns are:
 * How much does it add to the bundle size?
 * Does it add another thing intercepting our logs and if so does it play okay with everything else?
 * Hasn't been updated for 2 years and does pull in preact which gives it a bunch of dependencies, none of which have had updates in those 2 years.

toger5:
I spend some time searching for a more recent one but this seemed the only bigger project.
I basically had the same concerns. (preact, super old)
It also does not contain everythign we can find in the actual log of the browser console. It is still super helpful now.
Maybe just leafving it as a PR and rebasing it from time to time on the default branch is the best solution for now.

Then I can use this branch for debugging but its not really part of EC. (Until we find something newer with less deps.)

toger5:
Lets close this. I think we have enough ways to debug mobile nowadays.

PR #1854, opened 2023-10-31
Missing showControls

toger5:


dbkr (approved):
The nested ternary is more than my brain can handle, but yes, I think this makes sense.

PR #1891, opened 2023-11-10
Make E2EEConfig required

dbkr:
Previously it could be either undefined or type None which meant the same thing: no need to have both, just make it required.

This also means we can move the line to set e2ee enabled into a more sensible place rather than in the ActiveCall de-nulling wrapper.

### DIFF
--- a/src/room/GroupCallView.tsx
+++ b/src/room/GroupCallView.tsx
@@ -249,11 +249,13 @@ export const GroupCallView: FC<Props> = ({
     }
   }, [isJoined, rtcSession]);
 
-  const e2eeConfig = useMemo((): E2EEConfig | undefined => {
+  const e2eeConfig = useMemo((): E2EEConfig => {
     if (perParticipantE2EE) {
       return { mode: E2eeType.PER_PARTICIPANT };
     } else if (e2eeSharedKey) {
       return { mode: E2eeType.SHARED_KEY, sharedKey: e2eeSharedKey };
+    } else {
+      return { mode: E2eeType.NONE };
     }
   }, [perParticipantE2EE, e2eeSharedKey]);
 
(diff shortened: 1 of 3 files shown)


src/room/GroupCallView.tsx
```diff
@@ -249,11 +249,13 @@ export const GroupCallView: FC<Props> = ({
...
   }, [isJoined, rtcSession]);
 
-  const e2eeConfig = useMemo((): E2EEConfig | undefined => {
+  const e2eeConfig = useMemo((): E2EEConfig => {
     if (perParticipantE2EE) {
       return { mode: E2eeType.PER_PARTICIPANT };
     } else if (e2eeSharedKey) {
       return { mode: E2eeType.SHARED_KEY, sharedKey: e2eeSharedKey };
+    } else {
+      return { mode: E2eeType.NONE };
```
toger5, line 258: we also have the e2eeEnabled prop. maybe a check if `enableE2EE=false` should overrride this?

toger5 (commented):
looks good but I think in this context we should decide what we dot with `enableE2EE` https://github.com/vector-im/element-call/blob/04d41b8be1ac3b291db71afbf4fe7152293d1ba5/src/UrlParams.ts#L214-L215

It is somewhat redundant if missing perpartispant and missing shared key implies `enableE2EE` = false

src/room/GroupCallView.tsx
dbkr: ah, so this is a throwback from full mesh and passes the flag to the js-sdk to disable encryption of the events for calling, so not relevant for livekit. In fact, we should probably remove it altogether from the livekit branch as it won't do anything. This is probably not something for this PR though as it's not particularly dependant or related.

robintown (approved):

PR #1892, opened 2023-11-10
Keep screenshares when switching focus

### LINKED ISSUE #1413: Retain screenshare state on focus switch
### Your use case

When the call switches to a different livekit focus, any active screenshares should stay active (ie. resume in the new livekit session as they were).

I suspect this currently works by telling livekit to start screensharing and then it keeping the state internally which will get cleared on focus switch, so a good way of doing this might be to wrap it in a thing that takes such state via props and does whatever is necessary to the livekit room object to reconcile appropriately.

_No response_

### Additional context

_No response_

dbkr:
This also removes the use of the useLivekitRoom hook: we had reached the point where the only thing it was actually doing was disconnecting, so we now do that in the onClick handler for the leave button (I don't think we need to disconnect on unmount?). It was otherwise just getting in the way and causing tracks to be enabled/disabled when we didn't want them to be. This also removes the need for the blockAudio code.

Fixes https://github.com/vector-im/element-call/issues/1413
Based on: https://github.com/vector-im/element-call/pull/1891

### DIFF
--- a/src/livekit/useECConnectionState.ts
+++ b/src/livekit/useECConnectionState.ts
@@ -17,6 +17,7 @@ limitations under the License.
 import {
   AudioCaptureOptions,
   ConnectionState,
+  LocalTrack,
   Room,
   RoomEvent,
   Track,
@@ -55,8 +56,6 @@ async function doConnect(
   audioEnabled: boolean,
   audioOptions: AudioCaptureOptions,
 ): Promise<void> {
-  await livekitRoom!.connect(sfuConfig!.url, sfuConfig!.jwt);
-
   // Always create an audio track manually.
   // livekit (by default) keeps the mic track open when you mute, but if you start muted,
   // doesn't publish it until you unmute. We want to publish it from the start so we're
@@ -82,6 +81,8 @@ async function doConnect(
   }
   if (!audioEnabled) await audioTracks[0].mute();
 
+  logger.info("Pre-created microphone track");
+
   // check again having awaited for the track to create
   if (livekitRoom!.localParticipant.getTrack(Track.Source.Microphone)) {
     logger.warn(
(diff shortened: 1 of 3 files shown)


src/livekit/useLiveKit.ts
toger5, line 1: The `useLiveKit` file with the hook still seems to be here? Did I misunderstand the PR description? It sounded like this would be removed

dbkr: `useLivekitRoom` is the hook provided by Livekit components which is being removed. `useLivekit` is our own hook.

src/room/InCallView.tsx
```diff
@@ -200,8 +200,9 @@ export const InCallView: FC<InCallViewProps> = ({
   );
 
   const onLeavePress = useCallback(() => {
+    livekitRoom.disconnect();
     onLeave();
```
robintown, line 207: What about moving this call into `onLeave`? Would it make sense for it to live there?

src/livekit/useECConnectionState.ts
```diff
@@ -126,6 +167,31 @@ export function useECConnectionState(
...
+    const screenshareTracks: MediaStreamTrack[] = [];
+    for (const t of livekitRoom!.localParticipant.videoTracks.values()) {
+      if (t.track && t.source == Track.Source.ScreenShare) {
+        const newTrack = t.track.mediaStreamTrack.clone();
+        newTrack.enabled = true;
+        screenshareTracks.push(newTrack);
+      }
+    }
+
+    setSwitchingFocus(true);
```
robintown, line 182: Where does this get set back to false?

robintown (commented):
A couple questions

src/room/InCallView.tsx
dbkr: This is passed in from GroupCallView where we don't have the livekit room object (and is also called on a leave due to a disconnect) - commented.

src/livekit/useECConnectionState.ts
dbkr: Also commented.

robintown (approved):

PR #1895, opened 2023-11-14
Mark E2EEenabled flag as deprecated

toger5:
@dbkr If I understand your comment on the other PR right, this is the correct description for this flag.


dbkr (approved):
Yes, I think this is accurate.

PR #1902, opened 2023-11-16
Hide rooms we don't have the key for in recents list

dbkr:
This is an annoyingly large change duplicating a lot of the key management stuff into regular non-hooks functions since we can't call a hook in a loop. Feels like there should be a less terrible way of doing this.

### DIFF
--- a/src/e2ee/sharedKeyManagement.ts
+++ b/src/e2ee/sharedKeyManagement.ts
@@ -15,13 +15,18 @@ limitations under the License.
 */
 
 import { useEffect, useMemo } from "react";
+import { Room } from "matrix-js-sdk";
 
 import { setLocalStorageItem, useLocalStorage } from "../useLocalStorage";
 import { useClient } from "../ClientContext";
-import { useUrlParams } from "../UrlParams";
+import { UrlParams, getUrlParams, useUrlParams } from "../UrlParams";
 import { widget } from "../widget";
 
-export const getRoomSharedKeyLocalStorageKey = (roomId: string): string =>
+export function saveKeyForRoom(roomId: string, password: string): void {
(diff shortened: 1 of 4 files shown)


src/e2ee/sharedKeyManagement.ts
```diff
@@ -31,6 +36,21 @@ const useInternalRoomSharedKey = (roomId: string): string | null => {
...
+  const key = getRoomSharedKeyLocalStorageKey(roomId);
+  return localStorage.getItem(key);
+}
+
+function saveKeyFromUrlParams(urlParams: UrlParams): void {
+  if (!urlParams.password || !urlParams.roomId) return;
+
+  // We set the Item by only using data from the url. This way we
+  // make sure, we always have matching pairs in the LocalStorage,
+  // as they occur in the call links.
```
robintown: I must admit that this comment is confusing to read - it implies that there's some data we could use that we *avoid* using here? But what is that data?

src/e2ee/sharedKeyManagement.ts
```diff
@@ -74,12 +83,14 @@ export const useRoomSharedKey = (roomId: string): string | undefined => {
 
 export const useIsRoomE2EE = (roomId: string): boolean | null => {
```
robintown, line 85: Seeing `| null` on this return type makes me wonder what we were ever returning `null` for, and why we aren't doing that now

robintown (approved):

src/e2ee/sharedKeyManagement.ts
dbkr: Looks like we used to return null if we couldn't find the room: https://github.com/vector-im/element-call/blame/535712d108a3257501457ae24f497dc7d52ed671/src/e2ee/sharedKeyManagement.ts#L76 - unsure why that changed though.

PR #1908, opened 2023-11-20
Switch from Weblate to Localazy

t3chguy:
Fixes https://github.com/vector-im/customer-retainer/issues/158

And switch to translation keys over english-as-keys.

~Blocked on @dbkr adding the necessary repo secret as per DM~


robintown (approved):
Thanks, let's give this a go! I think given the potential for conflicts it's best to merge this ASAP, but I'll check briefly with the team

PR #1910, opened 2023-11-20
Fix joining calls with no media permission

### LINKED ISSUE #1661: Auto mute my audio and video when I decline mic/camera permissions

For calls when I don't expect to speak, I usually deny permissions to microphone and camera. But Element Call still has me unmuted on the pre-call screen.

So I don't have to remember to mute my audio and video and enter the call with black video and appearing unmuted.

Default me to muted on the pre-call screen after I decline mic/cam permissions.

_No response_

### Additional context

_No response_

### LINKED ISSUE #1907: EC fails if denied media permissions
### Steps to reproduce

1. Try to join a call (in a fresh browser so you get media permission prompts)
2. Deny media perms

#### What did you expect?
To be able to join the the call & see/hear people but not unmute mic/camera

#### What happened instead?
Sat on "waiting for participants" forever

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

Yes

dbkr:
You can still try to unmute your media in the preview screen and the button will show unmuted, but then mute again once you actually join, so not 100% perfect but better.

Fixes https://github.com/vector-im/element-call/issues/1907
Fixes https://github.com/vector-im/element-call/issues/1661

### DIFF
--- a/src/livekit/useLiveKit.ts
+++ b/src/livekit/useLiveKit.ts
@@ -245,11 +245,25 @@ export function useLiveKit(
               );
             }
           } catch (e) {
-            logger.error(
-              "Failed to sync audio mute state with LiveKit (will retry to sync in 1s):",
-              e,
-            );
-            setTimeout(() => syncMuteState(iterCount + 1, type), 1000);
+            if ((e as DOMException).name === "NotAllowedError") {
+              logger.error(
+                "Fatal errror while syncing mute state: resetting",
+                e,
+              );
+              if (type === MuteDevice.Microphone) {
+                audioMuteUpdating.current = false;
+                muteStates.audio.setEnabled?.(false);
+              } else {
+                videoMuteUpdating.current = false;
+                muteStates.video.setEnabled?.(false);
+              }
+            } else {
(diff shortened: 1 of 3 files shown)


src/livekit/useLiveKit.ts
```diff
@@ -245,11 +245,25 @@ export function useLiveKit(
...
-              e,
-            );
-            setTimeout(() => syncMuteState(iterCount + 1, type), 1000);
+            if ((e as DOMException).name === "NotAllowedError") {
+              logger.error(
+                "Fatal errror while syncing mute state: resetting",
+                e,
+              );
+              if (type === MuteDevice.Microphone) {
+                audioMuteUpdating.current = false;
```
dbkr, line 254: We also forgot to clear this flag on an exception so it would never update the mute state at all after an error(!)

robintown (approved):

src/livekit/useLiveKit.ts
robintown: Nice find

PR #1912, opened 2023-11-21
Show a warning on unencrypted media

robintown:
It looks like this:

Screenshot from 2023-11-20 10-58-04

Closes https://github.com/vector-im/voip-internal/issues/237

### DIFF
--- a/src/video-grid/VideoTile.tsx
+++ b/src/video-grid/VideoTile.tsx
@@ -36,13 +36,15 @@ import {
 } from "matrix-js-sdk/src/models/room-member";
 import MicOnSolidIcon from "@vector-im/compound-design-tokens/icons/mic-on-solid.svg?react";
 import MicOffSolidIcon from "@vector-im/compound-design-tokens/icons/mic-off-solid.svg?react";
-import { Text } from "@vector-im/compound-web";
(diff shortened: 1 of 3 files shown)


dbkr (approved):

src/video-grid/VideoTile.tsx
```diff
@@ -108,13 +114,22 @@ export const VideoTile = forwardRef<HTMLDivElement, Props>(
...
+    const videoInfo = useMediaTrack(
+      content === TileContent.UserMedia
+        ? Track.Source.Camera
+        : Track.Source.ScreenShare,
+      sfuParticipant,
+    );
+    const muted = audioInfo.isMuted !== false;
+    const encrypted =
+      audioInfo.publication?.isEncrypted !== false &&
+      videoInfo.publication?.isEncrypted !== false;
```
daniel-abramov, line 132: Perhaps a matter of style/preference, but I'd prefer to explicitly handle `undefined` instead of performing a comparison that would "convert" it into `boolean`.

```suggestion
    const muted = audioInfo.isMuted ?? false;
    const encrypted =
      audioInfo.publication?.isEncrypted ?? false &&
      videoInfo.publication?.isEncrypted ?? false;
```

fkwp (approved):

daniel-abramov (approved):

PR #1915, opened 2023-11-22
Show unencrypted media warnings in widget mode

robintown:
They weren't showing up because the video tiles didn't think that the call was encrypted in the per-participant keys case.


toger5 (approved):
Ah good catch. Thanks robin

PR #1918, opened 2023-11-22
I18n'ise hardcoded strings

t3chguy:


toger5 (approved):
Looks very much like replacing hardcoded strings with translated ones ;)
Thanks a lot!

PR #1920, opened 2023-11-24
Fix not disconnecting from livekit session.

toger5:
This happens if there is an error while
(due to opening the app in another tab)
the call is running. The lk session ends if we reroute to `/` (or quit the widget IFrame)
but we never explicitly disconnect from livekit.

This implementation pulls out the livekit hook from the `ActiveCall` view into the `GroupCallView` to be able to call `livekitRoom.disconnect` in in the corresponding effects.

An alternative would be to move the disconnect effects into the ActiveCall. I think this would be better but before doint the effor I would like to get another opinoin on this.

### DIFF
--- a/src/room/GroupCallView.tsx
+++ b/src/room/GroupCallView.tsx
@@ -44,9 +44,10 @@ import { useRoomAvatar } from "./useRoomAvatar";
 import { useRoomName } from "./useRoomName";
 import { useJoinRule } from "./useJoinRule";
 import { InviteModal } from "./InviteModal";
-import { E2EEConfig } from "../livekit/useLiveKit";
+import { E2EEConfig, useLiveKit } from "../livekit/useLiveKit";
 import { useUrlParams } from "../UrlParams";
 import { E2eeType } from "../e2ee/e2eeType";
+import { useOpenIDSFU } from "../livekit/openIDSFU";
 
 declare global {
   interface Window {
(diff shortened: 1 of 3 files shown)


daniel-abramov (approved):
The code has definitely changed a lot since I last checked in a detail, so I'm probably not the best person to review it, but from what I can see, the current PR primarily moves the usage of `useLiveKit` hook from one location to another, which does not look harmful (but we need to test it).

What I'm not sure is why it fixes the issue, i.e. if I got the description right, the issue only happens if there is an error during the call. I can very roughly assume that this caused a bug because in the previous version, the function that "ended" a call in case of it being open in a different tab, did not have access to the LiveKit room to also close it (it seems to be a big difference here).

If this is so, I think the fix may be sensible.

robintown (approved):
I'm a little worried that moving the invocation of the `useLiveKit` hook to GroupCallView is risky, but as long as we test this enough on call.element.dev it should be fine.

toger5:
> I'm a little worried that moving the invocation of the useLiveKit hook to GroupCallView is risky, but as long as we test this enough on call.element.dev it should be fine.

I fully agree.

> An alternative would be to move the disconnect effects into the ActiveCall. I think this would be better but before doint the effor I would like to get another opinoin on this.

Is this worthit to do this already. Or should we test on dev and then I will change it to this if there are issues coming up?

toger5:
I merged this now. We will refactor if this causes problems.

toger5:
We see problems that might be related -> 
> An alternative would be to move the disconnect effects into the ActiveCall. I think this would be better but before doint the effor I would like to get another opinoin on this.

This will be done

PR #1925, opened 2023-11-28
Refactor livekit disconnect to use an effect hook.

toger5:


fkwp (approved):

src/room/InCallView.tsx
```diff
@@ -85,29 +85,43 @@ import {
...
+
+  const cleanup = useCallback(() => {
+    if (connState === ConnectionState.Connected) {
+      livekitRoom?.disconnect();
+    }
+  }, [livekitRoom, connState]);
+
+  useEffect(() => {
+    return cleanup;
+  }, [cleanup, livekitRoom]);
```
daniel-abramov: Note that:
1. This clean-up function may be called multiple times.
2. The connection state could also be `Reconnecting` or `Connecting` in which case the `disconnect()` will not be called (meaning that we will connect to the room despite unmounting the `ActiveCall` leading to an unexpected bug for this edge-case).

I think the cleanest way to fix the problem is to do a clean effect that just disconnects the room on unmount without any further actions and checks (I don't think we need `connState` here), similarly to how LiveKit does it. I checked the LiveKit Meet example and how they do it. They use the `LiveKitRoom` component that relies on `LkRoomContext`. Here is how the clean-up is done: [link](https://github.com/livekit/components-js/blob/c5305c35cca3b8d08fc9f9652ef48acaa2f3db55/docs/storybook/.storybook/lk-decorators/RoomContext.tsx#L19C1-L19C1) which is just a typical React idiom for these sort of cases. Are there any strong reasons why we don't want to do it this way? 🙂

toger5: I had this initially. But was seeing it beeing called in the beginning, thats why I added:     `if (connState === ConnectionState.Connected) {`. But that was because i let copilot + tooling add the dependencies...

toger5: I now use exactly what you propose.

daniel-abramov (approved):

PR #1932, opened 2023-11-29
Make some corrections to the development guide

robintown:


toger5 (approved):
I wasnt even aware this is not in sync. Great you found/fixed it.

PR #1933, opened 2023-11-29
Fix the invite modal overflowing

robintown:
If the URL was the wrong shape, it could cause the modal to overflow, so it needs an extra nudge to line break in the right places.


toger5 (approved):
Good find!
Thanks Robin

PR #1961, opened 2023-12-01
Start refactoring some business logic into view models

robintown:
As Element Call grows in complexity, it has become a pain point that our business logic remains so tightly coupled to the UI code. In particular, this has made testing difficult, and the complex semantics of React hooks are not a great match for arbitrary business logic. Here, I show the beginnings of what it would look like for us to adopt the MVVM pattern. I've created a CallViewModel and TileViewModel that expose their state to the UI as rxjs Observables, as well as a couple of helper functions for consuming view models in React code.

This should contain no user-visible changes, but we need to watch out for regressions particularly around focus switching and promotion of speakers, because this was the logic I chose to refactor first.


src/livekit/MediaDevicesContext.tsx
```diff
@@ -83,14 +82,7 @@ function useMediaDevice(
...
   // kind, which then results in multiple permissions requests.
   const deviceObserver = useMemo(
-    () =>
-      createMediaDeviceObserver(
-        kind,
-        () => {
-          logger.error("Error creating MediaDeviceObserver");
-        },
-        requestPermissions,
-      ),
```
toger5: this looks like the livekit client got downgraded by one version. The current `createMediaDeviceObserver` has a on error calback as the second arg iirc.

toger5 (approved):
Looks good. Just a minor concern about the livekit client version

src/livekit/MediaDevicesContext.tsx
robintown: Good catch, have upgraded it

PR #1970, opened 2023-12-01
Fix a couple of visual glitches on Firefox

robintown:


toger5 (approved):
Cool you also found sth to fix the rounded corners.

Issue #1839, opened 2023-10-29
Rear camera is horizontally flipped on iOS


ara4n:
### Steps to reproduce

1. Start EC in EX as widget (yay!)
2. Switch to rear camera in settings
3. Observe the preview has left & right flipped, making it feel like you’re looking through a viewfinder on a twin lens reflex camera like a rolleiflex…

#### What did you expect?

left & right should not be flipped on rear camera previews

### Operating system

iOS 16 on iPhone 12 Pro Max

### Will you send logs?

No

PR #1989, opened 2023-12-07
Video tile redesign/refactor

### LINKED ISSUE #1764: 'Profile' button doesn't work after switching tabs in settings
### Steps to reproduce

1. Start on the home page as a logged in user
2. Open settings
3. Click the 'video' tab
4. Close settings
5. Open settings using the 'profile' option in the user menu

#### What did you expect?

Settings opens and takes me directly to the profile tab

#### What happened instead?

Settings opens and leaves me where I previously was: on the video tab

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

robintown:
Depends on https://github.com/vector-im/compound-web/pull/110

Closes https://github.com/element-hq/element-call/issues/1764
Fixes one half of https://github.com/element-hq/element-call/issues/1839 (local video is now mirrored correctly in the call view, but not in the lobby)


src/room/InCallView.module.css
```diff
@@ -28,9 +28,7 @@ limitations under the License.
   flex-direction: column;
   overflow: auto;
   overflow-inline: hidden;
-  /* There used to be a contain: strict here, but due to some bugs in Firefox,
-  this was causing the Z-ordering of modals to glitch out. It can be added back
-  if those issues appear to be resolved. */
+  contain: strict;
```
toger5, line 31: This is great news!

src/video-grid/VideoTile.tsx
```diff
@@ -242,44 +167,315 @@ export const VideoTile = forwardRef<HTMLDivElement, Props>(
...
+        const updateName = (): void => {
+          setDisplayName(vm.member!.rawDisplayName);
+        };
+
+        vm.member!.on(RoomMemberEvent.Name, updateName);
+        return (): void => {
+          vm.member!.removeListener(RoomMemberEvent.Name, updateName);
+        };
+      }
+    }, [vm.member, setDisplayName]);
```
toger5, line 440: Why is this not part of the view model?
(Just saw its mentioned in the TODO. What is holding back moving it?)

toger5 (approved):
Thanks. 
This is super satisfying to see how this also makes it a lot more readable (+ better to test)
Most of the code also was very striaght forward to review.

Playing around with it I found the following, is this intentional?
 - The user tile ("you") cannot be moved into the first position grid mode.
 
I will give this a approval since I am on holiday but before merging I think the "cannot be moved into the first position" should addressed (exept its by design. If so I would be curious why) and the TODO (move to vm in case that is easy enough for this PR)

robintown:
> The user tile ("you") cannot be moved into the first position grid mode.

Ah right, I've seen this happening in the application for some months, so I don't believe it's a regression introduced here. The plan for fixing this is that, the new grid layouts won't allow for dragging and dropping anyways, so I'd like to move ahead with implementing them.

src/video-grid/VideoTile.tsx
robintown: Hmm, I can't remember actually what the difficulty was! But regardless, I think it doesn't hurt to leave little things like this to be refactored later, in the spirit of working iteratively

PR #2014, opened 2023-12-19
Ensure that all our components have display names

robintown:
This turns on a lint rule to require display names for all of our components, which makes it a lot easier to find your way around the component tree in React's dev tools.


toger5 (approved):
Nice! This is a good idea.

PR #2083, opened 2024-01-20
Remove the gappy sync hack

robintown:
This hack was added in the early days of Element Call, back when we were doing call signaling using non-state room events, and missing part of a room's history could cause calls to fall apart. Nowadays we use state events for signaling, and all this hack is doing is making sync times unnecessarily long, so we can remove it.

### DIFF
--- a/src/matrix-utils.ts
+++ b/src/matrix-utils.ts
@@ -188,16 +188,8 @@ export async function initClient(
     await client.store.startup();
   }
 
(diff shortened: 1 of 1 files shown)


toger5 (approved):
Very nice you remember about this.
Indeed not required anymore. (even the encryption keys are resent every time so looking back in history is never required)

PR #2084, opened 2024-01-20
Fix builds on PRs from forks

robintown:
I've confirmed here in my fork that this now provides the correct URL: https://github.com/robintown/element-call/commit/24ec038297ea8070d38240c241c36ce75c1b6f00/checks#step:6:1


fkwp (approved):
nice

PR #2086, opened 2024-01-21
Add data models for the new call layouts

robintown:
This is a start at implementing the call layouts from the new designs. I've added data types to model the contents of each possible layout, and begun implementing the business logic to produce these layouts in the call view model.


src/state/CallViewModel.ts
```diff
@@ -65,9 +88,128 @@ export interface TileDescriptor<T> {
...
+export interface FullScreenLayout {
+  type: "full screen";
+  spotlight: Tile<MediaViewModel[]>;
+  pip?: Tile<UserMediaViewModel>;
+}
+
+export interface PipLayout {
+  type: "pip";
+  spotlight: Tile<MediaViewModel>;
+}
```
toger5: There seems to be enough overlap so that a Layout interface could be used and the one listed here could extend it.

src/state/CallViewModel.ts
```diff
@@ -65,9 +88,128 @@ export interface TileDescriptor<T> {
...
+enum GridBin {
+  SelfStart,
+  Presenters,
+  Speakers,
+  VideoAndAudio,
+  Video,
+  Audio,
+  NoMedia,
+  SelfEnd,
+}
```
toger5: why are they called **Grid**Bin. are they specific to the GridLayout? My first intuition was they are the bin categories for the Tiles independent of the layout.

src/state/CallViewModel.ts
```diff
@@ -165,9 +307,193 @@ export class CallViewModel extends ViewModel {
     },
   );
 
+  private readonly mediaItems = state(
```
toger5: I think the variables declared from here on: `mediaItems`, `userMedia`, `screenShares`, `spotlightSpeaker`, `grid`, `spotlight` are quicker to understand with explicit type annotations (if our linter settings allow this). So one can tell on first glance if we are already on the tile abstraction or are still dealing with media.

src/state/CallViewModel.ts
```diff
@@ -165,9 +307,193 @@ export class CallViewModel extends ViewModel {
...
+
+  private readonly gridMode = new BehaviorSubject<GridMode>("grid");
+
+  public setGridMode(value: GridMode): void {
+    this.gridMode.next(value);
+  }
+
+  // TODO: Make this react to changes in window dimensions and screen
+  // orientation
+  private readonly windowMode = of<WindowMode>("normal");
```
toger5, line 443: Any particular reason this is between two public class members?

src/state/CallViewModel.ts
```diff
@@ -165,9 +307,193 @@ export class CallViewModel extends ViewModel {
...
+        }
+      }
+    }).pipe(switchAll()),
+  );
+
   /**
    * The media tiles to be displayed in the call view.
    */
+  // TODO: Get rid of this field, replacing it with the 'layout' field above
+  // which keeps more details of the layout order internal to the view model
```
toger5, line 492: We still have it in here now because the layout is not yet used in the view itself right?

src/state/CallViewModel.ts
toger5: My initial idea of a type called bin would be a bin iteself. What do you think about calling this enum type `GridBinType` (Or do you argue the `Type` included by the fact its an enum?)

src/state/CallViewModel.ts
```diff
@@ -165,9 +307,193 @@ export class CallViewModel extends ViewModel {
...
+                  : speaker
+                    ? GridBin.Speakers
+                    : video
+                      ? audio
+                        ? GridBin.VideoAndAudio
+                        : GridBin.Video
+                      : audio
+                        ? GridBin.Audio
+                        : GridBin.NoMedia,
+            ] as const,
```
toger5: Just a proposal/my opinion.
Sadly this is a lot longer but I find this easier to read:
```suggestion
    (speaker, presenter, audio, video) => {
        let binType = GridBin.SelfStart;
        if (m.vm.local) {
            binType = GridBin.SelfStart
        }
        else if (presenter) {
            binType = GridBin.Presenters
        }
        else if (speaker) {
            binType = GridBin.Speakers
        }
        else if (video && audio) {
            binType = GridBin.VideoAndAudio
        }
        else if (video && !audio) {
            binType = GridBin.Video
        }
        else if (!video && audio) {
            binType = GridBin.Audio
        }
        else if (!video && !audio) {
            binType = GridBin.NoMedia
        }
        return [m, binType] as const,
    }
```
(The ternaries need the additional processing step that usually the value after the ? is returned except in the end.)

toger5 (approved):
This looks really good! Thx.
I think the overall structure can be made a bit clearer when using more detailed names.
It seems we have three levels:
 - The media item array
 - Filtered media item arrays
 - Tile arrays (Which can contain multiple MediaItems)
 - Layouts that contain bins of tiles + additional information.
 
 Calling the variables like that (e.g. `spotlight`->`spotlightTile`, `screenShares`->`screenSharesMedia`...) could help make these levels more obvious because there would be a set of variables called `somethingMedia` and `somethingTile/s`.

src/state/CallViewModel.ts
robintown: That's correct, it's because we're still using the legacy VideoGrid system for now

src/state/CallViewModel.ts
robintown: Renamed to SortingBin. I do kind of prefer keeping "bin" as the name as long as we don't have a need for a full-blown "bin" class which can be instantiated - it seems simplest to say that `SortingBin` is in fact how you reference a bin.

src/state/CallViewModel.ts
robintown: What benefit do you see in using an interface? My mental model of the new layouts is that they need to each use a different rendering path in InCallView, so I went for this way of doing a structured enum which allows for more type-safe destructuring than I think an interface would provide.

robintown: Merging this for now, but we can iterate on this

PR #2099, opened 2024-01-24
Widget return to lobby

toger5:
This is required for: https://github.com/matrix-org/matrix-react-sdk/pull/12057


src/room/GroupCallView.tsx
```diff
@@ -246,6 +244,9 @@ export const GroupCallView: FC<Props> = ({
 
   useEffect(() => {
     if (widget && isJoined) {
+      // set widget to sticky once joined.
+      widget!.api.setAlwaysOnScreen(true);
```
robintown, line 245: Can the other invocation of `setAlwaysOnScreen` in this file be removed now that this is here?

robintown (approved):
One non-blocking comment

PR #2100, opened 2024-01-24
Dont send register request in widget mode

### LINKED ISSUE #1897: EC still makes calls to the register endpoint in widget mode
### Steps to reproduce

Load EC in widget mode and see that it still makes a POST request to the register endpoint (which gets a 401). It does this because the `useInteractiveRegistration` hook does a register request on mount to fetch the recaptcha parameters and privacy policy URL, even though of course we don't go on to use this hook since we don't need to register a user.

It probably shouldn't do, but would need a little refactoring because hooks can't be called conditionally.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

toger5:
Closes https://github.com/element-hq/element-call/issues/1897

### DIFF
--- a/src/auth/useRegisterPasswordlessUser.ts
+++ b/src/auth/useRegisterPasswordlessUser.ts
@@ -21,6 +21,7 @@ import { useClient } from "../ClientContext";
 import { useInteractiveRegistration } from "../auth/useInteractiveRegistration";
 import { generateRandomName } from "../auth/generateRandomName";
 import { useRecaptcha } from "../auth/useRecaptcha";
+import { widget } from "../widget";
 
 interface UseRegisterPasswordlessUserType {
   privacyPolicyUrl?: string;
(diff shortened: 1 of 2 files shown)


robintown:
Closes https://github.com/element-hq/element-call/issues/1897 ?

src/auth/useRegisterPasswordlessUser.ts
```diff
@@ -39,6 +40,9 @@ export function useRegisterPasswordlessUser(): UseRegisterPasswordlessUserType {
       if (!setClient) {
         throw new Error("No client context");
       }
+      if (widget) {
+        throw new Error("We never register passwordless user in widget");
```
robintown: Instead of throwing an error, can we avoid calling this function in widget mode? I think I would find that cleaner.

toger5: Having unecassary errors in the consol is indeed not the cleanest.
I kept the error however but additionally added a condition where the method is called. This gives us insight in when we accidently call it in a widget at some point.

robintown (approved):

PR #2107, opened 2024-01-26
Add crop to fit toggle to context menu

### LINKED ISSUE #664: Display non 16:9 video tiles properly (currently crops them to 16:9)
### Steps to reproduce

2. What do you click?
3. More steps…

when you join the element-call call with a 4:3 camera, it crops it to 16:9 for the default "full view" of a participant. i think it would be better to make 4:3 cameras show the whole thing including the bottom. or if you like your camera getting chopped off the top and bottom maybe add a toggle for it, but i dont think this is optimal default behaviour imo. 

#### What did you expect?
i expected the camera to show itself in its entirety 4:3

#### What happened instead?
the camera was cropped to 16:9, cutting off a significant amount of viewage

### Operating system

affects all of them currently

### Browser information

affects all of them currently

### URL for webapp

affects all of them currently, usually use netlify and/or call.element.io

### Will you send logs?

Yes

toger5:
This adds a toggle to the context menu so we can switch between cropped (`object-fit: cover`) and the uncropped full video (`object-fit: contain`).

It defaults to `Crop to fit: true` (as it did before) Without using the context menu it should behave the same as before.
Fixes: https://github.com/element-hq/element-call/issues/664

Screenshot 2024-01-26 at 04 27 08
Screenshot 2024-01-26 at 04 27 22
Screenshot 2024-01-26 at 04 27 34

### DIFF
--- a/public/locales/en-GB/app.json
+++ b/public/locales/en-GB/app.json
@@ -143,6 +143,7 @@
   "unmute_microphone_button_label": "Unmute microphone",
   "version": "Version: {{version}}",
   "video_tile": {
+    "change_fit_contain": "Crop to fit",
     "exit_full_screen": "Exit full screen",
     "full_screen": "Full screen",
     "mute_for_me": "Mute for me",
--- a/src/state/TileViewModel.ts
+++ b/src/state/TileViewModel.ts
@@ -167,6 +167,12 @@ export class UserMediaTileViewModel extends BaseTileViewModel {
    */
   public readonly videoEnabled: StateObservable<boolean>;
 
+  private readonly _cropVideo = new BehaviorSubject(true);
+  /**
+   * Whether the tile video should be contained inside the tile or be cropped to fit.
+   */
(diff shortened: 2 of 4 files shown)


toger5:
How should we proceed with this?
If there are no strong concerns I would like to merge this. (Its quiet hidden and the default is still as it was before - its a minor/invisble change for the occasional user)

fkwp:
fine for me.

@americanrefugee WDYT?

americanrefugee:
I'm a bit confused as to what this feature is...

In the original concept, you can choose to always see you're own tile – whether in the grid, the list of participants, or full-screen mode. Is this what you mean by "Crop to fit" that, or something else?

americanrefugee:
Thanks for the explanation. Approved!

robintown (approved):
Code looks as I expected

notramo:
@toger5 thanks for implementing this, I needed this feature so much.

PR #2119, opened 2024-01-29
Lock livekit-client version to 1.15.2

toger5:


fkwp (approved):
lets test our on-scale regression :)

PR #2144, opened 2024-02-03
refactor: replace jest with vitest

### LINKED ISSUE #2019: Consider replacing Jest with Vitest
If we're to be serious about thoroughly testing new code, it might be worthwhile to look into improving the developer experience around testing.

[Vitest](https://vitest.dev/), which we're using successfully in Compound Web, and I also quite enjoy having used in personal projects, is a drop-in replacement for Jest that integrates nicely with your Vite config, so that you don't have to spend as much time in the long run getting your testing environment to behave like your runtime environment. It also runs in "watch mode" by default, where it listens for file changes and immediately re-runs the exact tests that could be affected by those changes, which makes the feedback loop between coding and testing that much shorter.

If this sounds like a good idea, we should switch to it sooner rather than later, while we have not that many tests to convert.

hereje:
# **Replace jest with vitest**

## Description
- Replace jest with vitest
- update test files
- remove jest references

### Fixes: #2019


yarn.lock
toger5, line 1: There seems to be something of with the formatting of the lock file.
I am not sure if just the ordering is off or something bigger happened to the yarn lock?

Maybe replacing it with the one on the livekit branch and running yarn again could be enough.

hereje: Done!

toger5: It seems like the issue reappeared. The changes to the yarn lock look like they are too significant.

.github/workflows/test.yaml
```diff
@@ -1,11 +1,11 @@
-name: Run jest tests
+name: Run vitest tests
```
robintown:
```suggestion
name: Run unit tests
```

Personally I think it's useful to specify what kind of tests these are, rather than the specific tool we use to run them

package.json
```diff
@@ -13,7 +13,9 @@
     "lint:types": "tsc",
     "i18n": "node_modules/i18next-parser/bin/cli.js",
     "i18n:check": "node_modules/i18next-parser/bin/cli.js --fail-on-warnings --fail-on-update",
-    "test": "jest",
+    "test": "vitest run",
+    "test:watch": "vitest watch",
```
robintown:
```suggestion
    "test": "vitest",
```

I think the watch mode is a useful default for developers, and vitest automatically switches to "run" mode when it's in CI, so can we replace these with just the one command?

vitest.config.js
```diff
@@ -0,0 +1,31 @@
...
+  defineConfig({
+  test: {
+    globals: true,
+    environment: 'jsdom',
+    alias: {
+      "\\.css$": "identity-obj-proxy",
+      "\\.svg\\?react$": "<rootDir>/test/mocks/svgr.ts",
+      "^\\./IndexedDBWorker\\?worker$": "<rootDir>/test/mocks/workerMock.ts",
+      "^\\./olm$": "<rootDir>/test/mocks/olmMock.ts"
+    },
```
robintown: Part of the advantage of using vitest is that because it shares configuration with Vite, you don't have to mock out quite as many things by hand. I believe we can safely remove this entire `alias` block (and those mock files as well).

robintown (commented):
Hey, this is really helpful, thanks for contributing this! I have just a couple of comments before we merge this:

.github/workflows/test.yaml
hereje: Totally!

package.json
hereje: Ok!

vitest.config.js
hereje: Done!

robintown:
For the record I regenerated the yarn.lock file because it looks like you had unintentionally upgraded a number of unrelated dependencies - this can happen if you're deleting yarn.lock and regenerating it from scratch

robintown (approved):

PR #2153, opened 2024-02-06
Improve focus switch logging

toger5:

### DIFF
--- a/src/room/useActiveFocus.ts
+++ b/src/room/useActiveFocus.ts
@@ -56,6 +56,9 @@ export function useActiveFocus(
     const newActiveFocus = getActiveFocus(rtcSession);
 
     if (!deepCompare(activeFocus, newActiveFocus)) {
+      logger.warn(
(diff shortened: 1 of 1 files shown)


src/room/useActiveFocus.ts
```diff
@@ -56,6 +56,9 @@ export function useActiveFocus(
     const newActiveFocus = getActiveFocus(rtcSession);
 
     if (!deepCompare(activeFocus, newActiveFocus)) {
+      logger.warn(
+        `got new active focus (focus switch) form: ${activeFocus} to ${newActiveFocus}`,
```
robintown:
```suggestion
        `got new active focus (focus switch) from: ${activeFocus} to ${newActiveFocus}`,
```

robintown (commented):
While we're at it, could we also reduce the spamminess of the log line in `getActiveFocus` somehow?

robintown (approved):
I still have the comment that I would like to reduce the log spam of `getActiveFocus`, though strictly speaking it's not a blocker

PR #2190, opened 2024-02-19
Add customHomeserver url Param

toger5:
This is an additional url param that allows to overwrite the homeserberUrl configured in EC.
This is usefult if a clinet that wants to create a call link wants users to end up on a specific homeserver.

### DIFF
--- a/src/UrlParams.ts
+++ b/src/UrlParams.ts
@@ -143,7 +143,7 @@ export interface UrlParams {
    * This can be user to configure a non default guest user server when
    * creating a spa link.
    */
-  customHomeserver: string | null;
+  homeserver: string | null;
 }
 
 // This is here as a stopgap, but what would be far nicer is a function that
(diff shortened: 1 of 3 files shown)


src/UrlParams.ts
```diff
@@ -229,6 +243,8 @@ export const getUrlParams = (
     perParticipantE2EE: parser.getFlagParam("perParticipantE2EE"),
     skipLobby: parser.getFlagParam("skipLobby"),
     returnToLobby: parser.getFlagParam("returnToLobby"),
+    viaServers: parser.getParam("viaServers"),
+    customHomeserver: parser.getParam("customHomeserver"),
```
robintown:
```suggestion
    homeserver: parser.getParam("homeserver"),
```

What do you think about this naming?

robintown (approved):
One minor suggestion

src/UrlParams.ts
toger5: I was also considering this.
I wanted to make sure its clear that this is not required. But I guess since its called `defaultHomeserver` in the rest of the code `homeserver` should be sufficient.

PR #2203, opened 2024-02-21
Allow Element Call theme configuration via url.

toger5:
This allows us to configure the EC theme based on a url parameter. `theme=light`...
It then sets the cpd theme.


src/App.tsx
```diff
@@ -58,6 +60,31 @@ const BackgroundProvider: FC<BackgroundProviderProps> = ({ children }) => {
   return <>{children}</>;
 };
 
+interface ThemeProviderProps {
+  children: JSX.Element;
+}
+
+const ThemeProvider: FC<ThemeProviderProps> = ({ children }) => {
```
robintown: Given that this component just outputs the children without adding anything, I think it would make more sense as a custom hook (`useTheme` or something)

src/App.tsx
```diff
@@ -58,6 +60,31 @@ const BackgroundProvider: FC<BackgroundProviderProps> = ({ children }) => {
...
+  const { theme } = useUrlParams();
+  const [previousTheme, setCurrentTheme] = useState<string | null>(
+    document.body.classList.item(0),
+  );
+  useEffect(() => {
```
robintown:
```suggestion
  useLayoutEffect(() => {
```

So that the UI doesn't flash from dark to light mode when first mounted

src/App.tsx
toger5: We need `useUrlParams` (which calls `useContext`). which errors if its run outside the react dom router.
Where would this effect then best be positined?

toger5: I changed it to keep the lines in App.tsx minimal with a external hook. I currently see the option of either squeesing it into antoher component (like the `BackgroundProvider`) or using the ThemeProvider solution.

src/useTheme.ts
```diff
@@ -0,0 +1,43 @@
...
+limitations under the License.
+*/
+
+import { useLayoutEffect, useState } from "react";
+
+import { useUrlParams } from "./UrlParams";
+
+export const useTheme = (): void => {
+  const { theme: themeName } = useUrlParams();
+  const [previousTheme, setPreviousTheme] = useState<string | null>(
```
robintown: This could be a ref rather than a state variable, which would help us avoid the extra render whenever `previousTheme` is changed

src/useTheme.ts
```diff
@@ -0,0 +1,43 @@
...
+    document.body.classList.item(0),
+  );
+  useLayoutEffect(() => {
+    // Don't update the current theme if the url does not contain a theme prop.
+    if (!themeName) return;
+    const theme = themeName.includes("light") ? "light" : "dark";
+    const themeHighContrast = themeName.includes("high-contrast") ? "-hc" : "";
```
robintown, line 28: This now doesn't match with the documentation in `UrlParams.ts`, which says it expects `-hc` rather than `high-contrast`

src/UrlParams.ts
```diff
@@ -131,7 +131,11 @@ export interface UrlParams {
    */
   returnToLobby: boolean;
   /**
-   * This defines the homeserver that is going to be used when joining a room.
+   * The theme to use for element call.
+   * can be "light", "dark", "light-hc" or "dark-hc".
+   */
+  theme: string | null;
```
robintown, line 137: This should be documented in the markdown file as well

src/App.tsx
```diff
@@ -57,14 +59,17 @@ const BackgroundProvider: FC<BackgroundProviderProps> = ({ children }) => {
 
   return <>{children}</>;
 };
+const ThemeProvider: FC<SimpleProviderProps> = ({ children }) => {
+  useTheme();
+  return <>{children}</>;
```
robintown:
```suggestion
  return children;
```

src/App.tsx
robintown: I didn't consider that this requires the context, that makes sense now

docs/url-params.md
```diff
@@ -1,57 +1,76 @@
...
-dedicated webview we send the postMessages same webview the widget lives in. Filtering is
-done in the widget so it ignores the messages it receives from itself)
+The url used to send widget action postMessages. This should be the domain of
+the client or the webview the widget is hosted in. (in case the widget is not
+in an Iframe but in a dedicated webview we send the postMessages same webview
+the widget lives in. Filtering is done in the widget so it ignores the messages
+it receives from itself)
 
-```
+```js
```
robintown: These are all TypeScript code snippets, right?

robintown (approved):

PR #2233, opened 2024-02-28
Theme loading without flickering

toger5:
The theme did flicker because there is a default theme shown before any JS is loaded.
When the layoutEffect from the useTheme hook is reached the theme gets updated -> If its a different theme ("light", "dark") one can see a very short flicker.

This is particularly annoying in widget mode.

To work around this we only send the contentLoaded action once the theme is set.
This relies on the js sdk not sending this action automatically AND the react sdk needs to be configured with waitForIframe=false, so that it does not consider the widget to be ready once the IFrame is loaded. 
Until the content_loadedaction is sent (which now is equivalent to: until EC is loaded), we will not show the widget. Instead the EW loading screen is shown.
This also saves us from the flickering: EW widget loading screen -> EC normal screen with header -> EC loading screen -> EC lobby.

This relies on:
Js-sdk PR that allows to disable sending the content loaded action right after the client is initialized but before the first react update (default behaviour of the js sdk):
https://github.com/matrix-org/matrix-js-sdk/pull/4086

EW PR that sets waitForIframe=false and shows the widget loading screen until the receiving the content_loaded action:
https://github.com/matrix-org/matrix-react-sdk/pull/12292

The type failiours will be fixed once https://github.com/matrix-org/matrix-js-sdk/pull/4086 is merged.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -13,7 +13,8 @@
     </script>
   </head>
 
-  <body class="nodisplay">
+  <!-- The default class is: .no-theme {display: none}. It will be overwritten once the app is loaded. -->
+  <body class="no-theme">
     <div id="root"></div>
   </body>
 </html>
--- a/public/index.html
+++ b/public/index.html
@@ -13,7 +13,7 @@
     </script>
   </head>
 
-  <body class="cpd-theme-dark">
+  <body class="nodisplay">
     <div id="root"></div>
   </body>
 </html>
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "i18next-http-backend": "^2.0.0",
     "livekit-client": "^2.0.2",
     "lodash": "^4.17.21",
-    "matrix-js-sdk": "github:matrix-org/matrix-js-sdk#8123e9a3f1142a7619758c0a238172b007e3a06a",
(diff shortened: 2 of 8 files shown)


public/index.html
```diff
@@ -13,7 +13,7 @@
     </script>
   </head>
 
-  <body class="cpd-theme-dark">
+  <body class="nodisplay">
```
robintown: Why do we need to hide the UI from within Element Call? Doesn't Element Web now hide the widget until the content is loaded?

toger5: Yes this is somewhat redundent. It is only used for the SPA where we dont have a good way to change the theme.
But it might come in handy once we change it to use the system theme for instance. Then we can have is as nodisplay (browser doing some default background. Which should also conform to system theme) and once js has started evaluating what theme is actually used we use the correct color from the start.

robintown: Could we rename this class to something like `no-theme` so that it's more apparent that it's involved in the theme code?

robintown (approved):

public/index.html
toger5: what about `invisible-loading-theme` (so its also clear that this class could be the reason why its not rendering anything.)

robintown: That's also fine by me

PR #2281, opened 2024-04-02
Knocking support

toger5:
This PR does three things seperated in three commits (review can be done per commit)
 - Add joining with knocking.
 - Refactor Encryption information using one hook.
 - Change recent list to matrixRTC (it was dependant on the `m.call` event before)

Overall this makes EC fully compatible with join links created with EW in rooms that use the join rule `Knock` (not public)
It also enables the EW user to kick or ban users from the call as desired.

To test this EW needs to have:
 - The `guest_spa_url` set. (`element_call:guest_spa_url`)
 - The following feature flags activated
     - "feature_element_call_video_rooms": true,
     - "feature_video_rooms": true,
     - "feature_new_room_decoration_ui": true,
     - "feature_group_calls": true,
     - "feature_ask_to_join": true
     
 (full config
 ```
     "features": {
        "feature_element_call_video_rooms": true,
        "feature_video_rooms": true,
        "feature_new_room_decoration_ui": true,
        "feature_group_calls": true,
        "feature_ask_to_join": true
    },

    "element_call": {
        "url": "https://localhost:3000",
        "guest_spa_url": "https://localhost:3000"
    }
```
)

### CODE REVIEW

File: src/settings/submit-rageshake.ts [hunk 1]
```diff
@@ -298,13 +298,13 @@ export function useRageshakeRequest(): (
 
   const sendRageshakeRequest = useCallback(
     (roomId: string, rageshakeRequestId: string) => {
+      // @ts-expect-error - org.matrix.rageshake_request is not part of `keyof TimelineEvents` but it is okay to sent a custom event.
```
robintown, line 301: The js-sdk's types disallow sending custom event types now by default? That's pretty surprising :confused:

src/e2ee/sharedKeyManagement.ts
```diff
@@ -68,30 +67,37 @@ const useKeyFromUrl = (): [string, string] | [undefined, undefined] => {
...
+export type EncryptionSystem = Unencrypted | SharedSecret | PerParticipantE2EE;
+
+export function useRoomEncryptionSystem(roomId: string): EncryptionSystem {
+  const { client } = useClient();
+
   // make sure we've extracted the key from the URL first
   // (and we still need to take the value it returns because
   // the effect won't run in time for it to save to localstorage in
   // time for us to read it out again).
   const [urlRoomId, passwordFormUrl] = useKeyFromUrl();
```
robintown:
```suggestion
  const [urlRoomId, passwordFromUrl] = useKeyFromUrl();
```

src/e2ee/sharedKeyManagement.ts
```diff
@@ -68,30 +67,37 @@ const useKeyFromUrl = (): [string, string] | [undefined, undefined] => {
...
-  return useMemo(() => !room || isRoomE2EE(room), [room]);
-};
-
-export function isRoomE2EE(room: Room): boolean {
-  // For now, rooms in widget mode are never considered encrypted.
-  // In the future, when widget mode gains encryption support, then perhaps we
-  // should inspect the e2eEnabled URL parameter here?
-  return widget === null && !room.getCanonicalAlias();
+  const room = client?.getRoom(roomId);
+  const e2eeSystem = useMemo(() => {
```
robintown:
```suggestion
  const e2eeSystem = useMemo<EncryptionSystem>(() => {
```

I think if you did this, you wouldn't have to cast all of the return values

AndrewFerr:
If a guest link is generated for a given `public`/`knock` room, and then the room's join rule is set to `private`, visiting the guest link still has ECall show the "Ask to join call" button.  Clicking it shows an error of `MatrixError: [403] You don't have permission to knock`.

It would be nice if visiting the guest link of a room that has since been changed to `private` access could have Element Call say that the room is now private, instead of suggesting to send a knock that will fail.

src/home/useGroupCallRooms.ts
```diff
@@ -80,38 +82,51 @@ function sortRooms(client: MatrixClient, rooms: Room[]): Room[] {
...
+  return joinRule === JoinRule.Knock || joinRule === JoinRule.Public;
+};
+
+const roomIsJoinedWithCall = (room: Room): boolean => {
+  const roomStateEvents = room
+    .getLiveTimeline()
+    .getState(EventTimeline.FORWARDS)?.events;
+  return (
+    room.getMyMembership() === KnownMembership.Join &&
+    !!roomStateEvents?.get(EventType.GroupCallMemberPrefix)
```
robintown, line 102: It took me a moment to understand that this condition represents whether a call has ever been placed in the room, rather than whether a call is currently ongoing, so a comment could help here

src/home/useGroupCallRooms.ts
```diff
@@ -80,38 +82,51 @@ function sortRooms(client: MatrixClient, rooms: Room[]): Room[] {
...
-        .map((groupCall) => groupCall.room)
+      const rooms = client
+        .getRooms()
+        .filter(roomIsJoinedWithCall)
         .filter(roomIsJoinable);
       const sortedRooms = sortRooms(client, rooms);
       const items = sortedRooms.map((room) => {
-        const groupCall = client.getGroupCallForRoom(room.roomId)!;
-
+        // const groupCall = client.getGroupCallForRoom(room.roomId)!;
```
robintown: Any reason to not just delete this?

src/home/useGroupCallRooms.ts
```diff
@@ -80,38 +82,51 @@ function sortRooms(client: MatrixClient, rooms: Room[]): Room[] {
...
+        const session = client.matrixRTC.getRoomSession(room);
+        session.memberships;
```
robintown, line 119: What's this doing here?

src/matrix-utils.ts
```diff
@@ -365,39 +358,58 @@ export async function createRoom(
...
+  })();
 
+  const roomIdPart = `roomId=${roomId}`;
+  const viaServersPart = viaServers
+    ? viaServers.map((s) => "&" + VIA_SERVERS_STRING + s).join("")
+    : "";
   return `/room/#${
     roomName ? "/" + roomAliasLocalpartFromRoomName(roomName) : ""
-  }?roomId=${roomId}${password ? "&" + PASSWORD_STRING + encodedPassword : ""}`;
+  }?${roomIdPart}${encryptionPart}${viaServersPart}`;
```
robintown: I don't like that we're expanding the PASSWORD_STRING hack to now also include these PER_PARTICIPANT_STRING and VIA_SERVERS_STRING constants. Maybe this is a sign that we should finally implement a proper function `UrlParams => string` or `UrlParams => URLSearchParams` or something similar, that would allow us to consolidate all information about the URL params into UrlParams.ts?

src/matrix-utils.ts
```diff
@@ -340,14 +341,6 @@ export async function createRoom(
 
   logger.log(`Creating group call in ${result.room_id}`);
```
robintown: This log line doesn't seem relevant anymore, though it's still useful to log the created room ID I guess

src/room/GroupCallLoader.tsx
```diff
@@ -51,27 +54,45 @@ export function GroupCallLoader({
   );
 
   switch (groupCallState.kind) {
+    case "loaded":
+    case "waitForInvite":
+    case "canKnock":
+      return <>{children(groupCallState)}</>;
```
robintown:
```suggestion
      return children(groupCallState);
```

src/room/useLoadGroupCall.ts
```diff
@@ -38,51 +45,196 @@ export type GroupCallLoading = {
...
   | GroupCallLoadFailed
-  | GroupCallLoading;
+  | GroupCallLoading
+  | GroupCallWaitForInvite
+  | GroupCallCanKnock;
 
-export interface GroupCallLoadState {
-  error?: Error;
-  groupCall?: GroupCall;
+export class CustomMessage extends Error {
```
robintown: Is there a name that would be less generic and make it clear that this class is for errors that arise when loading a group call?

src/room/useLoadGroupCall.ts
```diff
@@ -38,51 +45,196 @@ export type GroupCallLoading = {
...
+  public messageBody: string;
+  public reason?: string;
+  public constructor(
+    messageTitle: string,
+    messageBody: string,
```
robintown, line 83: I had to look further down into the code to tell whether these fields store translated strings, so that would be good to comment about

src/room/LobbyView.tsx
```diff
@@ -116,12 +121,18 @@ export const LobbyView: FC<Props> = ({
         <div className={styles.content}>
           <VideoPreview matrixInfo={matrixInfo} muteStates={muteStates}>
             <Button
-              className={styles.join}
-              size="lg"
-              onClick={onEnter}
+              className={classNames(
+                "cpd-theme-dark",
```
robintown: In the designs, this button doesn't appear to have dark mode forced on it

src/room/LobbyView.tsx
```diff
@@ -116,12 +121,18 @@ export const LobbyView: FC<Props> = ({
...
+                styles.join,
+                waitingForInvite ? styles.wait : "",
```
robintown:
```suggestion
                { [styles.wait]: waitingForInvite },
```

src/room/RoomPage.tsx
```diff
@@ -77,19 +80,85 @@ export const RoomPage: FC = () => {
...
+              muteStates={muteStates}
+            />
+          );
+        case "waitForInvite":
+        case "canKnock": {
+          const knock =
+            groupCallState.kind === "canKnock" ? groupCallState.knock : null;
+          const label =
+            groupCallState.kind === "canKnock" ? (
+              <>{t("lobby.ask_to_join")}</>
```
robintown:
```suggestion
              t("lobby.ask_to_join")
```

src/room/RoomPage.tsx
```diff
@@ -77,19 +80,85 @@ export const RoomPage: FC = () => {
...
+                    ? E2eeType.PER_PARTICIPANT
+                    : E2eeType.NONE,
+                },
+              }}
+              onEnter={(): void => knock?.()}
+              enterLabel={label}
+              waitingForInvite={groupCallState.kind === "waitForInvite"}
+              confineToRoom={confineToRoom}
+              hideHeader={hideHeader}
+              participantCount={0}
```
robintown: Maybe we should pass a `null` here rather than a number that might not be accurate, so it doesn't cause confusion in the future

src/room/useLoadGroupCall.ts
```diff
@@ -38,51 +45,196 @@ export type GroupCallLoading = {
...
+      return room;
+    };
+
+    const getRoomByKnocking = async (
+      roomId: string,
+      viaServers: string[],
+      onKnockSent: () => void,
+    ): Promise<Room> => {
+      let joinedRoom: Room | null = null;
+      await client.knockRoom(roomId);
```
robintown: We should provide the via servers here too, I believe

src/room/MuteStates.ts
```diff
@@ -20,10 +20,11 @@ import { MediaDevice, useMediaDevices } from "../livekit/MediaDevicesContext";
...
 
 /**
- * If there already is this many participants in the call, we automatically mute
- * the user
+ * If there already are this many participants in the call, we automatically mute
+ * the user.
  */
-const MUTE_PARTICIPANT_COUNT = 8;
+// TODO set back to 8. for demo purposes we use 1 here.
+export const MUTE_PARTICIPANT_COUNT = 8;
```
robintown: Comment seems inaccurate

src/room/LobbyView.module.css
```diff
@@ -24,6 +24,10 @@ limitations under the License.
   overflow: hidden;
   height: 100%;
 }
+.wait {
+  color: black !important;
+  background-color: rgba(255, 255, 255, 0.67) !important;
+}
```
robintown: I'm suspicious of overriding our component colors like this, especially when the colors aren't theme-reactive. Could we use a standard button state that has already been designed, like a disabled state?

src/room/GroupCallLoader.tsx
```diff
@@ -51,27 +54,45 @@ export function GroupCallLoader({
...
           </FullScreenView>
         );
+      } else if (groupCallState.error instanceof CustomMessage) {
+        return (
+          <FullScreenView>
+            <Heading>{groupCallState.error.message}</Heading>
+            <Text>{groupCallState.error.messageBody}</Text>
+            {groupCallState.error.reason && (
+              <>
+                Reason:
```
robintown: This isn't translated

src/room/GroupCallView.tsx
```diff
@@ -72,10 +75,23 @@ export const GroupCallView: FC<Props> = ({
...
   rtcSession,
+  muteStates,
 }) => {
   const memberships = useMatrixRTCSessionMemberships(rtcSession);
   const isJoined = useMatrixRTCSessionJoinState(rtcSession);
 
+  // The mute state reactively gets updated once the participant count reaches the threshold.
+  // The user then still is able to unmute again.
+  // The more common case is that the user is muted from the start (participant count is already over the threshold).
+  const [autoMuteHappened, setAutoMuteHappened] = useState(false);
```
robintown: This could be a ref rather than state, since it isn't used to render anything

src/room/GroupCallView.tsx
```diff
@@ -288,17 +291,23 @@ export const GroupCallView: FC<Props> = ({
 
   const { t } = useTranslation();
 
-  if (roomEncrypted && !perParticipantE2EE && !e2eeSharedKey) {
+  if (e2eeSystem.kind === E2eeType.NONE && !widget) {
+    // the url wants encryption, but we don't have a encryption system. (e.g. when joining a call without password)
```
robintown: How do we know this isn't supposed to be an unencrypted call? Does this change break unencrypted calls?

src/room/GroupCallView.tsx
```diff
@@ -288,17 +291,23 @@ export const GroupCallView: FC<Props> = ({
...
+    // TODO: we need to figure out what we do with encryption goning forward.
+    // In the SPA a room is unencrypted but uses a shared secret for the call (since we distribute via url that works)
+    // In embedded mode unencrypted rooms would make most sense to also call unencrypted.
+    // A call where one person joins with the spa and another person with the widget would be a problem.
     return (
       <ErrorView
         error={
           new Error(
-            "No E2EE key provided: please make sure the URL you're using to join this call has been retrieved using the in-app button.",
+            "No E2EE key or other encryption system provided: Element call calls are always encrypted, please make sure the URL you're using to join this call has been retrieved using the in-app button.",
```
robintown:
```suggestion
            `No E2EE key or other encryption system provided: ${import.meta.env.VITE_PRODUCT_NAME || "Element Call"} calls are always encrypted. Please make sure the URL you're using to join this call has been retrieved using the in-app button.`,
```

public/locales/en-GB/app.json
```diff
@@ -60,8 +60,16 @@
...
-  "group_call_loader_failed_text": "Calls are now end-to-end encrypted and need to be created from the home page. This helps make sure everyone's using the same encryption key.",
+  "group_call_loader": {
+    "banned_body": "You got banned from the room.",
+    "banned_heading": "Banned",
+    "call_ended_body": "You got removed from the call.",
+    "call_ended_heading": "Call Ended",
+    "failed_heading": "Call not found",
+    "failed_text": "Calls are now end-to-end encrypted and need to be created from the home page. This helps make sure everyone's using the same encryption key.",
+    "knock_reject_body": "The room members declined your request to join.",
+    "knock_reject_heading": "Not allowed to join"
```
robintown: Some minor copy suggestions:

```suggestion
    "banned_body": "You have been banned from the room.",
    "banned_heading": "Banned",
    "call_ended_body": "You have been removed from the call.",
    "call_ended_heading": "Call ended",
    "failed_heading": "Call not found",
    "failed_text": "Calls are now end-to-end encrypted and need to be created from the home page. This helps make sure everyone's using the same encryption key.",
    "knock_reject_body": "The room members declined your request to join.",
    "knock_reject_heading": "Not allowed to join"
```

robintown (commented):
Sorry for such a large review, there's just a lot to look through here. Here goes!

src/settings/submit-rageshake.ts
AndrewFerr: FTR this rule is relatively new, as it was applied under a month ago by matrix-org/matrix-js-sdk@97844f0e47697010247f606c000ac61a4bee3242

src/e2ee/sharedKeyManagement.ts
AndrewFerr: I checked that this is indeed the case

src/e2ee/sharedKeyManagement.ts
```diff
@@ -68,30 +67,37 @@ const useKeyFromUrl = (): [string, string] | [undefined, undefined] => {
...
+      return {
+        kind: E2eeType.SHARED_KEY,
+        secret: passwordFormUrl,
+      } as SharedSecret;
+    if (room.hasEncryptionStateEvent()) {
+      return { kind: E2eeType.PER_PARTICIPANT } as PerParticipantE2EE;
+    }
+    return { kind: E2eeType.NONE } as EncryptionSystem;
+  }, [passwordFormUrl, room, roomId, storedPassword, urlRoomId]);
+  return e2eeSystem;
```
AndrewFerr: Paired with the above:

```suggestion
    if (!room) return { kind: E2eeType.NONE };
    if (storedPassword)
      return {
        kind: E2eeType.SHARED_KEY,
        secret: storedPassword,
      };
    if (urlRoomId === roomId)
      return {
        kind: E2eeType.SHARED_KEY,
        secret: passwordFormUrl,
      };
    if (room.hasEncryptionStateEvent()) {
      return { kind: E2eeType.PER_PARTICIPANT };
    }
    return { kind: E2eeType.NONE };
  }, [passwordFormUrl, room, roomId, storedPassword, urlRoomId]);
  return e2eeSystem;
```

src/room/LobbyView.tsx
toger5: It is white in both designs though: https://www.figma.com/file/AjvSw4lI9iGO5aTe4WAl4r/Element-Call?type=design&node-id=3901-72407&mode=design&t=vTwhO0KU481UpZud-0

src/room/LobbyView.module.css
toger5: I played around with the standard button state and did not like it. I forgot the exact issues however.

toger5: The reason for this was, that the disabled button is the wrong color (dark in dark theme and light in light theme)

src/room/GroupCallView.tsx
toger5: there were not unencrypted call in the SPA we always joieded that it can only do encrypted calls. Since this change introduces per sender keys in the SPA it only allows non shared secret calls in the SPA with shared secret. With the `!widget` it does not break unenecrypted calls in embededded mode.

I was debating if we would like to allow unencrypted calls in the SPA in general. but it would make us end up in a world where users could join a call where they used the wrong link so they can participate without warning but only see encrypted streams.

I am now wondering if this is a smaller issue then not beeing able to share a call link to unencrypted rooms...?

src/room/LobbyView.module.css
toger5: This really should use compound color. I need transparent backgrounds however. I tried relative colors but this was not supported on firefox...

src/room/GroupCallView.tsx
robintown: I'm not sure I get everything you've said, but… all I'm worried about is whether we're breaking existing call links. So if a link like https://call.element.dev/room/#/robin-unencrypted?roomId=!FkGOYsKvcjIQJsNpiu:call-unstable.ems.host, which (currently) leads to a unencrypted SPA call, will continue to work, then we're on the same page

src/Header.tsx
```diff
@@ -150,15 +150,17 @@ export const RoomHeaderInfo: FC<RoomHeaderInfoProps> = ({
...
           <UserProfileIcon
             width={20}
             height={20}
             aria-label={t("header_participants_label")}
           />
           <Text as="span" size="sm" weight="medium">
-            {t("participant_count", { count: participantCount })}
+            {participantCount
+              ? t("participant_count", { count: participantCount })
+              : t("participant_count_unkonwn")}
```
robintown: Why would the "unknown" text ever be triggered? Won't it just be hidden due to the condition a few lines above? (I do think that hiding it is the better UX option here, for the record)

robintown (approved):

src/room/GroupCallView.tsx
toger5: from the initial error message i figured that unenecrypted calls are never supported...
So yea the link you shared would be broken.
A shared secret room is also an unencrypted room.
So i though if the room is not encrypted and we dont have a key from the url or local storage this already implies we reach a  failiour mode (since all calls have to be encrypted)
Otherwise the new hook basically covers all cases
encrypted room -> per sender keys
unenecrypted room 
   - with password in url -> shared secret
   - with password in local storage -> shared secret
   - no password -> unenecrypted call
   
So I think we can entirely skip the error screen.
This makes it also possible to open a guest link to an unencrypted room in the spa.

toger5: Updated it to allow it intentional unencrypted calls.

package.json
```diff
@@ -62,7 +62,7 @@
     "i18next-http-backend": "^2.0.0",
     "livekit-client": "^2.0.2",
     "lodash": "^4.17.21",
-    "matrix-js-sdk": "github:matrix-org/matrix-js-sdk#d55c6a36df539f6adacc335efe5b9be27c9cee4a",
+    "matrix-js-sdk": "github:AndrewFerr/matrix-js-sdk#msc-3266-compliance",
```
AndrewFerr: This can be reverted back to upstream, either on `#develop` or `#e874468ba3e84819cf4b342d2e66af67ab4cf804` (which is commit for matrix-org/matrix-js-sdk#4155).

AndrewFerr (changes requested):

AndrewFerr (approved):

PR #2286, opened 2024-04-05
Fix participants disappearing during focus switch (probably)

robintown:
I discovered that this hook was calling `complete` on the returned observable almost immediately when it gets mounted. This caused the call view model to never know when the application was switching focuses. At first I thought this was just because I forgot to move the call to `complete` to the effect's clean-up function, but even with that changed, React still calls the effect twice in strict mode. So, let's just remove the call entirely.

### DIFF
--- a/src/state/useObservable.ts
+++ b/src/state/useObservable.ts
@@ -1,5 +1,5 @@
 /*
-Copyright 2023 New Vector Ltd
+Copyright 2023-2024 New Vector Ltd
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


AndrewFerr (approved):
Should be fine, especially if it was tested to work

src/state/useObservable.ts
```diff
@@ -14,18 +14,17 @@ See the License for the specific language governing permissions and
...
- * replays its current value upon subscription, emits whenever the value
- * changes, and completes when the component is unmounted.
+ * replays its current value upon subscription and emits whenever the value
+ * changes.
  */
 export function useObservable<T>(value: T): Observable<T> {
   const subject = useRef<BehaviorSubject<T>>();
   subject.current ??= new BehaviorSubject(value);
   if (value !== subject.current.value) subject.current.next(value);
-  useEffect(() => subject.current!.complete(), []);
```
toger5, line 29: What about:
```ts
useEffect(() => subject.current!.complete, []);
```
Calling complete on unmount makes sense to me. The only issue seems to be the `()` after complete so that the effect fn does not return the complete method but instead runs it?

bblacher:
Could this be related to https://github.com/element-hq/element-web/issues/27114?

robintown:
@bblacher No, I don't believe that issue is related

src/state/useObservable.ts
robintown: You'd think that `useEffect(() => () => subject.current!.complete(), []);` would work as intended and only complete the observable on unmount, but due to React strict mode, [the effect will actually be run twice](https://react.dev/reference/react/StrictMode#fixing-bugs-found-by-re-running-effects-in-development): when mounted, the clean-up code is called almost immediately.

PR #2296, opened 2024-04-09
attempt to create docker pr images

fkwp:


.github/workflows/docker-pr.yaml
```diff
@@ -0,0 +1,61 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          github-token: ${{ secrets.ELEMENT_BOT_TOKEN }}
+          run_id: ${{ github.event.workflow_run.id }}
+          name: build
+          path: dist
+      
+      - name: Get current time
+        id: current-time
+        run: echo "unix_time=$(date +'%s')" >> $GITHUB_OUTPUT
```
robintown: It doesn't look like this value is used anywhere

robintown (approved):
Seems like it should work

.github/workflows/docker-pr.yaml
fkwp: right -- that's a leftover from the lastest_ci ... will remove

PR #2301, opened 2024-04-10
gh actions refactor

fkwp:


toger5 (approved):

robintown (dismissed):
I think this will break the Netlify PR workflow, since the name of the build artifact hasn't been updated there.

fkwp:
> I think this will break the Netlify PR workflow, since the name of the build artifact hasn't been updated there.

I think I duplicated the workflow for now. And if reusing workflows work as expected I can update the netlify pr as well

robintown:
Oh wait, though there are two build workflows now?

fkwp:
> Oh wait, though there are two build workflows now?

yes. build.yaml    & build_resuable.yaml

Issue #1613, opened 2023-09-26
Layout shift from 12 to >12 people is horrible


ara4n:
### Steps to reproduce

1. Be on a 12 person call, in a nice wide 4*3 layout.  Double clicking on a user zooms them to almost full screen, while leaving all other users visible.
2. 13th person joins
3. Layout reforms into a very small grid of thumbnails, which don't use all available space vertically /or/ horizontally (with large amounts of dead space to the left & right, if viewed on a typical fullscreen 1920x1080 style resolution laptop screen). Double-clicking on a user zooms them slightly (3x wider and 2x higher than other thumbnails).

#### What did you expect?

Rather than a huge layout shift from 12 to 13 users, keep the same pattern going - e.g. use grid layout for everything.

### URL for webapp

call.element.io

### Will you send logs?

No

nadonomy:
From discussion with @americanrefugee @fkwp: This should be solved when we move to a single grid engine.

robintown:
https://github.com/element-hq/element-call/pull/2325

Issue #1210, opened 2023-07-07
Prioritise making users with cameras enabled visible over users with cameras muted


jakewb-b:
### Your use case

As a lot of users join a call, some with cameras on and some with cameras off, we end up with a grid that has a mix of video feeds and blank avatars, and lots of users with cameras on are way down at the bottom. There is no way to resolve this except individually bringing all video-on users to the top, one by one, which is painful. 

There are a few things we could consider - some or all of the below might work, but this needs some UX thought.

- As a user joins, if their camera is on they should be sorted to top-left, if their camera is off, they should sort to bottom-right. But this could still gradually become mixed again as users turn cameras on/off.
- When a user turns their camera off, should they then be moved to bottom-right?
- Likewiser, when a user turns their camera *on* should they be moved to top? Moving them right to the top might draw more attention to them than they would want, so maybe just move into the visible area as if they were a speaker.
- Another option would be a 'sort all visible users to the top' button that just does it all for me as a one-off.

toger5:
I agree with the thought that we should not overexpose (pushing them to the top) ppl on unmuting the camera. 
I like putting ppl to the bottom on mute however.
It could be enough to implement:
 - on mute: move to bottom
 - on show: move next to last shown tile
 - on join muted: move to bottom
 - on join unmuted: move next to last shwon tile
 
This most likely wont be perfect in all scenarios but it should make it very comfortable to turn on the camera without the fear of beeing "featured" too agressivly and still result in most unmuted videos beeing in the upper secion.

I also really like the button option. But i feel like we might want to have a conversation about the general ux direction we want to go with:
 - all automatic
 - all maual with lots of tools
 - one view to rule them all / multiple views for each purpose
 - ...

robintown:
https://github.com/element-hq/element-call/pull/2325

Issue #1203, opened 2023-07-07
Make it easier for a user to locate their own tile in large grid layout


jakewb-b:
### Your use case

Even if I am not speaking, I may want to find my own tile and bring it into the visible space so I can keep an eye on how I appear. On very large calls it might be difficult to locate myself, and if lots of people are speaking I may eventually get pushed back down out of the visible area. We should consider how to:

- Locate my own tile easily
- Bring my own tile into the visible area easily
- Keep my own tile in the visible area even if other users are bring 'promoted'

robintown:
We now always keep your own tile in the very first position in the grid by default.

robintown:
Actually, this is waiting on a PR to land

robintown:
https://github.com/element-hq/element-call/pull/2325

Issue #1154, opened 2023-06-27
Ensure screenshares are brought into the visible space in freedom mode


jakewb-b:
Observed on the Monday sync so needs validating, but it seems as if screenshares can appear outside of the visible area of the screen?
It would make sense to:
- Treat the person screensharing as a speaker even if they haven't actually spoken - i.e. ensure they are promoted to the visible part of the screen if not already there.
- Start their screenshare next to their tile, making the connection between them clear and ensuring that the screenshare is also visible. 
- There is an edge case where the person sharing was on the far bottom right of the visible area of the screen, meaning they are already visible so won't be promoted, but if their screenshare is placed next to them it won't be visible. In this case, the speaker should be bumped up one slot so that both their camera and screenshare are visible.

Does this seem ok?

robintown:
This behavior is generally something the big grid layout already tries to do, but it's inadequate at it, so treating this as a defect

robintown:
https://github.com/element-hq/element-call/pull/2325

PR #2325, opened 2024-04-18
Unified grid layout

robintown:
Here I've implemented an MVP for the new unified grid layout, which scales smoothly up to arbitrarily many participants. It doesn't yet have a special 1:1 layout, so in spotlight mode and 1:1s, we will still fall back to the legacy grid systems.

Things that happened along the way:
- The part of VideoTile that is common to both spotlight and grid tiles, I refactored into MediaView
- VideoTile renamed to GridTile
- Added SpotlightTile for the new, glassy spotlight designs
- NewVideoGrid renamed to Grid, and refactored to be even more generic
- I extracted the media name logic into a custom React hook
- Deleted the BigGrid experiment

Closes https://github.com/element-hq/element-call/issues/1613
Closes https://github.com/element-hq/element-call/issues/1203
Closes https://github.com/element-hq/element-call/issues/1210
Closes https://github.com/element-hq/element-call/issues/1154
Closes https://github.com/element-hq/element-call/issues/1133
Closes https://github.com/element-hq/element-call/issues/597


src/grid/Grid.module.css
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2023 New Vector Ltd
+Copyright 2023-2024 New Vector Ltd
```
AndrewFerr, line 2: Nitpick: for consistency, this could be updated for all modified files.

src/useReactiveState.ts
```diff
@@ -44,7 +44,7 @@ export const useReactiveState = <T>(
   if (
     prevDeps.current === undefined ||
     deps.length !== prevDeps.current.length ||
-    deps.some((d, i) => d !== prevDeps.current![i])
+    deps.some((d, i) => !Object.is(d, prevDeps.current![i]))
```
AndrewFerr: What was the rationale for changing this?  As long as the entries of `deps` aren't expected to be `NaN` or signed zeros, using `!==` should be equivalent: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#description

src/room/InCallView.tsx
```diff
@@ -244,8 +260,21 @@ export const InCallView: FC<InCallViewProps> = subscribe(
       connState,
     );
     const items = useStateObservable(vm.tiles);
+    const layout = useStateObservable(vm.layout);
+    const hasSpotlight = layout.spotlight !== undefined;
+    // Hack: We insert a dummy "spotlight" tile into the tiles we pass to
+    // useFullscreen so that we can control the fullscreen state of the
+    // spotlight tile in the new layouts with this same hook.
+    const fullscreenItems = useMemo(
+      () => [...items, ...(hasSpotlight ? [dummySpotlightItem] : [])],
```
AndrewFerr: Nitpick
```suggestion
      () => hasSpotlight ? items.concat(dummySpotlightItem) : items,
```

src/room/InCallView.tsx
```diff
@@ -259,14 +288,109 @@ export const InCallView: FC<InCallViewProps> = subscribe(
...
+    );
+
+    const showSpeakingIndicators =
+      layout.type === "spotlight" ||
+      (layout.type === "grid" && layout.grid.length > 2);
+
+    const SpotlightTileView = useMemo(
+      () =>
+        forwardRef<HTMLDivElement, TileProps<MediaViewModel[], HTMLDivElement>>(
+          function SpotlightTileView(
```
AndrewFerr, line 347: Nitpick, I think this can be an anonymous function
```suggestion
          function (
```

src/grid/LegacyGrid.tsx
```diff
@@ -838,20 +839,19 @@ export interface ChildrenProperties<T> {
...
 }
 
-export interface VideoGridProps<T> {
+export interface LegacyGridProps<T, R extends HTMLElement> {
   items: TileDescriptor<T>[];
   layout: Layout;
   disableAnimations: boolean;
-  layoutStates: LayoutStatesMap;
-  children: (props: ChildrenProperties<T>) => ReactNode;
+  Tile: ComponentType<TileProps<T, R>>;
```
AndrewFerr, line 846: Nitpick, should `Tile` be lower-cased when it's a field name?  This would also avoid confusion with the `Tile` type on line 58.

src/state/CallViewModel.ts
```diff
@@ -154,11 +154,11 @@ class UserMedia {
     this.vm = new UserMediaViewModel(id, member, participant, callEncrypted);
 
     this.speaker = this.vm.speaking.pipeState(
-      // Require 1 s of continuous speaking to become a speaker, and 10 s of
+      // Require 1 s of continuous speaking to become a speaker, and 60 s of
```
AndrewFerr, line 157: Was 10 seconds really too short?  60 feels like a lot

src/useReactiveState.ts
robintown: I changed it because we do actually get NaN in the dependencies sometimes, now. (They come from the grid arrangement code when the width and height are not yet determined, giving aspect ratio = 0 / 0.)

Adding a comment.

src/room/InCallView.tsx
robintown: It can't unfortunately, because then the component will be missing a display name for React devtools (which the linter rightfully complains about)

src/grid/LegacyGrid.tsx
robintown: It seems to be a React convention to capitalize names that refer to components, even field names or prop names. Part of the reason is that a camel case name can't be used to refer to a component in JSX, or else React will think you're trying to reference an "intrinsic element" like `div`, `button`, etc.

src/state/CallViewModel.ts
robintown: It did seem to be too short in practice. In our team meetings for example, speakers often hold the floor for 30s at a time. I don't think we're at risk of the speaker section becoming overly large in large calls, because the number of people that speak during any given minute has a somewhat low upper limit - and I do think 60s has been feeling pretty good IMO!

AndrewFerr (approved):

PR #2327, opened 2024-04-19
Hotfix v0.5.14

fkwp:
Important Update: Upgrade LiveKit SDK for Chrome 124 Compatibility. With Chrome 124 releasing on April 16th, there's a breaking change in the setCodecPreferences API. Bumping livekit-client to v1.15.13.


robintown:
We ended up not finding this necessary

PR #2333, opened 2024-04-23
refactor run-id handling of GitHub Actions 

fkwp:


.github/workflows/pr-deploy.yaml
```diff
@@ -26,6 +26,7 @@ jobs:
       deployments: write
     uses: ./.github/workflows/netlify.yaml
     with:
+      artifact_run_id: ${{ github.event.workflow_run.id || github.run_id }}
```
robintown, line 29: Is the `|| github.run_id` part necessary if this parameter also has that as its default?

robintown (approved):

.github/workflows/pr-deploy.yaml
fkwp: tbd. I don't know. Not sure if it would handover undefined or similar. Hence, rather a little redundant

PR #2352, opened 2024-04-30
Remove unsupported LiveKit config option

robintown:
This config option is no longer supported by LiveKit (https://github.com/livekit/livekit/commit/10c8582a6b01c37fb2db7e343b5204c713fbb7cd) and will crash the server if specified.


AndrewFerr:
Haha, we both saw this at the same time :laughing: https://github.com/element-hq/element-call/pull/2349

PR #2355, opened 2024-05-02
Join knock rooms on page load if already invited

AndrewFerr:
When visiting the page for a knock room you are already invited to, join it right away instead of offering to knock (which will fail as long as you remain invited to the room).

### DIFF
--- a/src/room/useLoadGroupCall.ts
+++ b/src/room/useLoadGroupCall.ts
@@ -203,7 +203,8 @@ export const useLoadGroupCall = (
         //  - in SPA mode if the user already joined the room
         room = client.getRoom(roomId);
         activeRoom.current = room ?? undefined;
-        if (room?.getMyMembership() === KnownMembership.Join) {
+        const membership = room?.getMyMembership();
+        if (membership === KnownMembership.Join) {
           // room already joined so we are done here already.
           return room!;
         }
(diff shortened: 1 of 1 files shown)


AndrewFerr:
Issue: when visiting a knock room & auto-joining it, the room header shows the room ID, not the room name.  Does this mean a callback is missing somewhere?

robintown (approved):

robintown:
> Issue: when visiting a knock room & auto-joining it, the room header shows the room ID, not the room name. Does this mean a callback is missing somewhere?

Don't know about that one… in theory we update the name shown in the UI whenever the room emits a `RoomEvent.Name` event, so there could be a bug that matrix-js-sdk isn't signaling a name update in this case. Or, it's possible that there are multiple room objects floating around for the same room ID, and one of them is just a placeholder with only the room ID filled in. ([that has happened before](https://github.com/matrix-org/matrix-js-sdk/pull/3747))

AndrewFerr:
Eugh, there's something else funky about this.  Despite joining the room, a joined user forever sees "Waiting for other participants…" even if other users have joined the call.

Putting this in draft until that's solved.

AndrewFerr:
I did the sensible thing of testing the Netlify build of this instead of my local setup, and can reproduce _not_ hitting the problem of a missing room name & getting stuck waiting for other call members.

At this point I'm inclined to believe those issues are unrelated to this PR, especially given how simple this PR is.  So, I shall merge this.

PR #2358, opened 2024-05-03
Load focus information from well known and use client config only as a fallback.

toger5:
Instead of always requireig EC to configure a livekit service url we use the homeserver well-known to aquire the sfu to use.

Relies on: https://github.com/matrix-org/matrix-js-sdk/pull/4195

Discussion:
~~If the homserver does not promote a service url in the well-known we start searching for one on the other homeservers involved in the room. (check user domains)
The reason for this is, that we currently cannot start nor join a call without knowing a sfu service url since we wont just copy over what others have set in their state event.
Even if taht would be possible we could run into a situation where all members with a set foci_active disconnect and only members without foci active are left which would break the call immediately.
So making it a requirement to have a active foci makes sense.
Hence always loading one via the room member list might be a solution~~

In the current version we have not activated the option to load from other room members.
We add Foci from (in the order listed here)
 - the current rtc session
 - the hs well-known
 - the config
 - DEACTIVATED from other room members homeserver well known.


README.md
```diff
@@ -64,6 +64,28 @@ experimental_features:
 MSC3266 allows to request a room summary of rooms you are not joined.
 The summary contains the room join rules. We need that to decide if the user gets prompted with the option to knock ("ask to join"), a cannot join error or the join view.
 
+Element call requires a livekit SFU behind a livekit jwt service so work. The url to the livekit jwt service can either be configured in the config of element call (fallback/legacy configuration) or be configured by your homeserver via the `.well-known`.
```
AndrewFerr:
```suggestion
Element call requires a livekit SFU behind a livekit jwt service to work. The url to the livekit jwt service can either be configured in the config of element call (fallback/legacy configuration) or be configured by your homeserver via the `.well-known`.
```

README.md
```diff
@@ -64,6 +64,28 @@ experimental_features:
...
+        "type":"livekit",
+        "livekit_service_url":"https://someurl.com"
+    },
+     {
+        "type":"livekit",
+        "livekit_service_url":"https://livekit2.com"
+    },
+    {
+        "type":"another_foci",
+        "props_for_another_foci":"val"
```
AndrewFerr: Nitpick, put a space after colons:
```suggestion
        "type": "livekit",
        "livekit_service_url": "https://someurl.com"
    },
     {
        "type": "livekit",
        "livekit_service_url": "https://livekit2.com"
    },
    {
        "type": "another_foci",
        "props_for_another_foci": "val"
```

README.md
AndrewFerr, line 1: Nitpick, but capitalize "Element Call" and "Livekit", which looks more professional IMO.

README.md
```diff
@@ -3,7 +3,7 @@
 [![Chat](https://img.shields.io/matrix/webrtc:matrix.org)](https://matrix.to/#/#webrtc:matrix.org)
 [![Localazy](https://img.shields.io/endpoint?url=https%3A%2F%2Fconnect.localazy.com%2Fstatus%2Felement-call%2Fdata%3Fcontent%3Dall%26title%3Dlocalazy%26logo%3Dtrue)](https://localazy.com/p/element-call)
 
-Group calls with WebRTC that leverage [Matrix](https://matrix.org) and an open-source WebRTC toolkit from [LiveKit](https://livekit.io/).
+Group calls with WebRTC that leverage [Matrix](https://matrix.org) and an open-source WebRTC toolkit from [LiveKit](https://Livekit.io/).
```
AndrewFerr: Nitpick because domain names are case-insensitive, but I like lowercase for URLs/JSON/variables and uppercase just for natural language:
```suggestion
Group calls with WebRTC that leverage [Matrix](https://matrix.org) and an open-source WebRTC toolkit from [LiveKit](https://livekit.io/).
```

README.md
```diff
@@ -103,11 +125,13 @@ service for development. These use a test 'secret' published in this
...
-To use it, add SFU parameter in your local config `./public/config.json`:
+To use it, add a SFU parameter in your local config `./public/config.json`:
+(Be aware, that this is only the fallback Livekit SFU. If the homeserver
+advertises one in the client well-known, this will not be used.)
 
 ```json
-"livekit": {
-  "livekit_service_url": "http://localhost:8881"
+"Livekit": {
+  "Livekit_service_url": "http://localhost:8881"
```
AndrewFerr: AFAIK JSON is case-sensitive.
```suggestion
"livekit": {
  "livekit_service_url": "http://localhost:8881"
```

src/rtcSessionHelpers.ts
```diff
@@ -15,29 +15,92 @@ limitations under the License.
...
+): Promise<LivekitFocus[]> {
+  logger.log("Start building foci_preferred list: ", rtcSession.room.roomId);
+
+  const preferredFoci: LivekitFocus[] = [];
+
+  // Make the foci from the running rtc session the highest priority one
+  // This minimizes how often we need to switch foci during a call.
+  const focusFromMatrixRTC = rtcSession
+    .getOldestMembership()
+    ?.getPreferredFoci()[0];
```
AndrewFerr: Does this need to check for `.focus_selection === "oldest_membership"`?  It would be just for the sake of completion since there's no other selection type, but would this be the place to put that check to indicate that there could be more in the future?

src/rtcSessionHelpers.ts
```diff
@@ -15,29 +15,92 @@ limitations under the License.
...
+    FOCI_WK_KEY
+  ] as Focus[];
+  if (wellKnownFoci.length > 0) {
+    logger.log("Adding livekit focus from well known: ", wellKnownFoci);
+    preferredFoci.push(
+      ...wellKnownFoci
+        .filter(isLivekitFocusConfig)
+        .map((f) => ({ ...f, livekit_alias: livekitAlias })),
+    );
+  }
```
AndrewFerr: Since the well-known is essentially a user-provided value, this should check that `wellKnownFoci` really is an array.  There's also no point to typing it as a `Focus[]` thanks to the clever usage of `.filter(isLivekitFocusConfig)` providing type narrowing.

Also, small suggestion: only the valid objects from the well-known should be logged.  Either log them all after `filter`ing them, or hack at `map` to print them one at a time.

```suggestion
  const wellKnownFoci = rtcSession.room.client.getClientWellKnown()?.[
    FOCI_WK_KEY
  ];
  if (Array.isArray(wellKnownFoci)) {
    preferredFoci.push(
      ...wellKnownFoci
        .filter(isLivekitFocusConfig)
        .map((wellKnownFocus) => {
          logger.log("Adding livekit focus from well known: ", wellKnownFocus);
          return { ...wellKnownFocus, livekit_alias: livekitAlias };
        })
    );
  }
```

src/rtcSessionHelpers.ts
```diff
@@ -15,29 +15,92 @@ limitations under the License.
...
+
+  const urlFromConf = Config.get().livekit?.livekit_service_url;
+  if (urlFromConf) {
+    const focusFormConf = {
+      type: "livekit",
+      livekit_service_url: urlFromConf,
+      livekit_alias: livekitAlias,
+    } as LivekitFocus;
```
AndrewFerr: Nitpick
```suggestion
    const focusFormConf: LivekitFocus = {
      type: "livekit",
      livekit_service_url: urlFromConf,
      livekit_alias: livekitAlias,
    };
```

src/rtcSessionHelpers.ts
```diff
@@ -15,29 +15,92 @@ limitations under the License.
...
+    logger.log("Adding livekit focus from config: ", focusFormConf);
+    preferredFoci.push(focusFormConf);
+  }
+
+  if (preferredFoci.length === 0)
+    throw new Error(
+      `No livekit_service_url is configured so we could not create a focus
+    currently we skip computing a focus based on other users in the room.`,
```
AndrewFerr:
```suggestion
      `No livekit_service_url is configured so we could not create a focus.
    Currently we skip computing a focus based on other users in the room.`,
```

src/rtcSessionHelpers.ts
```diff
@@ -15,29 +15,92 @@ limitations under the License.
...
+
+  return preferredFoci;
+
+  // Currently we skip computing a focus based on other users in the room.
+  const focusOtherMembers = await focusFromOtherMembers(
+    rtcSession,
+    livekitAlias,
+  );
+  if (focusOtherMembers) preferredFoci.push(focusOtherMembers);
+}
```
AndrewFerr: Maybe it's a good idea to have a config flag to allow this code to run.  Otherwise, `focusFromOtherMembers` and its dependencies are unreachable & non-testable.  Alternatively, we could move it to another PR once this one is merged.

src/rtcSessionHelpers.ts
```diff
@@ -15,29 +15,92 @@ limitations under the License.
...
+
+  // Prioritize the client well known over the configured sfu.
+  const wellKnownFoci = rtcSession.room.client.getClientWellKnown()?.[
+    FOCI_WK_KEY
+  ] as Focus[];
+  if (wellKnownFoci.length > 0) {
+    logger.log("Adding livekit focus from well known: ", wellKnownFoci);
+    preferredFoci.push(
+      ...wellKnownFoci
+        .filter(isLivekitFocusConfig)
```
AndrewFerr, line 62: Because `isLivekitFocusConfig` (and the others) don't handle `null`/`undefined`, maybe this is a good idea:
```suggestion
        .filter(f => !!f)
        .filter(isLivekitFocusConfig)
```

src/rtcSessionHelpers.ts
```diff
@@ -74,3 +141,70 @@ export async function leaveRTCSession(
...
+
+async function fetchAndCacheLivekitFocus(
+  homeserverDomain: string,
+): Promise<LivekitFocusConfig | undefined> {
+  const cachedHomeserverFocus = livekitFocusCache.get(homeserverDomain);
+  if (cachedHomeserverFocus) return cachedHomeserverFocus;
+
+  const fetchedHomeserverFocus = (
+    await AutoDiscovery.getRawClientConfig(homeserverDomain)
+  )[FOCI_WK_KEY] as LivekitFocusConfig | undefined;
```
AndrewFerr: Won't this retrieve an array of focus configs instead of a single one?  If so, perhaps the typechecking+filtering done on the client well-known in `makePreferredFoci` could be split into a function & called here.

src/room/GroupCallView.tsx
```diff
@@ -318,7 +318,7 @@ export const GroupCallView: FC<Props> = ({
         client={client}
         matrixInfo={matrixInfo}
         muteStates={muteStates}
-        onEnter={(): void => enterRTCSession(rtcSession, perParticipantE2EE)}
+        onEnter={() => enterRTCSession(rtcSession, perParticipantE2EE)}
```
AndrewFerr: Should the type of `Props.onEnter` in LobbyView.tsx be changed to `() => Promise<void>`?

Or if `onEnter` is meant to ignore the `Promise<void>` return value of `enterRTCSession`, make it explicit with this:
```suggestion
        onEnter={() => void enterRTCSession(rtcSession, perParticipantE2EE)}
```

AndrewFerr (changes requested):

AndrewFerr (approved):

Issue #2348, opened 2024-04-30
Visiting a private room has UI provide a knock button or "Call not found"


AndrewFerr:
### Steps to reproduce

1. Create a guest link for a knock/public room
2. Change the room access to private
3. Visit the guest link in a new tab

#### What did you expect?
The guest link's ECall should report the room as being private.

#### What happened instead?
One of the two outcomes:

1) The guest link's ECall offers to send a knock to the room, which unsurprisingly fails & shows the server error of:
> MatrixError: [403] You don't have permission to knock (https://matrix.guest.ess-ecall-poc.ems-support.element.dev/_matrix/client/v3/knock/!xDTBDZXIBfULuWWyMZ%3Aess-ecall-poc.ems-support.element.dev?server_name=ess-ecall-poc.ems-support.element.dev)

2) The guest link's ECall reports the call as not found, and displays a message of:
> Call not found
> 
> Calls are now end-to-end encrypted and need to be created from the home page. This helps make sure everyone's using the same encryption key.

### URL for webapp

https://call.guest.ess-ecall-poc.ems-support.element.dev/

### Will you send logs?

No

AndrewFerr:
Outcome 1) must be due to ECall caching the room summary, because 2) is the fallback for when the room can't be found, which should be the case for a private room.

AndrewFerr:
As it turns out, even Element Web gives a "knock" button for private or non-existent rooms, though it gives a better error message for the former, of:

> Failed to join
>
> You need an invite to access this room.

PR #2360, opened 2024-05-03
Reword message for failure to join/find room

AndrewFerr:
The message originally focused on the old feature of being able to create a room with a custom URL. Instead, be more direct & say that the current URL is for an inaccessible room.

This new message is based on what Element Web says for this scenario.

### DIFF
--- a/public/locales/en-GB/app.json
+++ b/public/locales/en-GB/app.json
@@ -65,8 +65,8 @@
     "banned_heading": "Banned",
     "call_ended_body": "You have been removed from the call.",
     "call_ended_heading": "Call ended",
-    "failed_heading": "Call not found",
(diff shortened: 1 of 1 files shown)


AndrewFerr:
Partially addresses #2348.  The issue of a knock button sometimes appearing for private rooms is likely caused by a stale room membership cache, and won't be addressed by this PR.

AndrewFerr:
Also note that the condition for showing this message is an error response in getting the [room summary](https://github.com/deepbluev7/matrix-doc/blob/room-summaries/proposals/3266-room-summary.md), which returns `40x` for both private & non-existent rooms.

toger5 (approved):

PR #2361, opened 2024-05-03
Add more valid rooms to the recents list

AndrewFerr:
Include:
- all rooms you are a member of
- knock rooms you've knocked on and are waiting for an invite to
- knock rooms you've been invited to in response to a knock

### DIFF
--- a/src/home/useGroupCallRooms.ts
+++ b/src/home/useGroupCallRooms.ts
@@ -89,18 +89,44 @@ const roomIsJoinable = (room: Room): boolean => {
     return false;
   }
   // otherwise we can always join rooms because we will automatically decide if we want to use perParticipant or password
-  const joinRule = room.getJoinRule();
-  return joinRule === JoinRule.Knock || joinRule === JoinRule.Public;
+  switch (room.getJoinRule()) {
+    case JoinRule.Public:
+      return true;
+    case JoinRule.Knock:
+      switch (room.getMyMembership()) {
+        case KnownMembership.Join:
+        case KnownMembership.Knock:
+          return true;
+        case KnownMembership.Invite:
+          return (
+            room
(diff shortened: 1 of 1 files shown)


src/home/useGroupCallRooms.ts
```diff
@@ -89,18 +89,44 @@ const roomIsJoinable = (room: Room): boolean => {
...
+        case KnownMembership.Knock:
+          return true;
+        case KnownMembership.Invite:
+          return (
+            room
+              .getLiveTimeline()
+              .getState(EventTimeline.FORWARDS)
+              ?.getStateEvents(EventType.RoomMember, room.myUserId)
+              ?.getPrevContent().membership === JoinRule.Knock
+          );
```
toger5, line 107: I am wondering if a normal invite should also be in the room list.
In case I am inviting a guest user from callA to a new callB.

But then this would break the semantic of recents...

So I think were you are coming from and if you prefer to stick to what one would expect from recents but as a consequence breaking the usecase outlined above I am all for it.

(The above would just need to be handled like so: invite geust to callB send guest invite link -> with you additions that would result in an immediate join)

toger5 (approved):

src/home/useGroupCallRooms.ts
AndrewFerr: I agree on both points: it'd be useful to see normal invites, but not as an item in the recents list.

What's more, ECall tends to try to auto-join rooms, so showing general-invite rooms in the recents list would maybe auto-accept them if we don't add an exception for them.

As a compromise, we could add a dedicated section for "Invites" separate from the recents list.  Though that's something a new issue/PR should address.

AndrewFerr:
Force-pushed just to update the commit message.

PR #2368, opened 2024-05-08
Refactor settings to use observables

robintown:
Also removing some unused settings along the way.

Based on https://github.com/element-hq/element-call/pull/2325


src/settings/settings.ts
```diff
@@ -0,0 +1,98 @@
...
+ */
+export function useSetting<T>(setting: Setting<T>): [T, (value: T) => void] {
+  return [useObservableEagerState(setting.value), setting.setValue];
+}
+
+// TODO: This doesn't belong here
+export const isFirefox = (): boolean => {
+  const { userAgent } = navigator;
+  return userAgent.includes("Firefox");
+};
```
AndrewFerr: What's a better place for this?

src/settings/settings.ts
```diff
@@ -0,0 +1,98 @@
...
+// TODO: This setting can be disabled. Work out an approach to disableable
+// settings thats works for Observables in addition to React.
+export const useOptInAnalytics = (): [
+  boolean | null,
+  ((value: boolean | null) => void) | null,
+] => {
+  const setting = useSetting(optInAnalytics);
+  if (PosthogAnalytics.instance.isEnabled()) return setting;
+
+  return [false, null];
```
AndrewFerr:
```suggestion
  PosthogAnalytics.instance.isEnabled() ? useSetting(optInAnalytics) : [false, null];
```

robintown: Hooks can't be called conditionally (it's one of the more disappointing things about React, IMO), but sure, this could be abbreviated

AndrewFerr (approved):

PR #2369, opened 2024-05-08
Add a developer option to duplicate tiles

robintown:
This is useful for testing how the UI behaves with different numbers of participants.

Based on https://github.com/element-hq/element-call/pull/2368

### DIFF
--- a/public/locales/en-GB/app.json
+++ b/public/locales/en-GB/app.json
@@ -132,7 +132,7 @@
     "developer_settings_label": "Developer Settings",
     "developer_settings_label_description": "Expose developer settings in the settings window.",
     "developer_tab_title": "Developer",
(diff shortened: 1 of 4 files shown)


toger5 (approved):
Cool. I Like how simple this kind of change is!

public/locales/en-GB/app.json
```diff
@@ -132,6 +132,7 @@
     "developer_settings_label": "Developer Settings",
     "developer_settings_label_description": "Expose developer settings in the settings window.",
     "developer_tab_title": "Developer",
+    "duplicate_tiles_label": "Number of duplicate tiles",
```
toger5: I think we can phrase this less technical (even though its in the developer settings)

```suggestion
    "duplicate_tiles_label": "Number of additional tiles copies per participant.",
```

PR #2372, opened 2024-05-10
Wait for a full sync on startup

### LINKED ISSUE #2359: ECall can think you are joined to rooms that you've left
### Steps to reproduce

1. Open Element Web & Element Call, and log in with the same account for both
2. Accept an invite to a non-private, encrypted room with Element Web (knock rooms work best)
3. Join a call in that room in Element Call (just to confirm that the call works)
4. End the call in Element Call
5. Leave or be removed from the room in Element Web
6. Refresh the Element Call page

#### What did you expect?
Element Call shouldn't show the room in the recents list, because you are no longer a member of it. 

#### What happened instead?
Element Call may show the room & allow you to start a call in it, but doesn't actually join you to the room (as it thinks you have already joined).  This means you won't receive media streams of other users for calls in that room.

This doesn't always happen though, and if you wait on the ECall home page for long enough, the room might eventually disappear from the recents list.

In any case, it wouldn't hurt if ECall were to refresh is cache of joined rooms more often.

### Operating system

Fedora

### Browser information

Firefox 125.0.2

### URL for webapp

_No response_

### Will you send logs?

No

AndrewFerr:
This avoids missing any state not included in the latest cached sync.

Fixes #2359

### DIFF
--- a/src/matrix-utils.ts
+++ b/src/matrix-utils.ts
@@ -55,7 +55,10 @@ const SYNC_STORE_NAME = "element-call-sync";
 // (It's a good opportunity to make the database names consistent.)
 const CRYPTO_STORE_NAME = "element-call-crypto";
 
-function waitForSync(client: MatrixClient): Promise<void> {
+async function waitForSync(client: MatrixClient): Promise<void> {
+  // If there is a saved sync, wait for it to be restored before the first real sync,
+  // as the saved sync may be missing some state
+  let waitForSavedSync = !!(await client.store.getSavedSyncToken());
   return new Promise<void>((resolve, reject) => {
     const onSync = (
       state: SyncState,
@@ -63,8 +66,12 @@ function waitForSync(client: MatrixClient): Promise<void> {
       data?: ISyncStateData,
     ): void => {
       if (state === "PREPARED") {
-        client.removeListener(ClientEvent.Sync, onSync);
-        resolve();
+        if (waitForSavedSync) {
+          waitForSavedSync = false;
+        } else {
+          client.removeListener(ClientEvent.Sync, onSync);
+          resolve();
+        }
       } else if (state === "ERROR") {
         client.removeListener(ClientEvent.Sync, onSync);
         reject(data?.error);
@@ -190,8 +197,11 @@ export async function initClient(
 
   await client.initCrypto();
   client.setGlobalErrorOnUnknownDevices(false);
+  // Apply listener before starting the client.
+  // Otherwise, the first sync could resolve before the listener gets applied.
+  const syncPromise = waitForSync(client);
   await client.startClient();
-  await waitForSync(client);
+  await syncPromise;
 
   return client;
 }
--- a/src/matrix-utils.ts
+++ b/src/matrix-utils.ts
@@ -56,8 +56,11 @@ const SYNC_STORE_NAME = "element-call-sync";
 const CRYPTO_STORE_NAME = "element-call-crypto";
 
 async function waitForSync(client: MatrixClient): Promise<void> {
-  // If there is a saved sync, wait for it to be restored before the first real sync,
-  // as the saved sync may be missing some state
+  // If there is a saved sync, the client will fire an additional sync event
+  // for restoring it before it runs the first network sync.
+  // However, the sync we want to wait for is the network sync,
+  // as the saved sync may be missing some state.
+  // Thus, don't resolve on the first sync when we know it's for the saved sync.
   let waitForSavedSync = !!(await client.store.getSavedSyncToken());
   return new Promise<void>((resolve, reject) => {
     const onSync = (
@@ -197,8 +200,11 @@ export async function initClient(
 
   await client.initCrypto();
   client.setGlobalErrorOnUnknownDevices(false);
-  // Apply listener before starting the client.
-  // Otherwise, the first sync could resolve before the listener gets applied.
+  // Once startClient is called, syncs are run asynchronously.
+  // Also, sync completion is communicated only via events.
+  // So, apply the event listener *before* starting the client.
+  // Otherwise, a sync may complete before the listener gets applied,
+  // and we will miss it.
   const syncPromise = waitForSync(client);
   await client.startClient();
   await syncPromise;


src/matrix-utils.ts
```diff
@@ -55,16 +55,23 @@ const SYNC_STORE_NAME = "element-call-sync";
...
     const onSync = (
       state: SyncState,
       _old: SyncState | null,
       data?: ISyncStateData,
     ): void => {
       if (state === "PREPARED") {
-        client.removeListener(ClientEvent.Sync, onSync);
-        resolve();
+        if (waitForSavedSync) {
+          waitForSavedSync = false;
```
toger5, line 73: This variable update seems to never be used. we write to it after reading it after the only place where we read it.

src/matrix-utils.ts
```diff
@@ -190,8 +197,11 @@ export async function initClient(
 
   await client.initCrypto();
   client.setGlobalErrorOnUnknownDevices(false);
+  // Apply listener before starting the client.
+  // Otherwise, the first sync could resolve before the listener gets applied.
+  const syncPromise = waitForSync(client);
   await client.startClient();
-  await waitForSync(client);
+  await syncPromise;
```
toger5: instead of guaranteeing the listener gets setup I think we should check if it is already done syncing in `waitForSync` and resolve immediately in this case.

src/matrix-utils.ts
AndrewFerr: This indeed gets used: if the client has a cached sync result, `waitForSavedSync` gets set to `true` and the client will sync twice: first for the cached sync, and next for a "real" sync.  Each sync will fire a `ClientEvent.Sync` event, meaning `onSync` will get called twice.  Since `onSync` forms a closure over `waitForSavedSync`, it can be used to track whether `onSync` had been called or not; only if it has been called once before should the listener be removed (so that the listener can fire for both syncs) & the promise be resolved.

I also confirmed that removing `waitForSync = false` has the UI hang, as that prevents the promise from being able to resolve & thus has the UI think that no sync has finished.

src/matrix-utils.ts
AndrewFerr: As I understand it, no sync will happen until `startClient` is called, and since `startClient` [runs syncs asynchronously](https://github.com/matrix-org/matrix-js-sdk/blob/93d96281fd4cc8fd47c398eadac8b553662c85ac/src/client.ts#L1540) & communicates sync completion only via `ClientEvent.Sync` events, the only way to check if a sync has completed is to add a listener on that event -- and the only way to prevent a race condition between the event being fired & the listener being added is to add the listener before calling `startClient`.

I also don't think it would help to check if the client is done syncing first, because 1) if there is a cached sync state, the client will run two syncs on startup, and 2) after the initial sync, the client will be polling `/sync` forever.  I think this makes it hard to reason about when the client is really "done syncing", and it's much more straightforward to just listen for an expected sequence of `ClientEvent.Sync` events on startup.

robintown:
Possibly fixes https://github.com/element-hq/element-call/issues/2348

src/matrix-utils.ts
toger5: Oh missed the `onSync` connection inside the Promise.
I think it can then even be moved into the promise so the logic is as close together as possible
```
new Promise<void>((resolve, reject) => {
    let waitForSavedSync = !!(await client.store.getSavedSyncToken());
    const onSync = (
      state: SyncState,
      _old: SyncState | null,
      data?:
...
```

src/matrix-utils.ts
toger5: Thanks. Taking the cached sync into consideration this makes sense.
Could the comment be more detailed/explicit (mentioning the cached sync that happens right away on client start instead of calling it "first sync"?

toger5 (approved):
My comments were apparently just a result of not knowing enough about cached syncs.
If you could add comments with details on this would be great.

src/matrix-utils.ts
AndrewFerr: In order to do that, the executor function would have to be async, since `getSavedSyncToken` is async.

However, eslint advises against using async executors, and has a rule to forbid them: https://eslint.org/docs/latest/rules/no-async-promise-executor

src/matrix-utils.ts
AndrewFerr: a1105a8 adds some comments to explain this a bit better. Let me know if they need any improvements.

src/matrix-utils.ts
AndrewFerr: Given the above, I will leave this function as it is, unless you have another recommendation for improving it.

src/matrix-utils.ts
```diff
@@ -197,8 +200,11 @@ export async function initClient(
...
-  // Apply listener before starting the client.
-  // Otherwise, the first sync could resolve before the listener gets applied.
+  // Once startClient is called, syncs are run asynchronously.
+  // Also, sync completion is communicated only via events.
+  // So, apply the event listener *before* starting the client.
+  // Otherwise, a sync may complete before the listener gets applied,
+  // and we will miss it.
   const syncPromise = waitForSync(client);
   await client.startClient();
   await syncPromise;
```
toger5, line 210: if the syncs would only run asynchronously 
```
client.startClient();
await waitForSync(client);
```
should work right. the only reason why we are missing sync is because the initial one is synchronous and emitted before startClient returns.

AndrewFerr:
Re-requesting review to ask for feedback on:
- if the Promise in `waitForSync` [is fine as it is](https://github.com/element-hq/element-call/pull/2372#discussion_r1608648129)
- if the new comments in https://github.com/element-hq/element-call/commit/a1105a827333460e804088748080345a749cd18c [are okay](https://github.com/element-hq/element-call/pull/2372#discussion_r1608646788)

src/matrix-utils.ts
toger5: Sure that makes sense!

toger5:
The comment makes it very clear to me.

toger5 (approved):
Thanks for the changes

src/matrix-utils.ts
AndrewFerr: No, not even the first sync is synchronous.  `MatrixClient#startClient` doesn't `await` on any iteration of the sync loop, and instead resolves after just kicking off the loop (and having done some other async startup actions).

toger5: its actually suprising then that 
```
client.startClient();
await waitForSync(client);
```
does not work.
But in the end it depneds on the order of when async executions get added to the queue/loop.
So it also makes sense that the async code that emits is schduled before the even connection

PR #2380, opened 2024-05-16
Add toggle to always show yourself

robintown:
Based on https://github.com/element-hq/element-call/pull/2368

### DIFF
--- a/src/state/CallViewModel.ts
+++ b/src/state/CallViewModel.ts
@@ -61,9 +61,11 @@ import {
 } from "../livekit/useECConnectionState";
 import { usePrevious } from "../usePrevious";
 import {
+  LocalUserMediaViewModel,
   MediaViewModel,
-  UserMediaViewModel,
+  RemoteUserMediaViewModel,
   ScreenShareViewModel,
+  UserMediaViewModel,
 } from "./MediaViewModel";
 import { finalizeValue } from "../observable-utils";
 import { ObservableScope } from "./ObservableScope";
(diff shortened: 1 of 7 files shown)


src/state/CallViewModel.ts
```diff
@@ -129,14 +131,14 @@ export type WindowMode = "normal" | "full screen" | "pip";
...
-  SelfStart,
+  SelfAlwaysShown,
   Presenters,
   Speakers,
   VideoAndAudio,
   Video,
   Audio,
   NoMedia,
-  SelfEnd,
+  SelfNotAlwaysShown,
```
toger5: This does not immediately translate to a bin in my head.
In `SelfAlwaysShown` i would expect just myself.
What should i expect in a bin that includes participants that match: `SelfNotAlwaysShown`

src/state/CallViewModel.ts
```diff
@@ -405,10 +410,21 @@ export class CallViewModel extends ViewModel {
     switchMap((ms) => {
```
toger5: what does ms stand for?
it should be an element of the userMedia[] observable?

toger5 (approved):
This looks good. Thanks.
Can you help me understand the reason local an remote user are different classes instead of a property on the user tile?

src/state/CallViewModel.ts
toger5: What do you think about calling the bin enum sth like BinType since it is not the bin itself?

robintown: I would say that a `SortingBin` is the bin itself, on a conceptual level. In my mind, calling it `BinType` would imply that you can instantiate multiple distinct `Bin`s for any `BinType`, but a `Bin` doesn't really need to encode any more information than `BinType` does, so… a `SortingBin` already seems like a unique reference to the area that we see on screen.

Does that make sense at all? What is a 'bin', to you?

robintown:
The reason I've split up local and remote user media into different types is that they have distinct sets of properties (local media can be always shown/never shown, remote media has a volume, etc.), and I wanted to enforce that the views that consume them would be fully aware of their differences in functionality. It's the same reason why screen share media has its own type.

robintown:
Since you've approved this I'm going to merge it and continue working through the PR stack - I'm happy to iterate later on the `SortingBin` topic if you do have concerns there!

PR #2381, opened 2024-05-16
Replace react-rxjs with observable-hooks

robintown:
react-rxjs is the library we've been using to connect our React components to view models and consume observables. However, after spending some time with react-rxjs, I feel that it's a very heavy-handed solution. It requires us to sprinkle `<Subscribe />` and `<RemoveSubscribe />` components all throughout the code, and makes React go through an extra render cycle whenever we mount a component that binds to a view model. What I really want is a lightweight React hook that just gets the current value out of a plain observable, without any extra setup. Luckily the observable-hooks library with its `useObservableEagerState` hook seems to do just that—and it's more actively maintained, too!

Based on https://github.com/element-hq/element-call/pull/2380


toger5 (approved):
This is somewhat hard to review because the changes are small and the diff editor moves things all over the place.

The things I have manged to find the counter parts look like straight forward refactoring.
In general this makes things easier.
Cool you found this solution! Thanks!

Issue #2219, opened 2024-02-23
Spotlight layout is still jumpy and erratic


robintown:
When people are talking and they're not a screensharer, they can end up glitching up and down rapidly in the spotlight right-hand column or swapping out of the main spotlight area really quickly

dreamflasher:
Still unusable currently :(

toger5:
There is a concrete approach how this issue is tackled:
 - The root cause for this is known and understood.
 - A overall refactor of the spotlight view is in the making. It is designed so there shouldn't be any erratic layout changes
 - It is not worthit to spend time to fix the old implementation (doing so would postpone the transition to the new implementation)
 
I hope this summary gives an overview what is going on that makes sense to anyone beeing annoyed by the jumpy spotlight view.

robintown:
https://github.com/element-hq/element-call/pull/2382

Issue #1215, opened 2023-07-10
Animation issue in spotlight view


toger5:
### Steps to reproduce

If the reordering algorithm reorders the tiles on the right the spotlight tile gets teleported downwards by a little bit and afterwards animates back up to the top. Therefore resulting in occasional glitches.

### Will you send logs?

Yes

toger5:
This code will use the large grid layout for the base logic. This will most likely fix animation issues with the spotlight view.
pls get in contact with @robintown before working on this.

robintown:
https://github.com/element-hq/element-call/pull/2382

PR #2382, opened 2024-05-17
New spotlight layout

robintown:
Based on https://github.com/element-hq/element-call/pull/2381

Closes https://github.com/element-hq/element-call/issues/2219
Closes https://github.com/element-hq/element-call/issues/1215

### DIFF
--- a/src/grid/Grid.tsx
+++ b/src/grid/Grid.tsx
@@ -42,6 +42,7 @@ import { useMergedRefs } from "../useMergedRefs";
 import { TileWrapper } from "./TileWrapper";
 import { usePrefersReducedMotion } from "../usePrefersReducedMotion";
 import { TileSpringUpdate } from "./LegacyGrid";
+import { useInitial } from "../useInitial";
 
 interface Rect {
   x: number;
@@ -50,10 +51,13 @@ interface Rect {
   height: number;
 }
 
-interface Tile<Model> extends Rect {
+interface Tile<Model> {
   id: string;
   model: Model;
+  onDrag: DragCallback | undefined;
 }
+
+type PlacedTile<Model> = Tile<Model> & Rect;
 
 interface TileSpring {
   opacity: number;
@@ -73,23 +77,13 @@ interface DragState {
   cursorY: number;
 }
 
-interface SlotProps extends ComponentProps<"div"> {
-  tile: string;
+interface SlotProps<Model> extends Omit<ComponentProps<"div">, "onDrag"> {
+  id: string;
+  model: Model;
+  onDrag?: DragCallback;
   style?: CSSProperties;
   className?: string;
 }
-
-/**
- * An invisible "slot" for a tile to go in.
- */
-export const Slot: FC<SlotProps> = ({ tile, style, className, ...props }) => (
-  <div
-    className={classNames(className, styles.slot)}
-    data-tile={tile}
-    style={style}
-    {...props}
-  />
-);
 
 interface Offset {
   x: number;
@@ -113,9 +107,13 @@ function offset(element: HTMLElement, relativeTo: Element): Offset {
   }
 }
 
-export interface LayoutProps<Model, R extends HTMLElement> {
+export interface LayoutProps<LayoutModel, TileModel, R extends HTMLElement> {
   ref: LegacyRef<R>;
-  model: Model;
+  model: LayoutModel;
+  /**
+   * Component creating an invisible "slot" for a tile to go in.
+   */
+  Slot: ComponentType<SlotProps<TileModel>>;
 }
 
 export interface TileProps<Model, R extends HTMLElement> {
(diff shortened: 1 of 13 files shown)


robintown:
Not yet quite ready to review yet. I think I'd like to add some more comments before I consider this ready.

src/grid/Grid.tsx
```diff
@@ -91,6 +91,28 @@ export const Slot: FC<SlotProps> = ({ tile, style, className, ...props }) => (
...
+    element.offsetParent === relativeTo
+  ) {
+    return { x: element.offsetLeft, y: element.offsetTop };
+  } else {
+    const o = offset(element.offsetParent, relativeTo);
+    o.x += element.offsetLeft;
+    o.y += element.offsetTop;
+    return o;
+  }
+}
```
toger5, line 108: Do you know if this generally performs better or worse?
`getBoundingClientRect()`
`body.scrollLeft/scrollRight`
If we only want the offset at a set time T we also don't need to use the `body.scrollXXXXXX`
so a.getBoundingClientRect().left - b.getBoundingClientRect().left (and the same for right) should be enough?

src/tile/SpotlightTile.tsx
```diff
@@ -47,12 +51,63 @@ import { useObservableRef } from "../state/useObservable";
...
+  targetWidth: number;
+  targetHeight: number;
+  video: TrackReferenceOrPlaceholder;
+  member: RoomMember | undefined;
+  unencryptedWarning: boolean;
+  nameTag: string;
+  displayName: string;
+}
+
+interface SpotlightUserMediaItemBaseProps extends SpotlightItemBaseProps {
```
toger5, line 66: Do we use `Base` as to indicate that this is never used anywhere else except as `extends .*Base.*`?
Otherwise I would just remove the `Base` part. If it is just used for extending I like `Base` as a naming schema

src/grid/GridLayout.tsx
```diff
@@ -41,135 +41,144 @@ export const makeGridLayout: CallLayout<GridLayoutModel> = ({
...
+      }
+      if (slotHeight < slotMinHeight) slotHeight = slotMinHeight;
+      // Impose a minimum and maximum aspect ratio on the slots
+      const slotAspectRatio = slotWidth / slotHeight;
+      if (slotAspectRatio > slotMaxAspectRatio)
+        slotWidth = slotHeight * slotMaxAspectRatio;
+      else if (slotAspectRatio < slotMinAspectRatio)
+        slotHeight = slotWidth / slotMinAspectRatio;
+      // TODO: We might now be hitting the minimum height or width limit again
+
```
toger5, line 150: what do you think about moving all this computation into a simple descriptive funtion and keep theis method short?
e.g. somthing like `computeSlotHeight(`...

src/state/CallViewModel.ts
```diff
@@ -376,9 +376,9 @@ export class CallViewModel extends ViewModel {
       shareReplay(1),
     );
 
-  private readonly hasScreenShares: Observable<boolean> =
+  private readonly hasRemoteScreenShares: Observable<boolean> =
     this.screenShares.pipe(
-      map((ms) => ms.length > 0),
+      map((ms) => ms.find((m) => !m.vm.local) !== undefined),
```
toger5: `ms.some()` ?
I dont know the exat type of ms but if its an `Array` this is what you want?

toger5 (approved):
This is a fairly quick review.
This looks good and I could not find something that is an obvious problem.

src/grid/GridLayout.tsx
robintown: I've done exactly that in https://github.com/element-hq/element-call/pull/2417 :D
For the sake of time I won't try to reorder the commit stack to move that change into this PR. Hope that's fine!

src/tile/SpotlightTile.tsx
robintown: The name is horribly verbose but… yeah, that is the idea. There's already another interface named `SpotlightUserMediaItemProps`, because to get type safety in React you sometimes have to create a bajillion components (since there's no such thing as calling hooks conditionally… :pensive:)

src/state/CallViewModel.ts
robintown: Oops, of course

src/grid/Grid.tsx
robintown: Checking the client rect in combination with the body's scroll offset is a nice idea. I think to have the same correctness guarantees as this function though, you would have to traverse the DOM and subtract the scroll offsets for each scroll container that is a child of the DOM, as well. So in the end you'd still need a recursive process similar to this one. Unless I'm missing something?

I would note that in practice, `relativeTo` will only be 1-3 levels above the `element`, and I do think we want to account for non-body scroll containers in case we make some changes to how the call layouts are layered.

toger5: 3 levels should be totally fine then. 
The idea was that the scroll offset is the same. for all elements (but that is most likely incorrect if we have multiple scroll containers)
Then the scroll offset could be ignored. (it would get substracted away)

src/grid/GridLayout.tsx
toger5: oh definitely not!
Nice to have this updated in the final version.

hanthor:
I think I just tested this out on cell.element.dev. It doesn't seems to be giving the spotlight position to unmuted, talking participants. This is what I did:


1. start the call from my laptop at call.element.dev
2. send the invite link to myself
3. join on 2 different android phones
4. mute on all devices
5. my computer's video is in the spotlight and has had it the whole time
6. unmute and start talking on my phone 
7. the spotlight does not change, but the other participant get highlighted

PR #2386, opened 2024-05-24
package.json add name field

kilimnik:
I noticed, that the `package.json` is missing the name field. It is usually required and breaks some tools if it's missing.

For example I was packaging this for nixpkgs https://github.com/NixOS/nixpkgs/pull/300966, and had to patch it.

It would be great, if this would be upstreamed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "name": "element-call",
   "version": "0.0.0",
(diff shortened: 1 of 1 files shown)


toger5 (approved):
Makes sense,
Thanks a lot.

Issue #1582, opened 2023-09-25
Add a way to hide the top/bottom banner


Twi1ightSparkle:
#### What would you like to do?

Add a way to hide the top and bottom banner like Jitsi does. Currently ~20% of vertical screen real estate is wasted. Also, currently the banners does not get much smaller when reducing the vertical size of the browser. Down to the point where no video is shown, just the top and bottom banners.

#### Why would you like to do it?

To see more video on the screen

#### How would you like to achieve it?

Auto hide the top and bottom banner after n seconds. Maybe have this being an optional feature in settings?

### Have you considered any alternatives?

n/a

### Additional context

![Screenshot 2023-09-25 at 10 48 31 AM](https://github.com/vector-im/element-call/assets/19155609/2e0cf207-2d48-4ec3-9ab7-69349f5039bc)

---

![Screenshot 2023-09-25 at 11 03 36 AM](https://github.com/vector-im/element-call/assets/19155609/a0544bae-9ea9-4dec-afc4-6ea6eb42ac4a)

toger5:
There is a way: `&hideHeader=` e.g. `https://call.element.io/room/#/someCallName?roomId=!SOMEROOMID&hideHeader=` should do the trick.
I guess however you want something configurable in the ui. A button to hide it?
Or is this sufficient. (you can also add this to the links you are sharing with others)

Twi1ightSparkle:
That only hides the top header, leaving the largest bottom one. And I might want to be able to see the call name, number of participants, or access the share button during a call. I just don't need to see it all the time

nadonomy:
From brief discussion with @americanrefugee @fkwp: This sounds like a good enhancement but would want to figure out how it works generally (e.g. as an overlay?).

Twi1ightSparkle:
I like the way Jitsi does this. Show the overlays when mousing over the call, then hide after ~4 seconds

A recording for convenience:

https://github.com/vector-im/element-call/assets/19155609/692498a0-39ce-4ce7-952e-f7d93356d02d

dreamflasher:
Came here to make a feature request for this, wanting to suggest how it's done in Jitsi. This is already all there, thanks everyone!
So yes, imho it would be best to auto-hide top and bottom toolbar when the mouse is not moving – in order to see all speakers larger.

weeman1337:
[...]

PR #2397, opened 2024-06-01
Remove excessive footer space

dreamflasher:
Improves on https://github.com/element-hq/element-call/issues/1582 until the new layout is implemented.

### DIFF
--- a/src/room/InCallView.module.css
+++ b/src/room/InCallView.module.css
@@ -54,7 +54,7 @@ limitations under the License.
   grid-template-areas: "logo buttons layout";
   align-items: center;
   gap: var(--cpd-space-3x);
-  padding-block: var(--cpd-space-2x);
+  padding-block: var(--cpd-space-4x);
   padding-inline: var(--inline-content-inset);
   background: linear-gradient(
     180deg,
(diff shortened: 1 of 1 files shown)


toger5:
Thanks! Can you add before/after screenshots?

dreamflasher:
Happy to!

Before at 1366x768 resolution:
before

After at 1366x768 resolution:
after

fkwp:
@americanrefugee ^^

americanrefugee:
@fkwp The screenshot shows only a few pixels between the bottom edge of the bottoms and the bottom of the page, which is not enough.

Can you add a 24 pixel margin between the button row and page bottom?

americanrefugee:
Actually... I forgot about the tool tips on hover. So, we have 2 options:

1. Leave just enough space to show the tooltips below (as currently implemented)
2. Make tooltips show above the buttons, and 24 pixel margin between buttons and page edge

I prefer how it's currently implemented.

robintown:
Discussed this with the team, and we're going to go with Aaron's preference of leaving just enough space to show the tooltips below (about 40px), which strikes a nice compromise between space utilization and visual balance.

robintown (approved):

PR #2416, opened 2024-06-07
Show speaker in the spotlight in large grids

robintown:
Based on https://github.com/element-hq/element-call/pull/2382

### DIFF
--- a/src/state/CallViewModel.ts
+++ b/src/state/CallViewModel.ts
@@ -74,6 +74,13 @@ import { ObservableScope } from "./ObservableScope";
 // list again
 const POST_FOCUS_PARTICIPANT_UPDATE_DELAY_MS = 3000;
 
+// This is the number of participants that we think constitutes a "large" grid.
+// The hypothesis is that, after this many participants there's enough cognitive
+// load that it makes sense to show the speaker in an easy-to-locate spotlight
+// tile. We might change this to a scroll-based condition or do something else
+// entirely with the spotlight tile, if we workshop this further.
+const largeGridThreshold = 20;
+
 // Represents something that should get a tile on the layout,
 // ie. a user's video feed or a screen share feed.
 // TODO: This exposes too much information to the view layer, let's keep this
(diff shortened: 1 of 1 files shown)


src/state/CallViewModel.ts
```diff
@@ -504,7 +504,9 @@ export class CallViewModel extends ViewModel {
                     (grid, spotlight, screenShares): Layout => ({
                       type: "grid",
                       spotlight:
-                        screenShares.length > 0 ? spotlight : undefined,
+                        screenShares.length > 0 || grid.length > 20
```
toger5: the 20 seems somewhat random. I think this should either become a config or needs a comment what made us decide to hard code this to 20.

toger5: maybe also a "needs to show scrollbars" is the better condition?

toger5 (changes requested):
I think it would be best to use a non hardcoded condition here.

If we use the 20 limit as a stop gap is also okay. But then I think it would be nice to have comments explaining and justifying it (maybe even mentioning that this could be done with a "do we need scrollbars" condition

robintown:
I think I do have an argument against basing the decision on whether there is a scroll bar:

- Speakers are always promoted to the start of the grid, so it should be exceedingly rare that a speaker stays outside the visible area.
- So then, the purpose of this PR is not so much to bring obscured speakers into view, but rather to reduce the cognitive load of finding the speaker in a large grid of people.
- I think it's reasonable to assume that cognitive load is correlated to the number of participants in a call, but not necessarily to whether there is a scroll bar. (For example, if I have a large window showing 30 participants without a scroll bar, and then I make the window smaller so that it has a scroll bar and only shows the most important 15 participants, has the cognitive load really increased? I think it's gone down, if anything.)
- So, I think using a heuristic based on number of participants is appropriate.

However I totally get the desire to not use an arbitrary number. Basing the decision on whether there is a scroll bar would be pretty non-trivial, and this part of the designs is still subject to change (we want to workshop it further), so: I suggest that the scroll bar method is not worth the time investment at this stage, regardless of whether my argument above makes sense.

toger5:
> I suggest that the scroll bar method is not worth the time investment at this stage, regardless of whether my argument above makes sense.

I get behind this!
I was mostly thinking that the reason for this is to always have the speaker visible. With more ppl (or a scroll container) there is the chance (independent of the user being at the bottom or the top) to not see a user so it makes sense to at least make sure the speaker is visible.

But I like going with this first and making a decision that requires a lot of dev effort based on more usage experience!

toger5 (approved):

PR #2417, opened 2024-06-07
New one-on-one layout

### LINKED ISSUE #1244: Repace all remaining usages of VideoGrid with NewVideoGrid
Because the latter is more easily extended with new layouts, and will make it easier to adapt the call UI to changing design requirements.

It's also better at animation due to using `useTransition` rather than `useSprings`, and so doing this would resolve the following issues:

- https://github.com/vector-im/element-call/issues/1222
- https://github.com/vector-im/element-call/issues/1215

robintown:
Based on https://github.com/element-hq/element-call/pull/2416

Closes https://github.com/element-hq/element-call/issues/1244


robintown:
TODO: If you switch to spotlight while connecting, the app crashes

src/grid/OneOnOneLayout.tsx
```diff
@@ -0,0 +1,132 @@
...
+        >
+          <Slot
+            className={classNames(styles.slot, styles.local)}
+            id={localTileModel.vm.id}
+            model={localTileModel}
+            onDrag={onDragLocalTile}
+            data-block-alignment={pipAlignmentValue.block}
+            data-inline-alignment={pipAlignmentValue.inline}
+          />
+        </Slot>
```
toger5, line 128: Interesting. The slot for the local media feed is a child of the remote participant slot?
That makes actually.
So the local tile is always inside the parent/remote tile. Will this also hold true if we have a portrait remote media feed on a landscape screen. in that case i can imagine we would like the local tile to be outside the remote tile?
(currently we use the "show grey borders" (fit to frame) approach but in theory it would be nice to have the tile be the correct aspect ratio already)

toger5 (approved):
This looks good.
I might have missed it but as far as i can tell this does never render tiles in portrait mode.
I think in 1:1 that is the one situation where this is totally doable.
I think right now it behaves the other way around which is already great (landscape will be rendered as landscape on a portrait screen).
But portrait media being rendered as portrait on a landscape screen would also be nice. (maybe with a little bit of a cutoff to not make it too small 9:16 becomeing 2:3 or similar)

robintown:
Thanks for the suggestions, I'm pretty happy with how the 2:3 aspect ratio looks on mobile now:



Detecting when the media we're receiving should be landscape is something I don't know how to solve, currently :/ I think that'll have to be outside the scope of this PR. So the best I can do for now is assume that media should be portrait whenever the window is sufficiently narrow, which is the same as how the old one-on-one layout behaves.

src/grid/OneOnOneLayout.tsx
robintown: Yeah, making it be a child was the easiest way to get the right positioning. As I've commented below I don't know how to account for landscape media, currently (but neither does the old one-on-one layout).

PR #2435, opened 2024-06-19
Use correct DOM event name for visibility changes in useWakeLock

hughns:
The event should be `visibilitychange` not `visiblitychange` as per https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event.

I haven't tested this, but I assume that it means the useWakeLock doesn't work as intended today.


toger5 (approved):
Nice find.

Changes look super straight forward

PR #2453, opened 2024-06-21
Skip error screen for the issue, that the homeserver does not support the room summary endpoint.

toger5:
Add try inner try block to the room summary fetching and only throw after fetching and a "blind join" fails.
(blind join: call room.join without knowing if the room is public)

### DIFF
--- a/src/room/useLoadGroupCall.ts
+++ b/src/room/useLoadGroupCall.ts
@@ -242,17 +242,21 @@ export const useLoadGroupCall = (
             );
           }
           if (
-            roomSummary?.join_rule === JoinRule.Public ||
-            roomSummary === undefined
+            roomSummary === undefined ||
+            roomSummary.join_rule === JoinRule.Public
           ) {
             room = await client.joinRoom(roomId, {
               viaServers,
             });
           } else if (roomSummary.join_rule === JoinRule.Knock) {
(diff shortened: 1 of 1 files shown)


src/room/useLoadGroupCall.ts
```diff
@@ -225,29 +225,52 @@ export const useLoadGroupCall = (
...
+          let roomSummary: RoomSummary | undefined = undefined;
+          try {
+            roomSummary = await client.getRoomSummary(roomId, viaServers);
+          } catch (error) {
+            // If the room summary endpoint is not supported we let it be undefined and treat this case like
+            // `JoinRule.Public`.
+            // This is how the logic was done before: "we expect any room id passed to EC
+            // to be for a public call" Which is definitely not ideal but worth a try if fetching
+            // the summary crashes.
+            logger.error(
```
robintown:
```suggestion
            logger.warn(
```

A warning feels a little more appropriate since the application recovers gracefully and continues

robintown (approved):

src/room/useLoadGroupCall.ts
toger5: I wanted this to really standup so that there is a chance for this to be caught by users. In theory it should be considered as incompatible if a HS does not support the room summary endpoint.

But probably a warning is still enough.

PR #2457, opened 2024-06-27
Make widgets request the room creation event

AndrewFerr:
This allows the widget to check the room version, so it can know about version-specific auth rules (namely MSC3779).


AndrewFerr:
Related change: https://github.com/matrix-org/matrix-react-sdk/pull/12710
This allows granting the capability without interaction from the user.

toger5 (approved):
Thanks!

PR #2459, opened 2024-06-28
Let Element Call widget set session memberships

AndrewFerr:
Make Element Call widgets request permission to set device-specific session membership state events.


AndrewFerr:
Related change: https://github.com/matrix-org/matrix-react-sdk/pull/12713
This allows granting the capability without interaction from the user.

toger5 (approved):
Nice to see a TODO removed

PR #2460, opened 2024-06-28
Add config to send session membership state events

AndrewFerr:
If not set, legacy call membership state events are sent instead. Even if set, legacy events are sent in rooms with active legacy calls.

### DIFF
--- a/src/config/ConfigOptions.ts
+++ b/src/config/ConfigOptions.ts
@@ -74,8 +74,9 @@ export interface ConfigOptions {
     feature_group_calls_without_video_and_audio?: boolean;
     /**
      * Send device-specific call session membership state events
-     * instead of the legacy user-specific call membership state events,
-     * but not in rooms that contain active calls with legacy state events.
+     * instead of the legacy user-specific call membership state events.
+     * This setting has not effect when the user joins an active call with legacy state events.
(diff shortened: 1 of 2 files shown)


src/config/ConfigOptions.ts
```diff
@@ -65,11 +65,19 @@ export interface ConfigOptions {
...
   features?: {
-    feature_group_calls_without_video_and_audio: boolean;
+    /**
+     * Allow to join group calls without audio and video.
+     */
+    feature_group_calls_without_video_and_audio?: boolean;
+    /**
+     * Send device-specific call session membership state events
+     * instead of the legacy user-specific call membership state events,
+     * but not in rooms that contain active calls with legacy state events.
```
toger5:
```suggestion
     * Send device-specific call session membership state events
     * instead of the legacy user-specific call membership state events.
     * This setting has not effect when the user joins an active call with legacy state events.
     * For compatibility Element Call will always join running legacy calls with legacy state events.```

toger5 (approved):
I added an optional proposal. I think stating that this setting "has no effect" in certain conditions communicates better that this is a "preferred" like setting than the "but" sentence.

toger5 (approved):

PR #2463, opened 2024-07-03
Implement most of the remaining layout changes

robintown:
Includes the mobile UX optimizations and the tweaks we've made to cut down on wasted space, but does not yet include the change to embed the spotlight tile within the grid.

Based on https://github.com/element-hq/element-call/pull/2417


toger5:
This also should get a X-Blocked right? It also depends on other pr's

src/grid/SpotlightExpandedLayout.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { forwardRef, useCallback, useMemo } from "react";
+import { useObservableEagerState } from "observable-hooks";
+
+import { SpotlightExpandedLayout as SpotlightExpandedLayoutModel } from "../state/CallViewModel";
+import { CallLayout, GridTileModel, SpotlightTileModel } from "./CallLayout";
+import { DragCallback } from "./Grid";
+import styles from "./SpotlightExpandedLayout.module.css";
+import { useReactiveState } from "../useReactiveState";
+
+export const makeSpotlightExpandedLayout: CallLayout<
```
toger5, line 30: This could get a comment what exactly it is.

src/room/InCallView.tsx
```diff
@@ -369,52 +333,74 @@ export const InCallView: FC<InCallViewProps> = ({
...
+      minBounds: gridBoundsObservable,
+      spotlightAlignment,
+      pipAlignment,
+    };
+    return {
+      grid: makeGridLayout(inputs),
+      "spotlight landscape": makeSpotlightLandscapeLayout(inputs),
+      "spotlight portrait": makeSpotlightPortraitLayout(inputs),
+      "spotlight expanded": makeSpotlightExpandedLayout(inputs),
+      "one-on-one": makeOneOnOneLayout(inputs),
```
toger5: I think it would be nice to use a consistent style for all of those layouts:
```
      "spotlight-landscape": makeSpotlightLandscapeLayout(inputs),
      "spotlight-portrait": makeSpotlightPortraitLayout(inputs),
      "spotlight-expanded": makeSpotlightExpandedLayout(inputs),
      "one-on-one": makeOneOnOneLayout(inputs),
```

toger5 (approved):
Thanks!
I approve this already even though i think the comments on the less obvious layout names would be super nice/helpful!
The names (with `-` and without) don't have to change if there is a reason for it.

PR #2466, opened 2024-07-05
Clean up our tests in preparation for the testing sprint

robintown:
See the commit messages for the details!


AndrewFerr:
Regarding [f79ca02](https://github.com/element-hq/element-call/pull/2466/commits/f79ca02481e3cf958a97198a5cf1a9674239f87d), could we instead just change the linters' search paths to find test files?  Personally I prefer having test files in their own directory, as it makes it easier to separate them from source files & many of our projects put tests in a separate directory (aside from the Rust projects where tests are within source files).

robintown:
Hmm, my experience is that separation from the source files, at least for unit tests, just adds friction. When re-organizing files or directories, it becomes really easy to find the test files and make sure you're moving them at the same time as the code. I believe this is one of the reasons why you see Rust [and Vitest](https://vitest.dev/guide/in-source.html) even moving tests _into_ the source file nowadays.

@toger5 Do you have any preference here?

toger5:
I like it when the tests are on the same level where they test. I like if a function that does not require any context mocking but just validates a type for instance does not leave the scope of the file. So having the test in the file is nice here.
But as soon as there is mocking setup that might be useful for multiple files a dedicated test file makes sense. Then all mocking/helper code is at one spot and does not pollute the actual source.

This comment is not too helpful to find an immediate solution I think. It depends on what kind of tests we are mostly writing.

robintown:
Concluded in stand-up to keep the test files inside src, as I've done here. This is ready for review then

robintown:
Turning on a coverage gate for Codecov was pretty easy once I managed to find the right page of the docs: https://docs.codecov.com/docs/commit-status#patch-status

I've included that coverage gate configuration in this PR.

robintown:
Welp, I've rebased the PR and now the tests are throwing spurious errors in CI that I can't reproduce locally. You can see a bunch of commits above from me trying to fix it. Unfortunately, I'll need to leave that for someone else to figure out, as my week is about over.

(One possible idea: could it be that either the CI or my laptop has a bad Yarn cache? That's the simplest way I can imagine the behavior being different in CI.)

toger5:
> Welp, I've rebased the PR and now the tests are throwing spurious errors in CI that I can't reproduce locally. You can see a bunch of commits above from me trying to fix it. Unfortunately, I'll need to leave that for someone else to figure out, as my week is about over.

I am looking into this. Initially I was able to reproduce the issues locally.  After some time I also had them running without changing sth significantly. One finding is that having multiple async tests results in running the tests failing and having only one async seems to work. (but I am not sure if this is just random with a bias after all)

After some iterations I concluded the following:
 - The CI is probably not doing multithreaded tests and running `yarn vitest --no-file-parallelism` results in the same errors locally and on the remote.
 - All tests using the `userEvent` module cause errors in the test suit that make it fail (at least how we do them now)
 - Since my local test runner was suffering from multi thread flakyness I played around with component un-mounting. The result is the obvious, that a `const {unmount} = render(...)` behaves equivalent to a `afterEach(cleanup)`

toger5 (approved):
This is missing two tests now but has all the minor things fixed to get through ci.

PR #2467, opened 2024-07-05
Switch to Rust crypto

robintown:
We're not really using crypto at the moment, but with it being this easy, we should still make the switch!

Based on https://github.com/element-hq/element-call/pull/2466


toger5:
As you mention in the description, this is: Based on https://github.com/element-hq/element-call/pull/2466.
Should this PR be the standalone change?

AndrewFerr (commented):
This really is easy :smiley: 

If you want, you could rebase this onto `livekit` to make it independent of your other PRs, and it could be merged without conflicts much sooner.

robintown:
Yeah, I guess I could un-stack this one, but it's like… I know that it would create conflicts for some other things in the stack, requiring me to rebase N other changes, and it's not a very impactful or time-sensitive change, so I'd rather just keep it on the stack.

toger5:
Is this actually still blocked?

toger5:
Also blocked on missing rust encrypted to device sending https://github.com/matrix-org/matrix-js-sdk/issues/3304

robintown:
We're not using that API currently (though it being missing would slow our adoption of a new individual sender keys encryption scheme)

toger5:
I am wondering how we currently distribute megolm keys without toDevice? Is it just custom toDevice messages that are missing in the wasm bindings?

fkwp:
that's abstracted away by the js-sdk. the rust js-sdk is internally using toDevice msg but the API is not exposed to the js layer.

hughns:
As this is now on the critical path for looking at sender key distribution I have made https://github.com/element-hq/element-call/pull/2571 which has just the Rust crypto changes in it to make review easier.

PR #2473, opened 2024-07-09
Remove ResizeObserver polyfill

robintown:
All major browsers support it out of the box, nowadays.


fkwp (approved):
lgtm // needs conflict resolution

PR #2474, opened 2024-07-10
Don't update mute when reaching the user count threshold.

### LINKED ISSUE #2464: Magically mute without user input (automatically) after starting the call
### Steps to reproduce

For some users the call mutes automatically approx 5s into the call without user input.

toger5:
Fixes: https://github.com/element-hq/element-call/issues/2464

The idea is to mute whenever the participant count reaches 8 or more ppl. since there is a high chance for multiple users joining at the same time we update the mute state whenever we reach over 8ppl. So that a scheduled meeting still starts with everyone muted even though the meeting looked empty when the users were still in the lobby.

This changes it to only check this condition once at the beginningof the lobby and where the call just loaded (the client knows the user account)

### DIFF
--- a/src/room/GroupCallView.tsx
+++ b/src/room/GroupCallView.tsx
@@ -80,17 +80,14 @@ export const GroupCallView: FC<Props> = ({
   const memberships = useMatrixRTCSessionMemberships(rtcSession);
   const isJoined = useMatrixRTCSessionJoinState(rtcSession);
 
-  // The mute state reactively gets updated once the participant count reaches the threshold.
-  // The user then still is able to unmute again.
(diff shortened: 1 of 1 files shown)


robintown (approved):
This looks fine though I wonder if it would be more idiomatic as:

```typescript
useEffect(() => {
  if (memberships.length >= MUTE_PARTICIPANT_COUNT)
    muteStates.audio.setEnabled?.(false)
}, [])
```

without the `useCallback`

toger5:
> This looks fine though I wonder if it would be more idiomatic as:

I added it to allow for the upgrade path to `useEffectEvent`. But for now the simpler version is reasonable.

PR #2475, opened 2024-07-10
Enable lint rules for Promise handling to discourage misuse of them

hughns:
Whilst working on the codebase I noticed that promises are not always being handled/treated correctly.

The purpose of enabling these lint rules to force a conscious decision around how Promises are handled including the errors.

The `@typescript-eslint/promise-function-async` rule is a stylistic one, so less important.

I'm anticipating this my prompt some discussion rather than just being merged as is 😇.

### DIFF
--- a/src/analytics/PosthogSpanProcessor.ts
+++ b/src/analytics/PosthogSpanProcessor.ts
@@ -52,7 +52,9 @@ export class PosthogSpanProcessor implements SpanProcessor {
           this.onSummaryReportStart(span);
           return;
       }
-    });
+    }).catch((e) => {
+      // noop
+      });
   }
 
   public onEnd(span: ReadableSpan): void {
@@ -157,7 +159,8 @@ export class PosthogSpanProcessor implements SpanProcessor {
   /**
    * Shutdown the processor.
    */
-  public shutdown(): Promise<void> {
-    return Promise.resolve();
+  // eslint-disable-next-line @typescript-eslint/require-await
+  public async shutdown(): Promise<void> {
+    return;
   }
 }
--- a/src/analytics/PosthogSpanProcessor.ts
+++ b/src/analytics/PosthogSpanProcessor.ts
@@ -43,17 +43,19 @@ export class PosthogSpanProcessor implements SpanProcessor {
 
   public onStart(span: Span): void {
     // Hack: Yield to allow attributes to be set before processing
-    Promise.resolve().then(() => {
-      switch (span.name) {
-        case "matrix.groupCallMembership":
-          this.onGroupCallMembershipStart(span);
-          return;
-        case "matrix.groupCallMembership.summaryReport":
-          this.onSummaryReportStart(span);
-          return;
-      }
-    }).catch((e) => {
-      // noop
+    Promise.resolve()
+      .then(() => {
+        switch (span.name) {
+          case "matrix.groupCallMembership":
+            this.onGroupCallMembershipStart(span);
+            return;
+          case "matrix.groupCallMembership.summaryReport":
+            this.onSummaryReportStart(span);
+            return;
+        }
+      })
+      .catch((e) => {
+        // noop
       });
   }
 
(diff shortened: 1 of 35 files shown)


src/room/GroupCallView.tsx
```diff
@@ -192,21 +184,31 @@ export const GroupCallView: FC<Props> = ({
...
     if (widget && preload && skipLobby) {
       // In preload mode without lobby we wait for a join action before entering
-      const onJoin = async (
-        ev: CustomEvent<IWidgetApiRequest>,
-      ): Promise<void> => {
-        defaultDeviceSetup(ev.detail.data as unknown as JoinCallData);
-        await enterRTCSession(rtcSession, perParticipantE2EE);
-        await widget!.api.transport.reply(ev.detail, {});
+      const onJoin = (ev: CustomEvent<IWidgetApiRequest>): void => {
+        defaultDeviceSetup(ev.detail.data as unknown as JoinCallData)
```
hughns, line 188: Was it intended that `defaultDeviceSetup()` actually be run in parallel? I've assumed it should be serial.

src/room/GroupCallView.tsx
```diff
@@ -192,21 +184,31 @@ export const GroupCallView: FC<Props> = ({
...
       };
       widget.lazyActions.on(ElementWidgetActions.JoinCall, onJoin);
       return (): void => {
         widget!.lazyActions.off(ElementWidgetActions.JoinCall, onJoin);
       };
     } else if (widget && !preload && skipLobby) {
       // No lobby and no preload: we enter the rtc session right away
-      defaultDeviceSetup({ audioInput: null, videoInput: null });
-      enterRTCSession(rtcSession, perParticipantE2EE);
+      defaultDeviceSetup({ audioInput: null, videoInput: null })
```
hughns, line 204: Same as above.

src/room/GroupCallView.tsx
toger5: This might be the cause for the duplicated device requests we are getting!

toger5:
Note: We brought this up in the standup today.
 - The general acceptance was positive.
 - Concerned that this might result in lots of lint disable comments. (looking through the changes we saw that for now its definitely manageable)

src/analytics/PosthogSpanProcessor.ts
```diff
@@ -157,7 +161,8 @@ export class PosthogSpanProcessor implements SpanProcessor {
   /**
    * Shutdown the processor.
    */
-  public shutdown(): Promise<void> {
-    return Promise.resolve();
+  // eslint-disable-next-line @typescript-eslint/require-await
+  public async shutdown(): Promise<void> {
+    return;
```
robintown, line 166: I would prefer to remove the `eslint-disable` comment and return `Promise.resolve()` in a case like this. It seems more in the spirit of the require-await lint rule.

src/config/Config.ts
```diff
@@ -30,17 +30,21 @@ export class Config {
...
+    Config.internalInstance.initPromise = new Promise<void>(
+      (resolve, reject) => {
+        downloadConfig("../config.json")
+          .then((config) => {
+            Config.internalInstance.config = { ...DEFAULT_CONFIG, ...config };
+            resolve();
+          })
+          .catch(reject);
+      },
+    );
```
robintown, line 47: Could this be simplified to `downloadConfig("../config.json").then((config) => { ... })`?

src/rtcSessionHelpers.ts
```diff
@@ -36,6 +36,7 @@ export function makeActiveFocus(): LivekitFocusActive {
   };
 }
 
+// eslint-disable-next-line @typescript-eslint/require-await
```
robintown, line 39: I suggest `Promise.resolve(...)`-ing the return value here as well. That way, the thing that makes eslint happy is closer to the comment at the bottom of the function about how we want to await something.

robintown (approved):
Thanks, the improved safety and error logging will be very welcome. Just some nits:

toger5:
This has not moved to: https://github.com/element-hq/element-call/pull/2475

toger5:
Also robins review has been applied in the squashed +rebased PR version.

toger5:
Closing in favor of: #2607

PR #2482, opened 2024-07-17
Add DeviceMute widget action `io.element.device_mute`.

toger5:
This allows to send mute requests ("toWidget") and get the current mute state as a response.
And it will update the client about each change of mute states.

### DIFF
--- a/src/room/MuteStates.ts
+++ b/src/room/MuteStates.ts
@@ -14,7 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { Dispatch, SetStateAction, useEffect, useMemo } from "react";
+import {
+  Dispatch,
+  SetStateAction,
+  useCallback,
+  useEffect,
+  useMemo,
+} from "react";
 import { IWidgetApiRequest } from "matrix-widget-api";
 
 import { MediaDevice, useMediaDevices } from "../livekit/MediaDevicesContext";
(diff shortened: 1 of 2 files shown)


src/room/MuteStates.ts
```diff
@@ -74,5 +77,53 @@ export function useMuteStates(): MuteStates {
...
+      video_enabled: video.enabled,
+    });
+  }, [audio, video]);
+
+  useEffect(() => {
+    if (widget) {
+      const onMuteStateChangeRequest = (
+        ev: CustomEvent<IWidgetApiRequest>,
+      ): void => {
+        logger.log("aaaaa");
```
jmartinesp: aaaaa

toger5: Ahhhh!

src/widget.ts
```diff
@@ -46,6 +46,19 @@ export enum ElementWidgetActions {
   // host -> Element Call telling EC to stop screen sharing, or that
   // the user cancelled when selecting a source after a ScreenshareRequest
   ScreenshareStop = "io.element.screenshare_stop",
+  // This can be sent as form or to widget
```
stefanceriu:
```suggestion
  // This can be sent as from or to widget
```

stefanceriu: As EC automagically sets the muted state sometimes (based on room size? 🤔) it's important to be able to read the current state (without affecting it) when the CallKit session is established so that we keep the buttons in sync. How can we do that?

toger5:
```
  // (it is possible to sent an empty configuration
  // -> this will allow the client to only get the current state)
```
So you would send a `io.element.device_mute` action with the content `{}` then the reply would be:
```
response: {
 audio_enabled: true, // for example
 video_enabled: false // for example
}
```
with the format:
```
  // The data of the widget action request and the response are:
  // {
  //   audio_enabled?: boolean,
  //   video_enabled?: boolean
  // }
```

How could this be phrased to make more sense?

stefanceriu: Oh I see, perhaps `It is possible to sent an empty configuration to retrieve the current values or just one of the fields to update that particular value`

And I just noticed `An undefined field means that EC will not keep the mute state as is.`, I don't think that `not` belongs there, does it?

src/room/MuteStates.ts
```diff
@@ -74,5 +76,52 @@ export function useMuteStates(): MuteStates {
...
+      );
+
+      return (): void => {
+        widget!.lazyActions.off(
+          ElementWidgetActions.DeviceMute,
+          onMuteStateChangeRequest,
+        );
+      };
+    }
+  }, [audio, video]);
```
dbkr: I might need an explanation on what this is doing: whenever the audio or video mute state changes, we add a new listener for the device mute widget action? The listener presumably also fires (once?) to set the mute states and reply?

dbkr (commented):
In general, comments would be helpful: I say this as someone who's knowledge of this codebase is now a little rusty :)

src/room/MuteStates.ts
toger5: To my understanding the audio and video elements are not changing instead they mutate. So whenever they are set we connect the listener and when they are unset we remove it.

dbkr: I'm not sure it's that simple. Why not just use a callback and then the problem goes away? Also it will make the code less nested.

toger5: I updated this. So it is no creating the useCallback and recomputes the on/off when the callback changes.

dbkr (approved):
Looks good as far as I can see, comments make it much clearer, thanks!

fkwp (approved):
lgtm

PR #2483, opened 2024-07-17
Update sample config with livekit (rebase on livekit)

toger5:
replaces: https://github.com/element-hq/element-call/pull/1212

### DIFF
--- a/config/config.sample.json
+++ b/config/config.sample.json
@@ -5,5 +5,8 @@
       "server_name": "call.ems.host"
     }
   },
+  "livekit": {
+    "livekit_service_url": "http://localhost:7881"
+  },
   "eula": "https://static.element.io/legal/online-EULA.pdf"
 }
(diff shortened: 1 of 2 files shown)


fkwp:
do we need this key? going forward the information will be provided by .well-known

toger5:
I think for now its good since its still in the readme. Nowadays the best would be to properly updated the readme and explain everything.

toger5:
Just checked this is already explained in the readme: https://github.com/element-hq/element-call/blob/d0e848d90d8c29c868fa4bd0b0478554e41878a4/README.md#L127-L134

so this really just brings the example up to speed

config/config.sample.json
```diff
@@ -5,5 +5,8 @@
       "server_name": "call.ems.host"
     }
   },
+  "livekit": {
+    "livekit_service_url": "http://localhost:7881"
```
AndrewFerr, line 9: How about we update the port in the README example to match this?

AndrewFerr (approved):

PR #2486, opened 2024-07-18
Delete the unused full screen code

robintown:
We no longer allow individual tiles to be put in full screen, because we're seeing what it's like to just stretch the spotlight tile edge-to-edge and keep the margins minimal.


toger5 (approved):
Thanks for cleaning up.
Seems save, since this linter is not yelling about unable to find imports or unused imports.

PR #2493, opened 2024-07-22
Show layout switcher in widget/embedded mode

toger5:
We used `hideHeader` to decide if we want to hide the layout switcher (because the header is hidden in widget mode and we did not want the switcher in widget mode since it initially was part of the RoomHeader)

TLDR:  `hideHeader` -> show layout switch at the bottom right is a historic condition we dont need (doesnt make sense) anymore.


fkwp (approved):
lgtm

PR #2494, opened 2024-07-22
Quick and dirty fix to spotlight reactivity

robintown:


toger5 (approved):

robintown:
This is not a proper fix because in theory, changes to the `model` of each tile should automatically propagate through the `Grid` component without the need for a change in the grid's generation. However, incrementing the generation more often than necessary is perfectly safe, so this will do for the moment.

PR #2500, opened 2024-07-24
Update js-sdk to check for MSC3779 in room version

AndrewFerr:
This will tell EC to prefix non-legacy state event keys with an underscore to avoid 403 errors.


robintown:
The type error is fixed on `livekit` FYI

AndrewFerr:
That renders this PR obsolete, then, as the js-sdk used in `livekit` is new enough to get the room version checks.  Thanks!

PR #2501, opened 2024-07-24
Make layout reactivity less brittle

robintown:
Follow-up to ea2d98179cae73b46a5eeb4dfc5a1dc089027238

This took a couple of iterations to find something that works without creating update loops, but I think that by automatically informing Grid whenever a layout component is re-rendered, we'll have a much easier time ensuring that our layouts are fully reactive.

### DIFF
--- a/src/grid/Grid.tsx
+++ b/src/grid/Grid.tsx
@@ -128,10 +128,10 @@ const LayoutContext = createContext<LayoutContext | null>(null);
  * Enables Grid to react to layout changes. You must call this in your Layout
  * component or else Grid will not be reactive.
  */
-export function useLayout(): void {
+export function useUpdateLayout(): void {
   const context = useContext(LayoutContext);
   if (context === null)
-    throw new Error("useLayout called outside of a Grid layout component");
+    throw new Error("useUpdateLayout called outside a Grid layout context");
 
   // On every render, tell Grid that the layout may have changed
   useEffect(() =>
@@ -240,7 +240,7 @@ export function Grid<
   const [gridRoot, gridRef2] = useState<HTMLElement | null>(null);
   const gridRef = useMergedRefs<HTMLElement>(gridRef1, gridRef2);
 
-  const [layoutRoot, layoutRef] = useState<HTMLElement | null>(null);
+  const [layoutRoot, setLayoutRoot] = useState<HTMLElement | null>(null);
   const [generation, setGeneration] = useState<number | null>(null);
   const tiles = useInitial(() => new Map<string, Tile<TileModel>>());
   const prefersReducedMotion = usePrefersReducedMotion();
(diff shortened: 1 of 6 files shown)


src/grid/Grid.tsx
```diff
@@ -209,7 +240,7 @@ export function Grid<
   const [gridRoot, gridRef2] = useState<HTMLElement | null>(null);
   const gridRef = useMergedRefs<HTMLElement>(gridRef1, gridRef2);
 
-  const [layoutRoot, setLayoutRoot] = useState<HTMLElement | null>(null);
+  const [layoutRoot, layoutRef] = useState<HTMLElement | null>(null);
```
toger5: This is either wrong or there is some trick how forwardRef works under the hood. does it just call the Ref with the content so we can also use a state ?

src/grid/Grid.tsx
```diff
@@ -113,6 +118,27 @@ function offset(element: HTMLElement, relativeTo: Element): Offset {
...
+  setGeneration: Dispatch<SetStateAction<number | null>>;
+}
+
+const LayoutContext = createContext<LayoutContext | null>(null);
+
+/**
+ * Enables Grid to react to layout changes. You must call this in your Layout
+ * component or else Grid will not be reactive.
+ */
+export function useLayout(): void {
```
toger5: why does this get a use name?
It seems like it behaves more like a schedule recompute method?

toger5 (commented):
This generally makes sense. I am a bit confused about some naming.
I think its best if you give me a little bit of guidance on my comments. I think this could make it much more efficient to review.

src/grid/Grid.tsx
robintown: Oh right, this is making use of React's [callback ref](https://react.dev/reference/react-dom/components/common#ref-callback) feature, which allows any callback of type `(value: R) => void` to act as a `Ref<R>`. Since `setLayoutRoot` has type `(value: HTMLElement | null) => void`, it can be used as a `Ref<HTMLElement | null>`.

Would you find this clearer if the name was still `setLayoutRoot` instead of `layoutRef`?

src/grid/Grid.tsx
robintown: It gets a `use` name because it is a custom React hook (it calls `useContext` and `useEffect` internally). Do you think something other than `useLayout` would better communicate what it does though?

src/grid/Grid.tsx
toger5: It would be nice if they have a naming convention for this. (sth like a setRef)
But knowing this is enough so the name makes sense.

src/grid/Grid.tsx
toger5: `useLayoutEffect` maybe?

toger5 (approved):
Approving this since I am afraid that the names might be because of a lack of familiarity with the use-cases.
Thanks for your clarifications. If you feel like you have an idea how to make it more clear go for it otherwise I think it really should not be blocked by it and I can definite get accustomed to how it is now.

src/grid/Grid.tsx
robintown: That's the name of a built-in React hook, so it'd probably be confusing, but I went with `useUpdateLayout` which is hopefully an improvement

toger5:
Thanks Robin for taking the time to rename the vatiables.

PR #2502, opened 2024-07-24
Improve accessibility of keyboard shortcuts

robintown:
Ensure that they don't interfere with say, using spacebar to press a button, and also ensure that they won't do surprising things like scroll the page at the same time.


toger5 (approved):
You also added tests 🎉

PR #2505, opened 2024-07-25
Add future related widget capabilities

toger5:


src/widget.ts
```diff
@@ -77,6 +77,8 @@ export const widget = ((): WidgetHelpers | null => {
       logger.info("Widget API is available");
       const api = new WidgetApi(widgetId, parentOrigin);
       api.requestCapability(MatrixCapabilities.AlwaysOnScreen);
+      api.requestCapability("org.matrix.msc4157.send.delayed_event");
+      api.requestCapability("org.matrix.msc4157.update.delayed_event");
```
robintown: I would expect these capabilities to be requested by the RoomWidgetClient (like most of our capabilities related to MatrixClient usage are). Also, referring to the capabilities as `MatrixCapabilities.SendDelayedEvent` instead of using a string literal would allow us to stabilize them without that being a breaking change.

toger5: I was considering this but went for this for simplicity.
Was afraid about the overhead and complexity this might result in for andrews PR.
Maybe it makes most sense that I dont touch it and andrew adds those changes to the already existing future event related PR's

toger5:
@AndrewFerr I think your pr's contain a better solution for this.
Feel free to close if you can confirm this.

robintown (approved):

AndrewFerr (changes requested):
The lock file needs to be regenerated, then this will be ready

toger5:
Just forgot to commit it. committed again

AndrewFerr (approved):

PR #2509, opened 2024-07-25
Import Compound icons in the new recommended way

robintown:
The Compound design tokens package is now set up to generate React components for every icon, so we no longer need to use our more error-prone method of importing the SVGs.


fkwp (approved):
lgtm

PR #2513, opened 2024-07-25
Give tiles a minimum area rather than a minimum width and height

robintown:
This seems to result in more sensible cropping and allocation of space across the board, in my testing.

### DIFF
--- a/src/grid/CallLayout.ts
+++ b/src/grid/CallLayout.ts
@@ -93,7 +93,6 @@ export interface GridArrangement {
   columns: number;
 }
 
-const tileMinHeight = 130;
 const tileMaxAspectRatio = 17 / 9;
 const tileMinAspectRatio = 4 / 3;
 const tileMobileMinAspectRatio = 2 / 3;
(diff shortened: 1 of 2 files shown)


src/grid/CallLayout.ts
```diff
@@ -110,38 +109,33 @@ export function arrangeTiles(
   // use of screen space for n tiles without making those tiles too small or
   // too cropped (having an extreme aspect ratio)
   const gap = width < 800 ? 16 : 20;
-  const tileMinWidth = width < 500 ? 150 : 180;
+  const area = width * minHeight;
+  // Magic numbers that make tiles scale up nicely as the window gets larger
+  const tileArea = Math.pow(Math.sqrt(area) / 8 + 125, 2);
```
toger5, line 114: Is the 8 coming from what we allow as the max number of tiles next to each other?

src/grid/CallLayout.ts
```diff
@@ -110,38 +109,33 @@ export function arrangeTiles(
   // use of screen space for n tiles without making those tiles too small or
   // too cropped (having an extreme aspect ratio)
   const gap = width < 800 ? 16 : 20;
-  const tileMinWidth = width < 500 ? 150 : 180;
+  const area = width * minHeight;
```
toger5, line 112: width and minHeight are for the whole area not just the tile right?

src/grid/CallLayout.ts
```diff
@@ -110,38 +109,33 @@ export function arrangeTiles(
...
     // Don't create more columns than we have items for
-    tileCount,
+    tilesPerPage,
     // The ideal number of columns is given by a packing of equally-sized
     // squares into a grid.
     // width / column = height / row.
     // columns * rows = number of squares.
     // ∴ columns = sqrt(width / height * number of squares).
     // Except we actually want 16:9-ish tiles rather than squares, so we
     // divide the width-to-height ratio by the target aspect ratio.
```
toger5, line 126:
```suggestion
    // The relation between the tile aspect ratio and the available area aspect ratio
    // defines the optimal ratio between columns and rows.
    // areaAspectRatio / tileAspectRatio = columns / rows
    // substitute: rows = tilesPerPage / columns
    // columns^2 = areaAspectRatio / tileAspectRatio  * tilesPerPage
```
This might be an equally simple approach without the step of inserting `tileMinAspectRatio` at the correct location out of intuition.

toger5 (approved):
I added a comment but we will still merge this now without updating it.
@robintown if you feel like its an improvement you can add it in another PR.
I am afraid, that my version might be less clear to some ppl and more clear to others. (so maybe no obvious improvement even though it looks to me like it is cleaner/simpler.)

PR #2514, opened 2024-07-25
Improve the layouts on small mobile calls

robintown:
Due to an oversight of mine, https://github.com/element-hq/element-call/commit/244003763953db033b5e5350142b3f00b7c8910a actually removed the ability to see the one-on-one layout on mobile. This restores mobile one-on-one calls to working order and also avoids showing the spotlight tile unless there are more than a few participants.

Based on https://github.com/element-hq/element-call/pull/2512 and https://github.com/element-hq/element-call/pull/2511

### DIFF
--- a/src/grid/CallLayout.ts
+++ b/src/grid/CallLayout.ts
@@ -95,7 +95,6 @@ export interface GridArrangement {
 
 const tileMaxAspectRatio = 17 / 9;
 const tileMinAspectRatio = 4 / 3;
-const tileMobileMinAspectRatio = 2 / 3;
 
 /**
  * Determine the ideal arrangement of tiles into a grid of a particular size.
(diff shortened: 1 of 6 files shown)


toger5:
This is super nice. This was my main personal issue with the new layout.
Are those also the current designs?
I have in mind, that I saw a design that has a borderless video for 1:1 calls?
If that is true, is this the right place to update it?

Edit: borderless is only in landscape mode. Here we also have a feed orientation issue (the feed is still in portrait if the phone is in landscape mode)

I am wondering if this is missing though since there are no designs for portrait mobile with less than a handful if participants. 1 and 2 participants might make sense to show as borderless however? Like the lobby designs?

fkwp:
nice :-) I just observed in landscape mode I have two times self in 1:1 rather remote and self. Does that depend on the other PRs floating around?

toger5:
Blocked because of merge conflicts.

robintown:
@fkwp Fixed that in https://github.com/element-hq/element-call/pull/2528

> I am wondering if this is missing though since there are no designs for portrait mobile with less than a handful if participants. 1 and 2 participants might make sense to show as borderless however? Like the lobby designs?

@toger5 Updated the PR with this suggestion and rebased on livekit :heavy_check_mark:

toger5 (commented):
The code looks good.

I have a couple of comments about the behaviour:

I think we should not base the switch between borderless and with-border based on the aspect ratio.
Portrait on a small screen with only two participants should also be borderless. Otherwise we get some odd window resizing behaviour where it can switch between borderless and with-boarder rapidly.

I think sth like
"aspect videomedia track/ aspect screen about the same" and "screen area < threshold" -> no border 
would be simpler?

I also think lobby and inCallView should use the same logic.

I think if there is only one person in the call we should also only show one video tile. (currently we mirror our own tile into the main tile slot)

toger5 (approved):
Discussed in a meeting.
 - two tiles comment gets addressed in a different PR
 - It turns out that most of my feedback is addressable with some value tweaking

PR #2521, opened 2024-07-31
Change vp8 to h264

toger5:
This is an experiment that allows us to use https://pr2521--element-call.netlify.app/ to test h264.


toger5:
Superseeded by: https://github.com/element-hq/element-call/pull/3345

PR #2522, opened 2024-07-31
Backport deviceSetup await

toger5:
This is a change from https://github.com/element-hq/element-call/pull/2475 which currently has merge conflicts but I think we should at least get this finding on main.


fkwp (approved):
lgtm

toger5:
lets see if that makes any difference.

PR #2531, opened 2024-08-01
Add back some margins to the interface

robintown:
There were a couple of cases where the lack of margins after the new layout changes just looked odd. Specifically, when the header is hidden (as in embedded mode), there would be no margin at the top of the window. Also the floating tile would run directly up against the sides of the window.

After:
Screenshot from 2024-08-01 16-32-59

### DIFF
--- a/src/grid/GridLayout.module.css
+++ b/src/grid/GridLayout.module.css
@@ -40,7 +40,8 @@ limitations under the License.
   position: absolute;
   inline-size: 404px;
   block-size: 233px;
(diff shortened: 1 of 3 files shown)


toger5 (approved):
I was expecting this to just add a margin at the top but a fake header is much better! thanks
(in general a before picture also would be super helpful. In case one cannot remember the exact issue or is unsure what margin you are referring to exactly. In this case I was more then aware and a before was not necessary at all ;) )

PR #2532, opened 2024-08-01
Fix Renovate warning

robintown:
Apparently Renovate doesn't really like it when you use a `group:` preset inside `packageRules`, instead of at the top level of the config. We do want to apply `schedule:weekly` only to the "all non-major dependencies" group though, so we need to write the group definition out by hand.


fkwp (approved):
lgtm

PR #2541, opened 2024-08-06
Don't show local media on top of itself

robintown:
If you were the only one in the call, you could get a broken-looking view in which the local tile is shown in the spotlight, and it's also shown in the PiP. This is redundant.

### DIFF
--- a/src/state/CallViewModel.ts
+++ b/src/state/CallViewModel.ts
@@ -488,11 +488,17 @@ export class CallViewModel extends ViewModel {
         ? ([of(screenShares.map((m) => m.vm)), this.spotlightSpeaker] as const)
         : ([
             this.spotlightSpeaker.pipe(map((speaker) => [speaker!])),
-            this.localUserMedia.pipe(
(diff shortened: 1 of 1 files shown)


toger5 (approved):
Code makes a lot of sense. And also works.

toger5:
Maybe this is not the best place to mention it but I think there is still some logic issue with the borderless -> border video slot.
Screenshot from 2024-08-07 10-26-31
Screenshot from 2024-08-07 10-26-28

If I change the height just a tiny bit it switches between those two screenshots.
Do you want me to open another issue for it or do you think this observation is enough to find a fix. (I can also look into what is causing this.)

toger5:
(Lobby and in view call are also not equivalent. I think for this I will open another issue)

robintown:
Huh, that's unexpected. I'll fix that in another PR.

toger5:
Created the lobby issue: https://github.com/element-hq/element-call/issues/2543

PR #2545, opened 2024-08-08
Consider any sufficiently short window 'flat'

robintown:
This is because our layouts for flat windows are good at adapting to both small width and small height, while our layouts for narrow windows aren't so good at adapting to a small height.


robintown:
Fixes https://github.com/element-hq/element-call/pull/2541#issuecomment-2272918887

fkwp (approved):

PR #2546, opened 2024-08-08
Avoid duplicating the video of someone in the spotlight

robintown:
We've gotten feedback that it's distracting whenever the same video is shown in two places on screen. This fixes the spotlight case by showing only the avatar of anyone who is already visible in the spotlight. It also makes sense to hide the speaking indicators in spotlight layouts, I think, because this information is redundant to the spotlight tile.


AndrewFerr (approved):
LGTM

PR #2548, opened 2024-08-08
Show controls on tap/hover on small screens

robintown:
This changes the mobile landscape view to automatically hide the controls, giving more visibility to the video underneath, and show them on tap/hover.

### DIFF
--- a/src/room/InCallView.module.css
+++ b/src/room/InCallView.module.css
@@ -22,31 +22,6 @@ limitations under the License.
   overflow-y: auto;
 }
 
-.controlsOverlay {
-  position: relative;
-  flex: 1;
-  display: flex;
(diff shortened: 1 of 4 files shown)


robintown:
Sadly this change causes the controls overlay to glitch in and out of existence on Firefox. We should discuss what to do about this tomorrow.

robintown:
Added the check for Firefox as discussed, it doesn't totally fix the layering issue but it manifests very rarely now.

fkwp:
no glitches on iOS (Safari)

AndrewFerr (approved):
My React knowledge is still patchy, but this LGTM from what I can tell :slightly_smiling_face:

PR #2550, opened 2024-08-09
(EXPERIMENTAL) Use updated js-sdk containing encryption key enhancements

hughns:


AndrewFerr:
I believe the js-sdk version with the crypto fixes has been used by EC for some time now.

Please reopen if there's anything left to do here.

PR #2552, opened 2024-08-09
Don't keep someone in the spotlight if they've left the call

robintown:


src/state/CallViewModel.ts
AndrewFerr, line 1: Optional refactor for readability:

```typescript
      scan<(readonly [UserMedia, boolean])[], UserMedia, null>(
        (prev, mediaItems) => {
          // Only remote users that are still in the call should be sticky
          const [stickyPrevUserMedia, stickyPrevStillSpeaking] =
            (!prev?.vm.local && mediaItems.find(([m]) => m === prev)) || [];
          // Decide who to spotlight:
          // If the previous speaker is still speaking, stick with them rather
          // than switching eagerly to someone else
          return stickyPrevUserMedia && stickyPrevStillSpeaking
            ? stickyPrevUserMedia
            : // Otherwise, select any remote user who is speaking
              (mediaItems.find(([m, s]) => !m.vm.local && s)?.[0] ??
                // Otherwise, stick with the person who was last speaking
                stickyPrevUserMedia ??
                // Otherwise, spotlight an arbitrary remote user
                mediaItems.find(([m]) => !m.vm.local)?.[0] ??
                // Otherwise, spotlight the local user
                mediaItems.find(([m]) => m.vm.local)![0]);
        },
        null,
      ),
```

AndrewFerr (approved):

PR #2558, opened 2024-08-12
Proof of concept for spotlight PiP pop-out, webkitPresentationMode edition

robintown:


robintown:
https://github.com/element-hq/element-call/pull/2563

PR #2563, opened 2024-08-13
Support picture-in-picture on Element X iOS

robintown:


robintown:
https://github.com/element-hq/element-call/pull/2573

PR #2569, opened 2024-08-16
Fix long call names overflowing the interface

robintown:
|Before|After|
-|-
Screen Shot 2024-08-16 at 15 19 03|Screen Shot 2024-08-16 at 15 18 47


fkwp (approved):
lgtm

Issue #2537, opened 2024-08-05
Element X iOS Specific Picture in Picture


stefanceriu:
#### What would you like to do?
- take advantage of the system picture in picture behaviors while using ElementCall (especially when embedded in EX) 

#### Why would you like to do it?
- because it provides a better user experience than whatever custom solution we might come up with

#### How would you like to achieve it?
- allow picture in picture to be enabled from the web side on currently active speaker tile(?)
- iOS will intercept that and take the right steps to hide the main EC window so that the user can interact with the rest of the application
- web will change that particular video tag's source to be whoever is currently speaking and the iOS PiP will automatically update (as per our exploratory testing)

toger5:
how do we want to communicate this between app and webview?
I can imagine that EX will know when it gets pushed into the background and then informs the app that it should start PIP now?

stefanceriu:
EX will not get pushed into the background per se but the PiP will be presented on top of it. When that happens the system posts a `AVPictureInPictureControllerDidStart` notification which we can intercept and "minimise" the parent EC WebView

toger5:
What is the initial event that we need to use to start the PIP in EC in the first place?

stefanceriu:
I would assume that would sit completely on the Element Call side as a button on the video or similar.

toger5:
I would expect, that i can just use native multitasking from ios and still see the PIP.
(this is also what we need to reach feature parity with EXA)

stefanceriu:
I'm no longer sure we're talking about the same thing, what do you mean by native multitasking? How would you imagine this working?

There are different use cases for which we clearly need product input:
* does the user want to interact with the rest of the app and minimize element call? Element call currently controls all of the chrome.
* does the user want to minimize element x as a whole? would they expect element call to automagically enter pip mode?

YouTube for example automatically enters and exists PiP when backgrounding and foregrounding the app but using a completely custom interface for when minimizing within the app. We're trying to avoid coming up with a custom solution.

toger5:
I think that switching apps (minimizing apps) and automatically entering pip mode for the call is the primary use case. I dont see why we would want to show a pip view above the ongoing call. So making the user manually start a pip view before they switch to another app/screen on their device seems unnecessary.

This is also what we have on android.

So I would expect:
 - EX detecting that it will be moved to the BG.
  - EX sending an event to ElementCall
  - ElementCall starts a native PIP view
  - EX gets put into BG but the native PIP view is still visible to the user
  
[...]

PR #2573, opened 2024-08-20
Add simple global controls to put the call in picture-in-picture mode

robintown:
Our web and mobile apps (will) all support putting calls into a picture-in-picture mode. However, it'd be nice to have a way of doing this that's more explicit than a breakpoint, because PiP views could in theory get fairly large. Specifically, on mobile, we want a way to do this that can tell you whether the call is ongoing, and that works even without the widget API (because we support SPA calls in the Element X apps…)

To this end, I've created a simple global "controls" API on the window. Right now it only has methods for controlling the picture-in-picture state, but in theory we can expand it to also control mute states, which is current possible via the widget API only.

See the commits for details; I had to make a couple changes along the way to fix some leaks that were causing the APIs to not throw when outside a call.

For https://github.com/element-hq/element-call/issues/2537


src/controls.ts
```diff
@@ -0,0 +1,40 @@
...
+  canEnterPip(): boolean {
+    return pipEnable.observed;
+  },
+  enablePip(): void {
+    if (!pipEnable.observed) throw new Error("No call is running");
+    pipEnable.next();
+  },
+  disablePip(): void {
+    if (!pipDisable.observed) throw new Error("No call is running");
+    pipDisable.next();
```
bmarty: Noob question, but rather than having 2 Subject, would it be possible to have only 1 Subject which emits Booleans?

toger5 (approved):
This looks good to me.
I think we should meet up and discuss if we want to commit to one of the two:
 - custom js methods
 - custom widget api actions that get caught by the host client before making it to the widget api.

PR #2574, opened 2024-08-21
[WIP] Element Call dev view

toger5:
This is the work in progress of the element web view (this was done a week ago and I missed creating a PR.)

It is very rudamentary.
The idea is that we can easily debug key exchanges call member state and livekit state from the clients perspective.


toger5:
This has this has bitrotten and maybe can be done differently with rx-js.
Maybe there is a chance to add this in a smarter/simpler way as part of to-device encryption.

I think We will not continue on this pr however

PR #2575, opened 2024-08-22
Perform dead code analysis with Knip

### LINKED ISSUE #2544: Dead code analysis

Enforce in CI that there is no dead code, unused exports, unused dependencies, etc.

We shouldn't spend time maintaining things we don't use

https://knip.dev/
Element Web uses this project, and it apparently works well with very little configuration.

https://www.npmjs.com/package/find-unused-exports, but it flags so many false positives by default. It was probably meant for libraries.

### Additional context

_No response_

robintown:
See the commits for details.

Closes https://github.com/element-hq/element-call/issues/2544


toger5 (approved):
This looks good.
Even though I would love to understand the codecov token scenario.
It was added via the `token` field before. That failed in this PR (on robins fork) but worked on the copy of this PR I have done in on this repo: https://github.com/element-hq/element-call/pull/2584

I changed it to the `env` field as of: https://docs.codecov.com/docs/quick-start
and now it works. This currently makes little sense to me since this should still be the same token.

PR #2584, opened 2024-08-27
Perform dead code analysis with Knip (copied version)

toger5:
Seems I lost write access to robins repo. So this is a rebased version of: https://github.com/element-hq/element-call/pull/2575


toger5:
Okay it seems like this is now also not working anymore. This is good news since now both `Perform dead code analysis` are not working with the token being declared as `token:` instaed of `env.
See https://github.com/element-hq/element-call/pull/2575#pullrequestreview-2264665251.

toger5:
Closed in favor of https://github.com/element-hq/element-call/pull/2575

PR #2587, opened 2024-08-29
Fix rust crypto: Don't use legacy `cryptoStore` anymore.

toger5:
The logic with the legacy store was intercepting with the `rustCrypto that handles all the cases itself.

### DIFF
--- a/src/utils/matrix.ts
+++ b/src/utils/matrix.ts
@@ -16,9 +16,6 @@ limitations under the License.
 
 import { IndexedDBStore } from "matrix-js-sdk/src/store/indexeddb";
 import { MemoryStore } from "matrix-js-sdk/src/store/memory";
-import { IndexedDBCryptoStore } from "matrix-js-sdk/src/crypto/store/indexeddb-crypto-store";
-import { LocalStorageCryptoStore } from "matrix-js-sdk/src/crypto/store/localStorage-crypto-store";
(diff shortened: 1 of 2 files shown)


fkwp (approved):
lgtm

src/utils/matrix.ts
```diff
@@ -184,15 +130,13 @@ export async function initClient(
...
 
-  try {
-    await client.store.startup();
-  } catch (error) {
-    logger.error(
-      "Error starting matrix client store. Falling back to memory store.",
-      error,
-    );
-    client.store = new MemoryStore({ localStorage });
-    await client.store.startup();
```
richvdh, line 195: Ah. You've fallen into the trap of confusing `client.store` with `client.cryptoStore`. This is an understandable mistake, because the name is unclear and the documentation is, well, about as good as the rest of the documentation for the js-sdk.

`client.store` is used to stash things like sync results. It's totally independent of the cryptostore, and uses a separate indexeddb database.

Anyway, calling `client.store.startup()` is important, and you need to keep this, otherwise `client.startClient()` will hang. (You might reasonably ask why `client.startClient` doesn't call `store.startup()` itself. I don't know.)

toger5: Thank you so much Rich!!

This would have been a debugging session to fix this without your comment.

🙏

toger5: Added: https://github.com/element-hq/element-call/pull/2591

PR #2588, opened 2024-08-29
Remove shadow for layout switcher.

toger5:
So the layout switcher looks more aligned with the other buttons.

Now:
Screenshot from 2024-08-29 15-16-14
Before:
Screenshot from 2024-08-29 15-15-07


fkwp (approved):
lgtm

PR #2590, opened 2024-08-29
Make js-sdk imports consistent

richvdh:
We need to be consistent about whether we import matrix-js-sdk from `src` or `lib`, otherwise we get two copies of matrix-js-sdk, and everything explodes.

This fixes a long-standing bug which meant that element-call couldn't be successfully built against release versions of matrix-js-sdk. That bug recently became more acute due to https://github.com/matrix-org/matrix-js-sdk/pull/4357, which brought non-release versions of matrix-js-sdk into line with release versions.


AndrewFerr (approved):
This indeed fixes some build / import issues with EC, and unblocks it from being able to use newer js-sdk versions. Thanks for this!

PR #2591, opened 2024-08-30
 Fix for missing client store (caused by: #2587)

toger5:


src/utils/matrix.ts
```diff
@@ -139,7 +139,21 @@ export async function initClient(
     client.clearStores();
   }
 
+  // start the client store (totally independent to the crypto store)
```
fkwp, line 146:
```suggestion
  // start client store 
  // Note client.store is used to stash things like sync results. It's totally independent of the cryptostore, and uses a separate indexeddb database.
```

fkwp (approved):
lgtm

PR #2598, opened 2024-08-30
Test MediaViewModel

robintown:
This was the result of me playing around with [RxJS marble testing](https://rxjs.dev/guide/testing/marble-testing) to understand how to get things done with its `TestScheduler`. I discovered that it lacks a clear way to fire arbitrary actions during the test, so I built a small helper function called `schedule` which does this for us.

### DIFF
--- a/src/state/MediaViewModel.test.ts
+++ b/src/state/MediaViewModel.test.ts
@@ -1,0 +1,132 @@
+/*
+Copyright 2024 New Vector Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { RoomMember } from "matrix-js-sdk/src/matrix";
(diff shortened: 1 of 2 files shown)


toger5 (approved):
Cool this looks nice.
I am not sure where exactly it makes the most sense to do the abstraction cut.
creating the `new testScheduler` and using the `deep.equals` from vitest definitely should be abstracted.
I think we might also want a withTestScheduler that does not enforce the marbles -> action pattern. So I am wondering if we should already call it `withActionTestScheduler`? (easy to refactor afterwards however so I will approve for now.

robintown:
> I think we might also want a withTestScheduler that does not enforce the marbles -> action pattern.

What do you mean by this? Are you saying it would be nice to have a variant that lets you run observables and actions in real time? That's what you already get when writing a test without TestScheduler.

toger5:
I am wondering if this:
```
export function withTestScheduler(
  continuation: (helpers: OurRunHelpers) => void,
): void {
  new TestScheduler((actual, expected) => {
    expect(actual).deep.equals(expected);
  }).run((helpers) =>
    continuation(helpers),
  );
}
```
On its own is already helpful?
But thinking it through:
```
withTestScheduler((helpers)=>{
  // some custom rxjs testing
  helpers.cold("some_marbels").pipe(...);
  // some custom checks
  helpers.expectObservable()...;
})

VS

new TestScheduler((actual, expected) => {
  expect(actual).deep.equals(expected);
}).run((helpers) =>
  // some custom rxjs testing
  helpers.cold("some_marbels").pipe(...);
  // some custom checks
  helpers.expectObservable()...;
);
```
might not be enough of a simplification to justify the reduced explicity.

PR #2602, opened 2024-09-02
Display QR code when sharing invite link

### LINKED ISSUE #2495: QR support for sharing link
It would be nice if element call could generate a QR code for sharing the link to the call

Devices like smart TVs can work with element, but it can be really hard to get the link to the TV, or from the TV in many cases a QR code would work great

show a QR code when you click invite. 

I have a hard time finding other good ideas for alternatives, except for maybe an element client optimized for smart TVs

### Additional context

_No response_

Johennes:
Fixes: #2495

Not sure about the size and placement of the code but this looked fairly ok to me. Happy to try and add test coverage once it's confirmed that you actually want this feature. 😅

qr


src/QrCode.module.css
```diff
@@ -0,0 +1,20 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+.qrCode img {
+  max-width: 100%;
+  image-rendering: pixelated;
```
toger5, line 19: I think giving the qr code image a border radius would make it look more at home.

```suggestion
  image-rendering: pixelated;
  border-radius: var(--cpd-space-4x);
```

Screenshot from 2024-09-02 16-53-51

toger5 (approved):
Code looks good + thanks for the test.

I added a nonblocking comment about the styling.

src/QrCode.module.css
Johennes: Oh nice! 👍

PR #2603, opened 2024-09-02
Fix (rust crypto): Adjust login procedures to account for rust crypto behaviour.

toger5:
- If we have an invalid access token return `null` so that we get prompted with the registration (create call without being authenticated) view.
 - logout the guest user, before we try to login (interactive login with username and pwd)
 - clear (and await) the rust crypto store before we initialize it in case we do not want to restore a localstore session.


fkwp (approved):
lgtm

toger5:
We will merge this without meeting the coverage requirements since our time can be used for more useful tests.

PR #2607, opened 2024-09-03
Enable lint rules for Promise handling to discourage misuse of them. 

toger5:
This is using this PR as and rebases it ontop of the livekit branch. it squashes all the changes into one commit.
This cleans the history and simplifies rebasing.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,6 +24,7 @@ import {
 import * as Sentry from "@sentry/react";
 import { History } from "history";
 import { TooltipProvider } from "@vector-im/compound-web";
+import { logger } from "matrix-js-sdk/src/logger";
 
 import { HomePage } from "./home/HomePage";
 import { LoginPage } from "./auth/LoginPage";
@@ -70,11 +71,15 @@ interface AppProps {
 export const App: FC<AppProps> = ({ history }) => {
   const [loaded, setLoaded] = useState(false);
   useEffect(() => {
-    Initializer.init()?.then(() => {
-      if (loaded) return;
-      setLoaded(true);
-      widget?.api.sendContentLoaded();
-    });
+    Initializer.init()
+      ?.then(async () => {
+        if (loaded) return;
+        setLoaded(true);
+        await widget?.api.sendContentLoaded();
+      })
+      .catch((e) => {
+        logger.error(e);
+      });
   });
 
   const errorPage = <CrashView />;
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -77,9 +77,7 @@ export const App: FC<AppProps> = ({ history }) => {
         setLoaded(true);
         await widget?.api.sendContentLoaded();
       })
-      .catch((e) => {
-        logger.error(e);
-      });
+      .catch(logger.error);
   });
 
   const errorPage = <CrashView />;
--- a/src/analytics/PosthogAnalytics.ts
+++ b/src/analytics/PosthogAnalytics.ts
@@ -265,7 +265,7 @@ export class PosthogAnalytics {
         this.posthog.identify(analyticsID);
       } else {
         logger.info(
-          "No analyticsID is availble. Should not try to setup posthog",
+          "No analyticsID is available. Should not try to setup posthog",
         );
       }
     }
@@ -333,7 +333,9 @@ export class PosthogAnalytics {
   }
 
   public onLoginStatusChanged(): void {
-    this.maybeIdentifyUser();
+    this.maybeIdentifyUser().catch(() =>
+      logger.log("Could not identify user on login status change"),
+    );
   }
 
   private updateSuperProperties(): void {
@@ -382,20 +384,27 @@ export class PosthogAnalytics {
     }
   }
 
-  public async trackEvent<E extends IPosthogEvent>(
+  public trackEvent<E extends IPosthogEvent>(
     { eventName, ...properties }: E,
     options?: CaptureOptions,
-  ): Promise<void> {
+  ): void {
+    const doCapture = (): void => {
+      if (
+        this.anonymity == Anonymity.Disabled ||
+        this.anonymity == Anonymity.Anonymous
+      )
+        return;
+      this.capture(eventName, properties, options);
+    };
+
     if (this.identificationPromise) {
-      // only make calls to posthog after the identificaion is done
-      await this.identificationPromise;
+      // only make calls to posthog after the identification is done
+      this.identificationPromise.then(doCapture).catch((e) => {
+        logger.error("Failed to identify user for tracking", e);
+      });
+    } else {
+      doCapture();
     }
-    if (
-      this.anonymity == Anonymity.Disabled ||
-      this.anonymity == Anonymity.Anonymous
-    )
-      return;
-    this.capture(eventName, properties, options);
   }
 
   private startListeningToSettingsChanges(): void {
@@ -409,7 +418,9 @@ export class PosthogAnalytics {
     // won't be called (i.e. this.anonymity will be left as the default, until the setting changes)
     optInAnalytics.value.subscribe((optIn) => {
       this.setAnonymity(optIn ? Anonymity.Pseudonymous : Anonymity.Disabled);
-      this.maybeIdentifyUser();
+      this.maybeIdentifyUser().catch(() =>
+        logger.log("Could not identify user"),
+      );
     });
   }
 
--- a/src/analytics/PosthogAnalytics.ts
+++ b/src/analytics/PosthogAnalytics.ts
@@ -399,7 +399,7 @@ export class PosthogAnalytics {
 
     if (this.identificationPromise) {
       // only make calls to posthog after the identification is done
-      this.identificationPromise.then(doCapture).catch((e) => {
+      this.identificationPromise.then(doCapture, (e) => {
         logger.error("Failed to identify user for tracking", e);
       });
     } else {
--- a/src/analytics/PosthogSpanProcessor.ts
+++ b/src/analytics/PosthogSpanProcessor.ts
@@ -43,20 +43,19 @@ export class PosthogSpanProcessor implements SpanProcessor {
 
   public onStart(span: Span): void {
     // Hack: Yield to allow attributes to be set before processing
-    Promise.resolve()
-      .then(() => {
-        switch (span.name) {
-          case "matrix.groupCallMembership":
-            this.onGroupCallMembershipStart(span);
-            return;
-          case "matrix.groupCallMembership.summaryReport":
-            this.onSummaryReportStart(span);
-            return;
-        }
-      })
-      .catch((e) => {
-        // noop
-      });
+    try {
+      switch (span.name) {
+        case "matrix.groupCallMembership":
+          this.onGroupCallMembershipStart(span);
+          return;
+        case "matrix.groupCallMembership.summaryReport":
+          this.onSummaryReportStart(span);
+          return;
+      }
+    } catch (e) {
+      // log to avoid tripping @typescript-eslint/no-unused-vars
+      logger.debug(e);
+    }
   }
 
   public onEnd(span: ReadableSpan): void {
--- a/src/analytics/PosthogSpanProcessor.ts
+++ b/src/analytics/PosthogSpanProcessor.ts
@@ -161,8 +161,7 @@ export class PosthogSpanProcessor implements SpanProcessor {
   /**
    * Shutdown the processor.
    */
-  // eslint-disable-next-line @typescript-eslint/require-await
   public async shutdown(): Promise<void> {
-    return;
+    return Promise.resolve();
   }
 }
--- a/src/analytics/PosthogSpanProcessor.ts
+++ b/src/analytics/PosthogSpanProcessor.ts
@@ -43,16 +43,20 @@ export class PosthogSpanProcessor implements SpanProcessor {
 
   public onStart(span: Span): void {
     // Hack: Yield to allow attributes to be set before processing
-    Promise.resolve().then(() => {
-      switch (span.name) {
-        case "matrix.groupCallMembership":
-          this.onGroupCallMembershipStart(span);
-          return;
-        case "matrix.groupCallMembership.summaryReport":
-          this.onSummaryReportStart(span);
-          return;
-      }
-    });
+    Promise.resolve()
+      .then(() => {
+        switch (span.name) {
+          case "matrix.groupCallMembership":
+            this.onGroupCallMembershipStart(span);
+            return;
+          case "matrix.groupCallMembership.summaryReport":
+            this.onSummaryReportStart(span);
+            return;
+        }
+      })
+      .catch((e) => {
+        // noop
+      });
   }
 
   public onEnd(span: ReadableSpan): void {
@@ -157,7 +161,8 @@ export class PosthogSpanProcessor implements SpanProcessor {
   /**
    * Shutdown the processor.
    */
-  public shutdown(): Promise<void> {
-    return Promise.resolve();
+  // eslint-disable-next-line @typescript-eslint/require-await
+  public async shutdown(): Promise<void> {
+    return;
   }
 }
--- a/src/auth/useInteractiveLogin.ts
+++ b/src/auth/useInteractiveLogin.ts
@@ -58,9 +58,8 @@ export function useInteractiveLogin(
             password,
           }),
         stateUpdated: (): void => {},
-        requestEmailToken: async (): Promise<{ sid: string }> => {
-          return Promise.resolve({ sid: "" });
-        },
+        requestEmailToken: async (): Promise<{ sid: string }> =>
+          Promise.resolve({ sid: "" }),
       });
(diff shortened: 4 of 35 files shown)


toger5:
The remaining lint issues are already fixed in: https://github.com/element-hq/element-call/pull/2523
We should first merge: https://github.com/element-hq/element-call/pull/2523 and then rebase again to see what is left.
EDIT: this has happened now.

toger5:
I closed: https://github.com/element-hq/element-call/pull/2475 to minimize confusion.
Linking it here since it might be helpful when reviewing.

src/analytics/PosthogAnalytics.ts
```diff
@@ -382,20 +384,27 @@ export class PosthogAnalytics {
...
+      )
+        return;
+      this.capture(eventName, properties, options);
+    };
+
     if (this.identificationPromise) {
-      // only make calls to posthog after the identificaion is done
-      await this.identificationPromise;
+      // only make calls to posthog after the identification is done
+      this.identificationPromise.then(doCapture).catch((e) => {
```
AndrewFerr: Maybe to make it clear that `doCapture` doesn't return a Promise that needs catching:
```suggestion
      this.identificationPromise.then(doCapture, (e) => {
```

src/analytics/PosthogSpanProcessor.ts
```diff
@@ -43,16 +43,20 @@ export class PosthogSpanProcessor implements SpanProcessor {
...
+            this.onGroupCallMembershipStart(span);
+            return;
+          case "matrix.groupCallMembership.summaryReport":
+            this.onSummaryReportStart(span);
+            return;
+        }
+      })
+      .catch((e) => {
+        // noop
+      });
```
AndrewFerr: Not new, but what's the reason for the `Promise.resolve`? Since the called methods are synchronous, could this be something like the following:
```suggestion
    try {
      switch (span.name) {
        case "matrix.groupCallMembership":
          this.onGroupCallMembershipStart(span);
          return;
        case "matrix.groupCallMembership.summaryReport":
          this.onSummaryReportStart(span);
          return;
      }
    } catch (e) {
      // log to avoid tripping @typescript-eslint/no-unused-vars
      logger.debug(e);
    }
```

src/auth/useInteractiveLogin.ts
```diff
@@ -58,7 +58,7 @@ export function useInteractiveLogin(
             password,
           }),
         stateUpdated: (): void => {},
-        requestEmailToken: (): Promise<{ sid: string }> => {
+        requestEmailToken: async (): Promise<{ sid: string }> => {
           return Promise.resolve({ sid: "" });
         },
```
AndrewFerr: This can be a one-liner
```suggestion
        requestEmailToken: async (): Promise<{ sid: string }> => Promise.resolve({ sid: "" }),
```

src/App.tsx
```diff
@@ -70,11 +71,15 @@ interface AppProps {
...
-    });
+    Initializer.init()
+      ?.then(async () => {
+        if (loaded) return;
+        setLoaded(true);
+        await widget?.api.sendContentLoaded();
+      })
+      .catch((e) => {
+        logger.error(e);
+      });
```
AndrewFerr: If there are enough loggers that simply log the caught error variable without modifying it or adding some other text to the logged message, they could be made into one-liners:
```suggestion
      .catch(logger.error);
```

src/auth/useInteractiveRegistration.ts
```diff
@@ -87,17 +88,25 @@ export const useInteractiveRegistration = (
...
+              })
+              .catch((e) => {
+                logger.error(e);
+              });
           }
         },
-        requestEmailToken: (): Promise<{ sid: string }> => {
+        requestEmailToken: async (): Promise<{ sid: string }> => {
           return Promise.resolve({ sid: "dummy" });
         },
```
AndrewFerr: Can be a one-liner
```suggestion
        requestEmailToken: async (): Promise<{ sid: string }> => Promise.resolve({ sid: "dummy" }),
```

src/config/Config.ts
```diff
@@ -30,17 +30,17 @@ export class Config {
...
+
+    Config.internalInstance.initPromise = downloadConfig("../config.json").then(
+      (config) => {
         Config.internalInstance.config = { ...DEFAULT_CONFIG, ...config };
-        resolve();
-      });
-    });
+      },
+    );
     return Config.internalInstance.initPromise;
```
AndrewFerr: IMO this reads better if it has a single `return` statement and moves the preparation logic into a branch:
```typescript
    if (!Config.internalInstance?.initPromise) {
      const internalInstance = new Config();
      Config.internalInstance = internalInstance;

      Config.internalInstance.initPromise = downloadConfig("../config.json").then(
        (config) => {
          internalInstance.config = { ...DEFAULT_CONFIG, ...config };
        },
      );
    }
    return Config.internalInstance.initPromise;
```
Also, `internalInstance` can be declared as a `Config | undefined` for some extra typechecking (and this is why my snipped has `const internalInstance`).

src/e2ee/matrixKeyProvider.ts
```diff
@@ -55,19 +55,24 @@ export class MatrixKeyProvider extends BaseKeyProvider {
...
+        this.onSetEncryptionKey(keyMaterial, participantId, encryptionKeyIndex);
 
-    logger.debug(
-      `Sent new key to livekit room=${this.rtcSession?.room.roomId} participantId=${participantId} encryptionKeyIndex=${encryptionKeyIndex}`,
-    );
+        logger.debug(
+          `Sent new key to livekit room=${this.rtcSession?.room.roomId} participantId=${participantId} encryptionKeyIndex=${encryptionKeyIndex}`,
+        );
+      })
+      .catch((e) => {
```
AndrewFerr: Since the `then` callback doesn't return a Promise, I think this can be
```suggestion
      },
      (e) => {
```

src/grid/Grid.tsx
```diff
@@ -372,7 +372,7 @@ export function Grid<
     const { tileId, tileX, tileY } = dragState.current!;
     const tile = placedTiles.find((t) => t.id === tileId)!;
 
-    springRef.current
+    void springRef.current
```
AndrewFerr: Instead of returning nothing, could the result of this be caught & logged?

src/grid/Grid.tsx
```diff
@@ -362,7 +362,7 @@ export function Grid<
   // Because we're using react-spring in imperative mode, we're responsible for
   // firing animations manually whenever the tiles array updates
   useEffect(() => {
-    springRef.start();
+    void springRef.start();
```
AndrewFerr: Should the returned Promises be caught?
```suggestion
    springRef.start().forEach((p) => void p.catch(logger.error));
```

src/initializer.tsx
```diff
@@ -129,10 +133,14 @@ export class Initializer {
...
-      Config.init().then(() => {
-        this.loadStates.config = LoadState.Loaded;
-        this.initStep(resolve);
-      });
+      Config.init()
+        .then(() => {
+          this.loadStates.config = LoadState.Loaded;
+          this.initStep(resolve);
+        })
+        .catch((e) => {
```
AndrewFerr: Maybe
```suggestion
        },
        (e) => {
```

src/livekit/openIDSFU.ts
```diff
@@ -54,7 +54,9 @@ export function useOpenIDSFU(
         ? await getSFUConfigWithOpenID(client, activeFocus)
         : undefined;
       setSFUConfig(sfuConfig);
```
AndrewFerr: Instead of having to create an anonymous async function, how about something like this:
```typescript
  useEffect(() => {
    if (activeFocus) {
      getSFUConfigWithOpenID(client, activeFocus)
        .then((sfuConfig) => {
          setSFUConfig(sfuConfig);
        },
        (e) => {
          logger.error("Failed to get SFU config", e);
        });
    } else {
      setSFUConfig(undefined);
    }
  }, [client, activeFocus]);
```

src/room/GroupCallView.tsx
```diff
@@ -187,24 +179,31 @@ export const GroupCallView: FC<Props> = ({
...
+      const onJoin = (ev: CustomEvent<IWidgetApiRequest>): void => {
+        defaultDeviceSetup(ev.detail.data as unknown as JoinCallData)
+          .catch((e) => {
+            logger.error("Error setting up default devices", e);
+          })
+          .then(async () => enterRTCSession(rtcSession, perParticipantE2EE))
+          .then(() => widget!.api.transport.reply(ev.detail, {}))
+          .catch((e) => {
+            logger.error("Error entering RTC session", e);
+          });
```
AndrewFerr: The first `.then` callback (and its follow-ups) will be reached even if `defaultDeviceSetup` rejects, which I assume is not the intention here, given that the original implementation would early-exit if any of the called methods were to throw.

An easy way to get the same early-exit behaviour as the original while still returning `void` (at the cost of less granular error logging) is to wrap the `await`ed calls in an anonymous async function, which is a pattern I think I've seen elsewhere in this codebase:
```suggestion
        (async (): Promise<void> => {
          await defaultDeviceSetup(ev.detail.data as unknown as JoinCallData);
          await enterRTCSession(rtcSession, perParticipantE2EE);
          widget!.api.transport.reply(ev.detail, {});
        })().catch((e) => {
          logger.error("Error joining RTC session", e);
        });
```

The same can be done a few lines down.

src/room/GroupCallView.tsx
```diff
@@ -187,24 +179,31 @@ export const GroupCallView: FC<Props> = ({
...
-      join();
+      // No lobby and no preload: we enter the rtc session right away
+      defaultDeviceSetup({ audioInput: null, videoInput: null })
+        .catch((e) => {
+          logger.error("Error setting up default devices", e);
+        })
+        .then(async () => enterRTCSession(rtcSession, perParticipantE2EE))
+        .catch((e) => {
+          logger.error("Error entering RTC session", e);
+        });
```
AndrewFerr:
```suggestion
      (async (): Promise<void> => {
        await defaultDeviceSetup({ audioInput: null, videoInput: null });
        await enterRTCSession(rtcSession, perParticipantE2EE);
      })().catch((e) => {
        logger.error("Error joining RTC session", e);
      });
```

src/room/InCallView.tsx
```diff
@@ -470,8 +477,8 @@ export const InCallView: FC<InCallViewProps> = ({
     rtcSession.room.roomId,
   );
 
-  const toggleScreensharing = useCallback(async () => {
-    await localParticipant.setScreenShareEnabled(!isScreenShareEnabled, {
+  const toggleScreensharing = useCallback(() => {
+    void localParticipant.setScreenShareEnabled(!isScreenShareEnabled, {
```
AndrewFerr: Should this be non-`void` & followed by `.catch`?

src/room/RoomPage.tsx
```diff
@@ -68,9 +68,13 @@ export const RoomPage: FC = () => {
...
-      registerPasswordlessUser(displayName).finally(() => {
-        setIsRegistering(false);
-      });
+      registerPasswordlessUser(displayName)
+        .finally(() => {
+          setIsRegistering(false);
+        })
+        .catch((e) => {
+          logger.error("Failed to register passwordless user", e);
+        });
```
AndrewFerr: It probably doesn't make a difference here, but for better style, put the `.catch` before the `.finally`.
```suggestion
        .catch((e) => {
          logger.error("Failed to register passwordless user", e);
        })
        .finally(() => {
          setIsRegistering(false);
        });
```

src/room/useLoadGroupCall.ts
```diff
@@ -165,17 +165,21 @@ export const useLoadGroupCall = (
...
-              logger.log("Auto-joined %s", room.roomId);
-              resolve();
+              client
+                .joinRoom(room.roomId, { viaServers })
+                .then((room) => {
+                  joinedRoom = room;
+                  logger.log("Auto-joined %s", room.roomId);
+                  resolve();
+                })
+                .catch((e) => reject(e));
```
AndrewFerr: AFAICT there's never a case where `joinedRoom` will be left at `null` unless this promise rejects; and since the await on it isn't in a try-catch, the branch on `!joinedRoom` is unreachable.

How about making the promise simply resolve to the joined room, like so:
```typescript
    const getRoomByKnocking = async (
      roomId: string,
      viaServers: string[],
      onKnockSent: () => void,
    ): Promise<Room> => {
      await client.knockRoom(roomId, { viaServers });
      onKnockSent();
      return await new Promise<Room>((resolve, reject) => {
        client.on(
          RoomEvent.MyMembership,
          (room, membership, prevMembership): void => {
            if (roomId !== room.roomId) return;
            activeRoom.current = room;
            if (
              membership === KnownMembership.Invite &&
              prevMembership === KnownMembership.Knock
            ) {
              client
                .joinRoom(room.roomId, { viaServers })
                .then((room) => {
                  logger.log("Auto-joined %s", room.roomId);
                  resolve(room);
                },
                reject);
            }
            if (membership === KnownMembership.Ban) reject(bannedError());
            if (membership === KnownMembership.Leave)
              reject(knockRejectError());
          },
        );
      });
    };
```

src/rtcSessionHelpers.ts
```diff
@@ -128,13 +127,13 @@ const widgetPostHangupProcedure = async (
   // we need to wait until the callEnded event is tracked on posthog.
   // Otherwise the iFrame gets killed before the callEnded event got tracked.
   await new Promise((resolve) => window.setTimeout(resolve, 10)); // 10ms
-  widget.api.setAlwaysOnScreen(false);
+  await widget.api.setAlwaysOnScreen(false);
```
AndrewFerr: Without this await, errors thrown by `setAlwaysOnScreen` didn't cause the calling function to throw; is that still the desired behaviour?  If so, just use `.catch` on this.

src/settings/rageshake.ts
```diff
@@ -239,7 +237,7 @@ class IndexedDBLogStore {
       }
       // queue up a flush to occur immediately after the pending one completes.
       this.flushAgainPromise = this.flushPromise
-        .then(() => {
+        .then(async () => {
           return this.flush();
         })
```
AndrewFerr: Nitpick:
```suggestion
        .then(this.flush)
```
which is safe as long as `flush` is an arrow function (and thus gets the correct binding for its `this`).

src/useWakeLock.ts
```diff
@@ -28,19 +28,22 @@ export function useWakeLock(): void {
...
+            .then((newLock) => {
+              lock = newLock;
+              // Handle the edge case where this component unmounts before the
+              // promise resolves
+              if (!mounted)
+                lock
+                  .release()
+                  .catch((e) => logger.warn("Can't release wake lock", e));
+            })
+            .catch((e) => {
```
AndrewFerr: Maybe:
```suggestion
            },
            (e) => {
```

src/utils/matrix.ts
```diff
@@ -270,11 +270,17 @@ export async function createRoom(
...
+        .then((result) => {
+          if (room.roomId === result.room_id) {
+            resolve();
+            cleanUp();
+          }
+        })
+        .catch((e) => {
+          logger.error("Failed to wait for the room to arrive", e);
+        });
     };
```
AndrewFerr: Since the await on the inlined promise requires `createPromise` to settle, we can avoid having to await on `createPromise` again to pull the result out, by having the inlined promise resolve to that result:

```typescript
  const roomId = await new Promise<string>((resolve, reject) => {
    const onRoom = (room: Room): void => {
      createPromise
        .then((result) => {
          if (room.roomId === result.room_id) {
            resolve(room.roomId);
            cleanUp();
          }
        },
        (e) => {
          logger.error("Failed to wait for the room to arrive", e);
        });
    };
    // snip
```

Then `roomId` can be used instead of `result.room_id` later on.

AndrewFerr (changes requested):

AndrewFerr:
Perhaps it's also appropriate for this PR to update eslint-related dependencies:
```bash
yarn add -D @typescript-eslint/parser @typescript-eslint/eslint-plugin eslint@^8.57.0 typescript
```
(`eslint` must stay below v9 for compatibility with some `eslint-plugin` packages)

src/livekit/openIDSFU.ts
toger5: I like it.

src/rtcSessionHelpers.ts
toger5: I think we need both. We always want to send the hangup action but also need to await the setAlwaysOnScreen.

I opted for try catch because the formatter otherwise puts everything in its own line:
```js
  await widget.api
    .setAlwaysOnScreen(false)
    .catch((e) =>
      logger.error("Failed to set call widget `alwaysOnScreen` to false", e),
    );
```
so the try catch block syntax is easier to read. imo

AndrewFerr (approved):

PR #2608, opened 2024-09-03
Fix rageshakes

robintown:
We were relying on deprecated APIs that are not supported when using Rust crypto. Since this entire file was copied and pasted from matrix-react-sdk originally, I just copied and pasted some of its more recent code in.


fkwp (approved):
lgtm

PR #2611, opened 2024-09-05
Test header components

robintown:


robintown:
https://github.com/element-hq/compound-web/pull/246

PR #2612, opened 2024-09-06
Fkwp/change license

fkwp:


neilisfragile (commented):
LGTM other than removing references to 'The New Vector...'

PR #2614, opened 2024-09-06
Test a few React components

robintown:
Since so much of this code base is made up React components, we do need to test them, and I wanted to see how easy it would be to get some automated accessibility tests with [axe](https://github.com/dequelabs/axe-core/) up and running. Along the way I discovered the reason why we were seeing funny errors in our tests (it was because I had turned off Vitest's test isolation option), and the tests I wrote also caught a [real accessibility issue](https://github.com/element-hq/compound-web/pull/246) in Compound Web.

### DIFF
--- a/src/Modal.tsx
+++ b/src/Modal.tsx
@@ -98,6 +98,8 @@ export const Modal: FC<Props> = ({
               styles.drawer,
               { [styles.tabbed]: tabbed },
             )}
+            // Suppress the warning about there being no description; the modal
+            // has an accessible title
             aria-describedby={undefined}
             {...rest}
           >
@@ -121,6 +123,8 @@ export const Modal: FC<Props> = ({
           <DialogOverlay
             className={classNames(overlayStyles.bg, overlayStyles.animate)}
           />
+          {/* Suppress the warning about there being no description; the modal
+          has an accessible title */}
           <DialogContent asChild aria-describedby={undefined} {...rest}>
             <Glass
               className={classNames(
--- a/src/Modal.tsx
+++ b/src/Modal.tsx
@@ -98,6 +98,7 @@ export const Modal: FC<Props> = ({
               styles.drawer,
               { [styles.tabbed]: tabbed },
             )}
+            aria-describedby={undefined}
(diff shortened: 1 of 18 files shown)


vitest.config.js
```diff
@@ -12,7 +12,6 @@ export default defineConfig((configEnv) =>
             classNameStrategy: "non-scoped",
           },
         },
-        isolate: false,
```
toger5, line 15: That very much explains what I have observed.

src/Modal.tsx
```diff
@@ -98,6 +98,7 @@ export const Modal: FC<Props> = ({
               styles.drawer,
               { [styles.tabbed]: tabbed },
             )}
+            aria-describedby={undefined}
```
toger5, line 94: Why do we need this?

src/tile/SpotlightTile.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+          maximised={false}
+          expanded={false}
+          onToggleExpanded={() => {}}
+          showIndicators
+        />,
+      );
+      expect(await axe(container)).toHaveNoViolations();
+      // Name should be visible
+      screen.getByText("Alice");
+    },
```
toger5: We could add a lot more expectations here:
 - avatar url
 - expanded
 - indicators
 ...

src/utils/test.ts
```diff
@@ -58,3 +65,67 @@ export function withTestScheduler(
     }),
   );
 }
+
+export async function withLocalMedia(
```
toger5, line 83: Would it make sense to not require tests (ignore test utils in codecov) on our test utils?

src/tile/SpotlightTile.test.tsx
robintown: Right, I've added some more expectations now. I wasn't able to test the specific avatar URL, as that requires a full Matrix client setup, and the indicators are (intentionally, IIRC?) excluded from the accessibility tree. But, there were still some useful things to test, and I actually caught a couple more accessibility issues this way!

PR #2616, opened 2024-09-09
Include the hostname of where EC is running in rageshakes

hughns:
Or, are we excluding this for privacy reasons? I would be surprised if so given that hostnames are included in other places along with the full Matrix ID.


toger5 (approved):
I think it was mostly not included since it is already part of the matrix.id. But having it here explicitly makes things easier.

PR #2618, opened 2024-09-10
Add button to remove call from recents

### LINKED ISSUE #2243: Remove entries from previous call list
https://call.element.io/ shows a long list of previous calls, and I'd like to delete specific entries from that list.

The list of previous rooms is fantastic for recurring calls that use the same room, but outdated rooms clutter the list over time.

Have a delete button on each room entry.

Make a list of all rooms I still need, delete cookies/storage and reopen all still needed

### Additional context

_No response_

Johennes:
This adds a button on each recent call tile to allow removing it.

Fixes: #2243

Screenshot 2024-09-10 at 13 31 44

The trash can `IconButton` from compound is the least obtrusive option I could think of. It could be made even less obtrusive by only displaying it on hover but I wasn't sure if that's really needed.

I didn't set the `destructive` property because while semantically true, it makes the button very prominent, especially when you have several tiles on the screen. Maybe `destructive={true}` could be combined with only showing the button on hover though.

Screenshot 2024-09-10 at 13 37 59

One small annoyance is that because the `IconButton` is inside the `Link`, the browser will show the link target when hovering the button. This felt better than reducing the tile's click target though.


fkwp:
cool -- really nice. @amshakal wdyt ?

amshakal:
Oh hello there! :)
This is definitely a desirable feature. I was wondering if we could change the icon to the one shared below. The reason for that is to make it more cohesive with the rest of the element call designs. Happy to provide the icon in any file format that might be needed. SVG perhaps?
Screenshot 2024-09-10 at 2 15 42 PM

Close button

Johennes:
👋🙂 

> This is definitely a desirable feature. I was wondering if we could change the icon to the one shared below. The reason for that is to make it more cohesive with the rest of the element call designs. Happy to provide the icon in any file format that might be needed. SVG perhaps?

This looks like the close icon from compound. Just pushed a change to use that instead of the trash can. How does that look?

Screenshot 2024-09-10 at 16 22 10

amshakal:
Perfect 👌👌👌

src/home/CallList.tsx
```diff
@@ -9,7 +9,10 @@ import { Link } from "react-router-dom";
 import { MatrixClient } from "matrix-js-sdk/src/client";
 import { RoomMember } from "matrix-js-sdk/src/models/room-member";
 import { Room } from "matrix-js-sdk/src/models/room";
-import { FC } from "react";
+import { FC, useCallback, MouseEvent } from "react";
+import { t } from "i18next";
```
toger5: lets use the `const {t} = useTranslation()` hook here instead to have it reactive.

src/home/CallList.tsx
```diff
@@ -55,8 +58,16 @@ interface CallTileProps {
...
 }
 
-const CallTile: FC<CallTileProps> = ({ name, avatarUrl, room }) => {
+const CallTile: FC<CallTileProps> = ({ name, avatarUrl, room, client }) => {
   const roomEncryptionSystem = useRoomEncryptionSystem(room.roomId);
+  const onRemove = useCallback(
+    (e: MouseEvent) => {
+      e.stopPropagation();
+      e.preventDefault();
+      void client.leave(room.roomId);
```
toger5: This works since we listen to room member changes and update the list reactivity.
This also means, that on click we send the leave but the room is still rendered in the list.
The user can trigger the leave multiple times by pressing the button quickly?
What would be really pretty is disabling the remove button right on click and enabling it if an error occurred:
we would need to introduce a `[leaving, setLeaving] = useState(false)`
```suggestion
      setLeaving(true);
      client.leave(room.roomId).catch(()=>setLeaving(false));
```
And use leaving to change the IconButtons Enabled state. (If you want to go really fancy you can even change the transparency of the `CallTile`.
But this might want to get feedback from design @amshakal.
Disabling the button is almost non controversial imo.

toger5 (changes requested):
Very nice addition.
This is almost ready. Once the `useTranslation` change is done I can approve it.
The other change is optional but I would be really happy and thankful if invest the time because you also think its an improvement.

src/home/CallList.tsx
Johennes: Ah, I fell for the code completion in my editor there. 🤦‍♂️

src/home/CallList.tsx
Johennes: Good call. I've went a little further than this because the link itself has the same problem and is still clickable while leaving. Unfortunately, the `Link` component doesn't appear to support disabling so I've conditionally replaced it with a span. For the disabled look of the link, I simply reused the selection style (80% opacity).

toger5 (approved):

PR #2621, opened 2024-09-11
Give user feedback if the rageshake submission failed

hughns:
Otherwise it silently swallows errors like 50x.

The design isn't nice and the user might need to scroll to see the whole message.

image

n.b. the screenshot doesn't show the wording but does show the revised location of the error message.


fkwp (approved):
lgtm

PR #2622, opened 2024-09-11
Rageshake logging improvements

hughns:
Capture MatrixRTC related prefixed/child loggers from js-sdk.
Add explicit prefix to livekit log entries.


fkwp (approved):
lgtm

PR #2623, opened 2024-09-11
Intercept matrix_sdk logging via console and include in rageshake

hughns:


src/settings/rageshake.ts
```diff
@@ -467,13 +467,37 @@ declare global {
...
+  // this is nasty, but no logging hooks are provided
+  (
+    ["trace", "debug", "info", "warn", "error"] as (
+      | "trace"
+      | "debug"
+      | "info"
+      | "warn"
+      | "error"
+    )[]
+  ).forEach((level) => {
```
AndrewFerr: This is slightly less nasty:
```suggestion
  [
    "trace" as const,
    "debug" as const,
    "info" as const,
    "warn" as const,
    "error" as const,
  ].forEach((level) => {
```
And thank https://typescript-eslint.io/rules/prefer-as-const/ for teaching me about "as const" :smiley:

src/settings/rageshake.ts
```diff
@@ -467,13 +467,37 @@ declare global {
...
+    if (!window.console[level]) return;
+    const prefix = `${level.toUpperCase()} matrix_sdk`;
+    const originalMethod = window.console[level];
```
AndrewFerr: Nitpick, rearrange some lines to avoid a redundant lookup on `window.console`
```suggestion
    const originalMethod = window.console[level];
    if (!originalMethod) return;
    const prefix = `${level.toUpperCase()} matrix_sdk`;
```

src/settings/rageshake.ts
```diff
@@ -467,13 +467,37 @@ declare global {
...
+    window.console[level] = (...args): void => {
+      originalMethod(...args);
+      if (typeof args[0] === "string" && args[0].startsWith(prefix)) {
+        global.mx_rage_logger.log(LogLevel[level], "matrix_sdk", ...args);
```
AndrewFerr, line 497: Is passing the entire `args` array intentional, or should `args[0]` (i.e. the argument with the prefix) be excluded?

AndrewFerr (changes requested):

src/settings/rageshake.ts
hughns: This is intended. I've pushed some comments to try and make this clearer. 🤞

AndrewFerr (approved):

PR #2624, opened 2024-09-11
Replace typography components with Compound components

robintown:


toger5:
This seems to be a good place to also tackle this inconsistency:
image
Screenshot from 2024-09-12 10-47-25

Otherwise it looks good (but have only done a quick review yet)

src/button/Link.module.css
```diff
@@ -0,0 +1,10 @@
+/*
+Copyright 2024 New Vector Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only
+Please see LICENSE in the repository root for full details.
+*/
+
+.external {
+  color: inherit;
+}
```
toger5: Why is this needed. What is the color property set to otherwise.

Would a comment make sense here?

src/button/Link.tsx
```diff
@@ -38,15 +44,32 @@ export function useLink(
...
-  { to, ...props },
+  { to, state, ...props },
   ref,
 ) {
-  const [path, onClick] = useLink(to);
+  const [path, onClick] = useLink(to, state);
   return <CpdLink ref={ref} {...props} href={path} onClick={onClick} />;
 });
+
+export const ExternalLink = forwardRef<
```
toger5, line 65: I think a guiding comment/docstring here clarifying in what cases this should be used also would be nice.

Or maybe even a docstring for all link components listing what other optiosn there are so the devs stumble over a docsting that mentions that there are other types they might want to/need to consider.

toger5 (approved):
This is great.
definitely cleans up the code quiet a bit!
Ready to merge as is. Not sure you consider the changes I propose (explanatory comments only) as worth it.

robintown:
> This seems to be a good place to also tackle this inconsistency

I think those two footers are different for a valid reason: in the second situation, the user is prompted to join the call as a guest, and the question of "what if I don't have an account yet?" doesn't become relevant until after the app has suggested that you might want to log in.

What would you suggest they look like?

toger5:
I was only referring to the styling of the links: they both are internal app links with which both trigger a simple reroute but one uses the Link DOM element where the other uses the new standardized compound link component (or at least that is what it looks like)

I am suggesting to make both views use the bold+underline styling instead of the green look.


I agree the overall layout (question first or in the second line) make total sense to be different.

robintown:
Ah, I believe I fixed that in this PR then.

toger5:
> Ah, I believe I fixed that in this PR then.

Are you sure. I created the screenshots with the netlify build of this PR.



compare: https://pr2624--element-call.netlify.app/login

with: https://pr2624--element-call.netlify.app/ (after clearing local store and cookies)

PR #2625, opened 2024-09-12
Test useTheme hook

toger5:


src/useTheme.test.ts
```diff
@@ -0,0 +1,122 @@
...
+}));
+
+describe("useTheme", () => {
+  let originalClassList: DOMTokenList;
+  beforeEach(() => {
+    // Save the original classList so we can restore it later
+    originalClassList = document.body.classList;
+
+    vi.spyOn(document.body.classList, "add").mockImplementation(vi.fn());
+    vi.spyOn(document.body.classList, "remove").mockImplementation(vi.fn());
```
robintown: Generally `.mockImplementation(() => {})` or `.mockImplementation(vi.fn())` is a no-op, so these can likely be

```suggestion
    vi.spyOn(document.body.classList, "add");
    vi.spyOn(document.body.classList, "remove");
```

src/useTheme.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    vi.spyOn(document.body.classList, "item").mockImplementation(() => null);
```
robintown:
```suggestion
    vi.spyOn(document.body.classList, "item").mockReturnValue(null);
```

src/useTheme.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    expect(originalClassList.remove).toHaveBeenCalledWith(
+      "cpd-theme-light",
+      "cpd-theme-dark",
+      "cpd-theme-light-hc",
+      "cpd-theme-dark-hc",
+    );
+    expect(originalClassList.add).toHaveBeenCalledWith("cpd-theme-dark-hc");
+  });
+
+  test("should apply light-high-contrast theme when theme is set to light-high-contrast", () => {
```
robintown: Non-blocking, but this and the above tests are very similar, so `test.each` could make them a lot shorter

robintown (approved):

PR #2629, opened 2024-09-12
Test CallViewModel

robintown:
This adds tests for a couple of the less trivial bits of code in CallViewModel. Testing them helped me uncover why focus switches still weren't being smooth! (It was because I was using RxJS's [`sample`](https://rxjs.dev/api/index/function/sample) operator when I really wanted [`withLatestFrom`](https://rxjs.dev/api/operators/withLatestFrom).)

Based on https://github.com/element-hq/element-call/pull/2614


src/utils/test.ts
```diff
@@ -49,3 +62,139 @@ export function withTestScheduler(
     }),
   );
 }
+
+export function mockMember(member: Partial<RoomMember>): RoomMember {
```
toger5: We were discussing if we want to get those out of the js-sdk by moving it to an export.
I see how this would block this PR behind a js-sdk PR. Do you think a comment is a good idea to clarify that there might be a good chance to share code with the js-sdk here?

src/utils/test.ts
```diff
@@ -49,3 +62,139 @@ export function withTestScheduler(
...
+      return this;
+    },
+    removeListener() {
+      return this;
+    },
+    ...member,
+  } as RoomMember;
+}
+
+export function mockMatrixRoom(room: Partial<MatrixRoom>): MatrixRoom {
```
toger5, line 97: Can we have a generic `mockEventEmitter<T>(val: Partial<T>): T{}` and simplify those to:
```
export function mockMatrixRoom(room: Partial<MatrixRoom>): MatrixRoom {
    return mockEventEmitter(room);
}
```
(of course this is very much in conflict that this should be done in the js-sdk ... but still would shorten this a bit.)

src/utils/test.ts
```diff
@@ -49,3 +62,139 @@ export function withTestScheduler(
...
+
+export function mockRemoteParticipant(
+  participant: Partial<RemoteParticipant>,
+): RemoteParticipant {
+  return {
+    isLocal: false,
+    setVolume() {},
+    getTrackPublication: () =>
+      ({}) as Partial<RemoteTrackPublication> as RemoteTrackPublication,
+    on() {
```
toger5: This should also work:
```
export function mockRemoteParticipant( participant: Partial<RemoteParticipant>): RemoteParticipant {
    let emitterMock = mockEventEmitter(room);
    return {...emitterMock, isLocal: false, getTrack...}
}
```

src/tile/SpotlightTile.tsx
```diff
@@ -271,17 +284,20 @@ export const SpotlightTile = forwardRef<HTMLDivElement, Props>(
               targetWidth={targetWidth}
               targetHeight={targetHeight}
               intersectionObserver={intersectionObserver}
+              // This is how we get the container to scroll to the right media
+              // when the previous/next buttons are clicked: we temporarily
+              // remove all scroll snap points except for just the one media
+              // that we want to bring into view
               snap={scrollToId === null || scrollToId === vm.id}
+              aria-hidden={(scrollToId ?? visibleId) !== vm.id}
```
toger5, line 292: Oh i did not even know this is part of it! very nice

src/utils/test.ts
toger5: This is non blocking. (To merge this i think it is enough that the comment that we consider using js-sdk code here can also have a short sentence about a generic event emitter mock)

src/state/CallViewModel.ts
toger5, line 1: very nice you caught this!

toger5 (approved):
Thanks. This looks new and good to me.

Issue #2634, opened 2024-09-17
Potential race when knocking to join a call over federation


Half-Shot:
### Steps to reproduce

(N.B. I suspect this is more of a Synapse problem, but reporting at application level.

There are two users here. Alice is on Server A, and Bob is on Server B.

1. Alice creates a new call, and sends the link along to Bob
2. Bob requests to join the call (knocks)
3. Alice accepts and invites Bo,b
4. Bob sees an error about not being able to join
5. (workaround) Bob is able to refresh and immediately join the call.

#### What did you expect?

Upon being inviited, I should be joined to the call.

#### What happened instead?

![image](https://github.com/user-attachments/assets/73d5d61a-4af0-4159-b928-cf5d574e5ba7)

I suspect this is invites racing;

### Operating system

Arch Linux

### Browser information

Firefox

### Will you send logs?

No

Half-Shot:
I strongly suspect this is https://github.com/element-hq/synapse/issues/15012

Half-Shot:
#2637 reportedly makes this less of a problem, but this causes some delay in joining a room so it would be good to fix upstream.

PR #2637, opened 2024-09-18
Retry a join on invite if the response was M_FORBIDDEN

Half-Shot:
Workaround for #2634 

This simply retries a join if we got a M_FORBIDDEN from joining a room we know we were invited to. This is likely due to the delay mention in the linked issue of the linked issue. We attempt to join a few times, but eventually give up if the server doesn't yield.

The full solution is to fix Synapse so that it doesn't send an invite down sync until the room is ready to be joined to.


fkwp (approved):
lgtm

PR #2638, opened 2024-09-18
Bump js-sdk

hughns:
This brings in the change to only send the current/latest encryption keys when running in per-participant end-to-end encrypted media mode.

https://github.com/matrix-org/matrix-js-sdk/compare/169e8f86139111574a3738f8557c6fa4b2a199db...414ac9d8cc28330718236b90ad67a1507e146932

The pinned version is to resolve an eslint error: https://github.com/eslint/eslint/discussions/17215

```
Error: require() of ES Module /home/runner/work/element-call/element-call/node_modules/strip-ansi/index.js from /home/runner/work/element-call/element-call/node_modules/eslint/lib/cli-engine/formatters/stylish.js not supported.
Instead change the require of index.js in /home/runner/work/element-call/element-call/node_modules/eslint/lib/cli-engine/formatters/stylish.js to a dynamic import() which is available in all CommonJS modules.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,5 +109,8 @@
     "vite-plugin-html-template": "^1.1.0",
     "vite-plugin-svgr": "^4.0.0",
     "vitest": "^2.0.0"
+  },
+  "resolutions": {
+    "strip-ansi": "6.0.1"
   }
 }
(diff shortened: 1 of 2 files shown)


fkwp (approved):
lgtm

PR #2644, opened 2024-09-19
Make the cookie policy link show up again

robintown:
For some reason `yarn i18next` didn't catch that the string had to be changed? I removed the string and regenerated it, and now the cookie policy is back.


AndrewFerr (commented):
Does this need to be updated for other languages too? `git grep -F <5>` shows other translations for this that still use `<5>`.

robintown:
True, normally the workflow when a string has changed would be to wait for Localazy to ask translators for updated strings and remove those translations that haven't been updated, but in this case I agree that it makes sense to update the translation files directly.

AndrewFerr (approved):

PR #2646, opened 2024-09-23
Use non-deprecated method of re-processing MatrixRTCSession keys

hughns:
As per https://github.com/matrix-org/matrix-js-sdk/pull/4423 the key can be undefined and so we should handle this rather than waiting for SubtleCrypto.importKey() to fail.


src/e2ee/matrixKeyProvider.ts
```diff
@@ -41,7 +41,9 @@ export class MatrixKeyProvider extends BaseKeyProvider {
       encryptionKeys,
     ] of this.rtcSession.getEncryptionKeys()) {
       for (const [index, encryptionKey] of encryptionKeys.entries()) {
```
AndrewFerr: As it is currently written (and is sufficiently captured in type hints), the array returned by `getEncryptionKeys` is a [sparse array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Indexed_collections#sparse_arrays), which contains empty slots that are skipped by array iteration methods.

Thus, instead of using a `for...of` loop and having to check for `undefined`, this could simply use `forEach`:

```suggestion
      encryptionKeys.forEach((encryptionKey, index) => {
```

Note that this will work only if the array returned by `getEncryptionKeys` remains typed as `Array<Uint8Array>` instead of `Array<Uint8Array | undefined>`.

AndrewFerr (dismissed):

src/e2ee/matrixKeyProvider.ts
hughns: Well spotted. 👍 However, this is now redundant given the refactor from https://github.com/matrix-org/matrix-js-sdk/pull/4427

src/e2ee/matrixKeyProvider.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    const session = mockRTCSession();
+
+    keyProvider.setRTCSession(session);
+
+    expect(session.on).toHaveBeenCalledWith(
+      MatrixRTCSessionEvent.EncryptionKeyChanged,
+      expect.any(Function),
+    );
+    expect(session.off).not.toHaveBeenCalled();
+  });
```
toger5, line 43: I suppose we test that updating session (join/leave/receive new keys) also trigger a session `EncryptionKeyChanged` emit somewhere else?

toger5 (approved):
Very nice this has gotten simpler.

Added a comment since I am not too sure about our E2EE test coverage.

src/e2ee/matrixKeyProvider.test.ts
hughns: Yes, these would be at https://github.com/matrix-org/matrix-js-sdk/blob/develop/spec/unit/matrixrtc/MatrixRTCSession.spec.ts

PR #2648, opened 2024-09-26
Make it possible to use the JWT service with a local homeserver

robintown:
I recently tried to use the dev configs with a local homeserver and found that the configuration was not compatible with this. We need to relax the network isolation and TLS cert verification requirements for fully local development to be possible.

### DIFF
--- a/backend-docker-compose.yml
+++ b/backend-docker-compose.yml
@@ -32,10 +32,10 @@ services:
     # any…)
     network_mode: host
     # ports:
-    #   - "7880:7880"
-    #   - "7881:7881"
-    #   - "7882:7882"
-    #   - "50100-50200:50100-50200"
+    #   - "7880:7880/tcp"
+    #   - "7881:7881/tcp"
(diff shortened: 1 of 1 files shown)


backend-docker-compose.yml
```diff
@@ -23,11 +27,15 @@ services:
...
-      - "50100-50200:50100-50200"
+    # The SFU seems to work far more reliably when we let it share the host
+    # network rather than opening specific ports (but why?? we're not missing
+    # any…)
+    network_mode: host
+    # ports:
+    #   - "7880:7880"
+    #   - "7881:7881"
+    #   - "7882:7882"
+    #   - "50100-50200:50100-50200"
```
fkwp:
```suggestion
    #   - "7880:7880/tcp"
    #   - "7881:7881/tcp"
    #   - "7882:7882/tcp"
    #   - "50100-50200:50100-50200/upd"
```

fkwp (approved):

PR #2660, opened 2024-10-03
Set LiveKit keyring size to 256

hughns:
The default is 16 which doesn't match what we expect.

We will need to upgrade the LiveKit client before this merges because we need https://github.com/livekit/client-sdk-js/pull/1268.


fkwp (approved):

hughns:
n.b. i pulled the tests out of this and in to https://github.com/element-hq/element-call/pull/2646

toger5 (approved):
Thanks Hugh!

PR #2670, opened 2024-10-10
Keep tiles in a stable order

### LINKED ISSUE #2562: Grid view in v0.6.0 shifts people around constantly
### Steps to reproduce

With only 3-5 people in grid view, easily enough that can fit on a one screen, depending on whos talking people shift around constantly. It's very annoying and pointless as moving someone to the top has no real benefit.

#### What did you expect?
no shifting around without reason (someone leaves etc).
#### What happened instead?
everyong is moving around constantly.

### Operating system

All OSes did it

### Browser information

several

### URL for webapp

private server, v0.6.0

### Will you send logs?

No

robintown:
This introduces a new layer of abstraction on top of MediaViewModel: TileViewModel, which gives us a place to store data relating to _tiles_ rather than their media, and also generally makes it easier to reason about tiles as they move about the call layout. I have created a class called TileStore to keep track of these tiles.

This allows us to swap out the media shown on a tile as the spotlight speaker changes, and avoid moving tiles around unless they really need to jump between the visible/invisible regions of the layout.

Closes https://github.com/element-hq/element-call/issues/2562


hughns (approved):
There is a lot going on in this PR. I've reviewed it as much as I think I reasonably can. I understand it has been reasonably tested with real usage. As such I think it is good enough to merge.

It does need the merge conflicts resolving first.

PR #2681, opened 2024-10-18
Make the volume slider less silly

robintown:
Previously, dragging it all the way to the left would *not* mute the participant but rather bottom out at 10% volume, and people have found this unintuitive. Let's make it less silly by giving the slider a range of 0% to 100%, and making the mute toggle button have the same effect as dragging the slider to zero. When unmuting, it will reset to the last non-zero "committed" volume, similar to how the volume sliders in desktop environments work.

### DIFF
--- a/src/Slider.tsx
+++ b/src/Slider.tsx
@@ -16,6 +16,7 @@ interface Props {
   label: string;
   value: number;
   onValueChange: (value: number) => void;
+  onValueCommit?: (value: number) => void;
   min: number;
   max: number;
   step: number;
@@ -30,6 +31,7 @@ export const Slider: FC<Props> = ({
   label,
   value,
   onValueChange: onValueChangeProp,
+  onValueCommit: onValueCommitProp,
   min,
   max,
   step,
@@ -39,12 +41,17 @@ export const Slider: FC<Props> = ({
     ([v]: number[]) => onValueChangeProp(v),
     [onValueChangeProp],
   );
+  const onValueCommit = useCallback(
+    ([v]: number[]) => onValueCommitProp?.(v),
+    [onValueCommitProp],
+  );
 
   return (
     <Root
       className={classNames(className, styles.slider)}
       value={[value]}
       onValueChange={onValueChange}
+      onValueCommit={onValueCommit}
       min={min}
       max={max}
       step={step}
--- a/src/Slider.tsx
+++ b/src/Slider.tsx
@@ -16,6 +16,12 @@ interface Props {
   label: string;
   value: number;
   onValueChange: (value: number) => void;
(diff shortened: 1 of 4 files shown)


src/state/MediaViewModel.ts
```diff
@@ -232,18 +235,45 @@ export class LocalUserMediaViewModel extends BaseUserMediaViewModel {
...
+          : event === "commit"
+            ? { ...state, committedVolume: state.volume }
+            : // Volume adjustment
+              event === 0
+              ? // Dragging the slider to zero should have the same effect as
+                // muting: reset the volume to the committed volume, as if it were
+                // never dragged
+                { ...state, muted: true, volume: state.committedVolume }
+              : { ...state, muted: false, volume: event },
+    ),
```
AndrewFerr: IMO using a `switch` statement reads better:
```suggestion
      (state, event) => {
        switch (event) {
        case "toggle mute":
          return { ...state, muted: !state.muted };
        case "commit":
          return { ...state, committedVolume: state.volume };
        case 0: // Volume adjustment
          // Dragging the slider to zero should have the same effect as
          // muting: reset the volume to the committed volume, as if it were
          // never dragged
          return { ...state, muted: true, volume: state.committedVolume };
        default:
          return { ...state, muted: false, volume: event };
        }
      }),
```

src/Slider.tsx
```diff
@@ -30,6 +31,7 @@ export const Slider: FC<Props> = ({
   label,
   value,
   onValueChange: onValueChangeProp,
+  onValueCommit: onValueCommitProp,
```
AndrewFerr, line 40: OOI where are `onValueChangeProp` & `onValueCommit` prop defined?

src/Slider.tsx
```diff
@@ -16,6 +16,7 @@ interface Props {
   label: string;
   value: number;
   onValueChange: (value: number) => void;
+  onValueCommit?: (value: number) => void;
```
AndrewFerr, line 25: Why does this need to be optional?

AndrewFerr (approved):

src/Slider.tsx
robintown: Mm, you're looking at a function parameter being destructured here. So they're defined right on these lines. It works like this:

```typescript
const foo = ({ value: inputValue }) => { // Function accepting an object with some entry named 'value'
  console.log(inputValue) // 'value' has been bound to the variable 'inputValue'
}

foo({ value: 999 }) // Logs '999'
```

src/Slider.tsx
robintown: Commented

PR #2686, opened 2024-10-24
Don't cause the local tile to disappear when joining

robintown:
Before you're connected to the SFU the local participant object will have the empty string as its ID. This changes to your actual ID once you've connected. Apparently I tried to fix this by forcing the local ID to always be the string 'local' but then I just forgot to use it correctly :upside_down_face:


fkwp (approved):
LGTM

PR #2690, opened 2024-10-28
Show speaking indicators in spotlight during screen sharing

### LINKED ISSUE #2689: Speaking indicators not shown when a screen share is in the spotlight
### Steps to reproduce

1. Share your screen
2. Go into spotlight

#### What did you expect?

When others speak, their tiles light up with speaking indicators

#### What happened instead?

No speaking indicators, because the idea was that in spotlight mode any speakers would get moved to the spotlight, so indicators are redundant. But that's not actually the case when there's a screen share.

### Operating system

NixOS unstable

### Browser information

Firefox 131.0.2

### URL for webapp

call.element.dev

### Will you send logs?

No

robintown:
Closes https://github.com/element-hq/element-call/issues/2689

### DIFF
--- a/src/state/CallViewModel.ts
+++ b/src/state/CallViewModel.ts
@@ -724,9 +724,9 @@ export class CallViewModel extends ViewModel {
           return l.spotlight[0] instanceof ScreenShareViewModel;
         // In expanded spotlight layout, the active speaker is always shown in
         // the picture-in-picture tile so there is no need for speaking
-        // indicators
+        // indicators. And in one-on-one layout there's no question as to who is
+        // speaking.
         case "spotlight-expanded":
-        // In one-on-one layout there's no question as to who is speaking
         case "one-on-one":
           return false;
         default:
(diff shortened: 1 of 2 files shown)


src/state/CallViewModel.ts
```diff
@@ -714,7 +714,21 @@ export class CallViewModel extends ViewModel {
...
-    map((l) => l.type !== "one-on-one" && !l.type.startsWith("spotlight-")),
+    map((l) => {
+      switch (l.type) {
+        case "spotlight-landscape":
+        case "spotlight-portrait":
+          // If the spotlight is showing the active speaker, we can do without
+          // speaking indicators as they're a redundant visual cue. But if
+          // screen sharing feeds are in the spotlight we still need them.
+          return l.spotlight[0] instanceof ScreenShareViewModel;
+        case "spotlight-expanded":
```
hughns, line 729: Why is -expanded treated differently from -landscape/-portrait?

robintown: Added a comment

hughns (changes requested):
Thank you. Looks good other than the lint error.

hughns (approved):

PR #2693, opened 2024-10-28
Update the strings on the join as guest screen

robintown:
Hugh pointed out that the button labeled "Join call now" on this screen is misleading, because it does not join you to the call but rather send you to the lobby. Meanwhile, the designs call for this screen to use the words "Join as guest" and "Continue", so let's go with that.


fkwp (approved):
lgtm

PR #2694, opened 2024-10-28
Add a button to switch the camera on mobile

robintown:


toger5:
I think it makes sense to immediately think about > 2 cameras. I think nowadays, the amount of ===2 camera phones is very limited.
But if it is intentionally split into two PR's that also makes sense to me.
My camera is muted after a device switch (using the current selector + more than 2 devices, android, firefox)
Probably worth another issue right?

toger5 (commented):
Can you add a screenshot?
I am not sure what icon and css it is using.
But it seems its adding another entry in the bottom button list.
It does not handle clicks on your video tile rigth?

robintown:
> I think it makes sense to immediately think about > 2 cameras. I think nowadays, the amount of ===2 camera phones is very limited.

I don't have an iPhone to test with so I'll need someone to report what the device names are (so I can get an idea for whether analyzing them for their facing direction is possible) and also what the expected behavior is when you switch cameras in say, Facetime.

toger5:
Android Samsung S21
Firefox:
 - Facing front: 3
 - Facing front: 1
 - Facing front: 0
 - Facing back: 2
 
 Chrome:
 -  camera2 1, facing front
 -  camera2 3, facing front
 -  camear2 2, facing back
 -  camera2 0, facing back

fkwp:
tested with Pixle 7 Pro -> nice

is localization an issue? Safari on iOS has a German translation:
- Frontkamera
- Rückseitige Dual-Weitwinkelkamera
- Rückseitige Ultra-Weitwinkelkamera
- Rückkamera

robintown:
Haha, localization is what I feared :D

fkwp:
nice: just tested on
- iOS and it works 
- iPadOS and it works
- Android Pixel 7 Pro and it works

toger5:
Tested on S21
It works well in chrome.
On Firefox it works but it is very easy to get into a "broken" state

If i set it to one of the other cameras in the device drawer settings menu i dont get a stream anymore.

Playing around with it more:
It seems that there is just one camera that breaks everything. I select it once and it does not work anymore.
If I only use index 0 and 3 its all good. i can select the camerea from the dropdown and via the swap button.
On chrome and FF.

toger5 (commented):
Two review comments:
 - Does it make sense to use the same logic we have for the mute button? Each press independent if the camera has updated already swapts the target state. We always wait until the camera has finished updating/(track has sucessfully changed) only then we start the next update in case the user has pressed the button as many times so the current state is not the target state?
 - should we remove the video device selector on mobile knowing that my phone can get into a "broken" state? I also dont see any benefit to choose a specific device on mobile especially if the device descriptions are so poor.
 We can entirely drop the settings tab on mobile. On desktop i do see value in selecting the exact webcam one wants to use.

robintown:
> Does it make sense to use the same logic we have for the mute button? Each press independent if the camera has updated already swapts the target state.

In the mute button case, I think this was motivated by the immediate feedback that we have to provide in the form of changing the button's state. We don't have to provide any such feedback for the switch camera button, so I'd prefer to avoid complicating it. Or is there another reason to try that?

> should we remove the video device selector on mobile knowing that my phone can get into a "broken" state?

I'm not sure what you mean here. The video device selector does update when the camera is switched.

toger5:
> I'm not sure what you mean here. The video device selector does update when the camera is switched.

I was considering to entirely remove the video tab from the settings. So that the only way to select a camera device is by the switch button

toger5:
> In the mute button case, I think this was motivated by the immediate feedback that we have to provide in the form of changing the button's state.

What does happen if i press the button twice (quickly)
Will it ignore the second click because its in a "wait for camera switch" mode?
I mostly wanted to make sure its not colliding with the first switch.

src/room/InCallView.tsx
```diff
@@ -488,37 +491,41 @@ export const InCallView: FC<InCallViewProps> = ({
...
         );
       }
-      buttons.push(<SettingsButton key="4" onClick={openSettings} />);
+      buttons.push(<SettingsButton key="settings" onClick={openSettings} />);
     }
 
     buttons.push(
       <EndCallButton
-        key="6"
+        key="end_call"
```
toger5, line 528: That is much better then numbers.

src/room/LobbyView.tsx
```diff
@@ -89,6 +98,61 @@ export const LobbyView: FC<Props> = ({
...
+  );
+
+  const localTrackOptions = useMemo(
+    () => ({
+      // The only reason we request audio here is to get the audio permission
+      // request over with at the same time. But changing the audio settings
+      // shouldn't cause this hook to recreate the track, which is why we
+      // reference the initial values here.
+      // We also pass in a clone because livekit mutates the object passed in,
+      // which would cause the devices to be re-opened on the next render.
```
toger5, line 118: Why is this moved to the lobby file?

toger5 (approved):
This looks good.
I left a non blocking comment.

Thanks!

robintown:
> I was considering to entirely remove the video tab from the settings. So that the only way to select a camera device is by the switch button

I was more asking about what you mean by 'broken', as the state seemed to stay accurate in my testing
 
> What does happen if i press the button twice (quickly)
> Will it ignore the second click because its in a "wait for camera switch" mode?
> I mostly wanted to make sure its not colliding with the first switch.

It appears to simply ignore the second click, yeah, which is an acceptable behavior here IMO

src/room/LobbyView.tsx
robintown: Previously the only thing that needed the video track was the VideoPreview itself, but now the LobbyView wants to render a button which accesses the track as well, so I hoisted the video track and the code for starting it up to this component

PR #2700, opened 2024-10-30
Show encryption key status from LiveKit

hughns:
This is missing unit/automated tests, but is behaving okay with manual testing.


hughns:
Self approving after discussion with @fkwp.

PR #2701, opened 2024-10-30
Make video tiles be based on MatrixRTC member not LiveKit participants

toger5:
Currently a tile is shown whenever livekit has a participant.
This is wrong in a couple of situations:
 - If there are members in the livekit room that are not in the matrix room. (This can happen since we do not have any control over what the lk server is doing)
 - If a user is joined in matrix but they struggle to create a livekit connection that is a state we want to show to the user but currently we dont know about that. Since without a livekit participant we dont have any tile.
 - If someone leaves but still continue connects to livekit, we can remove their tile.
 - In an encrypted room, users without a matrix rtc membership cannot get the call keys, so their streams will be irrelevant anyways + they cannot do anything with the stream from others. -> there is no reason to add them as tiles.

image

image


src/state/CallViewModel.ts
```diff
@@ -668,8 +668,11 @@ export class CallViewModel extends ViewModel {
                             (
                               localUserMedia?.vm as LocalUserMediaViewModel
                             ).alwaysShow.pipe(
-                              map((alwaysShow) =>
-                                alwaysShow ? localUserMedia : undefined,
+                              map(
+                                (alwaysShow) =>
+                                  (alwaysShow
+                                    ? localUserMedia?.vm
+                                    : undefined) ?? null,
```
toger5: Oh wow that one is new right?
Its the same issue as we ran into with the other case but I dont remember it from our test-fix-session.

hughns: > Oh wow that one is new right? Its the same issue as we ran into with the other case but I dont remember it from our test-fix-session.

Yes. The compiler caught it, but our test cases did not.

hughns:
One "rough edge" on this is that when a remote participant leaves the call, they are shown as "Waiting for media..." whilst the animation to remove the tile from the screen is playing.

src/state/CallViewModel.ts
```diff
@@ -401,49 +435,86 @@ export class CallViewModel extends ViewModel {
     this.remoteParticipants,
     observeParticipantMedia(this.livekitRoom.localParticipant),
     duplicateTiles.value,
-    // Also react to changes in the list of members
-    fromEvent(this.matrixRoom, RoomStateEvent.Update).pipe(startWith(null)),
+    // Also react to changes in the MatrixRTC session list:
```
toger5:
```suggestion
    // Also react to changes in the MatrixRTC session list.
    // The session list will also be update if a room membership changes.
    // No additional RoomState event listener needs to be set up.
```

src/state/CallViewModel.ts
```diff
@@ -401,49 +435,86 @@ export class CallViewModel extends ViewModel {
...
+                  prevMedia ??
                     new UserMedia(
-                      userMediaId,
+                      indexedMediaId,
                       member,
-                      p,
+                      participant,
                       this.encryptionSystem,
                       this.livekitRoom,
+                      livekitParticipantId === "local",
```
toger5: It seems like this is redundant data?
`participant` is passed to UserMedia,
we set `livekitParticipantId` at the same time we set `participant`
```
              if (
                rtcMember.sender === room.client.getUserId()! &&
                rtcMember.deviceId === room.client.getDeviceId()
              ) {
                livekitParticipantId = "local";
                participant = localParticipant;
...
```
If `livekitParticipantId` is `"local"`, participant is not undefined since `localParticipant` is not optional.
Or do we have a type mistake where `{ participant: localParticipant },`
can be optional?

toger5: I do remember that we had a missing local participant but looking through the types here that seems impossible.

src/state/CallViewModel.ts
```diff
@@ -570,18 +635,34 @@ export class CallViewModel extends ViewModel {
...
+                                (alwaysShow) =>
+                                  (alwaysShow
+                                    ? localUserMedia?.vm
+                                    : undefined) ?? null,
+                              ),
+                            ) ?? of(null)
+                          );
+                        }),
+                      )
+                  : of(null),
```
toger5: I feel like this is at the edge of easy to grasp. It is very little and condense code for what its doing but we might want to consider:
 - returning a object with named properties
 - Be less LOC efficient and create some intermediate methods to:
 ```
 this.mediaItems.pipe(
  switchMap((mediaItems) => {
    const localUserMedia = mediaItems.find(
      (m) => m.vm instanceof LocalUserMediaViewModel,
    ) as UserMedia | undefined;
    return (
      (
        localUserMedia?.vm as LocalUserMediaViewModel
      ).alwaysShow.pipe(
        map(
          (alwaysShow) =>
            (alwaysShow
              ? localUserMedia?.vm
              : undefined) ?? null,
        ),
      ) ?? of(null)
    );
  }),
)
```

All of this could be wrapped in: `observePreferredPipItem(...)`
and we would just end up with:
```
([
  this.spotlightSpeaker.pipe(
    map((speaker) => (speaker ? [speaker] : [])),
  ),
  this.spotlightSpeaker.pipe(
    switchMap((speaker) =>
      speaker
        ? speaker.local
          ? of(null)
          : observePreferredPipItem(this)
    ),
  ),
] as const)
```

File: src/state/MediaViewModel.ts [hunk 5]
```diff
@@ -243,69 +250,72 @@ abstract class BaseMediaViewModel extends ViewModel {
...
-      distinctUntilChanged(),
-      shareReplay({ bufferSize: 1, refCount: false }),
+        (a?.publication?.isEncrypted === false ||
+          v?.publication?.isEncrypted === false),
+    ).pipe(this.scope.state());
+
+    this.encryptionStatus = this.participant.pipe(
+      switchMap((participant): Observable<EncryptionStatus> => {
+        if (!participant) {
+          return of(EncryptionStatus.Connecting);
```
toger5, line 265: If we include the `Connecting` state, I am wondering if `EncryptionStatus` is the right type name.
I think this could be considered the overall state of an rtc member once they have an rtc session. `MediaState`, `rtcMemberState`...?
```
  Connecting,
  Okay,
  KeyMissing,
  KeyInvalid,
  PasswordInvalid,
```
On the base rtc membership we have: the lk participant, the media key used by the participant and the keys we have received over matrix.
Thos are the possible states:
lk | matrix key | matches lk key | implies state
--- | --- | --- | ---
false | true/false | true/false | `Connecting`
true | false | false/false | `KeyMissing`
true | true | false | `KeyInvalid`
true | true | true | `Okay`

Looking at it like this I am wondering if the logic becomes much simpler if we first fo 3 bool observables (lk, matrix key, matches lk key) (or 1 observable with an object that has those 3 props) and then build one state observable based on those precomputed observables. The individual bool observables can also be helpful I think:
```
let gotKey = encryptionState === KeyInvalid || encryptionState === Okay
```
Would be used otherwise for checking for a received matrix key.

src/state/MediaViewModel.ts
```diff
@@ -421,6 +438,8 @@ export class LocalUserMediaViewModel extends BaseUserMediaViewModel {
  * A remote participant's user media.
  */
 export class RemoteUserMediaViewModel extends BaseUserMediaViewModel {
+  public readonly local = false;
```
toger5: We do only set this because of the screen share case right?
For remote and local mv's it would be enought to do an instance of check?

src/tile/SpotlightTile.tsx
```diff
@@ -210,7 +212,9 @@ export const SpotlightTile = forwardRef<HTMLDivElement, Props>(
     const ref = useMergedRefs(ourRef, theirRef);
     const maximised = useObservableEagerState(vm.maximised);
     const media = useObservableEagerState(vm.media);
-    const [visibleId, setVisibleId] = useState(media[0].id);
+    const [visibleId, setVisibleId] = useState<string | undefined>(
+      media.length > 0 ? media[0].id : undefined,
```
toger5: Any reason why this would not work?
```suggestion
      media[0]?.id,
```

src/tile/MediaView.module.css
```diff
@@ -74,9 +74,9 @@ unconditionally select the container so we can use cqmin units */
     calc(var(--media-view-border-radius) - var(--cpd-space-3x))
   );
   display: grid;
-  grid-template-columns: 1fr auto;
+  grid-template-columns: 30px 1fr 30px;
```
toger5, line 77: This should be one of the compound spaces i think?
8x most likely (32px)
```suggestion
  grid-template-columns: var(--cpd-space-8x) 1fr var(--cpd-space-8x);
```

toger5:
> One "rough edge" on this is that when a remote participant leaves the call, they are shown as "Waiting for media..." whilst the animation to remove the tile from the screen is playing.

This is fixed indirectly using the animated loading border,

src/tile/SpotlightTile.tsx
hughns: No. I always forget that JS has this behaviour rather than throwing an array index out of bounds exception.

src/tile/MediaView.module.css
hughns: This value is actually linked to the size of the reaction in top left and the options button in bottom right.

image

Top left is hard coded to 30px:
https://github.com/element-hq/element-call/blob/133dc26642e9e71c33634223f92fc1691d1ec15b/src/reactions/ReactionIndicator.module.css#L59

Bottom right appears to be hard coded to 20px (+ some margin):
https://github.com/element-hq/element-call/blob/133dc26642e9e71c33634223f92fc1691d1ec15b/src/tile/GridTile.tsx#L169

I'm very open to suggestions on how to properly handle this? I can do the compound spacing as suggested?

src/state/MediaViewModel.ts
hughns: Given that this value isn't currently being used, I think we can move this clean up to a new PR at a later date.

src/tile/MediaView.module.css
toger5: Oh.
I see how we then want that in favor of the spacing...

toger5: Lets not do this in here and try to remember bringing it up in a team sync.

src/state/MediaViewModel.ts
toger5: What about renaming `encryptionStatus` to `rtcMediaState`

src/state/MediaViewModel.ts
hughns: So, I tried doing this and ended up with a load of, IMHO, unintuitive instanceofs.

Which led me to want to do this:

```
export function isLocal(vm: BaseMediaViewModel): boolean {
  if (vm instanceof LocalUserMediaViewModel) {
    return true;
  } else if (vm instanceof RemoteUserMediaViewModel) {
    return false;
  } else if (vm instanceof ScreenShareViewModel) {
    return vm.local;
  }

  throw new Error("Unknown user media view model");
}
```

The end result would be that `vm.local` becomes `isLocal(vm)` which is basically the same.

What I have done in 3d0bfaefe8743bd90c9b0267faf3d490219f6496 is to make the purpose of `BaseMediaViewModel.local` be explicit (forcing it to be required for subclasses) and use an instanceof for the `BaseUserMediaViewModel` case.

src/state/CallViewModel.ts
hughns: > I feel like this is at the edge of easy to grasp. It is very little and condense code for what its doing but we might want to consider:
> 
> * returning a object with named properties

Done in 3d0bfaefe8743bd90c9b0267faf3d490219f6496.

hughns: > > I feel like this is at the edge of easy to grasp. It is very little and condense code for what its doing but we might want to consider:
> > 
> > * returning a object with named properties
> 
> Done in [3d0bfae](https://github.com/element-hq/element-call/commit/3d0bfaefe8743bd90c9b0267faf3d490219f6496).

Having done this an obvious thing to do was to split `spotlightAndPip` directly into `spotlight` and `pip`. I don't know why we want them combined?

So, I did that in 33f398ddcf09f2b183abd43671a56a17098aca1b.

Then, I tried replacing some of the nested ternaries with if statements. I think this is more legible and would provide a nice place to add comments if appropriate.

This is in 7fdcbb32922c62c1413b5db47666fe0ec1718f86.

I think this might be good enough, now?

src/state/CallViewModel.ts
hughns: I think you are right and have changed it in 7fdcbb32922c62c1413b5db47666fe0ec1718f86.

The might be something in future where we want to show a call layout before we have a local participant. But, we can tackle that later.

hughns:
> > One "rough edge" on this is that when a remote participant leaves the call, they are shown as "Waiting for media..." whilst the animation to remove the tile from the screen is playing.
> 
> This is fixed indirectly using the animated loading border,

The "animated loading border" is now in https://github.com/element-hq/element-call/pull/2858 for consideration.

hughns:
This is ready for the next round of review.

hughns:
With regards to the visual design/representation: I think we should get it merged as-is (with the "Waiting for media...") and then tackle the representation before we do a release.

src/state/CallViewModel.ts
```diff
@@ -576,34 +641,61 @@ export class CallViewModel extends ViewModel {
...
+                    return null;
+                  }),
+                );
+              }),
+            );
+          }),
+        );
+      }),
       this.scope.state(),
     );
```
toger5: Trying to follow that code I noticed that I am not a big fan of nested switchMaps.
Its hard to follow when things update:
 - the main switch map is done on Screenshares so each screenShare update does the condition and also reinitializes the observable generated by `sportlightSpeaker.pipe`.
 -  which then can give us the newly reconstructed of(null) or another observable based on `this.mediaItems.pipe`.

 It does make sense when thinking it through. I still would like to propose this alternative where we first combine our dependent observables.
 I like to see the clear list:`this.screenShares, this.spotlightSpeaker, this.mediaItems` of dependent observables in one place.
 It of course now will execute the whole logic more often. `screenShares.length > 0` will also be checked if mediaItems change.
 But the logic is easier to follow since we do not have the overhead that sections of the if tree are not reevaluated.
```suggestion
    private readonly pip: Observable<UserMediaViewModel | null> = combineLatest([
    this.screenShares,
    this.spotlightSpeaker,
    this.mediaItems,
  ]).pipe(
    switchMap(([screenShares, spoltlight, mediaItems]) => {
      if (screenShares.length > 0) {
        return this.spotlightSpeaker;
      }
      if (!spoltlight || spoltlight.local) {
        return of(null);
      }

      const localUserMedia = mediaItems.find(
        (m) => m.vm instanceof LocalUserMediaViewModel,
      ) as UserMedia | undefined;

      const localUserMediaViewModel = localUserMedia?.vm as
        | LocalUserMediaViewModel
        | undefined;

      if (!localUserMediaViewModel) {
        return of(null);
      }
      return localUserMediaViewModel.alwaysShow.pipe(
        map((alwaysShow) => {
          if (alwaysShow) {
            return localUserMediaViewModel;
          }

          return null;
        }),
      );
    }),
    this.scope.state(),
  );
```

src/state/CallViewModel.ts
toger5: the split of thwose two things is very nice!

src/state/CallViewModel.ts
toger5: That being said this is more then clear enough to merge it.
I am also curious about your take on this.

toger5 (commented):
Oh I cannot approve my own PR ;) So @hughns you need to approve on my behave

src/state/CallViewModel.ts
hughns: As per offline conversation just now, I was tempted to refactor it into a `combineLatest` when I was working on it earlier in the week. If it does actually work as before then I think it is much more readable 👍

robintown (approved):

robintown:
on Timo's behalf

PR #2708, opened 2024-11-01
Support for generic reactions

Half-Shot:
Requires https://github.com/element-hq/compound-design-tokens/pull/116 for the icon

This is the PoC for the generic-emoji-reactions feature. Currently you can:

- Send a reaction to a call via a new picker
- Play a specific sound for that reaction (or a generic sound otherwise).
- Incorporates the raise hand button into the same menu for space conservation.
- Enable/disable sounds/animations from the preferences menu.

TODO:

- [x] Add tests.
- [x] Appropriately specify licenses for sound effects.

image


toger5 (dismissed):
We played with it an observed an issue in widget mode:
Sending a raise hand does not result in the hand getting raised locally. The event gets send and other participants see the raised hand however.

What listener do we use to trigger the local hand raise.

The resulting issue is, that i cannot send a redact, since the widget is not in "raised hand state" and hence does not give the redaction option

Half-Shot:
> We played with it an observed an issue in widget mode: Sending a raise hand does not result in the hand getting raised locally. The event gets send and other participants see the raised hand however.
> 
> What listener do we use to trigger the local hand raise.
> 
> The resulting issue is, that i cannot send a redact, since the widget is not in "raised hand state" and hence does not give the redaction option

Should be fixed by https://github.com/element-hq/element-call/pull/2708/commits/a60d48965b0632a7f9300a473ef6ae8d2700afdd

src/button/ReactionToggleButton.tsx
```diff
@@ -0,0 +1,295 @@
...
+    async (reaction: ReactionOption) => {
+      const myMembership = memberships.find((m) => m.sender === userId);
+      if (!myMembership?.eventId) {
+        logger.error("Cannot find own membership event");
+        return;
+      }
+      const parentEventId = myMembership.eventId;
+      try {
+        setBusy(true);
+        // @ts-expect-error Trying to send a unspec'd event seems to miss the 3rd overload, need to come back to this.
```
Half-Shot: This needs https://github.com/matrix-org/matrix-js-sdk/pull/4497 in order to work.

src/sound/LICENCE.md
```diff
@@ -0,0 +1,22 @@
+# Sound effect licences
```
Half-Shot, line 1: Note, I've asked the legal team about whether it's acceptable to use these files https://github.com/element-hq/legal-compliance/issues/506.

src/reactions/index.ts
```diff
@@ -0,0 +1,163 @@
+/*
+Copyright 2024 New Vector Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only
+Please see LICENSE in the repository root for full details.
+*/
+
+import { RelationType } from "matrix-js-sdk/src/types";
+
+import catSoundOgg from "../sound/reactions/cat.ogg?url";
```
Half-Shot, line 10: These should be preloaded to prevent a delay between the reaction and the sound.

src/sound/LICENCE.md
Half-Shot: Legal resolved this.

src/Modal.tsx
```diff
@@ -27,8 +27,21 @@ import { useMediaQuery } from "./useMediaQuery";
...
   children: ReactNode;
   className?: string;
+  /**
+   * Class name to be used when in drawer mode (touchscreen). If `className` is given, this is ignored.
+   */
+  classNameDrawer?: string;
+  /**
+   * Class name to be used when in modal mode (desktop). If `className` is given, this is ignored.
+   */
+  classNameModal?: string;
```
robintown: Why not make `className` and `classNameDrawer`/`classNameModal` additive? This seems like a clearer pattern to me.

src/button/ReactionToggleButton.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+  return (
+    <div className={styles.reactionPopupMenu}>
+      <section className={styles.handRaiseSection}>
+        <Tooltip label={t("common.raise_hand")}>
+          <CpdButton
+            kind={isHandRaised ? "primary" : "secondary"}
+            aria-pressed={isHandRaised}
+            aria-label="Toggle hand raised"
+            onClick={() => toggleRaisedHand()}
```
robintown:
```suggestion
            onClick={toggleRaisedHand}
```

src/button/ReactionToggleButton.tsx
```diff
@@ -0,0 +1,324 @@
...
+  );
+
+  return (
+    <div className={styles.reactionPopupMenu}>
+      <section className={styles.handRaiseSection}>
+        <Tooltip label={t("common.raise_hand")}>
+          <CpdButton
+            kind={isHandRaised ? "primary" : "secondary"}
+            aria-pressed={isHandRaised}
+            aria-label="Toggle hand raised"
```
robintown: Needs to be translatable

src/button/ReactionToggleButton.tsx
```diff
@@ -0,0 +1,324 @@
...
+      <div className={styles.verticalSeperator} />
+      <section>
+        {isSearching ? (
+          <>
+            <Form.Root className={styles.searchForm}>
+              <Search
+                required
+                value={searchText}
+                name="reactionSearch"
+                placeholder="Search reactions…"
```
robintown: Untranslatable

src/button/ReactionToggleButton.tsx
```diff
@@ -0,0 +1,324 @@
...
+                onChange={onSearch}
+                onKeyDown={onSearchKeyDown}
+                // This is a reasonable use of autofocus, we are focusing when
+                // the search button is clicked (which matches the Element Web reaction picker)
+                // eslint-disable-next-line jsx-a11y/no-autofocus
+                autoFocus
+              />
+              <CpdButton
+                Icon={CloseIcon}
+                aria-label="close search"
```
robintown: Untranslatable

src/button/ReactionToggleButton.tsx
```diff
@@ -0,0 +1,324 @@
...
+                </CpdButton>
+              </Tooltip>
+            </li>
+          ))}
+        </menu>
+      </section>
+      {!isSearching ? (
+        <section>
+          <li key="search" className={styles.reactionPopupMenuItem}>
+            <Tooltip label="Search">
```
robintown: Untranslatable

src/button/ReactionToggleButton.tsx
```diff
@@ -0,0 +1,324 @@
...
+              <CpdButton
+                iconOnly
+                aria-label="Open reactions search"
```
robintown: Untranslatable

src/button/ReactionToggleButton.tsx
```diff
@@ -0,0 +1,324 @@
...
+            name: reaction.name,
+          },
+        );
+        // Do NOT close the menu after this.
+      } catch (ex) {
+        logger.error("Failed to send reaction", ex);
+      } finally {
+        setBusy(false);
+        setShowReactionsMenu(false);
+      }
```
robintown: Looks like this `finally` block is always closing the menu ..?

src/button/ReactionToggleButton.tsx
```diff
@@ -0,0 +1,324 @@
...
+    <>
+      <InnerButton
+        disabled={busy}
+        onClick={() => setShowReactionsMenu((show) => !show)}
+        raised={isHandRaised}
+        open={showReactionsMenu}
+      />
+      <Modal
+        open={showReactionsMenu}
+        title="Pick reaction"
```
robintown: Untranslatable

src/button/ReactionToggleButton.module.css
```diff
@@ -0,0 +1,66 @@
...
+/* These styles override the Modal styles to ensure our emoji
+   picker appears near the toolbar */
+
+.reactionPopupMenuModal {
+  width: fit-content !important;
+  top: 82vh !important;
+}
+
+@media not ((hover: none) or (pointer: coarse)) {
+  .reactionPopupMenuModal > div > div {
```
robintown: This is for styling the modal in the drawer case, right? Can the `classNameDrawer` prop help make this less brittle?

src/button/ReactionToggleButton.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+  );
+}
+
+test("Can open menu", () => {
+  const room = new MockRoom(memberUserIdAlice);
+  const rtcSession = new MockRTCSession(room, membership);
+  const { getByRole, container } = render(
+    <TestComponent rtcSession={rtcSession} room={room} />,
+  );
+  act(() => getByRole("button").click());
```
robintown: Could use @testing-library/user-event (throughout this test suite really)

src/button/ReactionToggleButton.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+  act(() => {
+    getByRole("button").click();
+  });
+  act(() => {
+    getByRole("button", {
+      name: "Search",
+    }).click();
+  });
+  await act(async () => {
+    getByPlaceholderText("Search reactions…").focus();
```
robintown: Likewise the user-event way would be to click on or tab into the search field with `user`, at which point you could remove the `act`

src/button/ReactionToggleButton.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+    getByRole("button").click();
+  });
+  act(() => {
+    getByRole("button", {
+      name: "Search",
+    }).click();
+  });
+  const searchField = getByPlaceholderText("Search reactions…");
+  act(() => {
+    fireEvent.keyDown(searchField, { key: "Escape" });
```
robintown, line 211: This should also have a user-event equivalent

src/button/__snapshots__/ReactionToggleButton.test.tsx.snap
```diff
@@ -0,0 +1,233 @@
...
+exports[`Can search for and send emoji 1`] = `
+<div
+  aria-hidden="true"
+  data-aria-hidden="true"
+>
+  <button
+    aria-disabled="false"
+    aria-expanded="true"
+    aria-labelledby=":r74:"
+    class="_button_i91xf_17 _has-icon_i91xf_66 _icon-only_i91xf_59"
```
robintown: Ideally these buttons would have `aria-haspopup` attributes

src/reactions/RaisedHandIndicator.tsx
```diff
@@ -76,29 +75,19 @@ export function RaisedHandIndicator({
...
+    <ReactionIndicator emoji="✋" miniature={miniature}>
       {showTimer && <p>{raisedHandDuration}</p>}
-    </div>
+    </ReactionIndicator>
   );
 
   if (onClick) {
     return (
       <button
         aria-label="lower raised hand"
```
robintown: Untranslatable

src/reactions/ReactionIndicator.tsx
```diff
@@ -0,0 +1,40 @@
...
+      className={classNames(styles.reactionIndicatorWidget, {
+        [styles.reactionIndicatorWidgetLarge]: !miniature,
+      })}
+    >
+      <div
+        className={classNames(styles.reaction, {
+          [styles.reactionLarge]: !miniature,
+        })}
+      >
+        <span role="img" aria-label="reaction indicator">
```
robintown: Untranslatable

src/reactions/ReactionIndicator.tsx
```diff
@@ -0,0 +1,40 @@
+/*
+Copyright 2024 New Vector Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only
+Please see LICENSE in the repository root for full details.
+*/
+
+import { PropsWithChildren, ReactNode } from "react";
+import classNames from "classnames";
+import "@formatjs/intl-durationformat/polyfill";
```
robintown: This import seems out of place since Intl.DurationFormat is only used in the other file

src/reactions/index.ts
```diff
@@ -0,0 +1,163 @@
...
+    ogg: genericSoundOgg,
+  },
+};
+
+// The first 6 reactions are always visible.
+export const ReactionSet: ReactionOption[] = [
+  {
+    emoji: "👍",
+    name: "thumbsup",
+    alias: ["+1", "yes", "thumbs up"],
```
robintown: This should get at least a 'TODO: i18n' if we're leaving i18n to a later iteration. (And when we do get to it, I imagine sourcing data from https://emojibase.dev/ would be better than translating names and aliases 1-to-1)

src/room/InCallView.module.css
```diff
@@ -144,3 +144,28 @@ Please see LICENSE in the repository root for full details.
...
+    translate: 100vw 0;
+    scale: 200%;
+  }
+
+  to {
+    opacity: 0;
+    translate: 100vw -100vh;
+    scale: 100%;
+  }
+}
```
robintown, line 172: Accessibility: Ideally this would skip the `translate` part of the animation (and maybe `scale` too) if `@media (prefers-reduced-motion)` matches

src/settings/PreferencesSettingsTab.tsx
```diff
@@ -21,11 +23,17 @@ export const PreferencesSettingsTab: FC = () => {
     showHandRaisedTimerSetting,
   );
 
+  const [showReactions, setShowReactions] = useSetting(showReactionsSetting);
+
+  const [playReactionsSound, setPlayReactionSound] = useSetting(
+    playReactionsSoundSetting,
+  );
+
   const onChangeSetting = useCallback(
```
robintown: The point of `useCallback` (referential stability) is kind of defeated if you're passing a new function on each render with `(e) => onChangeSetting(...)` below :) not that it matters much

src/sound/LICENCE.md
robintown, line 1: Where is 'deer' from?

src/useReactions.tsx
```diff
@@ -177,7 +191,53 @@ export const ReactionsProvider = ({
...
+
+        setReactions((reactions) => {
+          if (reactions[sender]) {
+            // We've still got a reaction from this user, ignore it to prevent spamming
+            return reactions;
+          }
+          setTimeout(() => {
+            // Clear the reaction after some time.
+            setReactions(({ [sender]: _unused, ...remaining }) => remaining);
+          }, REACTION_ACTIVE_TIME_MS);
```
robintown: React will log a warning if this timeout fires after the component has been unmounted

src/useReactions.tsx
```diff
@@ -177,7 +191,53 @@ export const ReactionsProvider = ({
...
+            (e) => e.eventId === membershipEventId && e.sender === sender,
+          )
+        ) {
+          logger.warn(
+            `Reaction target was not a membership event for ${sender}, ignoring`,
+          );
+          return;
+        }
+
+        const emoji = content.emoji?.split(/(?:)/u)?.[0];
```
robintown: `Intl.Segmenter` would be more appropriate here, as not all emoji are a single codepoint

robintown (commented):
Whew, lots of code = lots of comments. Feel free to tell me if there are certain things you would like to leave to later iterations.

src/useReactions.tsx
Half-Shot: Ah, this in theory does actually work although I'll take a more standard solution than my regex hack. I do remember that `Intl.Segmenter` caused us trouble with Firefox ESR in Element Web land, so I wonder if a polyfill may be appropriate here.

src/useReactions.tsx
Half-Shot: Fixed with a set that keeps track and clears any active timeouts.

src/settings/PreferencesSettingsTab.tsx
Half-Shot: oh yeah...duh

src/sound/LICENCE.md
Half-Shot: @anoadragon453, and I think he said it wasn't from a source we could license. I think we'll have to take it out unless we can find a better deer.

src/room/InCallView.module.css
Half-Shot: Excellent idea! And that's such an awesome media query <3

src/button/ReactionToggleButton.tsx
Half-Shot: https://github.com/element-hq/element-call/issues/2735 suggested so this doesn't happen by accident.

src/button/ReactionToggleButton.tsx
Half-Shot: Yeah, I don't know how best to present a failure right now. Asking in standup.

Half-Shot: Have added a generic error box so that we can raise the error state.

src/button/ReactionToggleButton.test.tsx
Half-Shot: I've instead opted to not click and ensure the autofocus does it's job. Seems to work.

src/button/ReactionToggleButton.test.tsx
Half-Shot: I couldn't find a way to fire a control character from user events, so had to resort to this. This seemed to be what Stack Overflow suggests :disappointed:

src/reactions/index.ts
Half-Shot: Hm, yeah I think we need to make this an improvement. I'll log a TODO and raise an issue once this is merged.

src/sound/LICENCE.md
anoadragon453: It was from a youtube video, but the channel has no licensing information...

Here's one I slightly modified [from pixabay](https://pixabay.com/sound-effects/baby-deer4-102323/) that's fun. Compressed so GitHub would accept it:

[deer-bleat.zip](https://github.com/user-attachments/files/17679507/deer-bleat.zip)

src/button/ReactionToggleButton.test.tsx
robintown: I believe it looks like: `await user.keyboard("[Escape]");`

src/button/ReactionToggleButton.tsx
```diff
@@ -236,12 +257,14 @@ export function ReactionToggleButton({
             name: reaction.name,
           },
         );
+        setErrorText(undefined);
+        setShowReactionsMenu(false);
         // Do NOT close the menu after this.
```
robintown: :thinking: This comment still looks inaccurate

package.json
```diff
@@ -116,5 +116,8 @@
   },
   "resolutions": {
     "strip-ansi": "6.0.1"
+  },
+  "dependencies": {
+    "@formatjs/intl-segmenter": "^11.7.3"
```
robintown: This can go in devDependencies

src/useReactions.tsx
```diff
@@ -285,6 +300,9 @@ export const ReactionsProvider = ({
       room.off(MatrixRoomEvent.Timeline, handleReactionEvent);
       room.off(MatrixRoomEvent.Redaction, handleReactionEvent);
       room.off(MatrixRoomEvent.LocalEchoUpdated, handleReactionEvent);
+      reactionTimeouts.forEach((t) => clearTimeout(t));
+      // If we're clearing timeouts, we also clear all reactions.
+      setReactions({});
     };
   }, [room, addRaisedHand, removeRaisedHand, memberships, raisedHands]);
```
robintown: Since this effect depends on `memberships` and `raisedHands` it seems it will clear the reactions as soon as any of those change. The pattern I usually go for in this situation is wrapping the data in refs to make the effect less "reactive":

```typescript
const latestMemberships = useLatest(memberships)
useEffect(() => {
   // do something with latestMemberships.current
}, [latestMemberships])

src/sound/LICENCE.md
robintown: This should still be noted in the list somewhere

fkwp: I think it was replaced by a free one in https://github.com/element-hq/element-call/pull/2708/commits/1062b982893e30152141f3263f6740ad327b9b2f

robintown (approved):

PR #2716, opened 2024-11-04
default to flat structre of m.call.member evenst (one event per sessi…

fkwp:
…on rather than an array)


toger5 (approved):
This seems like a good time to me where we tested it thoroughly.

PR #2724, opened 2024-11-06
Add rxjs/recommend eslint rules

hughns:
We will probably want to tweak these as we get more used to them.


fkwp (approved):

hughns:
I was hoping that these rules would catch some examples of `Observable<boolean>` being mistaken for `boolean` but it doesn't appear to do so out of the box if at all.

PR #2730, opened 2024-11-07
Fix lobby disappearing when using a lower 550 width window on Firefox.

toger5:
This can happen if the user is in embedded mode and uses the right panel chat + a large room list.

Happens only on Firefox.


toger5:
@robintown is this a reasonable fix? I dont know if there is a reasonable fix since this is definitely a Firefox bug.

fkwp (approved):

PR #2731, opened 2024-11-07
Add tests for Modal

Half-Shot:
Since #2708 uses the modal component, I felt compelled to write tests.

### DIFF
--- a/src/Modal.test.tsx
+++ b/src/Modal.test.tsx
@@ -5,11 +5,18 @@ SPDX-License-Identifier: AGPL-3.0-only
 Please see LICENSE in the repository root for full details.
 */
 
-import { afterAll, expect, test } from "vitest";
+import { expect, test } from "vitest";
 import { render } from "@testing-library/react";
 import { act, ReactNode, useState } from "react";
 
 import { Modal } from "./Modal";
+import { afterEach } from "node:test";
+import userEvent from "@testing-library/user-event";
+
+const originalMatchMedia = window.matchMedia;
+afterEach(() => {
+  window.matchMedia = originalMatchMedia;
+});
 
 test("that nothing is rendered when the modal is closed", () => {
   const { queryByRole } = render(
@@ -29,7 +36,7 @@ test("the content is rendered when the modal is open", () => {
   expect(queryByRole("dialog")).toMatchSnapshot();
 });
 
-test("the modal can be closed by clicking the close button", () => {
+test("the modal can be closed by clicking the close button", async () => {
   function ModalFn(): ReactNode {
     const [isOpen, setOpen] = useState(true);
     return (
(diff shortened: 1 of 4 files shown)


src/Modal.tsx
```diff
@@ -91,6 +91,7 @@ export const Modal: FC<Props> = ({
             )}
             // Suppress the warning about there being no description; the modal
             // has an accessible title
+            role="dialog"
             aria-describedby={undefined}
```
robintown:
```suggestion
            role="dialog"
            // Suppress the warning about there being no description; the modal
            // has an accessible title
            aria-describedby={undefined}
```

src/Modal.tsx
```diff
@@ -116,7 +117,12 @@ export const Modal: FC<Props> = ({
           />
           {/* Suppress the warning about there being no description; the modal
           has an accessible title */}
```
robintown: Could you move this down to the relevant line as well please?

src/Modal.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+    return (
+      <Modal title="My modal" open={isOpen} onDismiss={() => setOpen(false)}>
+        <p>This is the content.</p>
+      </Modal>
+    );
+  }
+  const { queryByRole, getByLabelText } = render(<ModalFn />);
+  act(() => {
+    getByLabelText("action.close").click();
+  });
```
robintown: I'd recommend using the @testing-library/user-event package for this sort of interaction:

```typescript
const user = userEvent.setup();
await user.click(getByRole("button", { name: "action.close" }));
```

(`getByRole` is also preferable to `getByLabelText`. I hope it does register as role="button"?)

src/Modal.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+  expect(queryByRole("dialog")).toBeNull();
+});
+
+const originalMatchMedia = window.matchMedia;
+
+afterAll(() => {
```
robintown: Is `afterEach` more appropriate here? If one added more tests to the bottom of this file it would be surprising if the `matchMedia` from the mobile test below affected those.

src/Modal.test.tsx
```diff
@@ -38,19 +45,14 @@ test("the modal can be closed by clicking the close button", () => {
       </Modal>
     );
   }
-  const { queryByRole, getByLabelText } = render(<ModalFn />);
-  act(() => {
-    getByLabelText("action.close").click();
+  const user = userEvent.setup();
+  const { queryByRole, getByRole } = render(<ModalFn />);
+  await act(async () => {
+    await user.click(getByRole("button", { name: "action.close" }));
```
robintown: There shouldn't need to be an `act` here as user-event handles calling `act` internally (though it has been known to be brittle at doing that in the past…)

robintown (approved):

PR #2737, opened 2024-11-08
Improve interactions to hide/show the footer

robintown:
This fixes a few different usability issues with the footer:

- When tapping one of the footer buttons, the footer would be dismissed rather than activating the button.
- When the footer was hidden, you could still tap the buttons.
- Interacting with controls in the footer would not reset the timer that hides it, leading to a feeling that the footer can disappear out from under you.


AndrewFerr (approved):
LGTM

PR #2738, opened 2024-11-08
Make the footer display more nicely on narrow screens

robintown:
It needed a bit of tweaking now that we have up to 6 buttons in the footer. I tried to do everything in CSS this time.

Screen Shot 2024-11-08 at 12 19 05


fkwp (approved):
tested on iOS Safari and Android

PR #2739, opened 2024-11-08
Ensure reactions from the same user across multiple devices shows in the correct Tile

Half-Shot:
This problem occurs because we mapped raised hand and reactions based upon the user's user ID. However this isn't unique and so when a reaction is sent from one device, it appears on multiple tiles.

This refactors so we use membership event IDs instead, however this isn't working yet because I haven't found a way to ID the correct GridTile for the right user.

This is going to need #2701 to work so we can hook off the membership event ID correctly.

### DIFF
--- a/src/button/ReactionToggleButton.test.tsx
+++ b/src/button/ReactionToggleButton.test.tsx
@@ -83,7 +83,7 @@ test("Can raise hand", async () => {
   expect(container).toMatchSnapshot();
 });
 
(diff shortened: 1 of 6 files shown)


hughns:
@Half-Shot #2701 has landed so can we now get this updated and ready for review?

Half-Shot:
Closing in favor of https://github.com/element-hq/element-call/pull/2885 which does this better.

PR #2752, opened 2024-11-11
Make one-on-one layout less prone to crashing

robintown:
The basic issue here, I think, was that the `oneOnOne` observable flag and the `onOnOneLayout` observable could become out of sync, as RxJS does *not* have atomicity guarantees. We can work around this by combining them into one observable.


fkwp (approved):
lgtm

PR #2756, opened 2024-11-11
Add matrix_rtc_session config options + bump js-sdk

hughns:
matrix-org/matrix-js-sdk@6971e7b...fcb69b1

Also brings in:

- matrix-org/matrix-js-sdk#4342
- matrix-org/matrix-js-sdk#4494

PR #2759, opened 2024-11-11
Bump matrix-js-sdk for MSC4222 support

robintown:
This bumps matrix-js-sdk to a branch https://github.com/matrix-org/matrix-js-sdk/tree/element-call-nov-preview that includes

- https://github.com/matrix-org/matrix-js-sdk/pull/4487
- https://github.com/matrix-org/matrix-js-sdk/pull/4498 (which is what the commit ref was pointing to previously as of https://github.com/element-hq/element-call/pull/2756)

in addition to everything we've gotten merged so far into develop. A smoke test indicates that the MSC4222 support appears to be working fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "livekit-client": "^2.5.7",
     "lodash": "^4.17.21",
     "loglevel": "^1.9.1",
-    "matrix-js-sdk": "matrix-org/matrix-js-sdk#fcb69b16ad8d170c67ea844f83543d467bbd7707",
+    "matrix-js-sdk": "matrix-org/matrix-js-sdk#8e9a04cdec0f88fc876bbbf406db55b0677f005d",
     "matrix-widget-api": "^1.10.0",
     "normalize.css": "^8.0.1",
     "observable-hooks": "^4.2.3",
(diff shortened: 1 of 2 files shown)


hughns:
n.b. this would impact the codepaths for EC running as an SPA but also for when it runs as embedded. So, both need testing.

toger5:
We do not need any additional changes to element call (embedded + spa)
As we hoped for it all works as it is now.

Context for why this works:
```
if (this.syncApi instanceof SyncApi)
    // sync V2
    // if the third (`stateAfterEventList`) parameter is undefined, `injectRoomEvents` will use the fourth (`timelineEventList`)
    // parameter with the legacy behaviour (using theimeline events as state events + emitting state changes)
    await this.syncApi.injectRoomEvents(this.room!, [], undefined, [event]);
} else {
    // SSS
    // not used in EC (even though matroska mode supports it) via client `opts.slidingSync`
    await this.syncApi!.injectRoomEvents(this.room!, [], [event]);
}
```

toger5 (approved):
As mentioned in a comment above this just work as it is because the js-sdk supporting MSC4222 will behave as before by default:
 - all timeline events will update the state.
  - The react sdk stop gap driver will only send state_after


So we get the bahaviour we want

PR #2761, opened 2024-11-12
Decrypt potential reaction events before checking their type

robintown:
By rights, this fix I had made to decrypt reaction events shouldn't have appeared successful, because I was requiring the event to have a certain type before asking matrix-js-sdk to decrypt it, and you can't know an event's type before it's decrypted. Probably what was happening is that another code path was requesting the events to be decrypted so that this mistake didn't matter.


Half-Shot (approved):
Makes a lot of sense!

fkwp (approved):

PR #2764, opened 2024-11-12
Init client with a memory store (for MSC4222)

Half-Shot:
A short term fix to see if we can figure out the storage issue. **Do not merge before testing if this helps**


toger5 (approved):
Thx!
Its based on the local echo and MSC4222 js sdk, perfect!

PR #2765, opened 2024-11-12
Bump js sdk to a branch supporting the state fix msc4222 changes and local-echo

toger5:


package.json
```diff
@@ -87,7 +87,7 @@
     "livekit-client": "^2.5.7",
     "lodash": "^4.17.21",
     "loglevel": "^1.9.1",
-    "matrix-js-sdk": "matrix-org/matrix-js-sdk#8e9a04cdec0f88fc876bbbf406db55b0677f005d",
+    "matrix-js-sdk": "matrix-org/matrix-js-sdk#0098c1a37d33123db5e5c54c6292a9cca7a87ca5",
```
Half-Shot, line 90: You can drop my change that just landed too :) #2764

toger5: sure. otherwise its not really testable right?

toger5: Lets not merge this for now but use the netlify version to test it
(so we have the save version on develop/livekit branch)

toger5:
We can close this since we did those changes as part of: https://github.com/element-hq/element-call/pull/2767

We will move them over to the LK branch once they are merged into the js-sdk

Issue #1281, opened 2023-07-20
[ER-59] E2EE: Send media keys via to_device message


dbkr:
# Add to-device key distribution

- After https://github.com/vector-im/element-call/issues/1278

In addition to sending media keys as events to the room, send them via to_device events encrypted with Olm as a primary way of sending the keys, then keep the room events as backup. Add a posthog event to track when the keys don't arrive via to_device. This will allow us to migrate away from using room events in the long term as they cause the media key to be saved in the room history.

## Prerequisits:
 - [x] To device widget MSC https://github.com/matrix-org/matrix-spec-proposals/pull/3819
 - [x] Make all platforms support to-device widget features
    - [x]  rust sdk https://github.com/matrix-org/matrix-rust-sdk/issues/4859
    - [x]   js-sdk (matrix-widget-api) https://github.com/matrix-org/matrix-widget-api/pull/57
 - [x] split the encryption key logic out from MatrixRTC session into an EncryptionManager (Done in: https://github.com/matrix-org/matrix-js-sdk/pull/4612)
 - [x]  Planning meeting what is needed and how it should be architectured (in regards to https://github.com/matrix-org/matrix-js-sdk/pull/4612) (results of planning meeting see **Production ready requirements** and **Tasks**

## Production ready requirements
 - [x] ~~Exclude untrusted devices (expected to need widget api MSC changes)~~ removed, now part of [ER-245](https://github.com/element-hq/voip-internal/issues/356)
 - [x] ~~Error handling on sending to-device events (expected to need widget api MSC changes)~~ removed, now part of [ER-244](https://github.com/element-hq/voip-internal/issues/355)
 - [x] ~~Performance: Investigate, test the performance and eventually add ratcheting (expected to need matrixRTC MSC changes)~~ removed, now part of [ER-244](https://github.com/element-hq/voip-internal/issues/355)

## Tasks

Iterations: are based on Temp: see: https://excalidraw.com/#room=b7556403b8947d5ee7ac,duF48bj2SbmRTsa3ivmLFA

### Iteration 1: "change transport" and "make it backwards compatible"
- [x] Use hughs work as reference
- [x] https://github.com/matrix-org/matrix-js-sdk/pull/4773
   - [x] move transport in its own class
   - [x] create a new transport for room event key sharing
   - [x] use that transport
   - [x] adjust tests to support the new architecture
- [x] #3155
   - [x] https://github.com/matrix-org/matrix-js-sdk/pull/4785 
[...]

PR #2766, opened 2024-11-12
Mitigation for gappy/limited sync responses in SPA mode

hughns:
We might not want to merge this and instead back paginate properly.


hughns:
Also, this is probably unnecessary when we are using to-device for distributing keys: https://github.com/element-hq/element-call/issues/1281

toger5:
This is not needed since the room message exchange in SPA is not desired anymore (If we use SPA per participant encryption we will use to-device nowadays).

For to-device messages we will not suffer from gappy syncs.

PR #2769, opened 2024-11-12
Backfill hand raised events once client has synced.

Half-Shot:
~~I *think* there is a race here where we might try to fetch reactions before the local state of the room has been synced, leading to missing hand raised reactions on newly joining clients to the room.~~

The easier solution here is to call `fetchRelations` on the client once we are ready, which will fetch any events in the case that we can't find the reaction we expect.

### DIFF
--- a/src/useReactions.test.tsx
+++ b/src/useReactions.test.tsx
@@ -5,7 +5,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 Please see LICENSE in the repository root for full details.
 */
 
-import { render } from "@testing-library/react";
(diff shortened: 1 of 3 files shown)


toger5:
This looks useful. What is missing to move it out of draft?
(Can the description get updated to explain the exact scenario in which this PR is needed. Iirc we might end up in the wrong toggle state if the related event is missing?)

BillCarsonFr:
stale, closing

robintown:
Presumably `fetchRelations` isn't implemented in the widget API, so that would be one of the blockers to fixing this for widgets.

PR #2774, opened 2024-11-14
Enable Codecov bundle analysis plugin

sandhose:
This is to help track changes to the output bundle size over time and in PRs.


hughns (approved):
Thank you very much 👍

PR #2775, opened 2024-11-14
Use optimized font from fontsource

sandhose:
This replaces the embedded font with a package from fontsource.

The @font-face declarations are way better, with the font being split depending on the font-weight, font-style, and unicode-range.

I also removed the custom font-family variables on desktop, so that we can get the carefully-crafted fallbacks from the compound-design-tokens.

Here is a side-by-side comparison for this patch:

https://github.com/user-attachments/assets/78ed4396-3b41-487b-95df-023fd50e7eff

You'll notice on the left that the font swap is barely noticeable (at 10s) compared to the right (at 22s) and that the fonts are way smaller to load.

I still kept the font overrides on Android/iOS, as I assume this is for better design in embedded mode on those platforms.


hughns:
The overall bundle size increases:

image

But are you saying this is okay because only the required fonts are actually loaded?

sandhose:
This is because public assets aren't accounted for in the bundle size, so the increase here is misleading.

This is the actual diff in size during loading:

image

hughns (approved):

PR #2776, opened 2024-11-14
Handle locales as Vite assets

sandhose:
This makes it so that locales are handled as assets by Vite.

This has a few advantages:

 - the assets names are hashed, so we can cache them
 - we have an exhaustive list of all locales in the app
 - so i18next can properly load the locales instead of trying a few
 - we reduce bundle size because we have a simple custom loader instead of i18next-http-backend

Here's a diff of the JSON files being loaded with this patch:

image


hughns (approved):
This looks sane to me but I'm not familiar with the localazy setup. @fkwp could you sanity check this too?

sandhose:
(Note that now there is a small conflict with #2778, depends which one lands first)

robintown (approved):

PR #2778, opened 2024-11-14
Load the Intl.Segmenter and Intl.DurationFormat polyfills only if needed

sandhose:
This removes around 200Kb of the main bundle.

It will still load the polyfill if needed (notably the previous Firefox ESR), but skips on modern browsers.

### DIFF
--- a/src/reactions/RaisedHandIndicator.tsx
+++ b/src/reactions/RaisedHandIndicator.tsx
@@ -12,12 +12,11 @@ import {
   useEffect,
   useState,
 } from "react";
-import { DurationFormat } from "@formatjs/intl-durationformat";
 import { useTranslation } from "react-i18next";
 
 import { ReactionIndicator } from "./ReactionIndicator";
 
-const durationFormatter = new DurationFormat(undefined, {
+const durationFormatter = new Intl.DurationFormat(undefined, {
   minutesDisplay: "always",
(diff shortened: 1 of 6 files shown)


src/reactions/RaisedHandIndicator.tsx
```diff
@@ -12,12 +12,11 @@ import {
   useEffect,
   useState,
 } from "react";
-import { DurationFormat } from "@formatjs/intl-durationformat";
 import { useTranslation } from "react-i18next";
 
 import { ReactionIndicator } from "./ReactionIndicator";
 
-const durationFormatter = new DurationFormat(undefined, {
+const durationFormatter = new Intl.DurationFormat(undefined, {
```
Half-Shot: This landed in Node 23, so we will need to polyfill the tests.

sandhose: Fixed! I also had to embed the translations in tests, as now the initializer waits for i18next to initialise now before continuing

Half-Shot: Oh but your polyfill code should actually be doing this, ugh??

sandhose: `initializeBeforeReact` isn't necessarily being run in tests

hughns (approved):

PR #2780, opened 2024-11-14
Always deal with MediaItem at the LayoutMedia layer

hughns:
That way we only convert to the view model when we do LayoutMedia => Layout

### DIFF
--- a/src/state/CallViewModel.ts
+++ b/src/state/CallViewModel.ts
@@ -91,37 +91,37 @@ const showFooterMs = 4000;
 
 export interface GridLayoutMedia {
   type: "grid";
-  spotlight?: MediaViewModel[];
-  grid: UserMediaViewModel[];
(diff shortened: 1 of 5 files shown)


toger5 (commented):
I see how this simplifies things. If I understand correctly one can phrase the PR as shifting the transition from Media to MediaViewModel into a lower level.
Passing the full MediaItem instead of just the ViewModel increases the access however. I think the reason for creating it like this is to have as limited access to the MediaItem as possible in the `CallViewModel`.

The vm should represents this "smallest set of data".

What is the win we get by passing the whole `MediaItem`?
If there is too much redundancy I think we might need to evaluate if MediaItem is already a View Model. And we should copy all `UserMediaViewModel` code into `UserMedia`.

Is there any data from the MediaItem that we need in the Layout classes?

PR #2784, opened 2024-11-14
Add keyboard shortcuts for raised hand / reactions

Half-Shot:
This is mostly moving more of the event sending code to useReactions, and then hooking up to the existing keyboard shortcuts code


robintown (commented):
Code looks fine but I have one design suggestion: could we use the `caption` field of tooltips here for the shortcuts? (cf Element Web)

Screenshot from 2024-11-19 10-46-07

robintown (approved):

PR #2786, opened 2024-11-14
Precompress assets and set the right cache headers

sandhose:
Looks like we're not serving gzipped assets off call.element.{dev,io}

This is because of the `Cache-Control: no-transform` header in the response.

This PR sets the right headers and also pre-compress the assets to serve them gzipped directly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,3 @@ FROM nginxinc/nginx-unprivileged:alpine
 
 COPY ./dist /app
 COPY config/nginx.conf /etc/nginx/conf.d/default.conf
-
-USER root
-
-RUN rm -rf /usr/share/nginx/html
-
-USER 101
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -3,23 +3,16 @@ server {
     server_name  localhost;
 
     root   /app;
+    gzip_static on;
 
     location / {
-        # disable cache entriely by default (apart from Etag which is accurate enough)
-        add_header Cache-Control 'private no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0';
-        if_modified_since off;
-        expires off;
-        # also turn off last-modified since they are just the timestamps of the file in the docker image
-        # and may or may not bear any resemblance to when the resource changed
-        add_header Last-Modified "";
-
         try_files $uri $uri/ /index.html;
+        add_header Cache-Control "no-store, no-cache, must-revalidate";
     }
 
     # assets can be cached because they have hashed filenames
     location /assets {
-        expires 1w;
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -2,9 +2,3 @@ FROM nginxinc/nginx-unprivileged:alpine
 
 COPY ./dist /app
 COPY config/nginx.conf /etc/nginx/conf.d/default.conf
-
-USER root
-
-RUN rm -rf /usr/share/nginx/html
```
sandhose, line 8: This is unnecessary, Docker has layers anyway, so it's not as if we were reducing the size of the image doing this, and we already replace the configuration where this is referenced from

config/nginx.conf
```diff
@@ -3,23 +3,16 @@ server {
     server_name  localhost;
 
     root   /app;
+    gzip_static on;
```
sandhose, line 6: this doesn't compress assets on the fly, but rather serve compressed assets if they exist (with a `.gz`)

config/nginx.conf
```diff
@@ -3,23 +3,16 @@ server {
...
-        # disable cache entriely by default (apart from Etag which is accurate enough)
-        add_header Cache-Control 'private no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0';
-        if_modified_since off;
-        expires off;
-        # also turn off last-modified since they are just the timestamps of the file in the docker image
-        # and may or may not bear any resemblance to when the resource changed
-        add_header Last-Modified "";
-
         try_files $uri $uri/ /index.html;
+        add_header Cache-Control "no-store, no-cache, must-revalidate";
```
sandhose, line 10: I think all of this was unnecessarily complicated

config/nginx.conf
```diff
@@ -3,23 +3,16 @@ server {
...
     }
 
     # assets can be cached because they have hashed filenames
     location /assets {
-        expires 1w;
-        add_header Cache-Control "public, no-transform";
+        add_header Cache-Control "public, immutable, max-age=31536000";
```
sandhose, line 15: The 'no-transform' was the culprit. Assets are always hashed in this folder, so it's good to have a very long TTL + mark them as immutable

vite.config.js
```diff
@@ -38,6 +39,10 @@ export default defineConfig(({ mode }) => {
       bundleName: "element-call",
       uploadToken: process.env.CODECOV_TOKEN,
     }),
+
+    compression({
+      exclude: [/config.json/],
```
sandhose, line 44: We exclude the config.json, as this is usually swapped at runtime, and might be there during the docker build for some reason

fkwp (approved):
nice

PR #2789, opened 2024-11-15
Small improvements to settings tab

Half-Shot:
This fixes a couple of bugs in the settings tab, namely the mixup on strings for the reactions settings and the profile element being off center


fkwp (approved):
lgtm

PR #2790, opened 2024-11-15
Pre-fetch the config.json to improve startup time

sandhose:
This does three things:

 - remove the ?cachebuster from the config URL, we instead rely on having correct cache headers
 - add a link rel=preload for the config file in the index.html
 - adjust the default cache-control to use a 'stale-while-revalidate' strategy. This way, the CDN and the browser will cache the index.html and the config.json for a short time, ensuring that we always get the latest version, but still allowing us to pre-fetch the config file.

The difference is most noticeable when the assets are already cached by the browser, as loading the app is basically two requests away: one for the index.html, which immediately triggers the preload, so we don't have to wait for the JS to be parsed and executed for that.

Comparison waterfall with the app already cached in both cases:

image

Notice how the config.json is loaded way earlier, and therefore unblocks loading of other resources

### DIFF
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -8,7 +8,7 @@ server {
 
     location / {
         try_files $uri $uri/ /index.html;
-        add_header Cache-Control "no-store, no-cache, must-revalidate";
+        add_header Cache-Control "public, max-age=30, stale-while-revalidate=30";
     }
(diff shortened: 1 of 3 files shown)


fkwp (approved):
lgtm

PR #2791, opened 2024-11-15
Update matrix-js-sdk including MSC4222 support

Half-Shot:
Draft PR to test the latest js-sdk changes.


Half-Shot:
We tried it just now and found that if you joined into an existing call then you are still missing some state.

toger5:
So we keep the updated js-sdk but keep the memory store?
Or was this blocked by florians 200room account which would make everything super slow?

Half-Shot:
> Or was this blocked by florians 200room account which would make everything super slow?

I would be wary, I'm also a few 10s of rooms in and this would likely spike costs for call.element.io (due to many initial syncs).

hughns:
Memory store was done in https://github.com/element-hq/element-call/pull/2811

hughns:
Superseded by https://github.com/element-hq/element-call/pull/2841

PR #2794, opened 2024-11-15
Add sound effect for call joined / left

### LINKED ISSUE #343: Sound effects for join/leave/end/etc
We should have sound effects for when certain actions happen in the call. imo, at a minimum we should have:
* An audio cue for when someone joins the call (to prevent the awkward case where you're waiting for someone in another tab and they stare at you for a minute before saying hello)
* An audio cue for when someone leaves the call (though potentially become inspired by Jitsi where after a certain size of call the join/leave effects no longer play)
* An audio cue for when you've clicked the hangup button, to give confidence that you've actually left the call and aren't broadcasting your camera to the world still

Half-Shot:
Fixes https://github.com/element-hq/element-call/issues/343 (aside from hangup sound, which is harder and explained below).

This adds the ability to play a sound when a new person joins or leaves a call, or when the current user joins a call. Technically this actually is when a tile is created or removed, so it might need a bit more finessing.

The hangup sound is a little more tricky to get right, and will probably land in a future PR. The problem is that a hangup results in a layout change (which we don't want to block on audio playing) that may either send the user to a `/` or to a signup page if the user isn't registered. It's tricky and the ideal state would be a universal hangup screen we could use for this.


src/room/CallEventAudioRenderer.tsx
```diff
@@ -0,0 +1,79 @@
...
+  vm,
+}: {
+  vm: CallViewModel;
+}): ReactNode {
+  const [shouldPlay] = useSetting(playReactionsSound);
+  const [effectSoundVolume] = useSetting(effectSoundVolumeSetting);
+  const memberIds = useObservableEagerState(vm.userMediaIds);
+  const previousMembers = useDeferredValue(memberIds);
+  const callEntered = useRef<HTMLAudioElement>(null);
+  const callLeft = useRef<HTMLAudioElement>(null);
```
toger5: As discussed already i think it would be cool to make the join/leave an observable as well.
This could be built with a `scan`.
sth like
```
let memberChanges = memberIds.pipe(scan((prev, ids)=>{
    const left = prev.ids.filter((id)=>ids.includes(id));
    const joined = ids.filter((id)=>prev.ids.includes(id));
    return {ids, joined, left}
}));
```
than this could be used like an event emitter that emits on join and leave:
```
memberChanges.pipe(
  map((m)=>m.left),
  filter((left)=>left.length === 0),
).subscribe((left)=>"play sond on leave")
```

src/room/CallEventAudioRenderer.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  useEffect(() => {
+    if (memberIds.length > MAX_PARTICIPANT_COUNT_FOR_SOUND) {
+      return;
+    }
```
toger5: this also would be super easy to add as another filter to the example above:
```
memberChanges.pipe(
  filter((change)=>change.ids.length>MAX_PARTICIPANT_COUNT_FOR_SOUND),
  map((m)=>m.left),
  filter((left)=>left.length === 0)
). subscribe((left)=>"play sond on leave")
```

src/room/CallEventAudioRenderer.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+    const memberLeft = !!previousMembers.filter((m) => !memberIds.includes(m))
+      .length;
+    const memberJoined = !!memberIds.filter((m) => !previousMembers.includes(m))
+      .length;
+
+    if (callEntered.current && callEntered.current?.paused && memberJoined) {
+      callEntered.current.volume = effectSoundVolume;
+      void callEntered.current.play();
+    }
```
toger5: if we have lots of leaves/joins per seconds this could sounds bad.
```
memberChanges.pipe(
  filter((change)=>change.ids.length>MAX_PARTICIPANT_COUNT_FOR_SOUND),
  map((m)=>m.left),
  filter((left)=>left.length === 0),
  // so we at most play two sounds at the same time
  throttleTime(SOUND_DURATION/2),
).subscribe(  (left)=>"play sound on leave" )
```

toger5 (commented):
As discussed i think we have a chance to go a step further where we use rxjs

src/room/CallEventAudioRenderer.tsx
Half-Shot: Due to the magic of \<audio />, we are currently limited to only playing the sound when it's not already playing so this isn't actually an issue at present.

However, I still think it's probably worth doing this for performance.

toger5: In this case I am almost inclined to see what a reasonable amount of parallel join sounds is (maybe 2-3) and create 2 or 3 players so that we can do the logic to only drop joines once all of them are already in playing state?

Half-Shot: I've added 2 players (configurable).

toger5: nice

toger5 (approved):
Thanks. The logic is very easy to follow.

PR #2802, opened 2024-11-19
Don't fall back as eagerly to unselected devices

### LINKED ISSUE #2770: Media devices reset on reload
### Steps to reproduce

1. Set your media devices to the ones you want
2. Reload the page

#### What did you expect?

The settings stay on the same devices.

#### What happened instead?

The settings reset to the default devices and the dropdowns in settings are blank (??)

### Operating system

NixOS unstable

### Browser information

Firefox 132.0.1

### URL for webapp

call.element.dev

### Will you send logs?

No

robintown:
Somewhere around version 131 or 132, Firefox has started being more paranoid about media device fingerprinting, and will not even give you the IDs of available devices until you've requested a media stream. Instead you only get a single audio input and video input each with the empty string as their device ID, representing the system's default device. We can recognize this case and avoid resetting the device selection.

Closes https://github.com/element-hq/element-call/issues/2770

### DIFF
--- a/src/livekit/MediaDevicesContext.tsx
+++ b/src/livekit/MediaDevicesContext.tsx
@@ -24,6 +24,7 @@ import {
   audioInput as audioInputSetting,
   audioOutput as audioOutputSetting,
   videoInput as videoInputSetting,
+  Setting,
 } from "../settings/settings";
 import { isFirefox } from "../Platform";
 
@@ -58,7 +59,7 @@ function useObservableState<T>(
 
 function useMediaDevice(
   kind: MediaDeviceKind,
-  fallbackDevice: string | undefined,
+  setting: Setting<string | undefined>,
   usingNames: boolean,
   alwaysDefault: boolean = false,
 ): MediaDevice {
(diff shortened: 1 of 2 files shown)


src/livekit/MediaDevicesContext.tsx
```diff
@@ -84,15 +85,19 @@ function useMediaDevice(
...
+      // If the selected device is available, use it. Or if every available
+      // device ID is falsy, the browser is probably just being paranoid about
+      // fingerprinting and we should still try using the selected device.
+      // Otherwise, fall back to the first available device.
+      devId =
+        available.some((d) => d.deviceId === selectedId) ||
+        available.every((d) => d.deviceId === "")
+          ? selectedId
           : available.at(0)?.deviceId;
     }
```
toger5: Where is the fallback happening?
If all devices are `""` we would still end up using selectedId. What if that is not available? If the browser is paranoid but also the mic has been disconnected?
Where do we catch that?

src/livekit/MediaDevicesContext.tsx
```diff
@@ -85,19 +85,21 @@ function useMediaDevice(
...
+      // If the preferred device is available, use it. Or if every available
       // device ID is falsy, the browser is probably just being paranoid about
-      // fingerprinting and we should still try using the selected device.
-      // Otherwise, fall back to the first available device.
-      devId =
-        available.some((d) => d.deviceId === selectedId) ||
+      // fingerprinting and we should still try using the preferred device.
+      // Worst case it is not available and the browser will gracefully fall
+      // back to some other device for us when requesting the media stream.
+      // Otherwise, select the first available device.
```
toger5, line 98: 🎉

toger5 (approved):

Issue #2022, opened 2024-01-03
Changing the audio output device is not working with Firefox


wolfito76:
#### Steps to reproduce
Click on "Settings" and select the tab "Audio"
The option to change the output device is not shown. Microphone is showing and changing the microphone is working.
![Bildschirmfoto vom 2024-01-02 15-03-38](https://github.com/element-hq/element-call/assets/14106486/80744bdc-ae30-4fa9-80ce-4cee567c40f3)

#### What did you expect?
The option to change the device for audio output.

#### What happened instead?
It's just missing.

### Operating system

Ubuntu 22.04 LTS

### Browser information

Firefox 121.0

### URL for webapp

call.element.io

### Will you send logs?

No

robintown:
Hi, the reason why this option is hidden is because Firefox had a bug with output devices which broke acoustic echo cancellation: https://bugzilla.mozilla.org/show_bug.cgi?id=1849108

This bug has been fixed in Firefox 123, so in a month or so we can look into turning this option back on.

wolfito76:
I opened the issue before I saw the exception for firefox in the code, but there was no comment explaining why.
Thanks for the clarification.

PR #2803, opened 2024-11-19
Replace device dropdowns with radio buttons

robintown:
This is closer to what the designs actually want device settings to look like, and it avoids the visual glitch in which the dropdown would render underneath the slider.

Screenshot from 2024-11-19 17-24-11

### DIFF
--- a/src/settings/SettingsModal.tsx
+++ b/src/settings/SettingsModal.tsx
@@ -5,10 +5,10 @@ SPDX-License-Identifier: AGPL-3.0-only
 Please see LICENSE in the repository root for full details.
 */
 
-import { ChangeEvent, FC, ReactNode, useCallback } from "react";
+import { ChangeEvent, FC, useCallback } from "react";
 import { Trans, useTranslation } from "react-i18next";
 import { MatrixClient } from "matrix-js-sdk/src/matrix";
-import { Dropdown, Separator, Text } from "@vector-im/compound-web";
+import { Root as Form, Text } from "@vector-im/compound-web";
 
 import { Modal } from "../Modal";
 import styles from "./SettingsModal.module.css";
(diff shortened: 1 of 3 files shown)


toger5:
Do you think this is the right place to also work on the default device bahaviour?
Iirc on chrome there is a virutal default device that is always in sync with the system default and on Firefox the order of the devices change in a ways o the first one is always the default device.
I could imagine that a checkbox `Default` could be added and the options will only become selectable once the checkbox is disabled?

src/settings/SettingsModal.tsx
```diff
@@ -125,33 +91,45 @@ export const SettingsModal: FC<Props> = ({
...
+            caption={t("common.microphone")}
           />
-        </div>
+          {!isFirefox() && (
+            <DeviceSelection
+              devices={devices.audioOutput}
+              caption={t("settings.speaker_device_selection_label")}
+            />
+          )}
+          <div className={styles.volumeSlider}>
```
toger5, line 105: I think this might not be needed. (anymore)
This is what i get with the WebRtc test page https://webrtc.github.io/samples/src/content/devices/input-output/

It seems like both platforms support input and output devices.

# Chrome
Screenshot from 2024-11-20 10-42-25
Screenshot from 2024-11-20 10-42-21

# Firefox
Screenshot from 2024-11-20 10-53-51
Screenshot from 2024-11-20 10-53-49

toger5: But I am not sure how much of the output selection even works. On linux It seems like I mostly just use the system level output/input.

robintown: Past me says that we were waiting for Firefox 123 in order to turn it back on: https://github.com/element-hq/element-call/issues/2022#issuecomment-1900528871 so I'll do so in a follow-up PR.

robintown:
> Do you think this is the right place to also work on the default device bahaviour? Iirc on chrome there is a virutal default device that is always in sync with the system default and on Firefox the order of the devices change in a ways o the first one is always the default device. I could imagine that a checkbox `Default` could be added and the options will only become selectable once the checkbox is disabled?

I'd like to avoid adding a switch that blocks you from changing your device until it's turned off. But perhaps we could normalize the behavior between Firefox and Chrome by adding a virtual 'Default' device in Firefox? If that sounds appropriate to you I could work on that in a later PR.

toger5 (approved):

toger5:
This works really great! Code looks good and I tested it a bit.

PR #2804, opened 2024-11-20
Remove redundant distinctUntilChanged

hughns:
Because this.scope.state() does this for us anyhow.


toger5 (approved):
So this saves us from numerus deep Eq's?
This is the code you are referring to (where its already happening):
https://github.com/element-hq/element-call/blob/9dbc43d692e7667991f6841fd308962d0ef7e9e4/src/state/ObservableScope.ts#L34-L39
?

PR #2805, opened 2024-11-20
Rxjs subjects should not be exposed

hughns:
This rule will encourage us to do the right thing and better abstract.

See https://github.com/cartant/eslint-plugin-rxjs/blob/main/docs/rules/no-exposed-subjects.md


toger5 (approved):
I am surprised CI just accepts this.
Do we actually never do this?

hughns:
> I am surprised CI just accepts this. Do we actually never do this?

Currently, no. But the was one in https://github.com/element-hq/element-call/pull/2701 until I refactored it.

PR #2807, opened 2024-11-20
Disambiguate between types of "member"

hughns:
We have Matrix room members and MatrixRTC session memberships. Livekit also has rooms.

So, this attempts to make it more obvious as to what type you are referring to.


toger5 (approved):
Yes!

PR #2809, opened 2024-11-20
Some tsdoc and explicit typing

hughns:
Pulled out of https://github.com/element-hq/element-call/pull/2701


toger5 (approved):
Simple but lovely!

PR #2813, opened 2024-11-20
Some simple initial tests for MediaView

hughns:
The purpose of moving videoMuted from CSS class to inline style is to make it more testable. I'm open to a more elegant way!

### DIFF
--- a/src/tile/MediaView.test.tsx
+++ b/src/tile/MediaView.test.tsx
@@ -1,0 +1,106 @@
+/*
+Copyright 2024 New Vector Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only
+Please see LICENSE in the repository root for full details.
+*/
+
+import { describe, expect, test } from "vitest";
+import { render, screen } from "@testing-library/react";
+import { axe } from "vitest-axe";
+import { TooltipProvider } from "@vector-im/compound-web";
+import { TrackReferencePlaceholder } from "@livekit/components-core";
+import { Track } from "livekit-client";
+
+import { MediaView } from "./MediaView";
+import { EncryptionStatus } from "../state/MediaViewModel";
+import { mockLocalParticipant } from "../utils/test";
+
+describe("MediaView", () => {
+  const participant = mockLocalParticipant({});
+  const trackReferencePlaceholder: TrackReferencePlaceholder = {
+    participant,
+    source: Track.Source.Camera,
+  };
+
+  test("is accessible", async () => {
+    const { container } = render(
+      <MediaView
+        displayName="Bob"
+        videoEnabled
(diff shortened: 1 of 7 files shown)


src/tile/MediaView.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+          encryptionStatus={EncryptionStatus.Connecting}
+          mirror={false}
+          unencryptedWarning={false}
+          video={trackReference}
+          member={undefined}
+        />,
+      );
+      expect(screen.getByTestId("name_tag")).toHaveTextContent("Bob");
+    });
+  });
```
toger5: Can we not combine those two tests?

hughns: Whilst the code is very similar and the expectations could be combined, I would argue that the purpose of each test case is clearer in the way that it is currently written.

If you are concerned about code duplication I think I can do something about that whilst keeping things separate. I will push an extra commit.

hughns:
I have tried to make the intent of each test case clearer in the latest commit by factoring out the common base props for each test case.

src/tile/MediaView.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+  test("is accessible", async () => {
+    const { container } = render(<MediaView {...baseProps} />);
+    expect(await axe(container)).toHaveNoViolations();
+  });
+
+  describe("placeholder track", () => {
+    test("neither video nor avatar are shown", () => {
+      render(<MediaView {...baseProps} video={trackReferencePlaceholder} />);
+      expect(screen.queryByTestId("video")).toBeNull();
+      expect(screen.getByTestId("avatar")).not.toBeVisible();
```
robintown: Can we use role-based queries for these rather than test IDs?

src/tile/MediaView.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  describe("unencryptedWarning", () => {
+    test("is shown and accessible", async () => {
+      const { container } = render(
+        <TooltipProvider>
+          <MediaView {...baseProps} unencryptedWarning={true} />
+        </TooltipProvider>,
+      );
+      expect(await axe(container)).toHaveNoViolations();
+      expect(screen.getByTestId("unencrypted_warning_icon")).toBeTruthy();
```
robintown: Could also use a role or label query here

src/tile/MediaView.test.tsx
hughns: I don't see a role for video elements so I am leaving that as a testid.

toger5 (approved):
The video with the test id makes sense to me.
I also did a quick search and nothing obvious came up.

PR #2814, opened 2024-11-21
Refactor the speaker detection logic into observeSpeaker and add tests

hughns:
@robintown the tests pass, but some of the values were off by 1ms from what I was expecting. Please can you sanity check them?

### DIFF
--- a/src/state/observeSpeaker.test.ts
+++ b/src/state/observeSpeaker.test.ts
@@ -19,8 +19,9 @@ const yesNo = {
 describe("observeSpeaker", () => {
   describe("does not activate", () => {
     const expectedOutputMarbles = "n";
-    test("no speaking", () => {
-      const speakingInputMarbles = " n";
+    test("starts correctly", () => {
+      // should default to false when no input is given
+      const speakingInputMarbles = "";
       withTestScheduler(({ hot, expectObservable }) => {
         expectObservable(
           observeSpeaker(hot(speakingInputMarbles, yesNo)).pipe(
@@ -30,8 +31,8 @@ describe("observeSpeaker", () => {
       });
     });
 
-    test("speaking for 1ms", () => {
-      const speakingInputMarbles = " y n";
+    test("after no speaking", () => {
+      const speakingInputMarbles = "n";
       withTestScheduler(({ hot, expectObservable }) => {
         expectObservable(
           observeSpeaker(hot(speakingInputMarbles, yesNo)).pipe(
@@ -41,8 +42,8 @@ describe("observeSpeaker", () => {
       });
     });
 
-    test("speaking for 999ms", () => {
-      const speakingInputMarbles = " y 999ms n";
+    test("with speaking for 1ms", () => {
+      const speakingInputMarbles = "y n";
       withTestScheduler(({ hot, expectObservable }) => {
         expectObservable(
           observeSpeaker(hot(speakingInputMarbles, yesNo)).pipe(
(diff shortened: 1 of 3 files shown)


src/state/observeSpeaker.ts
```diff
@@ -0,0 +1,27 @@
...
+  return isSpeakingObservable.pipe(
+    audit((s) =>
+      merge(
+        timer(s ? 1000 : 60000),
+        // If the speaking flag resets to its original value during this time,
+        // end the silencing window to stick with that original value
+        isSpeakingObservable.pipe(filter((s1) => s1 !== s)),
+      ),
+    ),
+    startWith(false),
```
toger5: Thinking through this a bit more seems like we might want to have a `distinctUntilChange` for this one.
Or in general I am wondering how it behaves when we feed multiple speaking=true (or false) events in a row?

I suspect it will emit `undefined` since we dont we will already send the latest audited event with the first timeout but since we starteed another timeout that will then not get any new queued event?

toger5: I think this suspicion is not true. But the filter is basically a manual `distinctUntilChanged`.
And we can still end up with double emissions. In `observeSpeaker`:
 - we start and **emit false** (or come from true and waited for 60s and emitted false. either option works)
 - set to true (won't emit for 10s because of audit)
 - we get a false so we do emit with `isSpeakingObservable.pipe(filter((s1) => s1 !== s)),` which will **emit false** (because this is the last thing the audit has seen.
 
 So we emit false twice.

toger5: Do you think the following is easier to understand:
```suggestion
export function observeSpeaker(
  isSpeakingObservable: Observable<boolean>,
): Observable<boolean> {
  const toggleObservable = isSpeakingObservable.pipe(distinctUntilChanged());

  return toggleObservable.pipe(
    audit((s) => merge(timer(s ? 1000 : 60000), toggleObservable)),
    startWith(false),
    distinctUntilChanged(),
```

toger5: With comments:
```suggestion
export function observeSpeaker(
  isSpeakingObservable: Observable<boolean>,
): Observable<boolean> {
  const toggleObservable = isSpeakingObservable.pipe(distinctUntilChanged());

  return toggleObservable.pipe(
    // Either change to the new value after the timer or re-emit the same value if it toggles back
    // (audit will return the latest (toggled back) value) before the timeout.
    audit((s) => merge(timer(s ? 1000 : 60000), toggleObservable)),
    // Filter the re-emissions (marked as: | ) that happen if we toggle quickly (<1s) from false->true->false|->..
    startWith(false),
    distinctUntilChanged(),
```

src/state/observeSpeaker.test.ts
```diff
@@ -0,0 +1,93 @@
...
+          observeSpeaker(hot(speakingInputMarbles, yesNo)).pipe(
+            distinctUntilChanged(),
+          ),
+        ).toBe(expectedOutputMarbles, yesNo);
+      });
+    });
+
+    test("speaking intermittently", () => {
+      const speakingInputMarbles =
+        " y 199ms n 199ms y 199ms n 199ms y 199ms n 199ms y 199ms n 199ms y 199ms n 199ms y 199ms n 199ms y 199ms n 199ms y 199ms n";
```
toger5: We should at least add one test that does multiple `y` in a row. I dont think this can happen but one never knows if there is some diamond'ing happening so we end up in a case where `speakingInputMarbles` emit true twice.

robintown:
The timings look fine to me.

src/state/observeSpeaker.test.ts
hughns: Added in https://github.com/element-hq/element-call/commit/7f2a7b00abdeb999e05f6c4a6619dcca384e05a6

src/state/observeSpeaker.ts
hughns: I originally didn't have the `distinctUntilChanged()` at the end because where the observer is used it is already piped into `this.scope.state()` (see https://github.com/element-hq/element-call/blob/a587956c759d0d58f777828a73f33097a5b1bf1e/src/state/CallViewModel.ts#L251) which does the distinct again (see https://github.com/element-hq/element-call/blob/a587956c759d0d58f777828a73f33097a5b1bf1e/src/state/ObservableScope.ts#L34-L39).

In https://github.com/element-hq/element-call/commit/a587956c759d0d58f777828a73f33097a5b1bf1e I have made this part of the observable itself.

However, this does mean that we do the `distinctUntilChanged()` twice.

hughns: I've incorporated your other suggestions too.

PR #2823, opened 2024-11-22
Add explicit code split on matrix-sdk-crypto-wasm to allow caching between deploys

hughns:


vite.config.js
```diff
@@ -82,6 +82,9 @@ export default defineConfig(({ mode }) => {
             // Default naming fallback
             return "assets/[name]-[hash][extname]";
           },
+          manualChunks: {
```
sandhose, line 85: This *should* be not necessary once https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/167 lands, as it will ship the WASM binary separately

hughns: and in the meantime we can use this 👌

hughns: I added an inline comment so that we don't forget.

PR #2828, opened 2024-11-22
Introduce intent URL param and make it change the default lobby behaviour

hughns:
Proposal: if the user is starting a call in widget mode then we should take them straight into the call rather than the lobby.

Otherwise, the widget host will have sent the call notify (which will have rung other people) but the person who started the call not actually joined yet... which is worked around in EW/ED at the moment.

Best viewed [without whitespace changes](https://github.com/element-hq/element-call/pull/2828/files?w=1).


fkwp:
sounds like a good plan

src/UrlParams.ts
```diff
@@ -263,10 +263,10 @@ export const getUrlParams = (
       "skipLobby",
       isWidget && intent === UserIntent.StartNewCall,
     ),
-    returnToLobby: parser.getFlagParam("returnToLobby"),
+    returnToLobby: isWidget ? parser.getFlagParam("returnToLobby") : true,
     theme: parser.getParam("theme"),
-    viaServers: parser.getParam("viaServers"),
-    homeserver: parser.getParam("homeserver"),
```
toger5: can we double check we do not use this for well know fetching (probably we just use the username for it but?)

robintown (approved):

src/UrlParams.ts
hughns: When working on another PR I noted that it came via `MatrixClient.getDomain()` which comes from the user ID at https://github.com/matrix-org/matrix-js-sdk/blob/349a86c119e6ac53d8d0677f6b6db5944c3ddcd1/src/client.ts#L1856-L1861

So, I think unaffected by this change.

PR #2831, opened 2024-11-22
Don't trigger keyboard shortcuts if modifiers are held

### LINKED ISSUE #2824: Keyboard shortcuts for sending emoji reactions breaks ability to change tab.
### Steps to reproduce

1. Open SPA in Chrome
2. Try to switch tab on macOS with cmd-1, cmd-2, etc
3. Discover they now send emoji reactions...

#### What did you expect?

Sending emoji reactions based on numbers would be fine. But if a modifer key is being pressed, it definitely shouldn't send, given it override basic things like changing tab.

### Operating system

macOS 15.1.1

### Browser information

Element Call version: v0.7.0

### URL for webapp

call.element.io

### Will you send logs?

No

robintown:
None of these keyboard shortcuts expect modifier keys, so they should in fact expect the absence of modifiers.

Closes https://github.com/element-hq/element-call/issues/2824

### DIFF
--- a/src/useCallViewKeyboardShortcuts.test.tsx
+++ b/src/useCallViewKeyboardShortcuts.test.tsx
@@ -93,6 +93,16 @@ test("reactions can be sent via keyboard presses", async () => {
   }
 });
 
+test("reaction is not sent when modifier key is held", async () => {
+  const user = userEvent.setup();
+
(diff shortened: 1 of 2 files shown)


hughns (approved):
👍

PR #2834, opened 2024-11-23
Mute state improvements including muting if skipLobby=true in SPA

hughns:
If skipLobby=true then always default to muted.
Use the mute states from the config file.
Refactor logic so that defaults are handled by useMuteStates().
Test cases.


docs/url-params.md
```diff
@@ -56,7 +56,7 @@ There are two formats for Element Call urls.
...
-| `skipLobby`               | `true` or `false`                                                                                    | No, defaults to `false` | No, defaults to `false` | Skips the lobby to join a call directly, can be combined with preload in widget.                                                                                                                                                                                                                                                                                                                    |
+| `skipLobby`               | `true` or `false`                                                                                    | No, defaults to `false` | No, defaults to `false` | Skips the lobby to join a call directly, can be combined with preload in widget. When `true` the audio and video inputs will be muted by default.                                                                                                                                                                                                                                                   |
```
toger5: Can we mention here:
Just so we immediately see this once we tackle unmuted DM call starts.
And whoever works on this does not get confused/stuck
```suggestion
| `skipLobby`               | `true` or `false`                                                                                    | No, defaults to `false` | No, defaults to `false` | Skips the lobby to join a call directly, can be combined with preload in widget. When `true` the audio and video inputs will be muted by default. (This means there currently is no way to start without muted video if one wants to skip the lobby. Also not in widget mode.)                                                                                                                                                                                                                                             |
```

src/room/GroupCallView.tsx
```diff
@@ -135,44 +135,41 @@ export const GroupCallView: FC<Props> = ({
...
+    }: JoinCallData): Promise<void> => {
       // XXX: I think this is broken currently - LiveKit *won't* request
       // permissions and give you device names unless you specify a kind, but
       // here we want all kinds of devices. This needs a fix in livekit-client
       // for the following name-matching logic to do anything useful.
       const devices = await Room.getLocalDevices(undefined, true);
-      const { audioInput, videoInput } = requestedDeviceData;
-      if (audioInput === null) {
-        latestMuteStates.current!.audio.setEnabled?.(false);
-      } else {
```
toger5, line 149: Why can we remove this call?
Do we know the default is false anyways in this situation.

I feel like the ref pointer here might also deserve a comment while we are at it:
```
  const latestMuteStates = useRef<MuteStates>();
  latestMuteStates.current = muteStates;
```

Explaining what is stored in latestMuteStates and why we not use `muteStates` directly.

toger5 (approved):
This looks good (and is also a good catch)

The comments on the latestMuteState are optional so I will approve already

src/room/GroupCallView.tsx
hughns: I'm not sure we can remove it. We don't know that the default is false if the input isn't matched.

I've added a TODO note for the comment. I don't actually know why we are using a ref.

PR #2838, opened 2024-11-27
Use correct key for matrix rtc foci in code comment.

toger5:


hughns (approved):

toger5:
It makes sense why this ended up as a PR on the tile fix but was not my intention. sorry...

toger5:
thanks for still merging it! Not worth doing the effort of closing and opening another pr for this small/obvious fix.

PR #2846, opened 2024-11-28
 Mute state improvements including muting if skipLobby=true in SPA (Same as #2834)

toger5:


toger5:
This bypassed reviewing since it is the same as #2834. Which already was reviewed.
It was merged on a different branch however so we recreated this PR.

PR #2848, opened 2024-11-28
Allow join unmuted in widget mode

toger5:


hughns (changes requested):
Three things:

- the test doesn't pass
- I think there is a simpler way to do this just at the UrlParams level
- the documentation update is missing.

As I am off today I will do a quick alternative PR.

src/room/MuteStates.ts
```diff
@@ -74,13 +74,14 @@ export function useMuteStates(): MuteStates {
   const devices = useMediaDevices();
 
   const { skipLobby } = useUrlParams();
-
+  // In SPA without lobby we need to protect from unmuted joins (Privacy).
+  const allowStartUnmuted = !skipLobby || widget !== null;
```
hughns:
```suggestion
  const allowStartUnmuted = !skipLobby || !!widget;
```

hughns:
Actually, I've run out of time to do something today. What I noticed was that the URL parameters were not being parsed from the mocked router URL.

toger5:
Thanks for the finding about the router URL.
The tests work locally lets see what i have setup differently!

src/room/MuteStates.ts
```diff
@@ -75,7 +75,7 @@ export function useMuteStates(): MuteStates {
 
   const { skipLobby } = useUrlParams();
   // In SPA without lobby we need to protect from unmuted joins (Privacy).
-  const allowStartUnmuted = !skipLobby || !!widget;
+  const allowStartUnmuted = !skipLobby || !widget;
```
toger5: This does not look right->
 - in SPA - `widget = undefined/null`
 - allowStartUnmuted will always be `true` (because !widget == true)
 - but we want it to be false in the spa.

hughns: Yes, I already reverted it.

toger5: I saw it too late sorry.

toger5:
This already checks that the widget is started in widget mode.
Hence the way it is now, it would not introduce the risk of allowing to share a unmuted link.

hughns:
Following discussion: marking draft until we revisit this topic as part of expected flows for telephone/DM use cases.

toger5:
This should NOT be done using url parameters since this always allows to "mock" them. The better solution is to check if we are doing an embedded build. Closing this in favor of: https://github.com/element-hq/element-call/issues/3442

PR #2856, opened 2024-12-02
Authenticate media requests when loading avatars

### LINKED ISSUE #2845: Support for authenticated media / profile pictures fail to load
### Steps to reproduce

0. Upload a fresh profile picture in Element (possibly optional)
1. Join an Element Call
2. See error:
image

### Since there doesn't seem to be an "extra information" text box I'll provide the extra information I know here
The browser version and Element X Android (probably, unverified) try to use `/_matrix/media/v3/thumbnail/{serverName}/{mediaId}` which is deprecated according to matrix-org/matrix-spec-proposals/pull/3916 and the [specification](https://spec.matrix.org/v1.12/client-server-api/#get_matrixmediav3thumbnailservernamemediaid). The desktop client (Element version 1.11.86, Linux) doesn't seem to have the same issue since it uses `/_matrix/media/v1/thumbnail/{serverName}/{mediaId}` (note the v1 instead of v3).

#### What did you expect?
My profile picture to load
#### What happened instead?
My profile picture doesn't load

### Operating system

Reproducible on Brave and Element X Android

### Browser information

Brave Version 1.73.91 (Chromium: 131.0.6778.85) and Element X version 1.6.22

### URL for webapp

_No response_

### Will you send logs?

No

Half-Shot:
Fixes #2845 

This introduces support for authenticated matrix media for avatars. All media requests are now made via `fetch` and then the result is stored as a local object url.


hughns (commented):
What is the intended behaviour when EC is running as a widget?

Half-Shot:
> What is the intended behaviour when EC is running as a widget?

:cry: , I suspect. There isn't an API presently to fetch media over the widget API and we don't have a token to use that could do it. I wish for a world where we get scoped OIDC tokens and can just use the actual API. https://github.com/matrix-org/matrix-widget-api/issues/115 would be useful.

I think for the *moment* we could either still support unauthenticated media for widgets while new avatars slowly replace the old ones and we lose access to them. Or we could just show placeholders for all avatars until we can properly support authenticated media.

hughns:
> > What is the intended behaviour when EC is running as a widget?
> 
> 😢 , I suspect. There isn't an API presently to fetch media over the widget API and we don't have a token to use that could do it. I wish for a world where we get scoped OIDC tokens and can just use the actual API. [matrix-org/matrix-widget-api#115](https://github.com/matrix-org/matrix-widget-api/issues/115) would be useful.
> 
> I think for the _moment_ we could either still support unauthenticated media for widgets while new avatars slowly replace the old ones and we lose access to them. Or we could just show placeholders for all avatars until we can properly support authenticated media.

If the PR can be updated to make it work with the widget mode fallback (with appropriate tests too), then I think we could merge this PR without having to figure out the whole widget API solution.

hughns:
@fkwp as this is currently proposed user profile thumbnails will no longer show when running in widget/embedded mode. Is this acceptable? or should EC be continuing to use non-authenticated media requests when running as a widget?

Marking as draft with X-Needs-Product label until answered.

Half-Shot:
https://matrix.to/#/!tDLCaLXijNtYcJZEey:element.io/$rljmYPXis_Ja28l-Z7_XIqkdgC7Y3TmqFFxssHj1o4Q?via=element.io&via=matrix.org&via=half-shot.uk

So I believe we were okay to merge this now, and then fix the thumbnail in widgets later as we didn't deem it critical enough. I was going to look at expanding the widget API to do so as a next-up item.

hughns:
> So I believe we were okay to merge this now, and then fix the thumbnail in widgets later as we didn't deem it critical enough. I was going to look at expanding the widget API to do so as a next-up item.

👌 I made https://github.com/element-hq/element-call/issues/2876 to add the feature back in.

src/Avatar.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+        src={member?.getMxcAvatarUrl()}
+      />
+    </TestComponent>,
+  );
+  const element = screen.getByRole("img", { name: "@alice:example.org" });
+  expect(element.tagName).toEqual("SPAN");
+  expect(client.mxcUrlToHttp).toBeCalledTimes(0);
+});
+
+test("should just render a placeholder when thumbnaisl are not supported", () => {
```
hughns:
```suggestion
test("should just render a placeholder when thumbnails are not supported", () => {
```

hughns (approved):
One typo. Otherwise looks good. 👍

dillfrescott:
Will this merge be in a release soon?

hughns:
> Will this merge be in a release soon?

There is currently a release blocker which is now being tracked at https://github.com/element-hq/element-call/issues/2972

dillfrescott:
Thank you!

PR #2858, opened 2024-12-02
Alternative design for "waiting for media..." state

hughns:
Currently there is a "waiting for media..." label:

image

This PR replaces it with an animated border:

Screen Recording 2024-12-06 at 12 07 07

Assuming the design is desirable then the following items are outstanding:

- turning off the "waiting for media..." label


hughns:
Some thoughts:

- I don't think this is obvious enough what it means.
- It isn't visible when in 1:1 mode.

toger5:
> I don't think this is obvious enough what it means.

I think the exact explanation what it is (a tile that is shown because there is a membership in the state but we are missing the associated Livekit participant) is too much information.
But I suppose its more about the gradient does mot make it clear that it is loading?
The issue might be that if its shown for less then 1 second its hard to grasp that there is an animation at all.
When its shown for longer it makes more sense i think.

hughns:
The loading spinner in https://github.com/element-hq/element-call/pull/2869 might be a better option.

hughns:
Closing as https://github.com/element-hq/element-call/pull/2869 is preferred.

PR #2812, opened 2024-11-20
Add background blur feature for supported devices

toger5:
Fixes https://github.com/element-hq/element-call/issues/121

This adds background blur via @livekit/track-processors for supported browsers.
The option can be found in the video tab of the settings.

The fallback will be a greyed out checkbox (on firefox for instance)
This way firefox users get informed why it does not work on their side if other users tell them where to set it up.

Sadly this is not exactly done the way we would want to set it up see: https://github.com/livekit/track-processors-js/issues/50

- [x] bundle local assets not CDN
- [x] stringify related hang on startup

PR #2859, opened 2024-12-02
backgroundBlur squashed

toger5:
@hughns this is the version we use for tomorrows demo testing (in one commit)
It does not include the new lk version since that does not seem to fix it.


hughns:
Closing in preference to https://github.com/element-hq/element-call/pull/2812

PR #2860, opened 2024-12-02
Add support for playing a sound when the user exits a call.

Half-Shot:
Implementation for https://github.com/element-hq/element-call/issues/343#issuecomment-2511975142

Depends on https://github.com/element-hq/element-call/pull/2863

This is the trickiest thing to get right by far. We can't use the view model as we do for the other sounds since the act of hanging up will kill that off first. The next problem is we have many states that can occur from hanging up depending on what mode the application is in:

- For SPA mode or video rooms / group call rooms, this is a trivial case of just playing a sound through the long lived audio context.
- For 1:1 calls where the widget closes when we hangup...we have a problem. We need to keep the widget alive long enough to play the sound but without making it hang around and feel unresponsive.

One option here is that for call leaves only, we make this the client's problem. The problem with this is we'd need to provide those audio files up to Element Web / EX etc, which is a nightmare if we decide to change the sound at any point. Supposedly we could make them part of compound, but that's a whole new design thing to consider.

So, this PR does the first two trivial things and currently holds the window open for 1:1s.

An alternative solution is not to play the hangup sound for 1:1s and just let the window drop. This would be asymmetric, but to be honest it's fairly visually obvious the call has dropped.


fkwp:
It seems that the playout device for the join/leave sounds is the OS default device rather than the configured audio device from the settings.

Half-Shot:
> It seems that the playout device for the join/leave sounds is the OS default device rather than the configured audio device from the settings.

#2863 will fix this.

hughns (commented):
This changes generally look sane. I will take another look once the dependent PR is merged.

src/Slider.tsx
```diff
@@ -48,7 +48,7 @@ export const Slider: FC<Props> = ({
   disabled,
 }) => {
   const onValueChange = useCallback(
-    ([v]: number[]) => onValueChangeProp(v),
+    ([v]: number[]) => onValueChangeProp?.(v),
```
hughns:
```suggestion
    ([v]: number[]) => onValueChangeProp(v),
```

hughns (commented):
Do we need a test to assert that the left sound gets played/scheduled?

src/Slider.tsx
Half-Shot: i wonder if this is a trivial linting rule

hughns (approved):
Fab 👍

PR #2863, opened 2024-12-03
Ensure call sound effects are played over the correct sink

Half-Shot:
This PR refactors much of the work I did on sound effects to make them play over an AudioContext instead, which let's us pick a preferred sync.

To fix:

- [x] Join/leave sounds
- [x] Hand raised sounds
- [x] Reaction sounds


src/useAudioContext.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { useAudioContext } from "./useAudioContext";
+import { FC } from "react";
+import { render } from "@testing-library/react";
+import { deviceStub, MediaDevicesContext } from "./livekit/MediaDevicesContext";
+import { afterEach } from "node:test";
+import { soundEffectVolumeSetting } from "./settings/settings";
+
+/**
+ * Test explanation.
+ * This test suite checks that the useReactions hook appropriately reacts
```
Half-Shot: this is wrong

src/useAudioContext.tsx
```diff
@@ -0,0 +1,156 @@
...
+  if (a.canPlayType("audio/ogg") === "maybe") {
+    return "ogg";
+  }
+  // Otherwise just assume MP3, as that has a chance of being more widely supported.
+  return "mp3";
+}
+
+type PrefetchedSounds<S extends string> = Promise<Record<S, ArrayBuffer>>;
+
+// We prefer to load these sounds ahead of time, so there
```
Half-Shot: Wrong comment?

hughns (changes requested):
Please merge/rebase the latest from `livekit` branch. The tests may then need a bit of refactoring to make sure that they work with the way that MatrixRTC session members are mocked instead of LiveKit participants.

src/Slider.tsx
```diff
@@ -16,6 +16,9 @@ interface Props {
   className?: string;
   label: string;
   value: number;
+  /**
+   * Event handler called when the value changes during an interaction.
+   */
   onValueChange: (value: number) => void;
```
hughns, line 22: Should this be made optional to match the `?.` below?

src/settings/SettingsModal.tsx
```diff
@@ -83,8 +83,11 @@ export const SettingsModal: FC<Props> = ({
 
   const devices = useMediaDevices();
   useMediaDeviceNames(devices, open);
-
   const [soundVolume, setSoundVolume] = useSetting(soundEffectVolumeSetting);
+  const [soundVolumeRaw, setSoundVolumeRaw] = useState(soundVolume);
+
+  // Debounce saving the sound volume as it triggers certain components to reload.
+  useEffect(() => {});
```
hughns: I read the comment, then I look at the line below and I don't get how it achieves this? What am I missing?

src/useAudioContext.tsx
```diff
@@ -0,0 +1,174 @@
...
+// is no delay on call join.
+const PreferredFormat = getPreferredAudioFormat();
```
hughns:
```suggestion
const preferredFormat = getPreferredAudioFormat();
```

src/useAudioContext.tsx
```diff
@@ -0,0 +1,174 @@
...
+  sounds: Record<S, SoundDefinition>,
+): PrefetchedSounds<S> {
+  const buffers: Record<string, ArrayBuffer> = {};
+  await Promise.all(
+    Object.entries(sounds).map(async ([name, file]) => {
+      const { mp3, ogg } = file as SoundDefinition;
+      // Use preferred format, fallback to ogg if no mp3 is provided.
+      // Load an audio file
+      const response = await fetch(
+        PreferredFormat === "ogg" ? ogg : (mp3 ?? ogg),
```
hughns:
```suggestion
        preferredFormat === "ogg" ? ogg : (mp3 ?? ogg),
```

src/useAudioContext.tsx
```diff
@@ -0,0 +1,174 @@
...
+  if (a.canPlayType("audio/ogg") === "maybe") {
+    return "ogg";
+  }
+  // Otherwise just assume MP3, as that has a chance of being more widely supported.
+  return "mp3";
+}
+
+type PrefetchedSounds<S extends string> = Promise<Record<S, ArrayBuffer>>;
+
+// We prefer to load these sounds ahead of time, so there
```
hughns:
```suggestion
```

src/useAudioContext.tsx
```diff
@@ -0,0 +1,174 @@
...
+// is no delay on call join.
```
hughns:
```suggestion
```

src/room/ReactionAudioRenderer.tsx
```diff
@@ -5,70 +5,55 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+
+const SoundMap = Object.fromEntries([
+  ...ReactionSet.filter((v) => v.sound !== undefined).map((v) => [
+    v.name,
+    v.sound!,
+  ]),
+  [GenericReaction.name, GenericReaction.sound],
+]);
+
+const Sounds = prefetchSounds(SoundMap);
```
hughns: Should we only be pre-fetching if the user has the play reaction sounds setting turned on? At the moment it loads even if I have it disabled.

hughns:
```suggestion
const sounds = prefetchSounds(soundMap);
```

src/room/ReactionAudioRenderer.tsx
```diff
@@ -5,70 +5,55 @@ SPDX-License-Identifier: AGPL-3.0-only
...
 
 export function ReactionsAudioRenderer(): ReactNode {
   const { reactions } = useReactions();
   const [shouldPlay] = useSetting(playReactionsSound);
-  const [effectSoundVolume] = useSetting(effectSoundVolumeSetting);
-  const audioElements = useRef<Record<string, HTMLAudioElement | null>>({});
+  const audioEngineCtx = useAudioContext({
+    sounds: Sounds,
```
hughns:
```suggestion
    sounds,
```

src/room/ReactionAudioRenderer.tsx
```diff
@@ -5,70 +5,55 @@ SPDX-License-Identifier: AGPL-3.0-only
...
-  playReactionsSound,
-  soundEffectVolumeSetting as effectSoundVolumeSetting,
-  useSetting,
-} from "../settings/settings";
+import { playReactionsSound, useSetting } from "../settings/settings";
 import { GenericReaction, ReactionSet } from "../reactions";
+import { prefetchSounds, useAudioContext } from "../useAudioContext";
+import { useLatest } from "../useLatest";
+
+const SoundMap = Object.fromEntries([
```
hughns:
```suggestion
const soundMap = Object.fromEntries([
```

hughns (changes requested):
I really like the direction of the abstraction that the refactor has taken. 👍 

A few questions and clean up needed.

src/Slider.tsx
Half-Shot: I think that's actually a mistake. We need to always handle change as otherwise the value of the slider doesn't update on input and then you can't actually move it. So change is always required but commit is optional.

I've updated the usage of `?.`

src/settings/SettingsModal.tsx
Half-Shot: Ah, I started writing this code and then found `onValueCommit` :)

src/room/ReactionAudioRenderer.tsx
Half-Shot: Hm, actually in the case of reactions it's probably fine to not preload them anyway since unlike call join/leaves you *probably* won't be trying to play them before you have loaded them.

Let's load on first component load.

hughns (approved):
I've pushed a settings wording tweak. But now looks good 👍

PR #2866, opened 2024-12-04
Rename `en-GB` language to `en`

### LINKED ISSUE #2865: No `en` language provided
If I have my browser preferences set like this:

image

I end up using German translations instead of English (which is what I expect). This is because we only have `en-GB` not `en` or `en-US`.

As reported by @stefanceriu

hughns:
Fixes https://github.com/element-hq/element-call/issues/2865

If we want to have GB and US specific terms later on we can then add then in as `en-GB` and `en-US`.


fkwp (approved):
lgtm

PR #2869, opened 2024-12-04
Inform user that their camera is starting in Lobby

hughns:
Instead of just showing a grey box.

image

n.b. it is expected that this will be visible for a second or so whilst the browser and OS set up access to the camera.


fkwp:
From internal discussion: I think text sounds apologetic. Maybe we can have a little spinner instead?

americanrefugee:
Maybe show the avatar of the room you're about to join. Also consider adding a small spinner icon next to / above the message "Camera is starting".

toger5:
Toggling the camera in the lobby it now looks like the avatar is appearing disappearing which i think is very nice feedback to switch from muted -> video.
I think putting a spinner in the center of the previous avatar would look really good.

americanrefugee:
[How about this?](https://www.figma.com/design/fq7cTHPNHZX3ttFW1gBEnv/Compound-Element-Call?node-id=3751-7739&t=Su4bHfSG2373vHCv-4)

Loading video

Also, I changed my mind. The avatar should be YOUR avatar, not the room.

- The room avatar is already in the header
- The lobby is a preview of you, not the call

toger5:
@americanrefugee does it look better with the greyed out image?
I think just the loader and the label would be enough?

americanrefugee:
> @americanrefugee does it look better with the greyed out image? I think just the loader and the label would be enough?

Yes, it should be grayed out (50% opacity). Otherwise it will hard/impossible to see the spinner against the image - especially if the user avatar is light/white.

hughns:
Benchmarking other applications:

Meet is showing the "Camera is starting" label:
image

Teams just goes to a grey tile until the video loads (a bit like what we have now):
image

Zoom doesn't provide feedback in the video tile (it continues to show the avatar), but has three states on the video button:

Off: image

Turning on: it's a grey intermediate version

On: image

hughns:
The avatar is currently so large that you can't fit text below it at most resolutions.

The colour in this is `text-secondary` with `space-10x` below the top:

image

I will try a spinner on top of the avatar next.

hughns:
It now looks like this which is an approximation of comments above:

image

fkwp:
fine for me

src/Avatar.module.css
```diff
@@ -0,0 +1,19 @@
...
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  opacity: 0.5;
+  color: var(--cpd-color-fg-primary);
```
robintown: This token doesn't exist (though it should, and hopefully will next time by the next time I look into tokens with Aaron: https://github.com/element-hq/compound/issues/313)

src/room/VideoPreview.tsx
```diff
@@ -53,6 +55,11 @@ export const VideoPreview: FC<Props> = ({
     };
   }, [videoTrack]);
 
+  const cameraIsStarting = useMemo(
+    () => muteStates.video.enabled && (!videoTrack || !videoEl.current),
+    [muteStates.video.enabled, videoTrack, videoEl],
+  );
```
robintown: The current value of a ref isn't generally something we can consume reactively. Maybe `videoEl.current` doesn't need to be part of the condition at all?

src/Avatar.tsx
```diff
@@ -123,14 +129,23 @@ export const Avatar: FC<Props> = ({
...
-      size={`${sizePx}px`}
-      src={avatarUrl}
-      style={style}
-      {...props}
-    />
+    <div>
+      {loading && (
+        <div className={styles.loading}>
+          <InlineSpinner
+            size={typeof sizePx === "number" ? sizePx / 3 : undefined}
```
robintown: If this loading indicator is designed to work with `size` being a number, maybe it's better to extract this bit from the Avatar component where it's more reasonable for it to be an ad-hoc thing. Then we could avoid the extra `<div>` wrapper being added to all avatars as well.

hughns:
I've done a refactor based on @robintown's feedback. My intention is that the design will be used within a call as well. But this will go in another PR.

toger5 (approved):
The code looks good. 
I am not sure if we can do anything about the character in the span not being 100% centered...
That is a bit annoying. not a merge blocker but sth one could try in a follow up PR.
Proabably the Span needs to become a div -> and the span needs to be added a child to the diff with some custom margins.

PR #2882, opened 2024-12-09
Ignore URL params where they are not applicable for the mode (widget vs SPA)

hughns:


robintown (approved):
I guess this is a little clearer about how they work?

hughns:
> I guess this is a little clearer about how they work?

It should reinforce what is in the url-params.md and also prevent people (ab)using unanticipated combinations.

A future step would be to actual error if bad param combinations are given rather than silently soldiering on like we do today.

PR #2883, opened 2024-12-09
Add missing full stops from end of settings descriptions

hughns:
For consistency


Half-Shot (approved):
Feels a bit like linter rule territory, but I'm not sure what you'd key off? "_body" / "_description"

Anyway, this fixes the things for now.

robintown (approved):

PR #2885, opened 2024-12-09
Refactor reactions / hand raised to use rxjs and start ordering tiles based on hand raised.

### LINKED ISSUE #2715: Port reactions/hand-raised code from a provider to the call view model.
https://github.com/element-hq/element-call/pull/2542 added the raise hands feature, but it's currently implemented as a provider and the logic is held outside the view model. 

Generally all state belonging to a call should be inside a view model, so this should be ported across. This should then allow us to change the sort order of tiles based on hands raised too.

### LINKED ISSUE #2843: Raising hand should move tile into view
### Steps to reproduce

If a user raises their hand, and their tile is off the screen, it should be moved into view.

#### What did you expect?

I haven't checked explicitly, but i'm assuming that particularly if you are in spotlight mode, if someone raises their hand and they're off the bottom of the list, they won't get moved into view.

#### What happened instead?

I suspect folks were raising hands on my call and i couldn't see them.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

call.element.io

### Will you send logs?

No

Half-Shot:
Fixes https://github.com/element-hq/element-call/issues/2715
Fixes https://github.com/element-hq/element-call/issues/2843

This change will adjust our code so that consumers of reactions and hand raised state use the CallViewModel as much as possible, which allows us to finally do fun stuff like order tiles by hand raised.

The changes are as follows:

- `useReactions` is no more. There is now `useReactionsReader` which consumes reaction/HR events, and emits them in `BehaviourSubjects`. The sending code has been refactored into a `ReactionsSenderProvider`.
- `CallViewModel` now consumes these two subjects, and includes a range of observables for our presentation components. The presentation components now contain very little selection logic, and focus on...presenting :)
- Raising your hand now causes your tile to rank higher than regular video participants, but below speakers and presenters.

And, dear reviewer, I'm very sorry for the diff :(. I hope it's worth it.


src/state/CallViewModel.ts
```diff
@@ -1112,6 +1145,128 @@ export class CallViewModel extends ViewModel {
...
+  }
+
+  public removeHandRaised(userId: string, date: Date | null) {
+    this.handsRaisedSubject.next([{ userId, date }]);
+  }
+
+  public addReaction(userId: string, reaction: ReactionOption, ttl: number) {
+    this.reactionsSubject.next([{ userId, reaction, ttl }]);
+  }
+
```
Half-Shot: This has been my approach so far where the VM has some subjects, and the useReactions provider has been updating these subjects with the latest values. The dependent components then subscribe to the relevant observable, and all is well.

The hook is still used to to *send* reactions / hand raised at the moment.

src/useReactions.tsx
```diff
@@ -92,19 +99,38 @@ export const ReactionsProvider = ({
     clientState?.state === "valid" && clientState.supportedFeatures.reactions;
   const room = rtcSession.room;
   const myUserId = room.client.getUserId();
+  const myDeviceId = room.client.getDeviceId();
```
Half-Shot: Potentially this hook could be split up, and listening for incoming reactions could be moved somewhere else (maybe even MatrixRTCSession?).

Half-Shot: I've split the sending and receiving logic pieces into a provider and a producer-style hook, using the call view model for state.

src/reactions/useReactionsReader.ts
```diff
@@ -1,120 +1,83 @@
...
+export default function useReactionsReader(rtcSession: MatrixRTCSession): {
+  raisedHands$: Observable<Record<string, RaisedHandInfo>>;
+  reactions$: Observable<Record<string, ReactionInfo>>;
+} {
+  const raisedHandsSubject$ = useRef(
+    new BehaviorSubject<Record<string, RaisedHandInfo>>({}),
+  );
+  const reactionsSubject$ = useRef(
+    new BehaviorSubject<Record<string, ReactionInfo>>({}),
+  );
```
robintown: These `useRef`s could be `useInitial` to avoid calling the BehaviorSubject constructor on every render

src/reactions/useReactionsReader.ts
```diff
@@ -1,120 +1,83 @@
...
+    .subscribe((reactions) => {
+      const date = new Date();
+      const nextEntries = Object.fromEntries(
+        Object.entries(reactions).filter(([_, hr]) => hr.expireAfter > date),
+      );
+      if (Object.keys(reactions).length === Object.keys(nextEntries).length) {
+        return;
+      }
+      reactionsSubject$.current.next(nextEntries);
+    });
```
robintown: It seems suspicious that `pipe` and `subscribe` are called on every render

src/reactions/useReactionsSender.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  const myRaisedHand = useObservableEagerState(
+    vm.handsRaised$.pipe(
+      map((v) =>
+        myMembershipIdentifier !== undefined
+          ? v[myMembershipIdentifier]
+          : undefined,
+      ),
+    ),
+  );
```
robintown: Likewise these `pipe`-d observables will be recreated on every render. `useObservable` from observable-hooks tries to support this use case of transforming observables in React code without recomputing them—but, as long as we're in React, and all you need to do is `map`, how about:

```typescript
const reactions = useObservableEagerState(vm.reactions$)
const myReaction = useMemo(() => ..., [reactions])
const myRaisedHand = useMemo(() => ..., [reactions])
```

src/state/CallViewModel.test.ts
```diff
@@ -782,3 +791,63 @@ it("should show at least one tile per MatrixRTCSession", () => {
...
+
+    withCallViewModel(
+      of([aliceParticipant, bobParticipant]),
+      of([aliceRtcMember, bobRtcMember]),
+      of(ConnectionState.Connected),
+      new Map(),
+      (vm, { raisedHands$ }) => {
+        schedule("ab", {
+          a: () => {
+            // We imagine that only three tiles (the first three) will be visible
```
robintown:
```suggestion
            // We imagine that only two tiles (the first two) will be visible
```

src/state/CallViewModel.ts
```diff
@@ -1171,12 +1203,94 @@ export class CallViewModel extends ViewModel {
...
+  );
+
+  public readonly handsRaised$ = this.handsRaisedSubject$.pipe();
+
+  /**
+   * Emits an array of reactions that should be visible on the screen.
+   */
+  public readonly visibleReactions$ = showReactions.value$
+    .pipe(switchMap((show) => (show ? this.reactions$ : of({}))))
+    .pipe(
```
robintown: Whenever you're calling `pipe` multiple times in a row you can combine them into one `pipe` call. In other words `observable.pipe(x).pipe(y)` = `observable.pipe(x, y)`.

src/utils/test-fixtures.ts
robintown, line 1: This file should be excluded from coverage stats in our vitest config if not already

src/utils/test-viewmodel.ts
robintown, line 1: Same with this file. Maybe it makes sense to put all the test utils together in a `src/utils/test` directory where it's easy to intuit that any new files will be excluded from coverage?

robintown (commented):
The `useReactionsReader` refactor looks like a good step forward. It's converted the data sources into observables which can now affect the view model, though while staying within React code.

I'm wondering if you feel up to the extra step of moving that hook's code fully out of React and into the view model. It would be a more suitable environment for creating these observables (as the class body runs once, rather than on every render), and it also gets us closer to a world in which there's a clear stratification between Model (matrix-js-sdk, LiveKit) → View Model (RxJS) → View (React), so that the View Models can become entirely external to React, with a unidirectional flow of data.

(That's something I haven't been very explicit about: that the current approach of letting a React effect create the CallViewModel is a bit of a hack, necessary because the app didn't start out with a view model architecture.)

Half-Shot:
> The `useReactionsReader` refactor looks like a good step forward. It's converted the data sources into observables which can now affect the view model, though while staying within React code.
> 
> I'm wondering if you feel up to the extra step of moving that hook's code fully out of React and into the view model. It would be a more suitable environment for creating these observables (as the class body runs once, rather than on every render), and it also gets us closer to a world in which there's a clear stratification between Model (matrix-js-sdk, LiveKit) → View Model (RxJS) → View (React), so that the View Models can become entirely external to React, with a unidirectional flow of data.
> 
> (That's something I haven't been very explicit about: that the current approach of letting a React effect create the CallViewModel is a bit of a hack, necessary because the app didn't start out with a view model architecture.)

I've half gone with your suggestion, but I factored the reaction logic into it's own class which exports it's own observables. My rationale is the code for doing reactions handling isn't trivial, and I find it easier to parse when you have an explicit class to handle and test, which then passes it's observables into the call view model. I think we could, if you felt very strongly here, port the code it the CVM but I worry about the size of that class..

Anyway! No more react which is the main thing :)

src/reactions/useReactionsReader.ts
Half-Shot: No longer react :+1:

src/reactions/useReactionsReader.ts
Half-Shot: Also fixed with a class.

src/reactions/useReactionsSender.tsx
Half-Shot: Fixed.

src/reactions/ReactionsReader.ts
```diff
@@ -0,0 +1,336 @@
...
+  /**
+   * The latest set of reactions.
+   */
+  public readonly reactions$ = this.reactionsSubject$.asObservable();
+
+  public constructor(private readonly rtcSession: MatrixRTCSession) {
+    // Hide reactions after a given time.
+    this.reactionsSubject$
+      .pipe(delay(REACTION_ACTIVE_TIME_MS))
+      .subscribe((reactions) => {
```
robintown: This subscription should be bound to some `ObservableScope` so that it doesn't run forever

robintown (approved):
Yay! Splitting call logic across multiple files is a good call.

PR #2886, opened 2024-12-09
Enable @typescript-eslint/consistent-type-imports lint rule

hughns:
This is to help ensure that we get proper vite/rollup lazy loading by not `import`ing more than we need to.


hughns:
I don't know if this will actually have a benefit to the bundle, but I don't think it will do any harm.

robintown (approved):

PR #2887, opened 2024-12-10
Add some quick-and-dirty debug info for TileStore

robintown:
I'm still in need of more detailed data in order to understand why big layout shifts happen in large calls. This adds a developer option to enable logging and a visual indicator for the state of the TileStore. The indicator should be useful for matching up the behavior I'm seeing in my recordings with the right timestamps.


hughns (approved):
I've rebased on latest and also changed the way that we check for debug being enabled so that it isn't so costly.

PR #2890, opened 2024-12-11
Configure i18next with all locales

hughns:
I noticed that we aren't i18next checking all locales. Should we be? I think it might help keep things consistent.

It also appears to be highlighting some non-visible characters in some entries. I suggest we consider fixing these in another PR rather than bloating this one.


robintown (changes requested):
The blank strings result in blank UI elements in the app rather than English fallbacks:

Screenshot 2024-12-12 at 19-26-41 Element Call Inicio

In general I think it may be a good thing that we as developers don't touch the translation files. Leaving them fully up to Localazy to manage means that there's no potential for merge conflicts. At worst there's just a little bit of extra data in the JSONs when we delete strings (until the next Localazy update happens).

Also, the non-printable characters I can see appear to be non-breaking spaces and zero-width joiners being used for their intended purposes.

hughns:
To get back to my original question. I think @robintown is saying: "No, we shouldn't have i18next checking all locales"

So, this PR isn't needed. @robintown If I have misinterpreted your answer then please do re-open the PR.

PR #2894, opened 2024-12-12
Determine which tiles are on screen in a more stable manner

### LINKED ISSUE #2892: Big layout shifts in large calls
### Steps to reproduce

1. Be in a call of 20+ participants
2. Watch as more people join the call

#### What did you expect?

Something like two tiles swapping places whenever someone joins

#### What happened instead?

Five or so unrelated tiles shift on screen and then shift back on screen with the newly joined user among them, in many cases

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

robintown:
Instead of tracking for each individual tile whether it's visible, just track the total number of tiles that appear on screen. This ought to make the whole thing a lot less dynamic, which is crucial given that our UI renders asynchronously and RxJS doesn't really support cyclic dependencies in any rigorous way.

In particular this ought to make the following kind of situation impossible:
1. There 3 tiles, ABC. A and B are on screen.
2. Now C becomes important. The requested order is now CAB.
3. To reduce the size of the layout shift, the algorithm selects to swap just B and C in the original order, giving ACB. However, the UI is blocked and doesn't render this order yet.
4. For whatever reason, a spurious update of the importance algorithm occurs. It once again requests CAB.
5. Now because the UI was blocked, the layout still thinks that A and B are on screen (rather than A and C). It thinks that C is some weird island of "off-screen territory" in the middle of the tile order. This confuses it into swapping A and C rather than keeping the layout stable.

The reality is that whenever we think N tiles are visible on screen, we're always referring to the first N tiles in the grid. It's best if the code reflects this assumption.

Closes https://github.com/element-hq/element-call/issues/2892

### DIFF
--- a/src/grid/Grid.tsx
+++ b/src/grid/Grid.tsx
@@ -24,7 +24,6 @@ import {
   createContext,
   forwardRef,
   memo,
-  useCallback,
   useContext,
   useEffect,
   useMemo,
@@ -54,7 +53,6 @@ interface Tile<Model> {
   id: string;
   model: Model;
   onDrag: DragCallback | undefined;
-  setVisible: (visible: boolean) => void;
 }
 
 type PlacedTile<Model> = Tile<Model> & Rect;
@@ -88,7 +86,6 @@ interface SlotProps<Model> extends Omit<ComponentProps<"div">, "onDrag"> {
   id: string;
   model: Model;
   onDrag?: DragCallback;
-  onVisibilityChange?: (visible: boolean) => void;
   style?: CSSProperties;
   className?: string;
 }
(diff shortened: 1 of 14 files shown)


robintown:
I'm not sure it's possible to write a unit test for this. The new `setVisibleTiles` callback doesn't allow you to create these "islands of off-screen territory" anymore, which are necessary to reproduce the bug.

robintown:
Well, I now have a test for a beneficial side effect of this PR, at least

hughns (approved):
There is a lot going on in this and I've not fully understood what is happening. The test case makes sense to me.

PR #2901, opened 2024-12-16
Wait for .well-known/matrix/client to load before determining MatrixRTC foci

hughns:
Otherwise we might not be populating `foci_preferred` as expected.


toger5 (approved):
Since there even is a dedicated method for it this seems super plausible that there is no guarantee that getting it will return what we expect.

In the initialization order system, there might be sth that tries to make it exist when reaching this.
But this is a obvious save-guard we want especially since its obviously not working atm. (or ever)

PR #2903, opened 2024-12-16
Force using the latest version of the WASM bindings

sandhose:
This should mean the WASM blob is finally shipped as WASM, and not a JS string! It massively helps with page load time

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,6 +121,7 @@
     "vitest-axe": "^1.0.0-pre.3"
   },
   "resolutions": {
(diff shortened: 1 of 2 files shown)


hughns (commented):
Whilst this appears to "just work", I think we should wait until https://github.com/matrix-org/matrix-js-sdk/pull/4589 is approved (or a least that they don't identify some issues that need to be worked through.

hughns:
The upstream PR has been merged. So this PR is superseded by https://github.com/element-hq/element-call/pull/2911

PR #2904, opened 2024-12-16
Add developer mode option to show RTC connection statistics

hughns:
New option in developer settings:

image

Which overlays some basic info on audio and video when available:

image

And you can click to get the full `RTCInboundRtpStreamStats | RTCOutboundRtpStreamStats` data in a modal:

image


toger5:
This introduces some odd horizontal scrolling on Firefox:
image

toger5:
Could be the tooltip provider?

hughns:
> This introduces some odd horizontal scrolling on Firefox:

Good spot. It's the same on Chrome too.

hughns:
I've moved the detailed data into a modal and I think it's probably good enough for now.

toger5 (changes requested):
It is hard to tell how much time and effort we want to put into this.
There are a couple of issues:
your own stream does not show reasonable data:
image

The styling is hard to read, the buttons are super small.
But I suspect this is just debug information and should not be used by users who want EC to be pretty.
I will give it a quick try to change the css to be a bit more fitting and you can check if you like it.

toger5:
Here are the proposed changes: https://github.com/element-hq/element-call/pull/2916

hughns:
> It is hard to tell how much time and effort we want to put into this.

Answer: Not a lot of effort.

This was intended to surface easily accessible information.

> your own stream does not show reasonable data: image

As this isn't a full feature this is fine for now. If/when we want own stats for debugging then we can wire more of them in.

> The styling is hard to read, the buttons are super small. But I suspect this is just debug information and should not be used by users who want EC to be pretty. I will give it a quick try to change the css to be a bit more fitting and you can check if you like it.

I've merged your proposal 👍

PR #2905, opened 2024-12-16
Use finnish notation for observables

hughns:
To help make our usage of the observables more readable/intuitive.


toger5 (approved):
Cool. I liked that when we talked about it for the first time. great that you jumped to action and executed on this.

PR #2906, opened 2024-12-16
Use AutoDiscovery.getRawClientConfig() instead of MatrixClient.getClientWellKnown() and wait for result

hughns:
I'm very open to suggestions on what is sensible to test around this...

This is a second attempt at https://github.com/element-hq/element-call/pull/2901

### DIFF
--- a/src/rtcSessionHelpers.ts
+++ b/src/rtcSessionHelpers.ts
@@ -13,6 +13,7 @@ import {
   isLivekitFocus,
   isLivekitFocusConfig,
 } from "matrix-js-sdk/src/matrixrtc/LivekitFocus";
+import { AutoDiscovery } from "matrix-js-sdk/src/autodiscovery";
 
 import { PosthogAnalytics } from "./analytics/PosthogAnalytics";
 import { Config } from "./config/Config";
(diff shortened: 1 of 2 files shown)


fkwp:
I tested it on EX iOS nightlies -- and it works

src/rtcSessionHelpers.ts
```diff
@@ -43,19 +44,28 @@ async function makePreferredLivekitFoci(
...
-        }),
-    );
+  // Prioritize the .well-known/matrix/client, if available, over the configured SFU
+  const domain = rtcSession.room.client.getDomain();
+  if (domain) {
+    // we use AutoDiscovery instead of relying on the MatrixClient having already
+    // been fully configured and started
+    const wellKnownFoci = (await AutoDiscovery.getRawClientConfig(domain))?.[
+      FOCI_WK_KEY
+    ];
```
toger5, line 54: Maybe we want to do a check if the client already knows about the well known? so we save the http request in that case. I suspect this is a race condition since it worked at some point.
So maybe at some point this works again.

I also think however that we might want to get to the root of where this race condition occurs. And when we find it, we can use the old logic again.
```suggestion
    const wellKnownFoci = rtcSession.room.client.getClientWellKnown()?.[FOCI_WK_KEY] ? (await AutoDiscovery.getRawClientConfig(domain))?.[
      FOCI_WK_KEY
    ];
```

toger5 (approved):
This is quiet a big issue and very much in conflict with the mission that users should use their own SFU configured on their home servers, I think we really should merge this asap.
Otherwise all our communication that users should not rely on the config in EC is wrong if it does not work atm.

So I will approve this and merge it. But Will also open an issue that this migh be fixable with a underling race condition.

PR #2908, opened 2024-12-17
Use afterEach from vitest for consistency

hughns:
This slipped through the PR review. Whilst not harmful, for consistency we use it from vitest.


Half-Shot (approved):
Whoops!

toger5 (approved):

PR #2910, opened 2024-12-18
React to theme changes in widget mode

robintown:
Part of the fix for https://github.com/element-hq/element-web/issues/27671
Depends on:

- https://github.com/matrix-org/matrix-widget-api/pull/117
- https://github.com/matrix-org/matrix-widget-api/pull/119

- [ ] created issue for the rust-sdk widget driver


src/useTheme.ts
```diff
@@ -5,17 +5,46 @@ SPDX-License-Identifier: AGPL-3.0-only
...
 
 export const useTheme = (): void => {
-  const { theme: themeName } = useUrlParams();
+  const [requestedTheme, setRequestedTheme] = useState(
+    () => getUrlParams().theme,
+  );
   const previousTheme = useRef<string | null>(document.body.classList.item(0));
+
+  useEffect(() => {
+    if (widget) {
```
hughns, line 22: As the whole body of this `useEffect()` is conditional on `widget` can we do something like the following so that we don't waste time on each render?

```js

if (widget) {
  useEffect(() => {
    const onThemeChange = ...
```

src/useTheme.ts
```diff
@@ -5,17 +5,46 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+      const onThemeChange = (
+        ev: CustomEvent<IThemeChangeActionRequest>,
+      ): void => {
+        ev.preventDefault();
+        if ("name" in ev.detail.data && typeof ev.detail.data.name === "string")
+          setRequestedTheme(ev.detail.data.name);
+        widget!.api.transport.reply(ev.detail, {});
+      };
+
+      widget.lazyActions.on(WidgetApiToWidgetAction.ThemeChange, onThemeChange);
```
hughns, line 32: Do we need to be (re-)subscribing on every render, or just once?

hughns (changes requested):
Need to fix the lint errors, which I think will then show the typings errors due to dependency on the unmerged PR in matrix-widget-api.

Please can you add the necessary version bump to package.json and yarn.lock?

Some comments inline.

src/useTheme.ts
robintown: The linter would complain as React hooks are not supposed to be called conditionally. Even though the condition happens to be constant here, doing this would confuse React dev tools as well (it can no longer analyze the hooks in use).

src/useTheme.ts
robintown: Just once, and that's what this does. A `useEffect` with `[]` as its dependencies runs only once in production (twice in dev, as a stress test).

robintown:
Lints fixed, I plan to bump the matrix-widget-api version once the linked PR is reviewed and released. (That's what I should've used the 'blocked' label for)

toger5:
The rust widget driver does not send the theme change action iirc.
I will check if it at least the url property is set by the rust sdk. (**edit:** url properties are supported by the rust sdk)
Can we create issues for the rust-sdk repo to add those features as port of this pr?

hughns:
I've merged https://github.com/matrix-org/matrix-widget-api/pull/119 so I think this PR is now unblocked.

hughns:
Marking as draft until updated with the upstream dep to fix the typechecking.

hughns (approved):
Looks good 👍

PR #2912, opened 2024-12-18
Use .ts files for vite config for type safety

hughns:
It turns out we've been using an unsupported sentry config option for some time.

By converting to TS it should help with maintainability.


tsconfig.json
```diff
@@ -18,6 +18,7 @@
     // Workaround for https://github.com/microsoft/TypeScript/issues/55132
     "useDefineForClassFields": false,
     "allowImportingTsExtensions": true,
+    "skipLibCheck": true,
```
toger5: This seems to be a step back in how strict we do our typing.
Are we sure we want this? Is there no way to fix the reason we add this with yarns resolutions?
Can it get a comment why we need this now?

toger5 (changes requested):

tsconfig.json
hughns: I concur that this is a backwards step. Unfortunately it's not just the case of some resolutions.

The main culprit appears to be @codecov/vite-plugin.

I will make this draft and lift the one material change over to a simpler PR.

toger5:
replaced by: https://github.com/element-hq/element-call/pull/3443

Issue #2871, opened 2024-12-04
Crypto-wasm asset changes hash between builds


hughns:
### Steps to reproduce

1. Build element-call.
2. Note the hash for the crypto-warm asset.
3. Change something in element-call but keep the same crypto-wasm version.
4. Build element-call again.
5. Check the crypto-wasm asset again to see what hash was allocated.
6. See that it is different than before.

#### What did you expect?

The asset hash would depend on the npm version. So, if the npm version hasn’t changed then the asset name would be the same and there is some chance that the browser will have cached the asset.

#### What happened instead?

The asset hash changed even though the npm version was the same. 

### Will you send logs?

No

hughns:
I had a look at this and the reproduction steps don't work locally.

However, I did look at what is different between asset versions and it is the Sentry integration. It appears that somewhere during the build two globals get set which are making their way tino the crypto-wasm:

```diff
[...]

PR #2913, opened 2024-12-18
Inject sentry release information at runtime not build time

hughns:
See https://github.com/element-hq/element-call/issues/2871 for context of impact.


hughns:
I'm not convinced that this is working as expected.

BillCarsonFr:
I don't understand, closing

PR #2917, opened 2024-12-19
Add screen share sound effect.

### LINKED ISSUE #2696: Provide an option to play a sound when a screenshare is started/stopped
I'd like to have element call play a sound when someone shares their screen. There could also be a sound when microphone mute is toggled. Discord and Jitsi have these sounds (at least some of them)

I like having the extra information that there is something relevant to look at. (Sometimes you have element call minimized)

There could be a setting/config option to enable this behavior.

_No response_

### Additional context

_No response_

Half-Shot:
Fixes https://github.com/element-hq/element-call/issues/2696

I felt the use of a stopped sound felt overkill, but we could add it if others feel strongly.


toger5 (approved):
Very easy to follow code 👌

toger5:
@Half-Shot thinking about this again is there sth we can do with an observable scope to not require the unsubscribe?

PR #2918, opened 2024-12-19
Disambiguate displaynames

### LINKED ISSUE #2746: EC doesn't disambiguate displaynames.
### Steps to reproduce

1. Start an EC video call in a room with two different people called Matthew
2. They both join
3. There's no way to see which video comes from which user.

#### What did you expect?

Like EW and EX (and all other Matrix clients should), users with the same-looking name should have their mxid appended after it in order to differentiate them from each other. 

matrix-js-sdk should give us this almost for free via unhomoglyph - we just need to hook it up in the UX.

#### What happened instead?

Ambiguity and risk of impersonation.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

Half-Shot:
Fixes #2746 

This adds a new observer for member displaynames and updates usages of the old hook to use the view model. I'm currently copying the `shouldDisambiguate` function from Element Web as we want to operate on call members, not all room members (and the changes felt too specific to us to incorporate upstream).

Either way, this works quite nicely and responsively.

That said, maybe this would be good to be included in `MatrixRTCSession`?


src/state/CallViewModel.ts
```diff
@@ -456,6 +496,56 @@ export class CallViewModel extends ViewModel {
...
     );
 
+  /**
+   * Displaynames for each member of the call. This will disambiguate
+   * any displaynames that clashes with another member. Only members
+   * joined to the call are considered here.
+   */
+  public readonly memberDisplaynames$ = fromEvent(
+    this.matrixRTCSession,
+    MatrixRTCSessionEvent.MembershipsChanged,
```
toger5: Do we emit this if a display name changes?
Very minor but otherwise we will not update the `memberDisplaynames` if a user changes their name.

src/state/CallViewModel.ts
```diff
@@ -456,6 +496,56 @@ export class CallViewModel extends ViewModel {
...
+          shouldDisambiguate(rawDisplayName, member.userId, memberships, room)
+        ) {
+          displaynameMap.set(
+            livekitParticipantId,
+            `${member.rawDisplayName} (${member.userId})`,
+          );
+        } else {
+          displaynameMap.set(livekitParticipantId, member.rawDisplayName);
+        }
+      }
```
toger5: Only this logic is really new. The rest is duplicated from `mediaItems$` right?
I think one of the following could work:
 - create a fn: `getRoomMemberFor(rtcMember)` and use this here and for mediaItems. (It also would need to return the participant?
 - create a BehaviorSubject for `memberDisplaynames$` and update it when updating the `mediaItems$`

src/state/CallViewModel.test.ts
```diff
@@ -54,15 +55,23 @@ vi.mock("@livekit/components-core");
 
 const localRtcMember = mockRtcMembership("@carol:example.org", "CCCC");
 const aliceRtcMember = mockRtcMembership("@alice:example.org", "AAAA");
+const aliceDopplegangerRtcMember = mockRtcMembership(
```
toger5:
```suggestion
const aliceDoppelgangerRtcMember = mockRtcMembership(
```

src/state/CallViewModel.test.ts
```diff
@@ -54,15 +55,23 @@ vi.mock("@livekit/components-core");
...
+  "@alice2:example.org",
+  "AAAA",
+);
 const bobRtcMember = mockRtcMembership("@bob:example.org", "BBBB");
 const daveRtcMember = mockRtcMembership("@dave:example.org", "DDDD");
 
-const alice = mockMatrixRoomMember(aliceRtcMember);
+const alice = mockMatrixRoomMember(aliceRtcMember, { rawDisplayName: "Alice" });
+const aliceDoppleganger = mockMatrixRoomMember(aliceDopplegangerRtcMember, {
```
toger5:
```suggestion
const aliceDoppelganger = mockMatrixRoomMember(aliceDopplegangerRtcMember, {
```

toger5: It makes more sense if you change/refactor this in an ide. All occurances of Doppleganger

src/state/CallViewModel.test.ts
```diff
@@ -792,6 +801,55 @@ it("should show at least one tile per MatrixRTCSession", () => {
...
+    const expectedLayoutMarbles = "abcd";
+
+    withCallViewModel(
+      of([]),
+      hot(scenarioInputMarbles, {
+        a: [],
+        b: [aliceRtcMember],
+        c: [aliceRtcMember, aliceDopplegangerRtcMember],
+        d: [aliceRtcMember, aliceDopplegangerRtcMember, bobRtcMember],
+        e: [aliceDopplegangerRtcMember, bobRtcMember],
```
toger5: How hard would it be to add the other cases: `LTR/RTL` and display name formatted like an impersonation?

src/state/CallViewModel.ts
Half-Shot: Yep, `MatrixRTCSessionEvent.MembershipsChanged` includes:

> // A member joined, left, or updated a property of their membership.

Displaynames are still part of membership at the moment, so we can trust this. I've also tested this one.

src/state/CallViewModel.test.ts
Half-Shot: Now covered!

src/state/CallViewModel.ts
Half-Shot: I decided to split out some of the common logic, but there isn't a huge amount from what I can see. I think this probably needs another review. We can pair on it if you're still unhappy :)

toger5 (approved):
Thanks
This has improved even more than what I reviewed. Thanks for moving the fixtures and the new test cases.

PR #2919, opened 2024-12-19
Upgrade to React Router v6

sandhose:
This is just a 'what would it take to do that' kind of thing.

And it looks like it works? I haven't tested the app thoroughly other than clicking around, but types are happy, tests as well


src/room/LobbyView.tsx
```diff
@@ -65,7 +64,6 @@ export const LobbyView: FC<Props> = ({
   waitingForInvite,
 }) => {
   const { t } = useTranslation();
-  useLocationNavigation();
```
sandhose, line 68: I'm pretty sure this wasn't doing anything, as it had 'enabled = false'?

hughns (approved):
Thank you very much for doing this! 👍

PR #2921, opened 2024-12-28
Add wave/hello reaction

davidegirardi:
Add a way to semi-quietly say hello when joining a meeting.


Half-Shot:
Failing test is due to `expect(container).toMatchSnapshot();` in one of the tests failing due to the extra reaction button. We can probably change/remove that expect without too much trouble.

hughns (commented):
It makes total sense to add a wave reaction. I'm not sure about the chosen sound (also where did the sound file come from so that we can confirm it has compatible license?).

I've added the design label to get input on the sound to use.

src/reactions/index.ts
```diff
@@ -25,6 +25,8 @@ import deerSoundOgg from "../sound/reactions/deer.ogg?url";
 import deerSoundMp3 from "../sound/reactions/deer.mp3?url";
 import rockSoundOgg from "../sound/reactions/rock.ogg?url";
 import rockSoundMp3 from "../sound/reactions/rock.mp3?url";
+import waveSoundOgg from "../sound/reactions/wave.ogg?url";
```
Half-Shot, line 28: If the sound is your own, can you add a note in https://github.com/element-hq/element-call/blob/dg/add-wave-reaction/src/sound/LICENCE.md so we have our backs covered?

davidegirardi: 9b42bc79dbd59be320540bdbb63b6064ae78452c

davidegirardi:
> I'm not sure about the chosen sound

I went for something that _could_ sound like an hello without having any words in it. It also sounds quite small by choice, to not be distracting if somebody uses it while someone else is talking (the opposite of the rock reaction so to speak :grimacing:).

hughns (approved):
Whilst I'm not sure about the exact sound, I've discussed with @fkwp and we will merge it as is for now. We can change the sound later on if needed.

I'm overruling the CLA bot as I know @davidegirardi works for Element.

davidegirardi:
Happy to hear about any kind of direction we want to take with sound design.

PR #2922, opened 2025-01-03
Enable Element Call to be consumed as a package within a larger app

robintown:
If we want to add Element Call as a dependency of Element Web and package it with the application, we should make `yarn pack` produce a built copy. Then it can be added as a simple Git dependency to projects using any modern package manager like Yarn Berry or NPM, and it will be saved and built in the local package cache.

Unfortunately, we don't use modern package managers yet for our web applications, we use Yarn Classic. This is in draft while I ask around about whether now is a good time to change that.

Part of https://github.com/element-hq/element-web/issues/23908


hughns:
I've been cribbing bits of this for https://github.com/element-hq/element-call/issues/2994 and some bits have already landed.

Things I've done:

- using `--frozen-lockfile` in CI
- using the embedded approach that you took and relocating `index.html`

The one thing I haven't used is the build during `yarn install`. Instead we only need to do a build for *releases* of Element Call.

Thank you!

PR #2926, opened 2025-01-06
Remove unnecessary dependency on @testing-library/react-hooks

hughns:
https://github.com/testing-library/react-hooks-testing-library#a-note-about-react-18-support

Required for react v19 support.


robintown (approved):
Oh nice, I thought the approach of testing hooks rather than components was simply deprecated now, or something. It's nice they're properly supporting it within the main React Testing Library.

PR #2928, opened 2025-01-06
Set the document title to the current call name.

### LINKED ISSUE #270: Improve page title on calls
Currently the page title when on a call is "Matrix Video Chat | [Call name]". I think it can be improved in two ways:

* "Matrix Video Chat" should be replaced with "Element Call"
* It would be helpful to have the call name appear first, in order to distinguish the Element Call tab from others

Half-Shot:
*nearly* f1xes https://github.com/element-hq/element-call/issues/270

The final version to fix #270 needs more design thought


hughns (commented):
The referenced issue (#270) also wanted the ordering of the title changed so that the call name appears leftmost. Did you intend to do this too?

Half-Shot:
> The referenced issue (#270) also wanted the ordering of the title changed so that the call name appears leftmost. Did you intend to do this too?

I'll do a design check first on this, since it's a bit of a change.

hughns (approved):

PR #2939, opened 2025-01-10
Lint rule to disallow deep imports from matrix-js-sdk/src/matrixrtc/*

hughns:


hughns:
Enabling this rule triggers some dependency issue from `eslint` to `strip-ansi`. We might need to move to eslint v9 to get this working. eslint upgrade is in https://github.com/element-hq/element-call/pull/2449

BillCarsonFr:
this done.

PR #2951, opened 2025-01-16
Make error screens more visually consistent

robintown:
This change gives our error screens a more consistent visual appearance and lets us easily present any new errors that we come up with in the future with custom React components. This will make it easy to, for example, throw an error if the JWT service is unreachable, present that error in non-technical language, while also letting admins dig deeper into the details on the error screen.

Here are some examples:

Screenshot 2025-01-15 at 20-05-22 Element Call
Screenshot 2025-01-15 at 19-43-00 Element Call
Screenshot 2025-01-15 at 20-17-20 Element Call The Testing Zone


locales/en/app.json
hughns, line 1: IIRC, when refactoring existing strings in the locale file, we should also update any non-English locale files.

hughns (changes requested):
Aside from refactoring the non-English translations this looks good to me 👍

locales/en/app.json
robintown: There were only two strings for which I didn't end up tweaking the copy. I've renamed those in the translation files now.

hughns (approved):

PR #2959, opened 2025-01-16
Prevent new devices from automatically starting unmuted in call

### LINKED ISSUE #2958: Plugging in device during a call can cause it to start unmuted
### Steps to reproduce

1. Have no webcam
2. Join a call
3. Plug in a webcam while you're on the call

#### What did you expect?

Media devices should only ever be automatically turned on if you're on the lobby page. The user's video should remain muted.

#### What happened instead?

The camera is unmuted automatically once plugged in.

A fix would be, instead of allowing the default mute state to be true [as long as `skipLobby` is false](https://github.com/element-hq/element-call/blob/3189bdba2a49368dce464202c86d270f427e2eed/src/room/MuteStates.ts#L86), we should only allow the default mute state to be true if the app is currently sitting on the lobby screen. That's the real important condition for which `skipLobby` is simply a proxy.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

robintown:
If the user plugs in a new media device mid-call, we don't want to enable it without user input.

Closes https://github.com/element-hq/element-call/issues/2958

### DIFF
--- a/src/room/GroupCallView.test.tsx
+++ b/src/room/GroupCallView.test.tsx
@@ -118,6 +118,7 @@ function createGroupCallView(widget: WidgetHelpers | null): {
         skipLobby={false}
         hideHeader={true}
         rtcSession={rtcSession as unknown as MatrixRTCSession}
+        isJoined
         muteStates={muteState}
         widget={widget}
       />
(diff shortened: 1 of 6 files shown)


hughns (changes requested):
This needs a merge/rebase.

hughns (approved):

PR #2966, opened 2025-01-23
Fix calculation of local membership identifier

### LINKED ISSUE #2948: Can't lower raised hand
### Steps to reproduce

1. Join a call on one device
2. Raise your hand
3. Join the call from another device
4. Raise your hand from that device
5. Lower your hand from that device

#### What did you expect?

Hand lowers.

#### What happened instead?

It doesn't, and you see the error message "some events could not be sent" in the timeline.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

Yes

Half-Shot:
Fixes #2948 

See if you can spot the obvious stupid.

### DIFF
--- a/src/reactions/useReactionsSender.tsx
+++ b/src/reactions/useReactionsSender.tsx
@@ -60,6 +60,7 @@ export const ReactionsSenderProvider = ({
   const room = rtcSession.room;
   const myUserId = room.client.getUserId();
   const myDeviceId = room.client.getDeviceId();
+  const myMembershipIdentifier = `${myUserId}:${myDeviceId}`;
 
   const myMembershipEvent = useMemo(
     () =>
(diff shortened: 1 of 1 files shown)


toger5 (approved):
It makes sense that there is not reason to use a Memo if all the requirements are already available anyways.
I am a bit surprised however that when we at some point update myMembershipIdentifier we do not recompute the raise hand state.

Half-Shot:
> It makes sense that there is not reason to use a Memo if all the requirements are already available anyways. I am a bit surprised however that when we at some point update myMembershipIdentifier we do not recompute the raise hand state.

Line 83, no? We have a memo that watches `myMembershipIdentifier` and `handsRaised`

PR #2967, opened 2025-01-23
Bump matrix-js-sdk to v36.1.0

### LINKED ISSUE #2973: Update matrix-js-sdk to v36.1.0
The is a breaking change in this update around the use of `randomString()`.

toger5:
Fixes #2973.

Full diff of matrix-js-sdk: https://github.com/matrix-org/matrix-js-sdk/compare/9134471dc72a14b29eb207f1c5ef207521f40bd3...v36.1.0


hughns:
Marking draft so that this doesn't accidentally get merged. Known to break EX integration.

toger5:
> Known to break EX integration.

This is not anymore the case because of:
https://github.com/matrix-org/matrix-js-sdk/pull/4652/commits/eaca58283432cbfc43a1703029ac3da75926285e

toger5:
Is this also blocked on: https://github.com/matrix-org/matrix-js-sdk/pull/4662

Since we reverted the changes that broke the awaiting on EX, I am wondering if this can get merged?
Is the initial state_update change part of the rc0

hughns:
> Is this also blocked on: [matrix-org/matrix-js-sdk#4662](https://github.com/matrix-org/matrix-js-sdk/pull/4662)
> 
> Since we reverted the changes that broke the awaiting on EX, I am wondering if this can get merged? Is the initial state_update change part of the rc0

I suggest waiting until v36.1.0 gets released and use that version rather the RC. Technically the reverted version is only in v36.1.0-rc.1 (not rc0) which hasn't actually been released.

hughns (approved):

toger5:
> Is this also blocked on: https://github.com/matrix-org/matrix-js-sdk/pull/4662
> 
> Since we reverted the changes that broke the awaiting on EX, I am wondering if this can get merged?
> Is the initial state_update change part of the rc0

This was partially true. This PR was indeed blocked on https://github.com/matrix-org/matrix-js-sdk/pull/4662. But I was missing the reason, which is: This also bumps the widget api.
Version v1.13.0 of the widget api introduces the UPDATE_STATE widget version and advocating that version changes what the element widget driver will send down to the widget. (in detail: Element Web will not send down state events from the state store when using the read_event action anymore)
So we need to lock the version to 1.10.0 for now (until we have https://github.com/matrix-org/matrix-js-sdk/pull/4662 merged)

This PR downgrades the widget-api version: https://github.com/element-hq/element-call/pull/2988

PR #2980, opened 2025-01-31
ADD info about element.json and client configuration to README

rezor92:


hughns:
Thank you very much for this PR.

We are actually working on removing the ability to configure the Element messenger clients in this way.

Instead we will have a static build of Element Call embedded directly within the Element messenger clients.

As such, I don't want to encourage others people to use these features in the interim only for them to be removed in the near future. So, I am going to close the PR.

PR #2983, opened 2025-02-03
Cleanup GitHub Actions

hughns:
This is mostly just about naming. Plus removing an unused end-to-end testing workflow.


fkwp (approved):
lgtm

PR #2988, opened 2025-02-05
Resolve version clash in matrix-widget-api dependency

toger5:
https://github.com/element-hq/element-call/pull/2967 introduced a version clash in our usage of matrix-widget-api and ended up running two version of matrix-widget-api.

We fix this by pinning matrix-widget-api to a version that matrix-js-sdk can use. Specifically v1.11.0


hughns (approved):
I've cleaned up the PR description as discussed, but this makes total sense otherwise 👍

PR #2992, opened 2025-02-05
Use --frozen-lockfile when install deps in CI

hughns:


toger5 (approved):
Oh this is great. Its a first step to prohibit unexpected dep updates like the one with the widget API.

Issue #2994, opened 2025-02-06
Support embedding of the Element Call assets within a widget host


hughns:
This only needs to support Element Call running as a widget (not standalone).

Specific scenarios that we want to support:

- a npm package for consumption in a web application such as Element Web
- a Maven package for consumption in an Android app such as Element X Android
- a SwiftPM package for consumption in an iOS app such as Element X iOS

Pull requests:
- https://github.com/element-hq/element-call/pull/2983
- https://github.com/element-hq/element-call/pull/2995
- https://github.com/element-hq/element-call/pull/3071
- https://github.com/element-hq/element-call/pull/3073
- https://github.com/element-hq/element-call/pull/3074
- https://github.com/element-hq/element-call/pull/3077
- https://github.com/element-hq/element-call/pull/3086
- https://github.com/element-hq/element-call/pull/3089

hughns:
Work in progress is visible at https://github.com/element-hq/element-call/compare/livekit...hughns:element-call:publishing

hughns:
[v0.9.0-rc.1](https://github.com/element-hq/element-call/releases/tag/v0.9.0-rc.1) has been published which has the embedded packages available for testing.

hughns:
[v0.9.0](https://github.com/element-hq/element-call/releases/tag/v0.9.0) has now been release which includes the embedded packages.

PR #2995, opened 2025-02-06
Don't produce compressed assets at build time and instead do it as part of the nginx build

hughns:
The core vite build shouldn't be responsible for implementing optimisations that are specific to a particular distribution mechanism. e.g. static compressed files for nginx.

This way the core build is "clean".

Instead, the packager/distributor of the build should be responsible for it. e.g. the nginx docker distribution should do any optimisations that it chooses.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,13 @@
+FROM alpine AS builder
+
+COPY ./dist /dist
+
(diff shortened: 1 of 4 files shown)


fkwp (approved):
lgtm

PR #2999, opened 2025-02-07
Bump JS SDK to temp version without my membership

toger5:
resending.

Full diff: https://github.com/matrix-org/matrix-js-sdk/compare/v36.1.0...2ce2bbc053803c258ebd9cb0bf8aad180d826075


toger5:
This PR's purpose is to build a version without without the event re-sending that allows us to test large calls without this feature.

PR #3001, opened 2025-02-10
Update config.sample.json

escix:


config/config.sample.json
```diff
@@ -1,12 +1,24 @@
...
+      "base_url":"https://mydomain.com",
+      "server_name":"mydomain.com"
     }
   },
-  "features": {
-    "feature_use_device_session_member_events": true
+  "livekit": {
+    "livekit_service_url":"https://livekit-jwt.mydomain.com"
   },
-  "eula": "https://static.element.io/legal/online-EULA.pdf"
```
hughns: I *think* the EULA field is currently a required field so the example should match. However, it will be made optional soon as it is not used when Element Call is running as an embedded widget.

config/config.sample.json
```diff
@@ -1,12 +1,24 @@
...
+     "url": "https://call.mydomain.com",
+     "participant_limit": 8,
+     "brand": "Element Call",
+     "use_exclusively": true
+    },
+  "features":{
+    "feature_video_rooms": true,
+    "feature_new_room_decoration_ui": true,
+    "feature_group_calls": true,
+    "feature_element_call_video_rooms": true,
```
hughns: These are Element Web/Desktop options not Element Call options. So, they shouldn't be in here at all.

```suggestion
```

config/config.sample.json
```diff
@@ -1,12 +1,24 @@
...
+  "livekit": {
+    "livekit_service_url":"https://livekit-jwt.mydomain.com"
   },
-  "eula": "https://static.element.io/legal/online-EULA.pdf"
+  "element_call": {
+     "url": "https://call.mydomain.com",
+     "participant_limit": 8,
+     "brand": "Element Call",
+     "use_exclusively": true
+    },
```
hughns: These are Element Web/Desktop options not Element Call options. So, they shouldn't be in here at all.

```suggestion
```

hughns (changes requested):
Thank you for contributing this.

Some of these options do not existing in Element Call and only exist in Element Web/Desktop. Please remove them and also check and fix the lint errors.

escix (commented):
Added EULA

config/config.sample.json
```diff
@@ -1,12 +1,15 @@
 {
-  "default_server_config": {
-    "m.homeserver": {
-      "base_url": "http://synapse.localhost:8008",
-      "server_name": "synapse.localhost"
+  "default_server_config":{
+    "m.homeserver":{
+      "base_url":"https://mydomain.com",
+      "server_name":"mydomain.com"
```
fkwp, line 5: If we want to rename those, lets create a dedicated config.sample.json which works out of the box with [dev-backend-docker-compose.yml](https://github.com/element-hq/element-call/blob/livekit/dev-backend-docker-compose.yml).

hughns (approved):
Thank you.

PR #3002, opened 2025-02-10
Update README.md

escix:


fkwp:
Hi @escix and thx for the PR. Sorry for the delay - we had a refactor of the README in progress which just landed. I think your contribution is best suited for a sample config file for caddy in a  /contrib directory. It would be great if you could adapt a PR accordingly.

escix:
> Hi @escix and thx for the PR. Sorry for the delay - we had a refactor of the README in progress which just landed. I think your contribution is best suited for a sample config file for caddy in a /contrib directory. It would be great if you could adapt a PR accordingly.


Hi fkwp, no need to apologies and you guys are doing a great job.

I am not a developer, but host matrix at home so thought the configs will help someone else as there are lots of queries re element-call.

As a non-developer I don't understand most of the github processes. I am happy for my commits to be in sample.config file.
Although I am not sure how to proceed with that.

PR #3007, opened 2025-02-14
Add Android assets packaging project and process

jmartinesp:
I believe this is everything we'd need to publish the Android assets as an AAR:

1. There is a new `platforms/android` folder which contains the Android project with the minimum of changes needed to publish the library as an [AAR](https://developer.android.com/studio/projects/android-library) into the GH gradle registry.
2. This new folder also includes a script to build the assets, copy them to the right folder and trigger the publishing gradle task.
3. And there is a new `publish_android` job that will call this script and automate the publishing.

There is still the need to update the version in the `gradle.properties` file before a release so it matches the EC one. That can probably be done later either manually, through some script that changes the version for both, or with a script that extracts the value from `package.json` and overrides it here (although the value there seems to be `0.0.0`?).

Note this PR still needs the relative paths for assets to work.


jmartinesp:
Closing this PoC PR.

PR #3011, opened 2025-02-17
Send a 'close' action when the widget is ready to close

robintown:
By keeping 'hangup' and 'close' as separate actions, we can allow Element Call widgets to stay on an error screen after the user has been disconnected without the widget completely disappearing from the host's UI. We don't have to request any additional capabilities to use a custom widget action like this one.

### DIFF
--- a/src/rtcSessionHelpers.ts
+++ b/src/rtcSessionHelpers.ts
@@ -144,11 +144,19 @@ const widgetPostHangupProcedure = async (
   // We send the hangup event after the memberships have been updated
   // calling leaveRTCSession.
   // We need to wait because this makes the client hosting this widget killing the IFrame.
-  await widget.api.transport.send(ElementWidgetActions.HangupCall, {});
+  try {
+    await widget.api.transport.send(ElementWidgetActions.HangupCall, {});
+  } catch (e) {
+    logger.error("Failed to send hangup action", e);
+  }
   // On a normal user hangup we can shut down and close the widget. But if an
   // error occurs we should keep the widget open until the user reads it.
   if (cause === "user") {
-    await widget.api.transport.send(ElementWidgetActions.Close, {});
+    try {
+      await widget.api.transport.send(ElementWidgetActions.Close, {});
+    } catch (e) {
+      logger.error("Failed to send close action", e);
+    }
     widget.api.transport.stop();
     PosthogAnalytics.instance.logout();
   }
--- a/src/rtcSessionHelpers.ts
+++ b/src/rtcSessionHelpers.ts
@@ -130,13 +130,9 @@ export async function enterRTCSession(
 
 const widgetPostHangupProcedure = async (
   widget: WidgetHelpers,
+  cause: "user" | "error",
   promiseBeforeHangup?: Promise<unknown>,
 ): Promise<void> => {
-  // we need to wait until the callEnded event is tracked on posthog.
-  // Otherwise the iFrame gets killed before the callEnded event got tracked.
-  await new Promise((resolve) => window.setTimeout(resolve, 10)); // 10ms
-  PosthogAnalytics.instance.logout();
-
   try {
     await widget.api.setAlwaysOnScreen(false);
   } catch (e) {
@@ -149,15 +145,23 @@ const widgetPostHangupProcedure = async (
   // calling leaveRTCSession.
   // We need to wait because this makes the client hosting this widget killing the IFrame.
   await widget.api.transport.send(ElementWidgetActions.HangupCall, {});
+  // On a normal user hangup we can shut down and close the widget. But if an
+  // error occurs we should keep the widget open until the user reads it.
+  if (cause === "user") {
+    await widget.api.transport.send(ElementWidgetActions.Close, {});
+    widget.api.transport.stop();
+    PosthogAnalytics.instance.logout();
+  }
 };
 
 export async function leaveRTCSession(
   rtcSession: MatrixRTCSession,
+  cause: "user" | "error",
   promiseBeforeHangup?: Promise<unknown>,
 ): Promise<void> {
   await rtcSession.leaveRoomSession();
   if (widget) {
-    await widgetPostHangupProcedure(widget, promiseBeforeHangup);
+    await widgetPostHangupProcedure(widget, cause, promiseBeforeHangup);
   } else {
     await promiseBeforeHangup;
   }
--- a/src/widget.ts
+++ b/src/widget.ts
@@ -21,10 +21,11 @@ import { getUrlParams } from "./UrlParams";
 import { Config } from "./config/Config";
 import { ElementCallReactionEventType } from "./reactions";
 
-// Subset of the actions in matrix-react-sdk
+// Subset of the actions in element-web
 export enum ElementWidgetActions {
   JoinCall = "io.element.join",
   HangupCall = "im.vector.hangup",
+  Close = "io.element.close",
   TileLayout = "io.element.tile_layout",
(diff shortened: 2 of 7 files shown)


robintown:
I bumped the TypeScript `module` version here because I needed top-level `await` syntax to write the tests, which is only available on `es2022` or later.

src/rtcSessionHelpers.ts
```diff
@@ -149,15 +145,23 @@ const widgetPostHangupProcedure = async (
   // calling leaveRTCSession.
   // We need to wait because this makes the client hosting this widget killing the IFrame.
   await widget.api.transport.send(ElementWidgetActions.HangupCall, {});
+  // On a normal user hangup we can shut down and close the widget. But if an
+  // error occurs we should keep the widget open until the user reads it.
+  if (cause === "user") {
+    await widget.api.transport.send(ElementWidgetActions.Close, {});
+    widget.api.transport.stop();
+    PosthogAnalytics.instance.logout();
```
toger5: As mentioned in the VoIP room I think that the general approach of two actions is correct. (It is the minimal set of information we need).

I do think we should not change the meaning of `ElementWidgetActions.HangupCall` however.
Changing it here is not an issue. But this change implies a change in EX and EW where we now do NOT close a widget anymore on receiving the hangup action. This make EC and EX incompatible with older EC versions which could be an issue on self hosted instances.
Also Jitsi might break.

I think the better would be to keep the action as they are and introduce the following:
 - a general purpose close action best with an MSC. `org.matrix.mscXXXX.close` action that any widget can use for requesting to get closed (This was sth that seemed like I wanted it for the matrixRTC demo also)
  - A dedicated element call hangup action `io.element.hangup` that is only used for element call hangups and does not imply any widget updating but only makes the call render as not connected. (Can you also update me again on the case in which this is different to the room state not including our own rtc memebrship anymore? I think that is not tracked yet in text)
  - replace `  HangupCall = "im.vector.hangup",` from he `ElementWidgetActions` enum with` io.element.hangup`

src/widget.ts
```diff
@@ -21,10 +21,11 @@ import { getUrlParams } from "./UrlParams";
 import { Config } from "./config/Config";
 import { ElementCallReactionEventType } from "./reactions";
 
-// Subset of the actions in matrix-react-sdk
+// Subset of the actions in element-web
 export enum ElementWidgetActions {
   JoinCall = "io.element.join",
   HangupCall = "im.vector.hangup",
```
toger5, line 27:
```suggestion
HangupCall = "io.element.hangup",
```

src/widget.ts
```diff
@@ -21,10 +21,11 @@ import { getUrlParams } from "./UrlParams";
...
+  Close = "io.element.close",
```
toger5, line 28:
```suggestion
  Close = "org.matrix.mscXXXX.close",
```

toger5 (changes requested):

src/rtcSessionHelpers.ts
```diff
@@ -149,15 +145,23 @@ const widgetPostHangupProcedure = async (
   // calling leaveRTCSession.
   // We need to wait because this makes the client hosting this widget killing the IFrame.
   await widget.api.transport.send(ElementWidgetActions.HangupCall, {});
+  // On a normal user hangup we can shut down and close the widget. But if an
+  // error occurs we should keep the widget open until the user reads it.
+  if (cause === "user") {
+    await widget.api.transport.send(ElementWidgetActions.Close, {});
```
toger5: Can we add a way to also emit this with a button on the error screen?
This is unnecessary for EW since we have the close button there. But on EX this is a better UX since we then can close the widget by pressing sth like "Ok" or "Close" inside the IFrame.

(for now this screen would of course not be visible but once we update the rust sdk (if we go this route where a rust sdk udpate is required) we dont introduce the isse that the user cannot exit the error screen except when using the os activity manager)

src/rtcSessionHelpers.ts
robintown: > I do think we should not change the meaning of `ElementWidgetActions.HangupCall` however. Changing it here is not an issue.

I'm still confused by this. This PR doesn't change the meaning of `im.vector.hangup` at all, it only adds more information.

> But this change implies a change in EX and EW where we now do NOT close a widget anymore on receiving the hangup action. This make EC and EX incompatible with older EC versions which could be an issue on self hosted instances.

It's always been the case that we expect EC+EW deployments to be on the latest EC release at the time of the EW release, otherwise it becomes unnecessarily hard to make forward progress when it comes to URL parameters or widget communication. I agree that this isn't the best situation, as admins might forget about this hidden assumption, but I suggest we continue accepting this for now since the embedding work (https://github.com/element-hq/element-call/issues/2994 / https://github.com/element-hq/element-web/issues/23908) is going to make the dependency explicit, resolving it for good.

We do need to keep the EW/EX PRs blocked until we're confident that we can release a new EC version though, that's a good thing to remember here.

> Also Jitsi might break.

Good thing to flag but it doesn't break, Jitsi video rooms are handled by a different class in Element Web and I've tested them.

> I think the better would be to keep the action as they are and introduce the following:
> * a general purpose close action best with an MSC. `org.matrix.mscXXXX.close` action that any widget can use for requesting to get closed (This was sth that seemed like I wanted it for the matrixRTC demo also)

I have some thoughts about this:

- Does this close action become part of the minimum implementation requirements of the widget API, or does it need to be requested as a capability?
- The benefit, I guess, of standardizing a close action would be to allow arbitrary, untrusted widgets to request their own closure. It doesn't change anything for deep widget integrations like EW/EX + EC, which could just as easily use proprietary actions.
- If you're a client, and you're presenting an untrusted widget, you have to either show your own close button in addition to the widget's, which could look awkward, or else risk that the widget won't offer any way of allowing the user to close it, permanently intruding the user's screen.

So generally it seems a bit hairy: see for example the restrictions that the Web platform places on [`window.close()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/close). I think sticking to a proprietary action here may just be a good thing.

> * A dedicated element call hangup action `io.element.hangup` that is only used for element call hangups and does not imply any widget updating but only makes the call render as not connected. (Can you also update me again on the case in which this is different to the room state not including our own rtc memebrship anymore? I think that is not tracked yet in text)
> * replace `  HangupCall = "im.vector.hangup",` from he `ElementWidgetActions` enum with` io.element.hangup`

How would such an action be any different from `im.vector.hangup`?

src/rtcSessionHelpers.ts
robintown: I would love to do that, yeah, but in a separate PR.

src/rtcSessionHelpers.ts
toger5: > How would such an action be any different from `im.vector.hangup`?

The win here would be that we could keep EX as is but still only show the error screens by sending the `im.vector` scoped action on close and communicate the hangup through a `io.element` scoped action.
`im.vector` would than be jitisi's scope and io.element EC's.

That was basically the main change I wanted to propose.
But the win is very small (EX current version compatible with error screens) and moving to a `io.element.hangup` scoped hangup action is not necessarily a win since it would need quiet a bit of documentation to not confuse ppl in the future.

src/rtcSessionHelpers.ts
toger5: Sounds good. I think we should try to mitigate the following situation:
 - Not being able to close the widget with a button on an error screen
 - EX updated to not close on `hangup` anymore and waiting for the close action.
 
 Then the use might feel stuck on an error screen until they close the activity.

toger5 (approved):

Issue #2955, opened 2025-01-16
Show error screens in embedded widget mode (aka Messenger Applications)


robintown:
### Implementation

This enhancement requires implementation in each of our Messenger clients.
- https://github.com/element-hq/element-web/pull/29254
- https://github.com/element-hq/element-x-android/pull/4297
- https://github.com/element-hq/element-x-ios/pull/3856

> [!NOTE]
> these PRs are blocked by until we are certain that we'll have a stable release up on call.element.io which sends 'close' actions within the current release cycles of EW and EX. To unblock and roll them out sooner, the plan is to [backport the sending of the 'close' action to a v0.7 release](https://github.com/element-hq/voip-internal/issues/304).
The dependencies are tracked by

which are blocked by:
- https://github.com/element-hq/voip-internal/issues/304  

### Your use case
https://github.com/element-hq/element-x-ios/pull/3813
#### What would you like to do?

Make sure that the error screens we're adding to the Element Call application can be shown in embedded mode as well, without the Messenger Client immediately hiding the widget.

#### Why would you like to do it?

Because we're putting effort into creating helpful error screens, and it would be a shame for the experience in embedded mode to still just be:

1. An error occurs
2. The widget closes immediately
3. ??? Confusion

 - [ ] Reconsider actionable error screen actions in widget mode. (if we are not connected to the RTC session EW should display a X icon)
   - Extracted to a new issue: https://github.com/element-hq/element-call/issues/3048

#### How would you like to achieve it?

When an error occurs, tell the widget to no longer be "always on screen", so that it can be closed, but at the same time make sure Element Web etc. won't just react to the disconnection by closing it immediately.

fkwp:
Is there a block diagram of the lifecycle of a widget?

hughns:
https://github.com/element-hq/element-call/pull/3011 changes Element Call so that it sends a `hangup` and `close` action to the widget host.

Old messenger clients will continue to behave the same with new EC (because EC is sending both actions).

We then change the messenger widget hosts so that they ignore the `hangup` and only close the widget on a `close`:
- https://github.com/element-hq/element-web/pull/29254
- https://github.com/element-hq/element-x-ios/pull/3813
- https://github.com/element-hq/element-x-android/pull/4297

When these PRs land older EC versions (pre https://github.com/element-hq/element-call/pull/3011) will then have a broken UX when the user hangs up. They get left on a blank EC screen.

[...]

PR #3021, opened 2025-02-24
Leave session when error occurs and show error screens in widget mode

robintown:
This is a refactor of how we handle and present errors that happen during a call. Rather than assuming that any errors will be manually passed to `onLeave`, this wraps the call in a React error boundary that takes care of leaving the session and presenting connection errors or crash screens as needed. Importantly, this now allows the 'connection lost' screen to be visible even in widget mode, since we aren't entangling errors with the 'call ended' component (which is deliberately hidden from widgets).

For https://github.com/element-hq/element-call/issues/2955

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -102,6 +102,8 @@ const canScreenshare = "getDisplayMedia" in (navigator.mediaDevices ?? {});
 
 const maxTapDurationMs = 400;
 
+export class ConnectionLostError extends Error {}
+
 export interface ActiveCallProps
   extends Omit<InCallViewProps, "vm" | "livekitRoom" | "connState"> {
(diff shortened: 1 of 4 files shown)


src/room/InCallView.tsx
```diff
@@ -173,7 +175,7 @@ export interface InCallViewProps {
   livekitRoom: Room;
   muteStates: MuteStates;
   participantCount: number;
-  onLeave: (error?: Error) => void;
+  onLeave: () => void;
```
hughns, line 179:
```suggestion
  /** Function to call when the user explicitly ends the call */
  onLeave: () => void;
```

hughns (approved):
👍

One minor comment suggestion.

PR #3022, opened 2025-02-24
Show an error screen when the SFU is at capacity

### LINKED ISSUE #2998: Detect when LiveKit Cloud SFU is at capacity and show an error to the user
### Current behaviour:

In SPA mode you get this error screen:

Image

In embedded widget mode:

The EC closes immediately with no feedback to the user at all.

### Desired behaviour

1. In SPA mode

Show a more specific error message. Something like:

**Insufficient capacity**

*The server has reached its maximum capacity and you cannot join the call at this time. Please try again later. If the problem persists please contact your server administrator.*

2. In embedded widget mode

Also have EC display the same error message. Keep EC open until the user has pressed a button. Add a new button instead of the "Return to home screen" which is the "Cancel" button. When Cancel is clicked then EC widget closes and the user goes back to their Matrix client/widget host.

### Implementation notes

The capacity being hit should manifest itself as a `429` error from the server which enters into [this codepath](https://github.com/livekit/client-sdk-js/blob/17aff2be6ba164563abf981910e6f8b78427f410/src/api/SignalClient.ts#L312-L325) in the LiveKit client-js-sdk.

So, in EC where we call `livekitRoom.connect()` we should catch and inspect the `ConnectionError.status` field to see if it is a 429.

I suspect that the widget host is closing the EC widget automatically when the call membership has been removed. So, in order to implement the desired EC embedded widget behaviour we might need to keep the call membership around for longer until the user presses Cancel (or successfully joins the call).

This capability is needed on the develop/`livekit` branch but also is a candidate to backport to `v0.7.x` due to the recent deployment of the sponsored LiveKit Cloud SFU for the Matrix Foundation and that a new release of Element Call is not imminent due to https://github.com/element-hq/element-call/issues/2972

robintown:
:warning: **Based on https://github.com/element-hq/element-call/pull/3021**, please review only the new commit

Screenshot from 2025-02-24 19-15-12

Closes https://github.com/element-hq/element-call/issues/2998

### DIFF
--- a/locales/en/app.json
+++ b/locales/en/app.json
@@ -83,7 +83,7 @@
     "generic": "Something went wrong",
     "generic_description": "Submitting debug logs will help us track down the problem.",
     "insufficient_capacity": "Insufficient capacity",
-    "insufficient_capacity_description": "The server has reached its maximum capacity and you cannot join calls at this time. Try again later, or contact your server admin if the problem persists.",
+    "insufficient_capacity_description": "The server has reached its maximum capacity and you cannot join the call at this time. Try again later, or contact your server admin if the problem persists.",
     "open_elsewhere": "Opened in another tab",
     "open_elsewhere_description": "{{brand}} has been opened in another tab. If that doesn't sound right, try reloading the page."
   },
--- a/locales/en/app.json
+++ b/locales/en/app.json
@@ -82,6 +82,8 @@
     "e2ee_unsupported_description": "Your web browser does not support encrypted calls. Supported browsers include Chrome, Safari, and Firefox 117+.",
     "generic": "Something went wrong",
     "generic_description": "Submitting debug logs will help us track down the problem.",
+    "insufficient_capacity": "Insufficient capacity",
+    "insufficient_capacity_description": "The server has reached its maximum capacity and you cannot join calls at this time. Try again later, or contact your server admin if the problem persists.",
     "open_elsewhere": "Opened in another tab",
     "open_elsewhere_description": "{{brand}} has been opened in another tab. If that doesn't sound right, try reloading the page."
   },
--- a/src/livekit/useECConnectionState.ts
+++ b/src/livekit/useECConnectionState.ts
@@ -7,6 +7,7 @@ Please see LICENSE in the repository root for full details.
 
 import {
   type AudioCaptureOptions,
+  ConnectionError,
   ConnectionState,
   type LocalTrack,
   type Room,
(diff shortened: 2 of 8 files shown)


locales/en/app.json
```diff
@@ -82,6 +82,8 @@
     "e2ee_unsupported_description": "Your web browser does not support encrypted calls. Supported browsers include Chrome, Safari, and Firefox 117+.",
     "generic": "Something went wrong",
     "generic_description": "Submitting debug logs will help us track down the problem.",
+    "insufficient_capacity": "Insufficient capacity",
+    "insufficient_capacity_description": "The server has reached its maximum capacity and you cannot join calls at this time. Try again later, or contact your server admin if the problem persists.",
```
hughns:
```suggestion
    "insufficient_capacity_description": "The server has reached its maximum capacity and you cannot join the call at this time. Try again later, or contact your server admin if the problem persists.",
```

hughns (commented):
Suggested wording change.

locales/en/app.json
hughns: Reason for this is that whilst you can't join *this* call, you might be able to join *another* call where a different SFU is being used.

src/livekit/useECConnectionState.ts
```diff
@@ -129,12 +132,20 @@ async function connectAndPublish(
...
+  try {
+    await livekitRoom!.connect(sfuConfig!.url, sfuConfig!.jwt, {
+      // Due to stability issues on Firefox we are testing the effect of different
+      // timeouts, and allow these values to be set through the console
+      peerConnectionTimeout: window.peerConnectionTimeout ?? 45000,
+      websocketTimeout: window.websocketTimeout ?? 45000,
+    });
+  } catch (e) {
+    // LiveKit uses 503 to indicate that the server has hit its track limits
+    // or equivalently, 429 in LiveKit Cloud
```
hughns, line 144:
```suggestion
    // or equivalently, 429 in LiveKit Cloud
    // For reference, the 503 response is generated at: https://github.com/livekit/livekit/blob/fcb05e97c5a31812ecf0ca6f7efa57c485cea9fb/pkg/service/rtcservice.go#L171

```

hughns (approved):
Approved, subject to wording change review from @fkwp.

hughns:
Please note, this new error won't be visible in clients until https://github.com/element-hq/element-call/issues/2955 is completed.

PR #3027, opened 2025-02-26
Error Handling: gracefully handle missing livekit service url in config

### LINKED ISSUE #2971: Show an error screen for when no MatrixRTC backend/SFU is exposed via .well-known and no fallback configured
By fallback we mean in the EC `config.json`.

BillCarsonFr:
Fixes https://github.com/element-hq/element-call/issues/2971

How to test:
- Update the local `config.json` in `public/` folder and remove the `livekit` entry

Will generate this screen:

I tried to put a clear and not technical message, with a call to action to contact the server admin.
The `error code` should be enough for the admin to know what to do

I tried to do a more complete test that would setup a lobby with a video preview and trigger this error, but I would need more time to do so. Any direction could help. 

## Update

Following internal discussions did a quick refactoring to improve error structure (add a category to match this [document](https://docs.google.com/document/d/15rcOVpfHRbYPDodLQf2xfHHPRTVuJzpY7VAJ3c1LDbM/edit?tab=t.0)), and also added more details on the RTC focus error: Include the domain and better title/description.

See this commit: https://github.com/element-hq/element-call/pull/3027/commits/2ba803fcefce79ae5fe040de9b7c110fa14248d1

Now Error is:
image


src/utils/ec-errors.ts
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Configuration problem due to no MatrixRTC backend/SFU is exposed via .well-known and no fallback configured.
+   */
+  MISSING_MATRIX_RTC_FOCUS = "MISSING_MATRIX_RTC_FOCUS",
+  CONNECTION_LOST_ERROR = "CONNECTION_LOST_ERROR",
+  // UNKNOWN_ERROR = "UNKNOWN_ERROR",
+}
+
+export enum ErrorCategory {
+  /** Calling is not supported, server miss-configured (JWT service missing, no MSC support ...)*/
```
hughns:
```suggestion
  /** Calling is not supported, server misconfigured (JWT service missing, no MSC support ...)*/
```

src/utils/ec-errors.ts
```diff
@@ -0,0 +1,73 @@
+/*
```
hughns: I'm not sure about this filename. Specifically the `ec-` prefix. Maybe it would be better as just `errors.ts`.

However, we could rename it in a subsequent PR.

src/utils/ec-errors.ts
```diff
@@ -0,0 +1,73 @@
...
+  /** Calling is not supported, server miss-configured (JWT service missing, no MSC support ...)*/
+  CONFIGURATION_ISSUE = "CONFIGURATION_ISSUE",
+  NETWORK_CONNECTIVITY = "NETWORK_CONNECTIVITY",
+  // SYSTEM_FAILURE / FEDERATION_FAILURE ..
+}
+
+/**
+ * Structure for errors that occur when using ElementCall.
+ */
+export class ElementCallError extends Error {
```
hughns: Should this be `abstract` to make it clear that we expect this to be sub-classed rather than using "as-is"?

src/RichError.tsx
```diff
@@ -65,3 +68,52 @@ export class InsufficientCapacityError extends RichError {
...
+    <ErrorView Icon={icon} title={title}>
+      <p>
+        {error.localisedMessage ? (
+          error.localisedMessage
+        ) : (
+          <Trans
+            i18nKey="error.unexpected_ec_error"
+            components={[<b />, <code />]}
+            values={{ errorCode: error.code }}
+          />
```
toger5: Any reason this would not work?
```suggestion
        {error.localisedMessage ??
          <Trans
            i18nKey="error.unexpected_ec_error"
            components={[<b />, <code />]}
            values={{ errorCode: error.code }}
          />
```

src/room/GroupCallView.tsx
```diff
@@ -165,6 +167,22 @@ export const GroupCallView: FC<Props> = ({
...
+    perParticipantE2EE: boolean,
+  ): Promise<void> => {
+    try {
+      await enterRTCSession(rtcSession, perParticipantE2EE);
+    } catch (e) {
+      if (e instanceof ElementCallError) {
+        // e.code === ErrorCode.MISSING_LIVE_KIT_SERVICE_URL)
+        setEnterRTCError(e);
+      } else {
+        logger.error(`Unknown Error while entering RTC session`, e);
```
hughns, line 186: Should we be re-throwing the error here which I *think* would make it be the same as the *current* error behaviour?

```suggestion
        logger.error(`Unknown Error while entering RTC session`, e);
        // TODO: rather than re-throw we probably want to display some generic
        // error message
        throw e;
```

locales/en/app.json
```diff
@@ -84,8 +85,10 @@
     "generic_description": "Submitting debug logs will help us track down the problem.",
     "insufficient_capacity": "Insufficient capacity",
     "insufficient_capacity_description": "The server has reached its maximum capacity and you cannot join the call at this time. Try again later, or contact your server admin if the problem persists.",
+    "matrix_rtc_focus_missing": "The server is not configured to work with {{brand}}. Please contact your server admin (Domain: {{domain}}, Error Code: {{ errorCode }}).",
     "open_elsewhere": "Opened in another tab",
-    "open_elsewhere_description": "{{brand}} has been opened in another tab. If that doesn't sound right, try reloading the page."
+    "open_elsewhere_description": "{{brand}} has been opened in another tab. If that doesn't sound right, try reloading the page.",
+    "unexpected_ec_error": "An unexpected error occurred (<0>Error Code:</0> <1>{{ errorCode }}</1>). Please contact your server administrator."
```
hughns: For consistency:

```suggestion
    "unexpected_ec_error": "An unexpected error occurred (<0>Error Code:</0> <1>{{ errorCode }}</1>). Please contact your server admin."
```

src/utils/ec-errors.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+}
+
+export class ConnectionLostError extends ElementCallError {
+  public constructor() {
+    super(
+      "Connection lost",
+      ErrorCode.CONNECTION_LOST_ERROR,
+      ErrorCategory.NETWORK_CONNECTIVITY,
+    );
```
hughns: Instead of falling back to the generic `unexpected_ec_error` error I think in totality the `ConnectionLostError` should say something like:

> **Connection lost**
> 
> The connection to the server has been lost. Please check your network connection and try reconnecting. If the problem persists please contact your server admin (<0>Error Code:</0> <1>{{ errorCode }}</1>).

hughns (changes requested):
One material thing to look at:

- wording for ConnectionLostError

Otherwise just nits.

src/room/GroupCallView.tsx
BillCarsonFr: No it depends on when this call is made. 
If you re-throw:
- and it is called during rendering it will crash the component thus triggering the Error Boundary
- and it is called on a button click, it will not crash rendering, so error boundary will not ne triggered

locales/en/app.json
BillCarsonFr: should be updated here https://localazy.com/p/element-call/source-language/_a7039230807923691922

Will do it

src/utils/ec-errors.ts
BillCarsonFr: Currently this not handled by the generic component, there is a specific case for this error (use direclty a custom `FullScreenView` instead of the `ErrorPage`)

https://github.com/element-hq/element-call/blob/7761e042fbf2a319a1fac73905a44ae071d9bf7d/src/room/GroupCallView.tsx#L355

This case has some callToAction / buttons not currently handled by the generic case, and also there is some down>up communication between component not supported yet (currently InCallView will crash, triggering the top ErrorBoundary error fallback defined in the child GroupCallView -parent of the crashing component- that is updating the internal state) => Need some more thinking

We need additional work and a refactoring to handle that in a more clean way.

src/room/GroupCallView.tsx
hughns: So, are you saying that the way it is written the behaviour is not changed in the case of a non-`ElementCallError` being encountered?

src/utils/ec-errors.ts
hughns: Are you saying that https://github.com/element-hq/element-call/blob/4701e413fd2a84f4b688096de26987312d379802/src/RichError.tsx#L88-L91 is not currently used? I think that is why my confusion stems from.

src/RichError.tsx
BillCarsonFr: right, this works. Thx

src/room/GroupCallView.tsx
BillCarsonFr: Depending on where it was called yes. But maybe I could just turn it into a new Unknown ElementCallError?
let me try something. Will reply with the commit

src/utils/ec-errors.ts
BillCarsonFr: Yes not yet used. Initially wanted to do refactoring in this PR but it needs more work (to handle the call to action -reconnect- and the leave call when happens.
Can remove for now? to make it clear?

src/room/GroupCallView.tsx
BillCarsonFr: @hughns 
 https://github.com/element-hq/element-call/pull/3027/commits/1fbf8e268622733b6569430859d16d924fee4759
 
 With that we will capture any kind of errors, it will be properly reported to Sentry, though the user facing message will not be very usefull

src/utils/ec-errors.ts
BillCarsonFr: @hughns removed for now https://github.com/element-hq/element-call/pull/3027/commits/f38adf12d9a91fc2f07bb1320fd6650fe3ca58e3

src/utils/ec-errors.ts
BillCarsonFr: Will tackle that in follow up refactoring PR if ok?

src/utils/ec-errors.ts
BillCarsonFr: done https://github.com/element-hq/element-call/pull/3027/commits/f8a5de604f17aa82df3c1b234fc30ccf3e662d15

src/utils/ec-errors.ts
BillCarsonFr: Also we can address connection lost message as a separate PR.
You want 
>  **Connection lost**
>
> The connection to the server has been lost. Please check your network connection and try reconnecting. If the problem persists please contact your server admin (<0>Error Code:</0> <1>{{ errorCode }}</1>).

Current is 

>  **Connection lost**
>
> You were disconnected from the call.
> **Reconnect**

There are already translations for that, need to update:
https://localazy.com/p/element-call/source-language/_a7577193556043316521
and
https://localazy.com/p/element-call/source-language/_a7085689531576827262

src/room/GroupCallView.tsx
hughns: Okay, this feels better than before. We can consider it further as we do more work in later PRs 👍

hughns (approved):
This is looking good to me 👍 

As we are using existing strings for the "connection lost" description I think leave the wording as is.

src/utils/ec-errors.ts
hughns: Right, thank you for clearing this up for me!

I think we leave the existing message as is. I suspect it will be reworked in the coming weeks anyhow 👍

PR #3038, opened 2025-03-03
Default vite config to support using local build of crypto wasm

BillCarsonFr:
Quick vite config to support running a local build of the crypto-wasm library.

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -5,7 +5,7 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 Please see LICENSE in the repository root for full details.
 */
 
-import { defineConfig, loadEnv } from "vite";
+import { defineConfig, loadEnv, searchForWorkspaceRoot } from "vite";
 import svgrPlugin from "vite-plugin-svgr";
 import { createHtmlPlugin } from "vite-plugin-html";
 import { codecovVitePlugin } from "@codecov/vite-plugin";
@@ -67,6 +67,14 @@ export default defineConfig(({ mode, packageType }) => {
   return {
     server: {
       port: 3000,
+      fs: {
+        allow: [
+          // search up for workspace root
+          searchForWorkspaceRoot(process.cwd()),
+          // Allow fs access to local build of wasm
+          "../matrix-rust-sdk-crypto-wasm/pkg",
+        ],
+      },
     },
     build: {
       sourcemap: true,
--- a/vite.config.js
+++ b/vite.config.js
@@ -70,7 +70,7 @@ export default defineConfig(({ mode, packageType }) => {
   // crypto (which could reside anywhere on their file system), Vite
   // needs to be told to recognize it as a legitimate file access.
   const allow = [searchForWorkspaceRoot(process.cwd())];
-  for (path of [
+  for (const path of [
     "node_modules/matrix-js-sdk/node_modules/@matrix-org/matrix-sdk-crypto-wasm",
     "node_modules/@matrix-org/matrix-sdk-crypto-wasm",
   ]) {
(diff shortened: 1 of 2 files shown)


toger5 (approved):
Thx.
Any chance we can have some docs on how to do this in general?
Does the js-sdk include this documentation? If so a link to the correct section in the js-sdk docs would be enough. Otherwise a very short section in the readme would be nice I think?

vite.config.js
```diff
@@ -58,6 +58,14 @@ export default defineConfig(({ mode }) => {
   return {
     server: {
       port: 3000,
+      fs: {
+        allow: [
+          // search up for workspace root
+          searchForWorkspaceRoot(process.cwd()),
+          // Allow fs access to local build of wasm
+          "../matrix-rust-sdk-crypto-wasm/pkg",
```
robintown: Every developer's file system and exact setup of linked packages could look different - for instance, I experience this same problem when I use a linked `../matrix-js-sdk` package. I think the following, resolving the symlinks with `realpath`, would work for us both:

```suggestion
          // The crypto WASM module is imported dynamically. Since it's common
          // for developers to use a linked copy of matrix-js-sdk or Rust
          // crypto (which could reside anywhere on their file system), Vite
          // needs to be told to recognize it as a legitimate file access.
          await realpath(
            "node_modules/matrix-js-sdk/node_modules/@matrix-org/matrix-sdk-crypto-wasm",
          ),
```

robintown:
> Does the js-sdk include this documentation?

I don't believe matrix-js-sdk tries to have any documentation today on how to use it with specific bundlers

vite.config.js
robintown: A complication: that path might not exist if not using a linked copy of matrix-js-sdk. (because the dependency will instead be hoisted to node_modules/@matrix-org/matrix-sdk-crypto-wasm)

toger5:
This is super helpful (almost required to have an okay linking experience)
I updated it to use robins realpath trick.
I ended up using the `Sync` version of `realpath` since otherwise it would not pass ci (embedded builds)

vite.config.js
```diff
@@ -63,10 +64,18 @@ export default defineConfig(({ mode, packageType }) => {
       }),
     );
   }
+  const allow = [searchForWorkspaceRoot(process.cwd())];
+  const matrixSdkCryptoWasmPath =
+    "node_modules/matrix-js-sdk/node_modules/@matrix-org/matrix-sdk-crypto-wasm";
```
robintown: We should additionally check for the existence of `node_modules/@matrix-org/matrix-sdk-crypto-wasm` to support cases where Yarn decides to hoist the package out of matrix-js-sdk (i.e. when you're not using a linked matrix-js-sdk, but simply a linked matrix-sdk-crypto-wasm)

toger5: I tried that and this works automatically
if i do not link it can load the rust crypto without any yarn adjustments

toger5: oh but if that is linked. got it
I had it like that before but reverted because i did not see the case where you only want to link crypto but the js-sdk

robintown (commented):
Cool, one last thing: could this get a comment like in https://github.com/element-hq/element-call/pull/3038#discussion_r1978584784?

vite.config.js
robintown: Yep, I recall Valere originally opened this PR because of that use case

vite.config.js
```diff
@@ -64,6 +64,11 @@ export default defineConfig(({ mode, packageType }) => {
       }),
     );
   }
+
+  // The crypto WASM module is imported dynamically. Since it's common
+  // for developers to use a linked copy of matrix-js-sdk or Rust
+  // crypto (which could reside anywhere on their file system), Vite
+  // needs to be told to recognize it as a legitimate file access.
   const allow = [searchForWorkspaceRoot(process.cwd())];
   for (path of [
```
robintown:
```suggestion
  for (const path of [
```

robintown (approved):

PR #3043, opened 2025-03-04
refactoring: Call Error management using error scope and component

BillCarsonFr:
**DRAFT**
Proposal for a more generic way of handling errors.
TODO: 
- [ ] Refactor `InsufficientCapacityError` to use the new components

### Actual: ErrorBoundary
Currently we are using ErrorBoundary to report errors.
That means that for "async" errors reporting we need to have a state to store the error then crash the rendering of the component when it is set so that the ErrorBoundary catch and report it.

I am not very familiar with react patterns, but it seems a bit off to turn an async error into a rendering error? Also on error the all component hierarchy is unmounted to display the error fallback UI.
We might want more flexibility by having a custom error management that would allow components to set errors but also actions and callbacks wihtout unmounting everything?

### Proposed: CallErrorStateProvider/CallErrorDisplay

CallErrorDisplay will display the call error if there is one (via FullScreenView).
Should be surrounded by a CallErrorStateProvider.
Any component in the hierarchy can access the call error context using useCallErrorDisplay. If the error is set then the ErrorPage will be displayed. A component can set up custom buttons/actions that will be displayed on the error page.
```
const { setCallErrorState } = useCallErrorDisplay();
// ...
    setCallErrorState({
         reason: new ConnectionLostError(),
         actions: [
               {
                labelKey: "call_ended_view.reconnect_button",
                onClick: reconnect,
              },
         ]
    }
// ...
  src/utils/CallErrorDisplay.tsx
```

Will still keep ErrorBoundary but this is strictly for rendering errors / un-expected should not happen errors.

Let's discuss


src/utils/CallErrorDisplay.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  const actionCallbacks = actions?.map((action) => {
+    return (): void => {
+      setCallErrorState(null);
+      action.onClick();
+    };
+  });
+  return (
+    <FullScreenView>
+      <ErrorView
```
robintown, line 81: ErrorView is already a form of FullScreenView, so this is redundant

src/utils/CallErrorDisplay.tsx
robintown, line 1: I would say that the src/utils directory is meant to not contain any React components. Maybe it makes sense to collect everything in an src/error directory at this point?

src/utils/errors.ts
```diff
@@ -66,9 +68,21 @@ export class MatrixRTCFocusMissingError extends ElementCallError {
 export class ConnectionLostError extends ElementCallError {
   public constructor() {
     super(
-      "Connection lost",
+      "ConnectionLostError",
```
robintown, line 71: Just noticing this: Is it possible for ElementCallError to determine its own name via [`obj.constructor.name`](https://stackoverflow.com/a/8715426), rather than requiring each subclass to specify it manually?

BillCarsonFr:
closed in favor of https://github.com/element-hq/element-call/pull/3058

src/utils/CallErrorDisplay.tsx
BillCarsonFr: If I remove the FullScreenView the content is not center anymore

PR #3046, opened 2025-03-04
Patch for v0.7: Send a 'close' action when the widget is ready to close

robintown:
This is a watered-down version of the changes in 832a5aaf5732aa6aaa07f9bacc741ddc961def0b which should be suitable as a patch to the v0.7 release series. It makes Element Call widgets send a 'close' action in addition to the original 'hangup' action, and nothing more.

The point is that we can then deploy this change to call.element.io with little effort so that clients are more immediately free to expect a 'close' action from Element Call (and thus we unblock https://github.com/element-hq/element-call/issues/2955 and https://github.com/element-hq/element-web/issues/29196).

For https://github.com/element-hq/voip-internal/issues/304

### DIFF
--- a/src/rtcSessionHelpers.ts
+++ b/src/rtcSessionHelpers.ts
@@ -135,7 +135,20 @@ const widgetPostHangupProcedure = async (
   // We send the hangup event after the memberships have been updated
   // calling leaveRTCSession.
   // We need to wait because this makes the client hosting this widget killing the IFrame.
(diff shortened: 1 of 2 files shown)


robintown:
I messed up the branches in https://github.com/element-hq/element-call/pull/3042 and forcibly closed the PR. Please can I have another review?

fkwp (approved):

PR #3047, opened 2025-03-04
Release v0.7.2 (for deployment preview only)

robintown:
For testing of https://github.com/element-hq/voip-internal/issues/304


robintown:
We now have https://call-next.element.io/ (yay)

PR #3049, opened 2025-03-04
Change LK log level to warn

toger5:
This limits the amount of log messages that especially impact safari.


robintown (approved):
Let's give it a try

PR #3051, opened 2025-03-04
Respect the returnToLobby flag

robintown:
See the commits for the details, but what happened is https://github.com/element-hq/element-call/pull/3021 regressed our handling of the `returnToLobby` URL parameter. This was causing video rooms on develop.element.io to show a blank screen after leaving the call rather than returning to the lobby view.

**Will need backporting** to `release-v0.7.2` because this implies a behavior change in the 'close' action that https://github.com/element-hq/element-web/pull/29254 would adapt to.

### DIFF
--- a/src/UrlParams.test.ts
+++ b/src/UrlParams.test.ts
@@ -110,8 +110,8 @@ describe("UrlParams", () => {
   });
 
   describe("returnToLobby", () => {
-    it("is true in SPA mode", () => {
-      expect(getUrlParams("?returnToLobby=false").returnToLobby).toBe(true);
+    it("is false in SPA mode", () => {
+      expect(getUrlParams("?returnToLobby=true").returnToLobby).toBe(false);
     });
 
     it("defaults to false in widget mode", () => {
--- a/src/UrlParams.ts
+++ b/src/UrlParams.ts
@@ -264,7 +264,7 @@ export const getUrlParams = (
       "skipLobby",
       isWidget && intent === UserIntent.StartNewCall,
     ),
-    returnToLobby: isWidget ? parser.getFlagParam("returnToLobby") : true,
+    returnToLobby: isWidget ? parser.getFlagParam("returnToLobby") : false,
     theme: parser.getParam("theme"),
     viaServers: !isWidget ? parser.getParam("viaServers") : null,
     homeserver: !isWidget ? parser.getParam("homeserver") : null,
--- a/src/UrlParams.ts
+++ b/src/UrlParams.ts
@@ -264,6 +264,8 @@ export const getUrlParams = (
       "skipLobby",
       isWidget && intent === UserIntent.StartNewCall,
     ),
+    // In SPA mode the user should always exit to the home screen when hanging
+    // up, rather than being sent back to the lobby
     returnToLobby: isWidget ? parser.getFlagParam("returnToLobby") : false,
     theme: parser.getParam("theme"),
     viaServers: !isWidget ? parser.getParam("viaServers") : null,
--- a/src/room/GroupCallView.tsx
+++ b/src/room/GroupCallView.tsx
@@ -495,9 +495,7 @@ export const GroupCallView: FC<Props> = ({
     }
   } else if (left && widget !== null) {
     // Left in widget mode:
-    if (!returnToLobby) {
-      body = null;
-    }
+    body = returnToLobby ? lobbyView : null;
   } else if (preload || skipLobby) {
     body = null;
   } else {
--- a/src/rtcSessionHelpers.test.ts
+++ b/src/rtcSessionHelpers.test.ts
@@ -6,7 +6,7 @@ Please see LICENSE in the repository root for full details.
 */
 
 import { type MatrixRTCSession } from "matrix-js-sdk/src/matrixrtc/MatrixRTCSession";
-import { expect, test, vi } from "vitest";
+import { expect, onTestFinished, test, vi } from "vitest";
 import { AutoDiscovery } from "matrix-js-sdk/src/autodiscovery";
 import EventEmitter from "events";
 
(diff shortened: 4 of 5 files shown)


src/rtcSessionHelpers.ts
```diff
@@ -149,14 +150,13 @@ const widgetPostHangupProcedure = async (
...
   // error occurs we should keep the widget open until the user reads it.
-  if (cause === "user") {
+  if (cause === "user" && !getUrlParams().returnToLobby) {
     try {
       await widget.api.transport.send(ElementWidgetActions.Close, {});
     } catch (e) {
       logger.error("Failed to send close action", e);
     }
     widget.api.transport.stop();
-    PosthogAnalytics.instance.logout();
```
hughns: How is this related? If not done here then where is it done?

src/rtcSessionHelpers.test.ts
```diff
@@ -15,11 +15,17 @@ import { mockConfig } from "./utils/test";
...
+const getUrlParams = vi.hoisted(() => vi.fn(() => ({})));
+vi.mock("./UrlParams", () => ({ getUrlParams }));
+
 const actualWidget = await vi.hoisted(async () => vi.importActual("./widget"));
 vi.mock("./widget", () => ({
   ...actualWidget,
   widget: {
-    api: { transport: { send: vi.fn(), reply: vi.fn(), stop: vi.fn() } },
+    api: {
+      setAlwaysOnScreen: () => {},
```
hughns:
```suggestion
      setAlwaysOnScreen: (): void => {},
```

or

```suggestion
      setAlwaysOnScreen: vi.fn(),
```

src/UrlParams.ts
```diff
@@ -264,7 +264,7 @@ export const getUrlParams = (
       "skipLobby",
       isWidget && intent === UserIntent.StartNewCall,
     ),
-    returnToLobby: isWidget ? parser.getFlagParam("returnToLobby") : true,
+    returnToLobby: isWidget ? parser.getFlagParam("returnToLobby") : false,
```
hughns, line 269: I don't understand why this is inverted. For SPA we always want to be showing the lobby which is achieved with `returnToLobby` set to `true`.

hughns (changes requested):

src/UrlParams.ts
robintown: My understanding is: `skipLobby` controls whether we show the lobby when joining the call, which we do want in SPA, and `returnToLobby` controls whether we show the lobby when leaving the call, which we don't want in SPA. The reality is we just don't check the flag at all in SPA mode (the behavior is more or less hardcoded) but, I thought it would be wise to still match its value to the expected behavior.

src/rtcSessionHelpers.ts
robintown: It's not related, I should extract to a separate PR and explain there

robintown: Force pushed to extract this without causing future merge conflicts

robintown: → https://github.com/element-hq/element-call/pull/3053

src/UrlParams.ts
hughns: Thank you for the comment to clarify. 👍

hughns (approved):

src/UrlParams.test.ts
```diff
@@ -110,8 +110,8 @@ describe("UrlParams", () => {
   });
 
   describe("returnToLobby", () => {
-    it("is true in SPA mode", () => {
-      expect(getUrlParams("?returnToLobby=false").returnToLobby).toBe(true);
+    it("is false in SPA mode", () => {
+      expect(getUrlParams("?returnToLobby=true").returnToLobby).toBe(false);
     });
```
toger5, line 115: Was this test just wrong initially? Or was there a time where we planned to always return to lobby in spa mode?

src/room/GroupCallView.tsx
```diff
@@ -495,9 +495,7 @@ export const GroupCallView: FC<Props> = ({
     }
   } else if (left && widget !== null) {
     // Left in widget mode:
-    if (!returnToLobby) {
-      body = null;
-    }
+    body = returnToLobby ? lobbyView : null;
   } else if (preload || skipLobby) {
```
toger5, line 499: How has the previous version worked if we need to set it to `lobbyView` in the widget case?

toger5, line 499: Could the whole block be changed to:
```
} else if (left && widget && !returnToLobby) {
    // Left in widget mode:
    body = null;
} else ...
  ```
  Then the final `else` block would set `body` to `lobbyView`

toger5: Not sure it makes it more readable however. The question about how the previous version worked is more important here.

toger5 (approved):
Just a couple of clarifications if changes are cosmetic or if it was broken before. I am not understanding how things have been working before.

src/UrlParams.test.ts
robintown: It seems `returnToLobby` was set to a misleading value in SPA mode, because the SPA doesn't actually do anything with the flag. I changed it here: https://github.com/element-hq/element-call/pull/3051/files#diff-d940aba1d89169ffa38dbeac7ec6282e01dbd906fdca5561fed8c0d7ee5b83a0R269

src/room/GroupCallView.tsx
robintown: It worked because the `else` branch at the bottom of this condition tree used to be [outside of the condition tree](https://github.com/element-hq/element-call/blob/9b59544db29c09d5953649c9b15a0b40744b62dd/src/room/GroupCallView.tsx#L424), so returnToLobby=true would cause this to fall through to the `else` case and return the lobby

robintown: The `preload || skipLobby` case seems to complicate this, enough that I wasn't sure whether an `else if (left && widget && !returnToLobby)` would really do the right thing. So I stuck to structuring this `left && widget !== null` branch in the same way as the `left && widget === null` branch, which I think brings its own kind of clarity.

toger5: oh I see so this was not working after the change and until now. Cool Thanks for the explanation!

PR #3053, opened 2025-03-05
Enable analytics only while authenticated

robintown:
We shouldn't be calling `PosthogAnalytics.instance.logout()` when leaving the call in widget mode, because all this does is reset your analytics ID. In the big picture this is probably inflating our user count metrics.

On the other hand, we should log out of PostHog and reset our analytics ID is when the user is logging out. This matches the [behavior in Element Web](https://github.com/element-hq/element-web/blob/e9a3625bd6e9a64f216e3caeabca66f48b649332/src/Lifecycle.ts#L887) and makes sense, I think, because logging out is essentially a request for the app to forget who you are. This means we should also start analytics at the point of logging in / reauthenticating.

I noticed while making this change that there was an unused branch in `setClient`, so I cleaned it up rather than making myself update it.


hughns (approved):
I've had a chance to review the context for this now and it does makes sense.

I did wonder if we should do call Posthog.logout() at the point that the user toggles the opt-in from on to off, such that if they toggle back on again they might get a new ID. But, on reflection, it seems reasonable that the same ID would be in use.

PR #3060, opened 2025-03-06
Fix multi subscription to displaynameMap with observableSubject

toger5:
fixes: https://github.com/element-hq/voip-internal/issues/300

This results in the map being created proportionally to the number of members in the call. Each change in memberships will call multiple resubscription to the map observable.

### DIFF
--- a/src/state/CallViewModel.ts
+++ b/src/state/CallViewModel.ts
@@ -69,7 +69,11 @@ import {
   ScreenShareViewModel,
   type UserMediaViewModel,
 } from "./MediaViewModel";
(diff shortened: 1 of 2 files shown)


robintown (changes requested):
This singleSubscriberSubject$ utility would be redundant to the usual 'share' and 'state' operators as you've picked up on in https://github.com/element-hq/element-call/pull/3062

toger5:
We prefer the solution from: https://github.com/element-hq/element-call/pull/3062

So closing this.

PR #3061, opened 2025-03-06
Refactor to simplify data flow from Matrix event listeners into call view logic

toger5:
On top of https://github.com/element-hq/element-call/pull/3062 a potential refactor for a clearer data flow.

Calls to `fromEvent()` are piped via `share()` to multiplex subscribers.

### DIFF
--- a/src/state/CallViewModel.ts
+++ b/src/state/CallViewModel.ts
@@ -40,6 +40,7 @@ import {
   of,
   race,
   scan,
+  share,
   skip,
   startWith,
   switchAll,
@@ -464,6 +465,16 @@ export class CallViewModel extends ViewModel {
       },
     );
 
+  private readonly membershipsChanged$ = fromEvent(
+    this.matrixRTCSession,
+    MatrixRTCSessionEvent.MembershipsChanged,
+  ).pipe(share());
+
+  private readonly roomMembers$ = fromEvent(
+    this.matrixRTCSession.room,
+    RoomStateEvent.Members,
+  ).pipe(share());
+
   /**
    * Displaynames for each member of the call. This will disambiguate
    * any displaynames that clashes with another member. Only members
@@ -471,9 +482,9 @@ export class CallViewModel extends ViewModel {
    */
   public readonly memberDisplaynames$ = merge(
     // Handle call membership changes.
-    fromEvent(this.matrixRTCSession, MatrixRTCSessionEvent.MembershipsChanged),
+    this.membershipsChanged$,
     // Handle room membership changes (and displayname updates)
-    fromEvent(this.matrixRTCSession.room, RoomStateEvent.Members),
+    this.roomMembers$,
   ).pipe(
     startWith(null),
     map(() => {
@@ -496,6 +507,7 @@ export class CallViewModel extends ViewModel {
       }
       return displaynameMap;
     }),
+    this.scope.state(),
   );
 
   /**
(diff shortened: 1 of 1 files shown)


src/state/CallViewModel.ts
```diff
@@ -496,6 +507,7 @@ export class CallViewModel extends ViewModel {
       }
       return displaynameMap;
     }),
+    this.scope.state(),
```
robintown: This, I take it, is the line that actually fixes things

robintown (commented):
I'm not sure what we gain by making the membershipsChanged$ and roomMembers$ observables hot, as a fromEvent observable is very lightweight to create and execute.

hughns:
> I'm not sure what we gain by making the membershipsChanged$ and roomMembers$ observables hot, as a fromEvent observable is very lightweight to create and execute.

The rationale for `membershipsChanged$` is to make it so the `CallViewModel` only "listens" once to the events from `MatrixRTCSession` and so it is clearer what the data flow is.

Because `roomMembers$` is only subscribed to once, the benefit is less so, other than making the data flow intent clearer.

src/state/CallViewModel.ts
hughns: Yes, this is the important step.

toger5:
Lets keep this open since it might be a good refactor independent to the actual fix (thanks for renaming it accordingly)

src/state/CallViewModel.ts
```diff
@@ -464,16 +465,36 @@ export class CallViewModel extends ViewModel {
...
+
+  /**
+   * Observable for changes to the Matrix Room member data.
+   *
+   * We do this to ensure that we only listen once to the event and then share internally.
+   */
+  private readonly roomMembers$ = fromEvent(
+    this.matrixRTCSession.room,
+    RoomStateEvent.Members,
+  ).pipe(share());
```
toger5, line 486: What would be the advantage to use share here instead of context.state()?

hughns: I don't think we need/want the `distinct()` operator. The same for `shareReplay()` vs `share()`.

It would probably make sense to make this explicit in a docstring for the `roomMember$` and `membershipsChanged$`.

toger5:
@robintown I cannot review this since I created the PR. It seems like it is easier to read to me.
From my pov we can merge it.

src/state/CallViewModel.ts
robintown, line 486: Any time we manually create a hot observable with `share`, we should also pipe it through `this.scope.bind` to ensure that its execution will terminate when the ViewModel is destroyed.

robintown (commented):
I still think that there's no need to be afraid of `fromEvent` observables being cold, and executing multiple times, because event emitters are already "hot" and designed to allow multiple subscribers to come and go without impacting each others' data flow. I won't block it if you both find this more readable.

src/state/CallViewModel.ts
robintown: Note that `this.scope.bind` needs to come before the `share` operator, like in the implementation for `ObservableScope.state`

toger5:
Where do we go with this? Should we close it since it does not introduce an obvious gain and reading through it again I feel like it is more consistent how we use fromEvent in other places without this change?

hughns:
> Where do we go with this? Should we close it since it does not introduce an obvious gain and reading through it again I feel like it is more consistent how we use fromEvent in other places without this change?

I'm fine with this too 👍

PR #3062, opened 2025-03-06
Reduce redundant calculations of display name map

toger5:
Simpler alternative to https://github.com/element-hq/element-call/pull/3060

Partial fix for: https://github.com/element-hq/voip-internal/issues/300

This means that we effectively "cache" the output of the `map()` section of `memberDisplaynames$` rather than recalculating it.

### DIFF
--- a/src/state/CallViewModel.ts
+++ b/src/state/CallViewModel.ts
@@ -497,7 +497,6 @@ export class CallViewModel extends ViewModel {
       }
       return displaynameMap;
     }),
-    share(),
     this.scope.state(),
   );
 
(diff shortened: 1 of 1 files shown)


src/state/CallViewModel.ts
```diff
@@ -496,6 +497,8 @@ export class CallViewModel extends ViewModel {
       }
       return displaynameMap;
     }),
+    share(),
```
robintown: `share` is redundant to `this.scope.state`; all state observables are hot

toger5:
@hughns I very much like this version with the comment and the removal of the redundant share (`share` + `shareReplay`)

toger5 (commented):
Approved hughs change

hughns (approved):

PR #3058, opened 2025-03-05
refactor: Centralize group call errors in custom GroupCallErrorBoundary

BillCarsonFr:
Centralize handling of group call error in a single `GroupCallErrorBoundary` component. 
This component handles all the different `ElementCallError` and offer callbacks allowing the `GroupCallView` to do leave/clean the call when needed

PR #3063, opened 2025-03-07
Error: use a close button instead of go to home when in widget mode

### LINKED ISSUE #3048: Add 'close' buttons to lobby and fatal error screens in widget mode

Add a 'close' / 'cancel' call-to-action button to the call lobby and all fatal error screens in widget mode, so that there's an obvious way for the user to dismiss the widget whenever they're outside the call.

Because otherwise it might not be clear that the user is free to dismiss the call UI once they've finished reading the error message, or have changed their mind about wanting to join. In Element Web we have a small and non-obvious 'X' icon in the room header which Design wants to get rid of.

Image

_No response_

### Additional context

https://github.com/element-hq/element-call/pull/3011#discussion_r1965221705

BillCarsonFr:
Depends on https://github.com/element-hq/element-call/pull/3058

Fixes: https://github.com/element-hq/element-call/issues/3048

The main change is in the ErrorView, if the widget API is available show a close button
```
      {widget ? (
        <CloseWidgetButton widget={widget} />
      ) : (
        !confineToRoom && <ReturnToHomeButton />
      )}
```

image

There is something a bit annoying, `widget` until then was always accessed via the global scope.
In that case we need it at the `ErrorView` level, so we have to pass it down. Hence all the changes

I suppose `widget` should be managed via a context instead. It would also make testing it a lot more easy.
Any reason for it to not be in a context?
Shoud I proceed and put it in the context?


robintown:
> I suppose `widget` should be managed via a context instead. It would also make testing it a lot more easy.
> Any reason for it to not be in a context?
> Shoud I proceed and put it in the context?

It seemed appropriate at the time for `widget` to be global, since there should actually be one instance of the widget communications per iframe. I guess some non-React code like PosthogAnalytics and rtcSessionHelpers would need to be instantiated/called differently, but if you're up for making that change for ease of testing, I would support it.

src/RichError.tsx
```diff
@@ -31,7 +31,11 @@ const OpenElsewhere: FC = () => {
   const { t } = useTranslation();
 
   return (
-    <ErrorView Icon={PopOutIcon} title={t("error.open_elsewhere")}>
+    <ErrorView
+      widget={null}
```
robintown: It shouldn't normally be possible for a widget to encounter this error but, just to avoid surprises:

```suggestion
      widget={widget}
```

src/room/GroupCallErrorBoundary.tsx
```diff
@@ -103,12 +107,14 @@ interface BoundaryProps {
   children: ReactNode | (() => ReactNode);
   recoveryActionHandler?: RecoveryActionHandler;
   onError?: (error: unknown) => void;
+  widget?: WidgetHelpers | null;
```
robintown: Why is it optional?

BillCarsonFr: The helper can be null but you still have to put the prop in the tag, I would have prefer it to be `| undefined`?
But anyhow I just put it like that for convenience in the tests, but that's probably a bad idea

robintown (approved):

hughns:
This was backported to v0.8.0 via https://github.com/element-hq/element-call/pull/3087

PR #3066, opened 2025-03-07
Cache calls to removeHiddenChars() to fix performance bottleneck in Safari

### LINKED ISSUE #3065: Display name disambiguation is causing main event loop on Safari to be slow
### Steps to reproduce

1. Join a call
2. Add some other participants
3. Check the console logs

#### What did you expect?

Log entries like this:

`MatrixRTCSession"[Perf]: emit MatrixRTCSessionEvent.MembershipsChanged took 5ms"`

#### What happened instead?

The milliseconds was 10x:

`MatrixRTCSession"[Perf]: emit MatrixRTCSessionEvent.MembershipsChanged took 50ms"`

### Operating system

macOS

### Browser information

Safari 18.3

### URL for webapp

call.element.dev

### Will you send logs?

No

hughns:
Fixes #3065

This PR caches the slow calls to unhomoglyph rather than attempting to debug why exactlt it so slow.

### DIFF
--- a/src/utils/displayname-integration.test.ts
+++ b/src/utils/displayname-integration.test.ts
@@ -1,0 +1,40 @@
+/*
+Copyright 2025 New Vector Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+Please see LICENSE in the repository root for full details.
+*/
+
+import { afterEach, beforeAll, describe, expect, test, vi } from "vitest";
+
+import { shouldDisambiguate } from "./displayname";
+import { alice } from "./test-fixtures";
+import { mockMatrixRoom } from "./test";
+
+// Ideally these tests would be in ./displayname.test.ts but I can't figure out how to
+// just spy on the removeHiddenChars() function without impacting the other tests.
+// So, these tests are in this separate test file.
+vi.mock("matrix-js-sdk/src/utils");
+
+describe("shouldDisambiguate", () => {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+  let jsUtils: typeof import("matrix-js-sdk/src/utils");
+
+  beforeAll(async () => {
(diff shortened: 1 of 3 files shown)


hughns:
The splitting of the tests feels like a hacky workaround. I would welcome a better way... 😇

hughns:
I've added a "backport candidate" label: I suggest this goes into a 0.8.x release as it is a performance regression from 0.7.x.

src/utils/displayname-integration.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { afterEach, beforeAll, describe, expect, test, vi } from "vitest";
+
+import { shouldDisambiguate } from "./displayname";
+import { alice } from "./test-fixtures";
+import { mockMatrixRoom } from "./test";
+
+// Ideally these tests would be in ./displayname.test.ts but I can't figure out how to
+// just spy on the removeHiddenChars() function without impacting the other tests.
+// So, these tests are in this separate test file.
```
toger5, line 16: Is the issue here, that `removeHiddenChars` is already mocked in the other tests and you would need to restore them back to the previous mocked/spyOn implementation?

Otherwise, `jsUtils.removeHiddenChars.restoreMocks()` should do the trick

toger5 (approved):

src/utils/displayname-integration.test.ts
hughns: The issue is that the other tests need the real implementation of the js-sdk for the tests to complete.

If I don't do `vi.mock()` then the spy on `removeHiddenChars` does not work.

If I do `vi.mock()` without implementation then the other tests fail.

If I do `vi.mock()` and provide an implementation of the mock using `vi.importActual()` then it fails in some way (I can't remember how right now).

toger5: I see vi.mock is the issue...

PR #3071, opened 2025-03-10
Use fetch() in a way that works for file URLs

hughns:
fetch returns a response code of 0 when it successfully loads a `file://` resource.

This means we can't just rely on `response.ok`.

Required for https://github.com/element-hq/element-call/issues/2994

### DIFF
--- a/src/config/Config.ts
+++ b/src/config/Config.ts
@@ -74,14 +74,15 @@ async function downloadConfig(
   configJsonFilename: string,
 ): Promise<ConfigOptions> {
   const url = new URL(configJsonFilename, window.location.href);
-  const res = await fetch(url);
+  const response = await fetch(url);
 
-  if (!res.ok || res.status === 404 || res.status === 0) {
+  // if we are running embedded on file:// origin then we can't trust the response status
+  if (!response.ok && (!response.url || !response.url.startsWith("file:"))) {
     // Lack of a config isn't an error, we should just use the defaults.
     // Also treat a blank config as no config, assuming the status code is 0, because we don't get 404s from file:
     // URIs so this is the only way we can not fail if the file doesn't exist when loading from a file:// URI.
     return DEFAULT_CONFIG;
   }
 
-  return res.json();
+  return response.json();
 }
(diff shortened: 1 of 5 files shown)


src/config/Config.ts
```diff
@@ -74,14 +74,15 @@ async function downloadConfig(
   configJsonFilename: string,
 ): Promise<ConfigOptions> {
   const url = new URL(configJsonFilename, window.location.href);
-  const res = await fetch(url);
+  const response = await fetch(url);
 
-  if (!res.ok || res.status === 404 || res.status === 0) {
+  // if we are running embedded on file:// origin then we can't trust the response status
+  if (!response.ok && (!response.url || !response.url.startsWith("file:"))) {
```
toger5: Are you sure `res.status === 404 || res.status === 0` implies `response.ok === false` (just double checking to make sure the or condition was actually unnecassary.)

toger5: why is it an `&&` condition if we cannot trust the response status?
It (`response.ok`) is still guaranteed to be false in a failure case but it is also false if it is a success?

toger5 (commented):
I think the condition
```
 !response.ok &&
        (!response.url || !response.url.startsWith("file:"))
```
Is un-intuitive enough so that is makes sense to move it into
```
/**
* Dedicated explanation why we need this for the file case here.
*/
isErrorNetworkOrFileResponse(response){
return  !response.ok &&
        (!response.url || !response.url.startsWith("file:"));
}
```

`isErrorNetworkOrFileResponse` can probably get a better name

src/config/Config.ts
hughns: `response.ok` is [defined](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok) as:

> The ok read-only property of the Response interface contains a Boolean stating whether the response was successful (status in the range 200-299) or not.

fetch returns a status of 0 for `file://` URLs.

I will put this into a function and clarify with comments.

toger5 (approved):
Thanks with this change i now fully get what is happening.

PR #3075, opened 2025-03-10
Remove default server config as it is not likely to be correct

hughns:
Originally part of https://github.com/element-hq/element-call/issues/2994 but is actually just cleanup.


hughns:
Draft as tests need fixing up.

BillCarsonFr:
closing this one, @robintown to create a new one for this quick cleaning

robintown:
Never mind, this looks okay to merge as-is

robintown (approved):

PR #3076, opened 2025-03-11
Error management: Handle fail to get JWT token

### LINKED ISSUE #3023: Show error screens when we can't get a JWT token
The current behavior when we can't obtain a token from the JWT service (whether due to a connection failure or a configuration error) is to do nothing. The call will sit in an unconnected state forever. As long as we know we won't retry the JWT token fetch, we should bubble this error up to the user as an error screen.

It's going to become more important to get this done as soon as we try to lock down the JWT service to only allow creation of LiveKit rooms for users on the same homeserver deployment as the SFU. (https://github.com/element-hq/voip-internal/issues/294)

BillCarsonFr:
Fixes https://github.com/element-hq/element-call/issues/3023

Draft as based on https://github.com/element-hq/element-call/pull/3058

Add JWT token error handling

image

Best reviewed commit by commit:
- c22412c04559a5c26c6c0e4c2441dae31278e85a refactoring to add API `showGroupCallErrorBoundary` to easily report async error
- 343da0db145a416a1115a7e7452dbcbb4e4372dd Add a utility to retry with backoff some mx API (`doNetworkOperationWithRetry`)
- 04a46ebabe66480e4e13526ae23aab98884b402c catch and handle getOpenId errors


src/room/useCallErrorBoundary.ts
```diff
@@ -0,0 +1,58 @@
...
+export type UseErrorBoundaryApi = {
+  showGroupCallErrorBoundary: (error: ElementCallError) => void;
+};
+
+export function useGroupCallErrorBoundary(): UseErrorBoundaryApi {
+  const context = useContext(GroupCallErrorBoundaryContext);
+
+  if (!context)
+    throw new Error(
+      "useGroupCallErrorBoundary must be used within an GoupCallErrorBoundary",
```
robintown:
```suggestion
      "useGroupCallErrorBoundary must be used within an GroupCallErrorBoundary",
```

src/room/GroupCallErrorBoundaryContextProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+import type { ElementCallError } from "../utils/errors.ts";
+import {
+  GroupCallErrorBoundaryContext,
+  type GroupCallErrorBoundaryContextType,
+} from "./GroupCallErrorBoundaryContext.tsx";
+
+export const GroupCallErrorBoundaryContextProvider: FC<PropsWithChildren> = ({
+  children,
+}) => {
+  const subscribers = useRef<Set<(error: ElementCallError) => void>>(new Set());
```
robintown: `useInitial` should work here and avoid constructing a new Set on each render

src/room/GroupCallErrorBoundaryContext.tsx
```diff
@@ -0,0 +1,18 @@
...
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+Please see LICENSE in the repository root for full details.
+*/
+
+import { createContext } from "react";
+
+import { type ElementCallError } from "../utils/errors.ts";
+
+export type GroupCallErrorBoundaryContextType = {
+  subscribe: (cb: (error: ElementCallError) => void) => () => void;
```
robintown: I find it difficult to understand what this function does from the name alone. It's about subscribing to requests to reset the error boundary, right?

BillCarsonFr: This is my understanding:
If there are 2 child components that call `useGroupCallErrorBoundary` so they both have a "local" state to store the error.
If the local error state is set it will throw an error.
This error is then caught by the ErrorBoundary. If the error boundary is closed it will clear the reset the state of the ErrorBoundary, but not the one of the child component (the component local error state), so it will throw again?

So I need when the ErrorBoundary is reset to reset all the child component useGroupCallErrorBoundary local state.

Initially I wanted just to expose the ErrorBoundary resetError into the context but I cannot (?) as it is not exposed in the sentry error boundary. it is only giving access to it in the fallbackRenderer callback (and it is not possible to update a ref in such callbacks?)

So now when calling `const { showGroupCallErrorBoundary } = useGroupCallErrorBoundary()` in a component it will call `subscibre` so the context has a list of all the "components" reporting errors, so when the errorFallback renderer is called it will then notify all the component to clear there local state if needed

robintown: That makes sense. Could this get a comment and perhaps a more descriptive name?

BillCarsonFr: Sry, I did some more tests, and I am wrong on that. When error boundary is "triggered" it unmounts the child component and replace it with the child renderer, so it clears the state of the local error. No needs to subscribe to do it.
Looks like this case: https://react.dev/learn/preserving-and-resetting-state#different-components-at-the-same-position-reset-state

playwright/errors.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+      route.continue();
+      hasRetriedCallback();
+    }
+  });
+
+  // Join the call
+  await page.getByTestId("lobby_joinCall").click();
+  // Expect that the call has been retried
+  await hasRetriedPromise;
+  await expect(page.getByText("Something went wrong")).not.toBeVisible();
```
robintown: I'm wary of async assertions that expect something to *not* exist - it seems there is opportunity for the assertion to succeed before the page actually updates to the state it's meant to be testing. Is there an equivalent, positive assertion that this could be replaced with?

BillCarsonFr: Agree, I wanted to do something like that but I didn't see anything obvious in the UI showing that I have or not the token.

But maybe I could create a fixture where the tester is invited to an existing call, and then I could check that there are 2 participants?

BillCarsonFr: by adding an inviter and multiple page to the test it was malking it a bit flacky. So I just tested that the API is called a second time and then that the video element is displayed

robintown (approved):

src/utils/errors.ts
```diff
@@ -113,6 +113,19 @@ export class UnknownCallError extends ElementCallError {
   }
 }
 
+export class FailToGetOpenIdToken extends ElementCallError {
+  public constructor(error: Error) {
+    super(
+      t("error.generic"),
+      ErrorCode.OPEN_ID_ERROR,
+      ErrorCategory.CONFIGURATION_ISSUE,
```
robintown, line 121: I just noticed that this is always `CONFIGURATION_ISSUE`, even when the error is due to network connectivity. Is this something you'd like to fix?

BillCarsonFr: > I just noticed that this is always `CONFIGURATION_ISSUE`, even when the error is due to network connectivity. Is this something you'd like to fix?

I added retry for the `request_token` with the built in backoff. So the transiant connectivity problems should be handled properly. I believe that we would need something else that detects when you are offline and handles that globally, WDYT?

There is not much info about the possible failures for `request_token` in the [spec](https://spec.matrix.org/v1.13/client-server-api/#post_matrixclientv3useruseridopenidrequest_token), it only specify 429 error :/
The category for that is up for discussion though, it looked as a sensible default

PR #3077, opened 2025-03-11
Vite embedded build of Element Call

hughns:

This creates a new "embedded" build (vs "full" build) at the vite level. It will be used by a later PR that actually provides platform specific packages.

Embedded build:

- Uses relative URLs
- Uses relative config.json path and other resource loading
- Has a config.json built in
- Doesn't include the public folder (e.g. favicon)

Out of scope:

- this doesn't attempt to exclude SPA functionality, so technically the build could be used in SPA
- the above means that the crypto-wasm binary is included in the build


package.json
```diff
@@ -114,7 +118,8 @@
     "unique-names-generator": "^4.6.0",
     "vaul": "^1.0.0",
     "vite": "^6.0.0",
-    "vite-plugin-html-template": "^1.1.0",
+    "vite-plugin-generate-file": "^0.2.0",
+    "vite-plugin-html": "^3.2.2",
```
robintown, line 122: Out of curiosity what was the reason for switching HTML plugins?

robintown (approved):
You asked whether the term 'embedded' could cause confusion — I personally can't think of a more appropriate name for this packaging mode. I've been making an effort since https://github.com/element-hq/element-call/pull/1497 to refer only to "widget mode" or specific URL parameters in the code, rather than the more nebulous "embedded mode", so I think that finally giving that term a new, limited meaning will help.

package.json
hughns: I can't remember, but will check. You also did it in https://github.com/element-hq/element-call/pull/2922/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R80!

.github/workflows/build.yaml
```diff
@@ -61,3 +63,15 @@ jobs:
...
         type=raw,value=latest-ci
         type=raw,value=latest-ci_{{date 'X' }}
+  build_embedded_element_call:
+    # Use the embedded package vite build
+    uses: ./.github/workflows/build-element-call.yaml
+    with:
+      package: embedded
+      vite_app_version: ${{ github.event.release.tag_name || github.sha }}
+    secrets:
+      SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
```
BillCarsonFr, line 73: Aren't these already in the nuild-element-call.yaml?

BillCarsonFr (approved):
Just a minor comment

src/config/Config.ts
```diff
@@ -29,7 +29,20 @@ export class Config {
...
       Config.internalInstance = internalInstance;
 
-      Config.internalInstance.initPromise = downloadConfig("/config.json").then(
+      let fetchTarget: string;
+
+      if (
+        window.location.pathname.endsWith("/room/") ||
+        window.location.pathname.endsWith("/room")
+      ) {
+        // it looks like we are running as an SPA so use the config at the root
```
hughns:
```suggestion
        // it looks like we are running in standalone mode so use the config at the root
```

package.json
hughns: If I try and achieve the same output with `vite-plugin-html-template` I get these failure modes:

- if `index.html` stays inside `public` then the generated assets parts are all `../assets/foo`
- if you move `index.html` to the project root (to fix the `..`) and set `pages.index.template` to be `index.html` then the EJS compilation errors

Converting it to use `vite-plugin-html` "just works".

.github/workflows/build.yaml
hughns: It does feel a bit convoluted to me, but I think `build-element-call.yaml` is intended to be a generic build step.

Then, the secrets are treated as a parameterised input to the generic build step.

So, depending on where `build-element-call.yaml` is called from you could pass different secrets in.

In reality we don't use this.

PR #3086, opened 2025-03-12
Publish embedded package for releases of Element Call

hughns:

This PR:

- Publishes embedded builds as Tarball, NPM, AAR, SwiftPM for releases
- Publishes full builds as Tarball for releases
- Adds comments to release notes with the built artifact locations
- Does a dry-run (as much as is possible) of the artifact generation on every build

An example of a complete release run is for: https://github.com/element-hq/element-call/releases/tag/v0.9.0-release-test.10
Embedded packaging action run: https://github.com/element-hq/element-call/actions/runs/13904564030
Full packaging action run: https://github.com/element-hq/element-call/actions/runs/13904564028

Release notes look like this:

image


hughns:
Marked as blocked to make sure it doesn't accidentally get merged.

hughns:
I've requested review from folks on each platform team. Please can you review the contents of the respective `/embedded` folder?

embedded/web/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "version": "0.0.0",
+  "files": [
+    "README.md",
+    "LICENSE-AGPL-3.0",
+    "LICENSE-COMMERCIAL",
+    "dist"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/element-hq/element-call.git"
```
t3chguy:
```suggestion
    "url": "git+https://github.com/element-hq/element-call.git"
```

You can run `npm pkg fix` to fix it, otherwise it'll get fixed on every release for the release artifact only

.github/workflows/publish-embedded-packages.yaml
```diff
@@ -0,0 +1,228 @@
...
+          run-id: ${{ github.event.workflow_run.id || github.run_id }}
+          name: build-output-embedded
+          path: ${{ env.FILENAME_PREFIX}}
+      - name: Create Tarball
+        run: tar --numeric-owner -cvzf ${{ env.FILENAME_PREFIX }}.tar.gz ${{ env.FILENAME_PREFIX }}
+      - name: Create Checksum
+        run: find ${{ env.FILENAME_PREFIX }} -type f -print0 | sort -z | xargs -0 sha256sum | tee ${{ env.FILENAME_PREFIX }}.sha256
+      - name: Upload
+        if: ${{ env.DRY_RUN == 'false' }}
+        uses: softprops/action-gh-release@v2
```
t3chguy: Should be pinned to avoid supply chain attack

.github/workflows/publish-embedded-packages.yaml
```diff
@@ -0,0 +1,228 @@
...
+        run: |
+          npm version ${{ env.VERSION }} --no-git-tag-version
+          echo "ARTIFACT_VERSION=$(jq '.version' --raw-output package.json)" >> "$GITHUB_ENV"
+          npm publish --provenance --access public ${{ env.DRY_RUN == 'true' && '--dry-run' || '' }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_RELEASE_TOKEN }}
+
+      - name: Add release note
+        if: ${{ env.DRY_RUN == 'false' }}
+        uses: softprops/action-gh-release@v2
```
t3chguy: Should be pinned to avoid supply chain attack

.github/workflows/publish-embedded-packages.yaml
```diff
@@ -0,0 +1,228 @@
...
+          EC_VERSION: ${{ env.ARTIFACT_VERSION }}
+          ORG_GRADLE_PROJECT_mavenCentralUsername: ${{ secrets.MAVEN_RELEASE_USERNAME }}
+          ORG_GRADLE_PROJECT_mavenCentralPassword: ${{ secrets.MAVEN_RELEASE_PASSWORD }}
+          ORG_GRADLE_PROJECT_signingInMemoryKey: ${{ secrets.GPG_SIGNING_KEY }}
+          ORG_GRADLE_PROJECT_signingInMemoryKeyPassword: ${{ secrets.GPG_SIGNING_KEY_PASSWORD }}
+        run: embedded/android/publish_android_package.sh -s
+
+      - name: Add release note
+        if: ${{ env.DRY_RUN == 'false' }}
+        uses: softprops/action-gh-release@v2
```
t3chguy: Should be pinned to avoid supply chain attack

.github/workflows/publish-embedded-packages.yaml
```diff
@@ -0,0 +1,228 @@
...
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          path: element-call
+
+      - name: 📥 Download built element-call artifact
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4
```
t3chguy, line 150: The inconsistency between pinning some `actions/*` actions and not others is concerning, the security stance should be consistent

.github/workflows/publish-embedded-packages.yaml
```diff
@@ -0,0 +1,228 @@
...
+          git tag -a ${{ env.ARTIFACT_VERSION }} -m "${{ github.event.release.html_url }}"
+
+      - name: Push
+        working-directory: element-call-swift
+        run: |
+          git push --tags ${{ env.DRY_RUN == 'true' && '--dry-run' || '' }}
+
+      - name: Add release note
+        if: ${{ env.DRY_RUN == 'false' }}
+        uses: softprops/action-gh-release@v2
```
t3chguy: Should be pinned to avoid supply chain attack

.github/workflows/publish.yaml
```diff
@@ -1,60 +1,57 @@
...
-        run: |
-          tar --numeric-owner --transform "s/dist/element-call-${TARBALL_VERSION}/" -cvzf element-call-${TARBALL_VERSION}.tar.gz dist
+        run: tar --numeric-owner --transform "s/dist/${{ env.FILENAME_PREFIX }}/" -cvzf ${{ env.FILENAME_PREFIX }}.tar.gz ${{ env.FILENAME_PREFIX }}
+      - name: Create Checksum
+        run: find ${{ env.FILENAME_PREFIX }} -type f -print0 | sort -z | xargs -0 sha256sum | tee ${{ env.FILENAME_PREFIX }}.sha256
       - name: Upload
-        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
-        env:
-          GITHUB_TOKEN: ${{ github.token }}
+        uses: softprops/action-gh-release@v2
```
t3chguy: Should be pinned to avoid supply chain attack

.github/workflows/publish.yaml
```diff
@@ -64,3 +61,23 @@ jobs:
...
         type=semver,pattern=v{{version}}
+  add_docker_release_note:
+    needs: publish_docker
+    name: Add docker release note
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get artifact version
+        run: echo "ARTIFACT_VERSION=${VERSION:1}" >> "$GITHUB_ENV"
+      - name: Add release note
+        uses: softprops/action-gh-release@v2
```
t3chguy: Should be pinned to avoid supply chain attack

t3chguy (approved):
Looks sane otherwise

.github/workflows/publish-embedded-packages.yaml
hughns: I was waiting for renovate bot to do it for me, but I will force it as part of this PR 👍

fkwp (approved):

.github/workflows/publish-embedded-packages.yaml
```diff
@@ -0,0 +1,228 @@
...
+        with:
+          distribution: "temurin"
+          java-version: "17"
+
+      - name: Get artifact version
+        run: echo "ARTIFACT_VERSION=${VERSION:1}" >> "$GITHUB_ENV"
+
+      - name: Publish AAR
+        if: ${{ env.DRY_RUN == 'false' }}
+        # TODO: can we do some kind of dry run?
```
jmartinesp: Gradle can be run with a `--dry-run` parameter to check which tasks would be run. But that would have to go inside the `publish_android_package.sh` script.

.github/workflows/publish-embedded-packages.yaml
```diff
@@ -0,0 +1,228 @@
...
+        with:
+          append_body: true
+          body: |
+            ## Android AAR embedded package
+
+            This package provides a build of Element Call that can be used as an embedded widget within another application.
+
+            ```
+            dependencies {
+              implementation 'io.element.android:element-call-embedded:${{ env.ARTIFACT_VERSION}}'
```
jmartinesp: Missing whitespace here, I guess it doesn't affect the variable inlining?

embedded/ios/LICENSE-AGPL-3.0
jmartinesp, line 1: Does the Android subproject need a couple of license files too? 🤔

jmartinesp (approved):
The Android part LGTM, just a couple of notes.

embedded/ios/Package.swift
```diff
@@ -0,0 +1,18 @@
+// swift-tools-version: 6.0
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "EmbeddedElementCall",
```
Velin92, line 7: I would add a platform and version specification below here
` platforms: [.iOS(.v17_6)],`

This sets the minimum iOS version that should build, and I think it should match EL-X which currently supports 17.6 as the minimum supported iOS version.
MacOS min versions can be inferred by the iOS version. But if you want to add it to match EL-X, the minimum is 14.6

.github/workflows/publish-embedded-packages.yaml
hughns:
```suggestion
              implementation 'io.element.android:element-call-embedded:${{ env.ARTIFACT_VERSION }}'
```

embedded/ios/LICENSE-AGPL-3.0
hughns: Unlike the npm and SwiftPM package, the ARR manifest (pom.xml) listed the licenses with remote links to GitHub: https://github.com/element-hq/element-call/pull/3086/files#diff-d8e693c868746be17186dfa1ffe8ab1436c2b101c94e79604f8ae5a2f6f9134cR41-R52

Do you think the licenses should be included in the AAR directly? If so I will add the license files and then change the pom.xml to reference the license files in the AAR.

embedded/ios/Package.swift
hughns:
```suggestion
    name: "EmbeddedElementCall",
    platforms: [.iOS(.v17_6)],
```

.github/workflows/publish-embedded-packages.yaml
hughns:
```suggestion
        uses: softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # v2
```

.github/workflows/publish-embedded-packages.yaml
hughns:
```suggestion
        uses: softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # v2
```

.github/workflows/publish-embedded-packages.yaml
hughns:
```suggestion
        uses: softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # v2
```

.github/workflows/publish-embedded-packages.yaml
hughns:
```suggestion
        uses: softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # v2
```

.github/workflows/publish.yaml
hughns:
```suggestion
        uses: softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # v2
```

.github/workflows/publish.yaml
hughns:
```suggestion
        uses: softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # v2
```

embedded/ios/Package.swift
Velin92: Actually just using `iOS v17` should be sufficient, macOS isn inferred, and is easier to test on older sim versions since 17.6 is not provided as a valid sim by Apple

embedded/ios/Package.swift
```diff
@@ -0,0 +1,19 @@
...
+    platforms: [.iOS(.v17_6)],
```
hughns, line 8:
```suggestion
    platforms: [.iOS(.v17)],
```

hughns: @Velin92 like this?

embedded/ios/LICENSE-AGPL-3.0
hughns: I checked the other packages on maven central published for the io.element.android group and I see this approach of putting the URLs in pom.xml is standard.

As such, I think it is fine that the license files are not within the embeded/android subfolder.

.github/workflows/publish-embedded-packages.yaml
hughns: Thanks, I've added this in. albeit by bypassing the shell script and making that just for dev usage.

PR #3091, opened 2025-03-13
First integration tests with playwright

BillCarsonFr:
Adds the first playwright integration tests.

~~**Includes a failing test reproducing the regression discovered in 0.8.0-rc2**~~
**Fixed since https://github.com/element-hq/element-call/commit/46e7e8e9cd3bcd690b1105eaf4026f5fcbbd46a9**

image

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -18,7 +18,7 @@ jobs:
         run: yarn playwright install --with-deps
       - name: Run backend components
         run: |
-          docker-compose -f dev-backend-docker-compose.yml up -d
+          docker compose -f dev-backend-docker-compose.yml up -d
           docker ps
       - name: Run Playwright tests
         run: yarn playwright test
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -11,9 +11,10 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: lts/*
+          cache: "yarn"
+          node-version-file: ".node-version"
       - name: Install dependencies
-        run: npm install -g yarn && yarn
+        run: yarn install --frozen-lockfile
       - name: Install Playwright Browsers
         run: yarn playwright install --with-deps
       - name: Run backend components
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -5,7 +5,7 @@ on:
     branches: [livekit, full-mesh]
 jobs:
   test:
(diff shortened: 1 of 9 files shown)


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,32 @@
...
+    branches: [livekit, full-mesh]
+jobs:
+  test:
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
```
hughns: To match the other workflows:

```suggestion
          cache: "yarn"
          node-version-file: ".node-version"
```

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Install dependencies
+        run: npm install -g yarn && yarn
```
hughns: Yarn is alread installed, so we can just do:

```suggestion
        run: yarn install --frozen-lockfile
```

README.md
```diff
@@ -189,6 +189,64 @@ yarn backend
...
+```
+
+The easiest way to develop new test is to use the codegen feature of Playwright:
+
+```shell
+npx playwright codegen
+```
+
+This will record your action and write the test code for you. Use the tool bar to test visibility, text content,
+clicking..
```
hughns:
```suggestion
clicking.
```

playwright.config.ts
```diff
@@ -0,0 +1,77 @@
+import { defineConfig, devices } from "@playwright/test";
```
hughns: I think this needs the copyright header.

playwright.config.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// import dotenv from 'dotenv';
+// import path from 'path';
+// dotenv.config({ path: path.resolve(__dirname, '.env') });
```
hughns:
```suggestion
```

?

hughns (changes requested):

playwright.config.ts
BillCarsonFr: right, let me see why it was not caught up by CI

playwright.config.ts
BillCarsonFr: it's boilerplate added by the default playwrigth setup file. Removed

hughns (approved):
One question: does it matter that tsconfig.json doesn't include `playwright/**/*.ts`?

PR #3092, opened 2025-03-13
Fix: js-sdk version that uses expected value for `isJoined` (`isJoined` will now be false before we left as expected by EC)

toger5:


toger5:
Oh okay so our ci is able to transpile the changes. its just local dev which is not working (as robin has said before iirc)

fkwp (approved):

PR #3094, opened 2025-03-13
Backport: Fix: js-sdk version that uses expected value for isJoined (isJoined will now be false before we left as expected by EC)

toger5:


fkwp (approved):

toger5:
Ugh it seems it cannot deploy...

toger5:
Is this normal for non livekit branch PR's?

PR #3093, opened 2025-03-13
Transition from consuming js-sdk source to `lib` (build output)

toger5:
Closes https://github.com/element-hq/element-call/issues/3081

As of https://github.com/element-hq/element-call/pull/3095 being merged, this allows us to switch between linked and unlinked copies of matrix-js-sdk while consuming its code in a uniform way (via the recommended `matrix-js-sdk/lib` build output).

Issue #3081, opened 2025-03-11
Cannot develop with linked matrix-js-sdk


robintown:
### Steps to reproduce

1. Pull the latest version of `livekit`
2. `yarn install`
3. `yarn link matrix-js-sdk`
4. `yarn dev`
5. Open Element Call in a browser

#### What did you expect?

Element Call starts

#### What happened instead?

The page crashes on a white screen because there is a decorator used in matrix-js-sdk/src/http-api/fetch.ts, which Babel does not transpile. I thought https://stackoverflow.com/a/69665198 might be the solution here but it still doesn't transpile the decorator. (why?)

### Operating system

NixOS 24.11

### Browser information

Firefox 135.0.1

### Will you send logs?

No

robintown:
This is blocking my work on https://github.com/element-hq/element-web/issues/29429

toger5:
Syncing with the web team gave the following result:
https://docs.google.com/document/d/1Qr2ocRQFSx4Vp9qPUu4rxP5f8j1JLrdxcPVZHvjynZk/edit?tab=t.0#heading=h.m8zbp5oc221e

We seek for confirmation if it makes sense to move to matrix-js-sdk/lib imports (currently we import src)
 - src imports were never supported. The react sdk used it since then but that also implied headaches (EW still uses it as of now. Its planned to change (use lib imports) eventually)
 - Watch mode on the js-sdk to auto-run tsc
 - Mixing lib+scr will make us end up with two versions.
 - A specific git commit hash in package.json is not supported with lib imports anymore.
Any other recommendations/tips?
 - Vite uses rollup under the hood and ts is able to build transpile decorators natively nowadays -> maybe we can get Vite develop mode working by adjusting the ts config.


I tested what it would take to move over to lib imports.
This is the changes for element-call: https://github.com/element-hq/element-call/pull/3093

(I previously needed a bunch of changes in the js-sdk as well. There seemed to be workarounds however)


I gave this a bit more touching up. I think it might make sense to make a draft PR out of the branch.
@robintown do you think you have time to mention this next week and eventually move it forward?

PR #3095, opened 2025-03-14
Upgrade to Yarn v4 (Berry)

robintown:
https://github.com/element-hq/element-call/pull/3093 presents a dilemma: We want to consume matrix-js-sdk via its build artifacts, but we also want to continue using development versions of matrix-js-sdk (pinned to Git commits) rather than releases. Development versions, of course, don't come with build artifacts. This threatens to complicate the build process for developers, packagers, CI, everyone. (Now you would always need to clone & link matrix-js-sdk, cross-reference our package.json, check out the right commit, and build it separately.)

The good thing is, modern package managers have figured out a far more elegant way to support this use case: if a Git dependency contains a `prepack` script, they will run that script before installing the package into `node_modules`, making this the perfect way to automatically perform a build step. In fact, matrix-js-sdk [already has](https://github.com/matrix-org/matrix-js-sdk/blob/8395919f0fd1af7cab1e793d736f2cdf18ef7686/package.json#L9) this exact script! We just can't benefit from it as long as we're stuck using Yarn Classic, which hasn't been actively developed for years.

I suggest that now would be as good a time as any to upgrade from Yarn Classic to Yarn v4 (Berry). The Yarn developers put a lot of thought into making the `yarn` command automatically invoke the right version of the tool depending on the repository you're in, so switching between Yarn Classic and Yarn Berry repositories is a pretty seamless experience. The only obstacle to upgrading that I encountered is that Yarn Berry [regresses the developer experience of working with linked packages](https://github.com/yarnpkg/berry/issues/90#issuecomment-547411228). However, I believe I've remedied this with a plugin I threw together.

Any reservations on making such an upgrade?

For https://github.com/element-hq/element-call/issues/3081
Depends on https://github.com/element-hq/element-call-webapp-deployments/pull/3


toger5:
> Any reservations on making such an upgrade?

This sounds really nice. Since this is basically a full package manager change I feel like it would be appropriate to do a bit of research and collect pro/cons to other alternatives and not make the decision an artifact of us not being able to compile with vite dev.

docs/linking.md
```diff
@@ -0,0 +1,12 @@
+# Developing with linked packages
+
+If you want to make changes to a package that Element Call depends on and see those changes applied in real time, you can create a link to a local copy of the package. Yarn has a command for this (`yarn link`), but it's not recommended to use it as it ends up modifying package.json with details specific to your development environment.
+
+Instead, you can use our little 'linker' plugin. Create a file named `.links.yaml` in the Element Call project directory, listing the names and paths of any dependencies you want to link. For example:
+
+```yaml
+matrix-js-sdk: ../path/to/matrix-js-sdk
+"@vector-im/compound-web": /home/alice/path/to/compound-web
```
toger5, line 9: I am loving this.

Do I understand it correctly, that it just allows us to put this part of the package.json into its own file, so that we can git-ignore it?
pnpm has the same behaviour and I was wondering how this could be elegantly solved.
This is great!

src/analytics/PosthogAnalytics.ts
```diff
@@ -297,7 +296,7 @@ export class PosthogAnalytics {
     const posthogIdMaterial = "ec" + accountAnalyticsId + client.getUserId();
     const bufferForPosthogId = await crypto.subtle.digest(
       "sha-256",
-      Buffer.from(posthogIdMaterial, "utf-8"),
+      new TextEncoder().encode(posthogIdMaterial),
```
toger5, line 298: is this needed for the yarn berry switch or just randomly here?

hughns (commented):
n.b. This would need a corresponding PR for https://github.com/element-hq/element-call-webapp-deployments/blob/main/.github/workflows/deploy.yml to enable corepack

docs/linking.md
robintown: Yes, though I just realized that this method will still modify yarn.lock, sigh. I don't know of a better solution right now.

src/analytics/PosthogAnalytics.ts
robintown: I was trying out Yarn's plug-n-play mode, and it needed this change. I reverted to the normal node_modules linker so I don't think it's strictly necessary. Would you prefer I exclude it?

robintown:
This depends on https://github.com/element-hq/element-call-webapp-deployments/pull/3 then, so marking as blocked.

robintown:
@toger5 
> This sounds really nice. Since this is basically a full package manager change I feel like it would be appropriate to do a bit of research and collect pro/cons to other alternatives and not make the decision an artifact of us not being able to compile with vite dev.

I had a look at whether switching to npm or pnpm would be possible. At first, I was under the impression that either would work for us, but it turns out that their handling of Git dependencies is subtly broken.

- npm just [does not execute `prepack` scripts](https://github.com/npm/cli/issues/5641#issuecomment-1751681172) for Git dependencies, even though its docs claim that it does. It will not build an unlinked matrix-js-sdk from source.
- pnpm circumvents Git and downloads dependencies directly from GitHub's CDN, meaning that the cache entry for the package lacks a .git directory. matrix-js-sdk assumes .git will exist in order to include a version identifier in the build, which given that we're requesting a Git dependency, is a very reasonable assumption to make. The matrix-js-sdk build fails.

So Yarn appears to be the only package manager out there that handles Git dependencies in a reasonable way, as of now. I think this is a good example of the Yarn project's commitment to stability & correctness around edge cases, which is a reason I would've advocated for it even if npm/pnpm were equally viable. It's a package manager's best feature :)

Bun is another alternative, but I haven't tried it out because switching to an entirely different JS runtime seems like a far deeper rabbit hole.

toger5:
@robintown Thank you for going the extra mile of looking into this and providing us with a specific reason/reasons why we want to go that path.

docs/linking.md
toger5: Can a plugin also create the a `.links.lock` file and add compose the actual lock file based on `yarn.lock`+`.links.lock`?

src/analytics/PosthogAnalytics.ts
toger5: No I am fine keeping it like this. Just wanted to know its not accidental.

toger5: Can we (did you already?) make sure we use `TextEncoder` everywhere?

.yarn/plugins/linker.cjs
```diff
@@ -0,0 +1,87 @@
...
+            `portal:${path.resolve(config.projectCwd, link)}`,
+          ]),
+        );
+        const overrideIdentHashes = new Set();
+        for (const name of Object.keys(overrides))
+          overrideIdentHashes.add(
+            structUtils.parseDescriptor(`${name}@*`, true).identHash,
+          );
+
+        registerPackageExtension(selfDescriptor, { dependencies: overrides });
```
toger5, line 63: Can this line get a comment or a link to the `registerPackageExtension` docs?

docs/linking.md
toger5: This is resolved with the git hooks.

toger5 (approved):
Really looking forward to work with this change!
I only have doc comments (which would be nice to have but not required) and the things I already committed:
ff6365cfed6eddadec48cbda272965c5598d0c47...6a5b79143e0b09598b415b01b88f4e9eea9c92fd

PR #3096, opened 2025-03-14
Integration Test: Add test for register, login, logout and guest create and join call

BillCarsonFr:
Based on https://github.com/element-hq/element-call/pull/3091

Point of intereset: https://github.com/element-hq/element-call/commit/83198c061a53ffc7642e6a012a68df3aa9e44d0f
I have disabled rate limits for the env setup, but maybe we want a specific config for tests with no rate limits and keep some limits for the dev setup?

image

### DIFF
--- a/backend/dev_homeserver.yaml
+++ b/backend/dev_homeserver.yaml
@@ -41,10 +41,23 @@ max_event_delay_duration: 24h
 #   - burst_count: number of requests a client can send before being throttled.
 
 rc_message:
-  # This needs to match at least the heart-beat frequency plus a bit of headroom
-  # Currently the heart-beat is every 5 seconds which translates into a rate of 0.2s
-  per_second: 0.5
-  burst_count: 30
+  per_second: 10000
+  burst_count: 10000
+
+rc_login:
+  address:
+    per_second: 10000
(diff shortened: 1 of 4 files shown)


hughns (approved):
Brilliant 👍

backend/dev_homeserver.yaml
```diff
@@ -41,10 +41,23 @@ max_event_delay_duration: 24h
 #   - burst_count: number of requests a client can send before being throttled.
 
 rc_message:
-  # This needs to match at least the heart-beat frequency plus a bit of headroom
-  # Currently the heart-beat is every 5 seconds which translates into a rate of 0.2s
-  per_second: 0.5
-  burst_count: 30
+  per_second: 10000
```
fkwp, line 44: If we need to change rate limits... wondering if we would like to add a dedicated playwright synapse config?

BillCarsonFr: We discussed about that. Keep the default config for local dev and create a new automated test config with no ratelimit. The decision was to keep it simple. But I think I will just after look at it so we keep some rate limit on local dev to run in more real setup (particularly given we have already some rate limiting problems)

PR #3098, opened 2025-03-17
Keep rate limits for dev backend, disable only for playwright tests

BillCarsonFr:
Follow up on @fkwp comment https://github.com/element-hq/element-call/pull/3096#discussion_r1996342138

Using a specific config for integration tests without rate limiting.
Keep default config for dev env


fkwp (approved):
lgtm

Issue #3107, opened 2025-03-18
PR preview deployments for backport PRs don't work


hughns:
Since https://github.com/element-hq/element-call/pull/3077 was merged to `livekit` branch, PRs to the 0.7 and 0.8 release branches no longer have the Netlify PR preview working.

For example https://github.com/element-hq/element-call/pull/3106 fails with:

<img width="1367" alt="Image" src="https://github.com/user-attachments/assets/94fe088a-e145-4140-9d45-96d19a84cc7d" />

This is because the PR preview workflow runs off the `livekit` branch instead of the release branch.

PR #3108, opened 2025-03-18
Rename built artifact name on v0.8.0 release branch for compatibility with Netlify PR preview

hughns:
Fixes [#3107](https://github.com/element-hq/element-call/issues/3107) by updating the artifact name to be the same as on the `livekit` branch


hughns:
The action run of this PR [here](https://github.com/element-hq/element-call/actions/runs/13920831605) validates that this is now working.

BillCarsonFr (approved):

PR #3118, opened 2025-03-21
Fix join errors mistakenly showing a generic error screen

robintown:
We were previously using the `useGroupCallErrorBoundary` hook to surface errors that happened during joining, but because that part is outside the `GroupCallErrorBoundary` it just ended up sending them to the app-level error boundary where they got displayed with a more generic message.

Regression was due to https://github.com/element-hq/element-call/commit/c22412c04559a5c26c6c0e4c2441dae31278e85a


hughns:
I think this is a candidate to backport to 0.9.0 branch.

robintown:
@fkwp Do you consider this a release blocker for v0.9.0?

robintown:
There's my answer :+1:

BillCarsonFr (approved):
Thx!
Nice to have the refactoring as a first commit before the fix. 
Great to have a test for that now 👍

PR #3135, opened 2025-03-26
Rename EULA to SSLA

toger5:
The linked "End User License Agreement" is actually a "Software and Services License Agreement" so we should call it that.

fixes: https://github.com/element-hq/voip-internal/issues/320
 
- rename i18n fields to reference Software and Services License Agreement
- rename the config property `eula` to `ssla`

### DIFF
--- a/src/auth/RegisterPage.tsx
+++ b/src/auth/RegisterPage.tsx
@@ -216,8 +216,7 @@ export const RegisterPage: FC = () => {
                   apply.
                   <br />
                   By clicking "Register", you agree to our{" "}
-                  {/* if the deprecated eula field is setup we still show it. */}
-                  <ExternalLink href={Config.get().eula ?? Config.get().ssla}>
+                  <ExternalLink href={Config.get().ssla}>
                     Software and Services License Agreement (SSLA)
                   </ExternalLink>
                 </Trans>
--- a/src/auth/RegisterPage.tsx
+++ b/src/auth/RegisterPage.tsx
@@ -204,7 +204,7 @@ export const RegisterPage: FC = () => {
                 />
               </FieldRow>
               <Text size="sm">
-                <Trans i18nKey="recaptcha_caption">
+                <Trans i18nKey="recaptcha_ssla_caption">
                   This site is protected by ReCAPTCHA and the Google{" "}
                   <ExternalLink href="https://www.google.com/policies/privacy/">
                     Privacy Policy
@@ -216,8 +216,9 @@ export const RegisterPage: FC = () => {
                   apply.
                   <br />
                   By clicking "Register", you agree to our{" "}
-                  <ExternalLink href={Config.get().eula}>
-                    End User Licensing Agreement (EULA)
+                  {/* if the deprecated eula field is setup we still show it. */}
+                  <ExternalLink href={Config.get().eula ?? Config.get().ssla}>
+                    Software and Services License Agreement (SSLA)
                   </ExternalLink>
                 </Trans>
               </Text>
--- a/src/config/ConfigOptions.ts
+++ b/src/config/ConfigOptions.ts
@@ -78,12 +78,6 @@ export interface ConfigOptions {
 
   /**
    * A link to the software and services license agreement (SSLA)
-   * @deprecated renamed to ssla
-   */
-  eula?: string;
-
-  /**
-   * A link to the software and services license agreement (SSLA)
    */
   ssla?: string;
 
--- a/src/config/ConfigOptions.ts
+++ b/src/config/ConfigOptions.ts
@@ -158,7 +158,7 @@ export const DEFAULT_CONFIG: ResolvedConfigOptions = {
   features: {
     feature_use_device_session_member_events: true,
   },
-  ssla: " https://static.element.io/legal/element-software-and-services-license-agreement-uk-1.pdf",
+  ssla: "https://static.element.io/legal/element-software-and-services-license-agreement-uk-1.pdf",
   media_devices: {
     enable_audio: true,
     enable_video: true,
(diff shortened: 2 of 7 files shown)


toger5:
@yousefhashaishi 

How do we deal with EC setups that have a eula config field 
 - do we show it in the now renamed ssla links?
 - do we fallback to the default element ssla?
 
 it is now setup as following:
  - if no eula config is provided use the default ssla
  - if ssla is provided in the config use it
  - if eula is provided use the config eula instead of the ssla
  - if ssla and eula is provided the person setting it up did a mistake since eula is marked as deprecated but eula will be used.

yousefhashaishi:
sorry i'm not sure i fully understand some of this; 

if eula is provided use the config eula instead of the ssla
if ssla and eula is provided the person setting it up did a mistake since eula is marked as deprecated but eula will be used.

as far as possible references to EULA need to be replaced by SSLA - not clear on what is meant by "use eula instead of ssla in case it is provided in the config."

toger5:
@yousefhashaishi 
Thanks for the response.
Its good to know, that EULA is not needed anymore and we now only use ssla.

There is one situation I would like to get you approval.
An EC host that updates EC and does not change their config.json will not use the default SSLA from element (https://static.element.io/legal/element-software-and-services-license-agreement-uk-1.pdf) instead of their own configured EULA.

It is their responsibility to read the changelog and change their config to include a SSLA instead of EULA.

How do we do it in elements messanger apps (what are the options (for example for EW) to configure the EULA/SSLA via the config.json)

hughns:
> How do we do it in elements messanger apps (what are the options (for example for EW) to configure the EULA/SSLA via the config.json)

@toger5 do you mean with the new embedded Element Call feature that is being worked on?

If so, then the answer is that the messenger apps simply do not set the eula field in Element Call config. This is because the config value is only shown in the user/guest registration flows of Element Call which are not used when the embedded Element Call package is used.

src/config/ConfigOptions.ts
```diff
@@ -152,7 +158,7 @@ export const DEFAULT_CONFIG: ResolvedConfigOptions = {
   features: {
     feature_use_device_session_member_events: true,
   },
-  eula: "https://static.element.io/legal/online-EULA.pdf",
+  ssla: " https://static.element.io/legal/element-software-and-services-license-agreement-uk-1.pdf",
```
hughns:
```suggestion
  ssla: "https://static.element.io/legal/element-software-and-services-license-agreement-uk-1.pdf",
```

src/auth/RegisterPage.tsx
```diff
@@ -216,8 +216,9 @@ export const RegisterPage: FC = () => {
                   apply.
                   <br />
                   By clicking "Register", you agree to our{" "}
-                  <ExternalLink href={Config.get().eula}>
-                    End User Licensing Agreement (EULA)
+                  {/* if the deprecated eula field is setup we still show it. */}
+                  <ExternalLink href={Config.get().eula ?? Config.get().ssla}>
```
hughns:
```suggestion
                  <ExternalLink href={Config.get().ssla ?? Config.get().eula}>
```

src/home/UnauthenticatedView.tsx
```diff
@@ -185,10 +185,11 @@ export const UnauthenticatedView: FC = () => {
               </Text>
             )}
             <Text size="sm" className={styles.notice}>
-              <Trans i18nKey="unauthenticated_view_eula_caption">
+              <Trans i18nKey="unauthenticated_view_ssla_caption">
                 By clicking "Go", you agree to our{" "}
-                <ExternalLink href={Config.get().eula}>
-                  End User Licensing Agreement (EULA)
+                {/* if the deprecated eula field is setup we still show it. */}
+                <ExternalLink href={Config.get().eula ?? Config.get().ssla}>
```
hughns:
```suggestion
                <ExternalLink href={Config.get().ssla ?? Config.get().eula}>
```

src/room/RoomAuthView.tsx
```diff
@@ -80,10 +80,11 @@ export const RoomAuthView: FC = () => {
               />
             </FieldRow>
             <Text size="sm">
-              <Trans i18nKey="room_auth_view_eula_caption">
+              <Trans i18nKey="room_auth_view_ssla_caption">
                 By clicking "Join call now", you agree to our{" "}
-                <ExternalLink href={Config.get().eula}>
-                  End User Licensing Agreement (EULA)
+                {/* if the deprecated eula field is setup we still show it. */}
+                <ExternalLink href={Config.get().eula ?? Config.get().ssla}>
```
hughns:
```suggestion
                <ExternalLink href={Config.get().ssla ?? Config.get().eula}>
```

hughns (changes requested):
I think the `ssla` value should have priority over the deprecated `eula`, not the other way around.

toger5:
@hughns 
> I think the ssla value should have priority over the deprecated eula, not the other way around.

The thinking here is that since the ssla has a default value it is never undefined. either it uses the one from the config file or the fallback from (`DEFAULT_CONFIG`) the default.

What i tried to achieve is that a admin who updates without adjusting the config file will not end up with the element ssla by surprise.

toger5:
maybe we should just remove the eula config and expect admins to read the change log.
-> make this a breaking change.

It is confusing otherwise and putting a eula link behind a ssla ui button is not a good situation in the first place.

toger5:
I created a commit (I can easily revert) that breaks compatibility and removes the eula field entirely.

yousefhashaishi:
just looking around to see if anything changed - not sure if this screenshot is in the scope of changes being made

image

toger5:
This is how it looks for me. are you testing with: https://pr3135--element-call.netlify.app/
image
Screenshot From 2025-03-28 17-24-34

yousefhashaishi:
must have used a different link it looks good now, thanks Timo!!

hughns (approved):

PR #3136, opened 2025-03-27
Fix example JWT service .well-known endpoints in self-hosting.md

fkwp:


hughns (commented):
Please can you confirm that the intended net result of this is that as per https://github.com/hughns/matrix-spec-proposals/blob/hughns/matrixrtc-livekit/proposals/4195-matrixrtc-livekit.md#sfu-authentication-with-livekit the MatrixRTC client (i.e. Element Call) would be making the `POST` request to:

`https://matrix-rtc.example.com/livekit/jwt/sfu/get` and
`https://matrix-rtc-2.example.com/livekit/jwt/sfu/get` for these examples?

fkwp:
> Please can you confirm that the intended net result of this is that as per https://github.com/hughns/matrix-spec-proposals/blob/hughns/matrixrtc-livekit/proposals/4195-matrixrtc-livekit.md#sfu-authentication-with-livekit the MatrixRTC client (i.e. Element Call) would be making the `POST` request to:
> 
> `https://matrix-rtc.example.com/livekit/jwt/sfu/get` and `https://matrix-rtc-2.example.com/livekit/jwt/sfu/get` for these examples?

Yup that is how it behaves -- I did double-check with the provided nginx config and had a look in the nginx logs verifying the concatenation of .../livekit/jwt & /sfu/get

hughns (approved):
I originally hadn't realised that this was basically a bug in the current docs. I've updated the name to reflect this 👍

Issue #3143, opened 2025-03-28
Expose embedded package version at runtime in AAR and SwiftPM


hughns:
#### What would you like to do?

Make a version string available at runtime from a messenger app consuming the SwiftPM and AAR embedded packages.

#### Why would you like to do it?

For including in messenger app debug logs, and as an additional check to make sure the embedded library actually compiles.

#### How would you like to achieve it?

TBD

PR #3144, opened 2025-03-28
Expose version string in embedded SwiftPM and AAR

hughns:
Implements https://github.com/element-hq/element-call/issues/3143


hughns:
The AAR part of this isn't working and needs input from @jmartinesp to figure out.

BillCarsonFr (approved):
read it

PR #3164, opened 2025-04-03
Fix lint and playwright errors caused by matrix-js-sdk/src import

toger5:


BillCarsonFr (approved):
Fixes the problem for me!

toger5 (commented):
Remove element call config from ew. Its not used anymore now

PR #3165, opened 2025-04-03
Update readme

fkwp:


toger5 (changes requested):
This pulled in some not yet merged changes from the to-device branches.

Either we need to drop those commits or wait until the to-device things are merged.

README.md
```diff
@@ -45,30 +45,34 @@ organize the flow of the meeting.
...
-conferencing web application, allowing users to join or host calls without
+In Standalone mode, Element Call operates as an independent, full-featured video
+conferencing web application, enabling users to join or host calls without
 requiring a separate Matrix client.
 
-### Widget mode embedded in Messenger Apps
+### 📲 In-App Calling (Widget Mode in Messenger Apps)
+
+As a widget 🧩, Element Call utilizes only the core calling (MatrixRTC)
+components, while authentication, event handling, and room state updates are
```
toger5: > event handling, and room state updates

nitpicking: I think this makes more sense to just mention "matrix cs api requests, hs syncing" (or similar) otherwise this might be confused with managing the state events which is done in the widget logic.

README.md
```diff
@@ -80,6 +84,30 @@ and voice calls within Matrix rooms.
 > [!IMPORTANT]
 > Embedded packaging is recommended for Element Call in widget mode!
 
+### 📦 Element Call Packaging
+
+Element Call offers two packaging options: one for standalone or widget
+deployment, and another for widget-based seamless integration into messenger
```
toger5:
```suggestion
deployment, and another for seamless widget-based integration into messenger
```

toger5 (changes requested):
Can the images be included as svg's?

Just a general question not sure if this is even supported by the drawing tool you use.

README.md
fkwp: rephrased the paragraph

README.md
```diff
@@ -45,30 +45,35 @@ organize the flow of the meeting.
...
-In Standalone mode Element Call operates as an independent, full-featured video
-conferencing web application, allowing users to join or host calls without
+In Standalone mode, Element Call operates as an independent, full-featured video
+conferencing web application, enabling users to join or host calls without
 requiring a separate Matrix client.
 
-### Widget mode embedded in Messenger Apps
+### 📲 In-App Calling (Widget Mode in Messenger Apps)
+
+When used as a widget 🧩, Element Call relies solely on the core calling
```
toger5: I like this new version.
I do understand `relies` as this is the only thing EC relies on from the js-sdk.
I think its more direct like this:
```suggestion
When used as a widget 🧩, Element Call is solely responsible on the core calling
```

toger5 (approved):
Approving it since the last proposal is optional

PR #3167, opened 2025-04-04
Add option to enable to-device-encryption

toger5:
This only adds a developer option to use the to-device work. It hence is a non-breaking change.
Blocked on: https://github.com/matrix-org/matrix-js-sdk/pull/4785

### DIFF
--- a/src/utils/matrix.ts
+++ b/src/utils/matrix.ts
@@ -251,6 +251,7 @@ export async function createRoom(
         "m.room.encryption": 100,
         "m.room.name": 50,
         "m.room.message": 0,
+        // revert this once we do not rely on room messages for perSenderKeys anymore
         "m.room.encrypted": 0,
         "m.sticker": 50,
         "org.matrix.msc3401.call.member": 0,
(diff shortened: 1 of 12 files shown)


toger5 (commented):
Dont merge. this is only for testing purposes.

toger5:
I think this is ready to merge now. @BillCarsonFr can you review?

src/utils/matrix.ts
```diff
@@ -251,14 +251,24 @@ export async function createRoom(
...
     },
+    initial_state: [
+      {
+        type: "m.room.encryption",
+        state_key: "",
+        content: {
+          algorithm: "m.megolm.v1.aes-sha2",
+        },
+      },
+    ],
```
toger5: This is the one critical change in this PR.

toger5: It will change how rooms in the spa are configured on creation.

toger5: I moved this out into a seperate PR to make this one less controversial: https://github.com/element-hq/element-call/pull/3184

BillCarsonFr (approved):
LGTM

PR #3171, opened 2025-04-04
Add PR- label requirement action

toger5:


.github/workflows/changelog-label.yml
```diff
@@ -0,0 +1,14 @@
+name: PR changelog label
+
+on:
+  pull_request:
+    types: [labeled, unlabeled, opened, edited, synchronize]
```
robintown: This should use `pull_request_target` as the event type so that it runs even when the branch has conflicts. Also I think running on `edited` and `synchronize` isn't strictly necessary for a label check.

.github/workflows/changelog-label.yml
```diff
@@ -0,0 +1,14 @@
...
+jobs:
+  pr-changelog-label:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: yogevbd/enforce-label-action@2.1.0
+        with:
+          REQUIRED_LABELS_ANY: "PR-Bug-Fix,PR-Documentation,PR-Task,PR-Feature,PR-Enhancement,PR-Develeoper-Experience,dependencies"
```
robintown:
```suggestion
          REQUIRED_LABELS_ANY: "PR-Bug-Fix,PR-Documentation,PR-Task,PR-Feature,PR-Improvement,PR-Developer-Experience,dependencies"
```

fkwp (approved):

PR #3173, opened 2025-04-07
Add yarn and bundled element call renovate group.

toger5:
See: https://github.com/element-hq/element-call/pull/3154#issuecomment-2774090298

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -48,6 +48,20 @@
       "groupName": "Vaul",
       "matchDepNames": ["vaul"],
       "prHeader": "Please review modals on mobile for visual regressions."
+    },
+    {
+      "groupName": "bundled element-call dependencies",
+      "matchPackageNames": [
+        "com.vanniktech.maven.publish",
+        "com.android.library",
+        "com.android.tools.build",
+        "gradle"
+      ]
+    },
+    {
+      "groupName": "Yarn",
+      "matchPackageNames": ["yarn"],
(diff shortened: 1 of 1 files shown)


robintown (approved):

renovate.json
```diff
@@ -48,6 +48,20 @@
       "groupName": "Vaul",
       "matchDepNames": ["vaul"],
       "prHeader": "Please review modals on mobile for visual regressions."
+    },
+    {
+      "groupName": "bundled element-call dependencies",
```
robintown, line 53: Perhaps we could stick to the term "embedded" here?

toger5: I dont like embedded. It is what we use for `embedded.js` and spa vs widget in general so I think it might lead to confusion to overload the word?

But maybe this is a discussion for a meeting.
wdyt?

toger5: I added the topic to the meeting notes.

toger5: I will create a follow up pr in case we want it to be called embedded. (otherwise i can create a follow up pr to rename it bundled consistently in the EC repo)

hughns:
Rather than matching on individual package names which could change, an alternative implementation I would suggest matching based on package descriptor filename.

Something like this:

```json
{
  "groupName": "embedded package dependencies",
  "matchFileNames": ["embedded/**/*"]
}
```

PR #3177, opened 2025-04-08
Improve the reliability of state changes in widget mode

robintown:
This makes the mechanism by which an Element Call widget can receive changes to room state more reliable by updating matrix-js-sdk and matrix-widget-api, which now support the `update_state` action from [the latest version of MSC2762](https://github.com/matrix-org/matrix-spec-proposals/pull/4237).

Depends on https://github.com/matrix-org/matrix-js-sdk/pull/4790
For https://github.com/element-hq/voip-internal/issues/289

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "livekit-client": "^2.11.3",
     "lodash-es": "^4.17.21",
     "loglevel": "^1.9.1",
-    "matrix-js-sdk": "github:matrix-org/matrix-js-sdk#f3f4a1f7e2f5f2c26cf4fce89a729a9197f619b5",
+    "matrix-js-sdk": "github:matrix-org/matrix-js-sdk#7fa9195e398764749b0195f14d3cf0190ad10253",
     "matrix-widget-api": "1.11.0",
     "normalize.css": "^8.0.1",
     "observable-hooks": "^4.2.3",
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "livekit-client": "^2.11.3",
     "lodash-es": "^4.17.21",
     "loglevel": "^1.9.1",
-    "matrix-js-sdk": "github:matrix-org/matrix-js-sdk#7fa9195e398764749b0195f14d3cf0190ad10253",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -102,8 +102,8 @@
     "livekit-client": "^2.11.3",
     "lodash-es": "^4.17.21",
     "loglevel": "^1.9.1",
-    "matrix-js-sdk": "github:matrix-org/matrix-js-sdk#f3f4a1f7e2f5f2c26cf4fce89a729a9197f619b5",
-    "matrix-widget-api": "1.11.0",
+    "matrix-js-sdk": "github:matrix-org/matrix-js-sdk#7fa9195e398764749b0195f14d3cf0190ad10253",
```
toger5: In this particular instance it might make sense to mention where exactly the js-sdk commit comes from.

I would almost argue we should make this a procedure rule:
 - If one uses a commit NOT from the develop (main) branch, a link to the related branch with a short summary has to be added as a comment.
 ```suggestion
    "NONMAIN-matrix-js-sdk":"This commit is from: https://github.com/matrix-org/matrix-js-sdk/pull/4847, the PR combines state-after and widget driver to-device fallback featurres.",
    "matrix-js-sdk": "github:matrix-org/matrix-js-sdk#7fa9195e398764749b0195f14d3cf0190ad10253",
```

toger5 (approved):
Looks like a simple patch PR should look like.
I am wondering if we can improve on the "custom js-sdk hash" situation?

package.json
toger5: I forgot its not that simple. NONMAIN-... will raise an error as the npm package cannot be found.

toger5: So we might need to find a good place somewhere else. package.json.md?

robintown: Different, possibly simpler idea:

Yarn has the ability to reference a Git dependency by branch name rather than commit. We can just use that! Then it's not too difficult to observe that we're using a non-`develop` branch and go over to https://github.com/matrix-org/matrix-js-sdk to dig into the details of the branch if you really need them.

e4924df8fba113b0cf1f047ceffe65ea5b30c2f5

I think Yarn has always had this feature, but in Yarn Classic it caused the package cache to be a bit broken and that's why we always avoided it. But now we're of course on Yarn Berry.

PR #3182, opened 2025-04-10
[TEST] to device key transport enabled by default

toger5:


toger5:
This is now on call-next.element.io no need for this testin pr anymore

PR #3184, opened 2025-04-10
Update default power levels for room and encrypt rooms on creation

toger5:
This change is pulled out from: https://github.com/element-hq/element-call/pull/3167
So we can decide if we really want it. Its mostly independent except that if we want room encryption event fallbacks this is needed.
For spa guest links we use rooms defined by EW so they should be encrypted already. So this is only for todevice (or room key) transport in the spa with rooms created in the spa.

### DIFF
--- a/src/utils/matrix.ts
+++ b/src/utils/matrix.ts
@@ -251,7 +251,8 @@ export async function createRoom(
         "m.room.encryption": 100,
         "m.room.name": 50,
         "m.room.message": 0,
-        "m.room.encrypted": 50,
+        // revert this once we do not rely on room messages for perSenderKeys anymore
+        "m.room.encrypted": 0,
         "m.sticker": 50,
         "org.matrix.msc3401.call.member": 0,
       },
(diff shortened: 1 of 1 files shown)


BillCarsonFr (commented):
What is the rationale for this custom power level override ?
Like only allow what is strictly needed out of caution?

BillCarsonFr (approved):

toger5:
After a large call test we noticed high server load. It has to be concluded if this is caused by the room encryption being enabled or the toDevice key distribution.

toger5:
Maybe we should only merge this once we changed it so reactions will only be send unencrypted. Otherwise we will rapidly create a large megolm session in large calls where ppl use reactions.

toger5:
We do not want to merge this since it will impact all spa rooms (and make them heavier: full megolm session needed for each room to send reactions)

Instead the following is planned:
https://github.com/element-hq/element-call/issues/3264

PR #3185, opened 2025-04-11
Add `RoomAndToDeviceKeyTransport`

toger5:
**Warning**: This enables the to-device transport developer flag by default with the reasoning, that it will autmatically fallback to room events and hence is backwards compatible. 

Needs thorough testing.

This replaces the current dev option of to-device key sharing with a different option that starts with toDevice but will fallback to room events once another member sends a room event.


BillCarsonFr (approved):
ok

PR #3187, opened 2025-04-11
Update renovate.json

toger5:


renovate.json
```diff
@@ -60,7 +60,7 @@
     },
     {
       "groupName": "Yarn",
-      "matchPackageNames": ["yarn"],
+      "matchDepNames": ["yarn"],
       "extends": ["schedule:weekly"]
```
hughns: Quick thought: surely we only need to update yarn on the default (monthly) schedule rather than anything faster?

hughns (approved):
I don't know if the `yarn` part will work, but I expect that the embedded package deps are caught.

fkwp (approved):

PR #3190, opened 2025-04-11
Fix Docker image tag in release notes

AndrewFerr:
The version tag of published Docker images includes a leading "v", so include it in release notes.

---

Example:
```
$ docker pull ghcr.io/element-hq/element-call:0.9.0
Error response from daemon: failed to resolve reference "ghcr.io/element-hq/element-call:0.9.0": ghcr.io/element-hq/element-call:0.9.0: not found
$ docker pull ghcr.io/element-hq/element-call:v0.9.0
v0.9.0: Pulling from element-hq/element-call
...
Digest: sha256:f13a30cb527a788b9058fd568073b380fe42ff417b58955c9cd8c4a4a8a65849
Status: Downloaded newer image for ghcr.io/element-hq/element-call:v0.9.0
ghcr.io/element-hq/element-call:v0.9.0
```

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -66,8 +66,6 @@ jobs:
     name: Add docker release note
     runs-on: ubuntu-latest
     steps:
-      - name: Get artifact version
-        run: echo "ARTIFACT_VERSION=${VERSION:1}" >> "$GITHUB_ENV"
       - name: Add release note
         uses: softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # v2
         with:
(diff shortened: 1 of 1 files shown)


toger5 (approved):

hughns:
I don't think this is a complete fix.

For some reason we have ended up with some docker tags with a `v` prefix and others without. e.g. https://github.com/element-hq/element-call/pkgs/container/element-call/383943369?tag=v0.9.0 vs https://github.com/element-hq/element-call/pkgs/container/element-call/382187098?tag=0.9.0-rc.4

It looks like the RCs are missing the version number. This should be fixed.

I suggest doing:

- [ ] Fix the inconsistency between Release Candidates and Releases in CI: https://github.com/element-hq/element-call/issues/3192
- [x] Manually fix the existing release notes to reference the correct tag (e.g. https://github.com/element-hq/element-call/releases/tag/v0.9.0): https://github.com/element-hq/element-call/issues/3191

PR #3196, opened 2025-04-11
Explicitly set docker image version for releases

hughns:
Otherwise we end up with inconsistency across RCs and Releases.


toger5 (approved):

hughns:
I finally tested this as working in https://github.com/element-hq/element-call/actions/runs/14407661230/job/40408054915

PR #3197, opened 2025-04-11
Set the correct license on docker images

hughns:
Previously the docker/metadata action was autodetecting the license and only getting AGPL not dual license.

### DIFF
--- a/.github/workflows/build-and-publish-docker.yaml
+++ b/.github/workflows/build-and-publish-docker.yaml
@@ -46,6 +46,8 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: ${{ inputs.docker_tags}}
+          labels: |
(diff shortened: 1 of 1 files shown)


hughns:
Currently I have no way to test this PR in our CI.

This is because the `pr-deploy.yaml` workflow (which in turn calls the `build-and-publish-docker.yaml` workflow) is triggered by the completion of the `build.yaml` workflow. This means that `pr-deploy.yaml` is always run from the default (i.e. `livekit`) branch and in turn the `build-and-publish-docker.yaml` is run from `livekit`.

This is sub-optimal.

To fix could:

- combine the pr-deploy.yaml steps into the build.yaml directly
- call pr-deploy.yaml from build.yaml explicitly

In the meantime I'm going to just merge this PR and test on `livekit` branch.

fkwp (approved):
lgtm

hughns:
This is working in so far as the expected licenses label is appearing in https://github.com/element-hq/element-call/actions/runs/14408255425/job/40410011160

PR #3202, opened 2025-04-14
Update self-hosting.md

ArkadSt:
Added msc4140_enabled: to homeserver config as well as /sfu/get|/healthz endpoint to nginx config. - this is the setup that worked for me. Also removed json tags from code blocks as they are not jsons.

### DIFF
--- a/docs/self-hosting.md
+++ b/docs/self-hosting.md
@@ -42,6 +42,8 @@ homeserver, you'll need to additionally add the following config items to
 experimental_features:
   # MSC3266: Room summary API. Used for knocking over federation
   msc3266_enabled: true
+  # MSC4140: Delayed events are required for proper call participation signalling. If disabled it is very likely that you end up with stuck calls in Matrix rooms
+  msc4140_enabled: true
   # MSC4222 needed for syncv2 state_after. This allow clients to
   # correctly track the state of the room.
   msc4222_enabled: true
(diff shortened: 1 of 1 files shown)


docs/self-hosting.md
```diff
@@ -88,9 +90,19 @@ the example above, this results in:
 
 Using Nginx, you can achieve this by:
 
-```jsonc
+```
 server {
     ...
+
+    location ~ ^(/sfu/get|/healthz) {
```
fkwp, line 97: this location is already handled by `location ^~ /livekit/jwt/`

you just need to adapt your .well-known/matrix/client org.matrix.msc4143.rtc_foci key accordingly 

where `livekit_service_url` does need the suffix `/livekit/jwt`

`"livekit_service_url": "https://matrix-rtc.example.com/livekit/jwt"`

docs/self-hosting.md
```diff
@@ -42,6 +42,8 @@ homeserver, you'll need to additionally add the following config items to
 experimental_features:
   # MSC3266: Room summary API. Used for knocking over federation
   msc3266_enabled: true
+  # MSC4140: Delayed events are required for proper call participation signalling. If disabled it is very likely that you end up with stuck calls in Matrix rooms
```
fkwp, line 45: this one is not required as delayed events are automatically enabled if you set

`max_event_delay_duration: 24h`

docs/self-hosting.md
ArkadSt: Oh, you are right. It was a mistake then, sorry for that. And thank you for pointing it out!

PR #3203, opened 2025-04-14
Improve tagging of Release versions of platform specific embedded packages

### LINKED ISSUE #3199: Embedded npm package should only have latest tag set for Releases not RCs
### Steps to reproduce

1. Go to https://www.npmjs.com/package/@element-hq/element-call-embedded?activeTab=versions
2. Look at the list of available versions

#### What did you expect?

The `latest` version to be `v0.9.0`.

#### What happened instead?

The `latest` version is `v0.10.0-hugh.2` (or previously it was `v0.10.0-rc.1` which is also incorrect).

The could be a `next` tag which applies to RCs but would not apply to test releases.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

### LINKED ISSUE #3200: Pre-releases of embedded AAR package should be published as SNAPSHOTs
### Steps to reproduce

1. Go to https://central.sonatype.com/artifact/io.element.android/element-call-embedded/versions
2. Look at the list of versions

#### What did you expect?

The pre-release versions to be [SNAPSHOTs](https://maven.apache.org/guides/getting-started/#What_is_a_SNAPSHOT_version.3F).

#### What happened instead?

All versions look like final releases.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

hughns:
Fixes #3200 and #3199. Best reviewed commit by commit.

This should make it easier to consume the actual Releases of the NPM and AAR packages.

Changes are:

- **Web NPM**: Set [dist tags](https://docs.npmjs.com/cli/v11/commands/npm-dist-tag) of:
  - `latest` for Releases
  - `other` for anything else

- **Android AAR**: Mark anything other than Releases as SNAPSHOT:
  - `0.9.0` for Releases
  - `0.9.0-rc.1-SNAPSHOT` for Release Candidates
  - `0.9.0-asdsad.1-SNAPSHOT` for anything else 

This should be backported to be v0.10.0 release branch.

### DIFF
--- a/.github/workflows/publish-embedded-packages.yaml
+++ b/.github/workflows/publish-embedded-packages.yaml
@@ -39,14 +39,11 @@ jobs:
         run: echo "UNPREFIXED_VERSION=${VERSION:1}" >> "$GITHUB_OUTPUT"
       - id: tag
         # latest = a proper release
-        # next = a release-candidate
         # other = anything else
         name: Set tag
         run: |
           if [[ "${VERSION}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
             echo "TAG=latest" >> "$GITHUB_OUTPUT"
-          elif [[ "${VERSION}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+-rc\.[0-9]+$ ]]; then
-            echo "TAG=next" >> "$GITHUB_OUTPUT"
           else
             echo "TAG=other" >> "$GITHUB_OUTPUT"
           fi
--- a/.github/workflows/publish-embedded-packages.yaml
+++ b/.github/workflows/publish-embedded-packages.yaml
@@ -19,6 +19,7 @@ jobs:
       DRY_RUN: ${{ steps.dry_run.outputs.DRY_RUN }}
       PREFIXED_VERSION: ${{ steps.prefixed_version.outputs.PREFIXED_VERSION }}
       UNPREFIXED_VERSION: ${{ steps.unprefixed_version.outputs.UNPREFIXED_VERSION }}
+      TAG: ${{ steps.tag.outputs.TAG }}
     steps:
       - name: Calculate VERSION
         # We should only use the hard coded test value for a dry run
@@ -36,6 +37,19 @@ jobs:
         name: Set UNPREFIXED_VERSION
         # This just strips the leading character
         run: echo "UNPREFIXED_VERSION=${VERSION:1}" >> "$GITHUB_OUTPUT"
+      - id: tag
+        # latest = a proper release
+        # next = a release-candidate
+        # other = anything else
+        name: Set tag
+        run: |
+          if [[ "${VERSION}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "TAG=latest" >> "$GITHUB_OUTPUT"
+          elif [[ "${VERSION}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+-rc\.[0-9]+$ ]]; then
+            echo "TAG=next" >> "$GITHUB_OUTPUT"
+          else
+            echo "TAG=other" >> "$GITHUB_OUTPUT"
+          fi
 
   build_element_call:
     needs: versioning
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-embedded-packages.yaml
```diff
@@ -148,7 +162,13 @@ jobs:
           java-version: "17"
 
       - name: Get artifact version
-        run: echo "ARTIFACT_VERSION=${{ needs.versioning.outputs.UNPREFIXED_VERSION }}" >> "$GITHUB_ENV"
+        # Anything that is not a final release will be tagged as a snapshot
+        run: |
+          if [[ "${{ needs.versioning.outputs.TAG }}" == "latest" ]]; then
+            echo "ARTIFACT_VERSION=${{ needs.versioning.outputs.UNPREFIXED_VERSION }}" >> "$GITHUB_ENV"
+          else
+            echo "ARTIFACT_VERSION=${{ needs.versioning.outputs.UNPREFIXED_VERSION }}-SNAPSHOT" >> "$GITHUB_ENV"
```
jmartinesp, line 167: We could append any suffix to `-SNAPSHOT` (`-SNAPSHOT-foo`) if we want to have different versions to test instead of always overriding the latest one. If we're ok with just using a single `-SNAPSHOT` release this is fine.

jmartinesp (approved):
The Android side of this LGTM, thanks!

.github/workflows/publish-embedded-packages.yaml
```diff
@@ -32,15 +64,15 @@ jobs:
...
+    needs: [build_element_call, versioning]
     if: always()
     name: Publish tarball
     runs-on: ubuntu-latest
     permissions:
       contents: write # required to upload release asset
     steps:
       - name: Determine filename
-        run: echo "FILENAME_PREFIX=element-call-embedded-${VERSION:1}" >> "$GITHUB_ENV"
+        run: echo "FILENAME_PREFIX=element-call-embedded-${{ needs.versioning.outputs.PREFIXED_VERSION }}" >> "$GITHUB_ENV"
```
t3chguy: is it fine that this now the `v` prefix of the version?

.github/workflows/publish-embedded-packages.yaml
```diff
@@ -11,19 +11,51 @@ on:
...
+        # other = anything else
+        name: Set tag
+        run: |
+          if [[ "${VERSION}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "TAG=latest" >> "$GITHUB_OUTPUT"
+          elif [[ "${VERSION}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+-rc\.[0-9]+$ ]]; then
+            echo "TAG=next" >> "$GITHUB_OUTPUT"
+          else
+            echo "TAG=other" >> "$GITHUB_OUTPUT"
+          fi
```
t3chguy: I don't think this behaves quite right, or at least inconsistently with our other packages.

`next` may be pointing to something older than `latest`

whereas for e.g. in js-sdk `next` points to `latest` or to an RC newer than `latest`, never to something older than `latest` otherwise you end up with a confusing testing experience of having to manually figure out whether to install `<pkg>@next` or `<pkg>[@latest]`

.github/workflows/publish-embedded-packages.yaml
hughns: Well spotted, this wasn't intentional and I will revert.

.github/workflows/publish-embedded-packages.yaml
hughns: I was going to make this consistent (by adding an extra `npm dist-tag add` step when required), but I'm not sure that is quite "right" either...

A circumstance that we have encountered already is: 

- we are working on a new release and doing a series of RCs e.g. `10.0.0-rc.X`
- during the RC series we find an issue with the previous release that needs a hot-fix style minor release e.g. `9.0.1`

If we had a `next` tag then I think it should stay with `10.0.0-rc.X` RCs even though the `latest` would point to `9.0.1`.

As such, I'm going to move to having two tags: `latest` and `other`

The `other` only exists because I don't see a way to tell npm to not tag a published package.

t3chguy: Fair enough, I don't have a need for `next` but being inconsistent between projects within the company just makes it easier to make mistakes

hughns:
Blocked as the AAR snapshot publishing is failing.

hughns:
Unblocked as the AAR publishing is now working.

toger5 (approved):

PR #3204, opened 2025-04-14
Build Docker image on slim base

AndrewFerr:
This reduces the size of the runtime image by about 50%.

depends on https://github.com/element-hq/element-call/pull/3294


toger5:
Can we track what tests we did with the slim-base to make sure its not breaking anything.
 - [ ] deploy and run EC from the slim-base version

robintown (approved):

AndrewFerr:
> Can we track what tests we did with the slim-base to make sure its not breaking anything.

There weren't any existing tests that used Docker, so I added 073baca to run the Playwright tests against a freshly-built Docker container of EC.  This should ensure that not only does the Docker image build properly, but that running containers can successfully serve EC.

playwright/fixtures/widget-user.ts
```diff
@@ -63,16 +63,27 @@ const CONFIG_JSON = {
...
-  });
-}
+const setDevToolElementCallDevUrl = process.env.USE_DOCKER
+  ? async (page: Page): Promise<void> => {
+      await page.evaluate(() => {
+        window.mxSettingsStore.setValue(
+          "Developer.elementCallUrl",
+          null,
+          "device",
+          "https://localhost:3000/room",
```
AndrewFerr: For some reason, this wasn't able to use a variable, so I've defined two different versions of the function for each value of the URL.

AndrewFerr:
Not sure why the end-to-end tests are hanging in CI, and they run properly for me in an Ubuntu VM...

toger5:
@AndrewFerr they are currently broken.
We need to get them fixed before merging (We did a couple of exceptions for the release relevant PR's)
We will merge it in your name once we have them fixed

AndrewFerr:
I noticed "playwright" in some recent commit messages, so here's a merge to try tests again :slightly_smiling_face: 

EDIT: Still no luck.

AndrewFerr:
Now we're getting somewhere.  Increasing the test job timeout revealed [the errors that the tests are hitting](https://github.com/element-hq/element-call/actions/runs/15470239945/job/43552420476?pr=3204#step:8:2142), which do not reproduce when running them in a local Ubuntu VM.

The tests are failing on expected page elements not being visible.  Any idea what could be causing that?

fkwp (changes requested):
Nice catch https://github.com/element-hq/element-call/pull/3204/commits/5a849360131b9efd764c405f2124dbcb77878171

I think to get it over the line removing the debugging logs to track down https://github.com/element-hq/element-call/pull/3204/commits/5a849360131b9efd764c405f2124dbcb77878171 would be nice.

scripts/playwright-webserver-command.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+if [ -n "$USE_DOCKER" ]; then
+    set -ex
+    yarn build
+    IMAGE_NAME=localhost:5000/element-call:testing
```
fkwp: What is the purpose of localhost:5000 here as `IMAGE_NAME` prefix?

AndrewFerr: This was to explicitly use the local image registry for the built image, just in case `docker run element-call:testing`  defaulted to pulling from docker.io.  It doesn't, so this can be removed.

AndrewFerr: Reverted in d8ec2e6

fkwp (approved):

PR #3208, opened 2025-04-14
Fix to-device encryption info label

### LINKED ISSUE #3201: The `using to Device key transport` label is shown when there is no to-device key sharing happening
The label at the top needs revisiting and more logic when it is shown. (now it just mirrors the setting and listens to the event that gets emitted when the fallback to room key happens)
The logic has to be:

- are we using perParticipantE2EE
- is the dev flag enabled
- are we still in to-device mode or already fallen back to room keys

toger5:
fixes: https://github.com/element-hq/element-call/issues/3201
The label was shown also without checking that we use PerParticipantE2EE. Which is a prerequisite for toDevice transport. As a result the label was shown when not desired.

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -233,17 +233,21 @@ export const InCallView: FC<InCallViewProps> = ({
     RoomAndToDeviceEvents.EnabledTransportsChanged,
     (enabled) => setDidFallbackToRoomKey(enabled.room),
   );
-  const [toDeviceEncryptionSetting] = useSetting(
+  const [useExperimentalToDeviceTransport] = useSetting(
     useExperimentalToDeviceTransportSetting,
   );
   const encryptionSystem = useRoomEncryptionSystem(rtcSession.room.roomId);
 
   const showToDeviceEncryption = useMemo(
     () =>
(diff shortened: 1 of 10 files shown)


src/room/InCallView.tsx
```diff
@@ -220,19 +223,21 @@ export const InCallView: FC<InCallViewProps> = ({
...
-  );
-  useEffect(() => {
-    setShowToDeviceEncryption(toDeviceEncryptionSetting);
-  }, [toDeviceEncryptionSetting]);
+  const [didFallbackToRoomKey, setDidFallbackToRoomKey] = useState(false);
   useTypedEventEmitter(
     rtcSession,
     RoomAndToDeviceEvents.EnabledTransportsChanged,
-    (enabled) => setShowToDeviceEncryption(enabled.to_device),
+    (enabled) => setDidFallbackToRoomKey(enabled.room),
```
toger5, line 234: This is the actual bug: I used `enabled.to_device` but of course `enabled.room` is better.

BillCarsonFr (approved):
we can't have a test?

toger5:
@BillCarsonFr yes, I will add a test.

PR #3210, opened 2025-04-15
Add a .links.temp-disabled.yaml state explicitly for when the pre-commit hook disabled the yaml.

toger5:
Otherwise we end up with a linked repo after every commit.


toger5:
I bypassed merge this since it is just the git hooks which are opt in. And I really need this fix to not need to wait for `yarn` to complete on each commit.

PR #3214, opened 2025-04-15
Ecryption key ratcheting dev tools

toger5:
This sets the ratchedWindowSize to >0 so the livekit client will try to ratchet keys before logging a unable to decrypt error.
It also exposes a ratchet button in the developer ui. Clients that do not support the ratched window > 0 will fail to decrypt once that button is pressed.

Supporting clients will jump to the next index.

This is a first step to support ratcheting and explore how it would work.
Since this just improves our devtools with a potentially helpful feature, we can consider merging it even though its mostly intended as a test.

A finding is, that the key material is `extractable: false` so we need to see if there is a solution how we could send the derived keys from the ratcheting to new joiners if we cannot extract it from the matrerial...

### DIFF
--- a/src/e2ee/matrixKeyProvider.ts
+++ b/src/e2ee/matrixKeyProvider.ts
@@ -5,7 +5,11 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 Please see LICENSE in the repository root for full details.
 */
 
-import { BaseKeyProvider, createKeyMaterialFromBuffer } from "livekit-client";
+import {
+  BaseKeyProvider,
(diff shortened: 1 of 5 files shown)


toger5:
Do we want this in the v0.10.0 release to make it possible to introduce ratcheting without breaking compatibility with v0.10.0?
If yes, then we should also do a backport for this and get it reviewed+merged.
To track it I will add it to the release tracker issue.

toger5:
@BillCarsonFr from the testing today this seems to still be a okay idea right?
Should this be adapted to match the new encryption manager an lk patch?

toger5:
@BillCarsonFr I think your emoji debug tool might superseed this?

PR #3221, opened 2025-04-22
only show to-device transport label in dev mode

BillCarsonFr:
Draft because based on https://github.com/element-hq/element-call/pull/3208

image


toger5 (approved):
Looks sane.
The test might fail because its not merging into main.

This could also get a test. But we will probably add a test in the parent PR and then only adjust the test here?

toger5:
I hope its fine If i take over this PR since it is a bit of rebase work I hope you were not eager to do?
(and the tests)

If not just force push your version from your local branch.

toger5 (approved):
@BillCarsonFr I wanted to get this off the table today so I have space for different things tomorrow.
It is basically still exactly your change.
Maybe you can still give it a quick scan and ping me tomorrow if there is any concern.

PR #3224, opened 2025-04-22
Set `ratchetWindowSize` to > 0 enable rachet compatibility.

toger5:
This is a subset of: https://github.com/element-hq/element-call/pull/3214

It will allow us to be compatible with any version of EC that does ratcheting.


BillCarsonFr (approved):
Ok for me. There might be a behavior change regarding how the `failureTolerance` works and when it will stop trying a bad key, but that's fine?

PR #3238, opened 2025-04-29
Test: Temporary disable widget test on firefox

BillCarsonFr:


BillCarsonFr:
It is fixed already 🤣

PR #3247, opened 2025-05-05
Keep the HTML lang attribute in sync with the i18next language

sandhose:
Small thing that has been bothering me for a bit: Firefox always offers me to translate the page, because it thinks it's in English, even though it's already in French.


toger5 (approved):
Thanks @sandhose,
I have not yet observed that but its because my browser is configured differently.

You tested, that it fixes your situation though right?  It does not break anything on my side so I am happy to merge.

PR #3248, opened 2025-05-05
Send notification events when starting a call

robintown:
Previously this has been the responsibility of the hosting application (Element Web / Element X), but I would like to move this responsibility to Element Call itself to make it even more lightweight to integrate Element Call into a widget-capable client.

Depends on https://github.com/matrix-org/matrix-js-sdk/pull/4826


toger5:
I remember the idea being that this is part of the hosting application is that we can have different behaviours on mobile vs desktop and that we might introduce custom ui to send a ring/notify call.
A modal for instance asking you if you really want to ring `LargeNumber` of users. Or a notice, that you do not have the right to ring in this room.

robintown:
I think we would still have the opportunity to do that, as the widget here gets to decide the notification behavior at the point that it enters the session, and the hosting application can always give extra context to the widget about how it's being used (intent etc.) through the URL params

src/rtcSessionHelpers.ts
```diff
@@ -18,8 +18,9 @@ import { AutoDiscovery } from "matrix-js-sdk/lib/autodiscovery";
 import { PosthogAnalytics } from "./analytics/PosthogAnalytics";
 import { Config } from "./config/Config";
 import { ElementWidgetActions, widget, type WidgetHelpers } from "./widget";
-import { MatrixRTCFocusMissingError } from "./utils/errors.ts";
-import { getUrlParams } from "./UrlParams.ts";
+import { MatrixRTCFocusMissingError } from "./utils/errors";
+import { getUrlParams } from "./UrlParams";
```
toger5, line 22: The URL params should include a flag if it should ring or not.

toger5 (approved):
Reviewed in pair-programming session

package.json
```diff
@@ -108,7 +108,7 @@
     "livekit-client": "^2.13.0",
     "lodash-es": "^4.17.21",
     "loglevel": "^1.9.1",
-    "matrix-js-sdk": "github:matrix-org/matrix-js-sdk#head=develop",
+    "matrix-js-sdk": "github:matrix-org/matrix-js-sdk#49e6985694c98d9ab42c1008a5e2afbd13b16145",
```
robintown: Ideally we return this to `head=develop` before merging

alexander-potemkin:
@robintown , @toger5 , thank you for that! Am I correct that this feature has to be enabled in the configs? If so - could you please, advice on how to enable it? And when it might be available in production?

fkwp:
> @robintown , @toger5 , thank you for that! Am I correct that this feature has to be enabled in the configs? If so - could you please, advice on how to enable it? And when it might be available in production?

It needs to be passed as an URL parameter by the Messenger Apps. Its on the agenda next:
- ship new EC release
- get PRs in the messenger Apps
- release the messenger Apps

alexander-potemkin:
> It needs to be passed as an URL parameter by the Messenger Apps. Its on the agenda next:
> 
> * ship new EC release
> * get PRs in the messenger Apps
> * release the messenger Apps

Thank you! From what I recall, from the chat on matrix.org, Element X developers has been resistant to enable group call notifications - so there are chances that messenger just won't enable that (unless I'm missing something!).

If so, is there a possibility to regulate Element Call behavior via server side configs somehow?

PR #3249, opened 2025-05-05
Add custom audio renderer for iPhone earpiece and only render joined participants

### LINKED ISSUE #2927: Only subscribe to audio from MatrixRTC members
### Steps to reproduce

1. Be on a call with someone with a flaky internet connection
2. Wait until the person is automatically disconnected from MatrixRTC due to their connection dropping, but still has a LiveKit connection

#### What did you expect?

I should neither see nor hear the participant since they are disconnected from MatrixRTC.

#### What happened instead?

You can hear someone without seeing their tile.

Their tile disappears, but I can still hear them. This is because we render audio from everyone in the LiveKit room without verifying that they are part of the MatrixRTC session.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

toger5:
This does two things:
 - It will filter playback tracks by matrixRTC members (only actual participants will ever be audible) https://github.com/element-hq/element-call/issues/2927
 - It will add a new "virtual device" option for ios to enable earpice speaker
 
Fixes https://github.com/element-hq/element-call/issues/2927

### DIFF
--- a/src/livekit/MatrixAudioRenderer.test.tsx
+++ b/src/livekit/MatrixAudioRenderer.test.tsx
@@ -1,0 +1,104 @@
+/*
+Copyright 2023, 2024 New Vector Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+Please see LICENSE in the repository root for full details.
+*/
+
+import { afterEach, beforeEach, expect, it, vi } from "vitest";
+import { render } from "@testing-library/react";
+import { type CallMembership } from "matrix-js-sdk/lib/matrixrtc";
+import {
+  getTrackReferenceId,
+  type TrackReference,
+} from "@livekit/components-core";
+import { type RemoteAudioTrack } from "livekit-client";
+import { type ReactNode } from "react";
+import { useTracks } from "@livekit/components-react";
+
+import { testAudioContext } from "../useAudioContext.test";
+import * as MediaDevicesContext from "./MediaDevicesContext";
+import { MatrixAudioRenderer } from "./MatrixAudioRenderer";
+import { mockTrack } from "../utils/test";
+
+export const TestAudioContextConstructor = vi.fn(() => testAudioContext);
+
+beforeEach(() => {
+  vi.stubGlobal("AudioContext", TestAudioContextConstructor);
+});
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+  vi.clearAllMocks();
+});
+
+vi.mock("@livekit/components-react", async (importOriginal) => {
+  return {
+    ...(await importOriginal()), // this will only affect "foo" outside of the original module
+    AudioTrack: (props: { trackRef: TrackReference }): ReactNode => {
+      return (
+        <audio data-testid={"audio"}>
+          {getTrackReferenceId(props.trackRef)}
+        </audio>
+      );
+    },
+    useTracks: vi.fn(),
+  };
+});
+
+const tracks = [mockTrack("test:123")];
+vi.mocked(useTracks).mockReturnValue(tracks);
+
+it("should render for member", () => {
+  const { container, queryAllByTestId } = render(
+    <MatrixAudioRenderer
+      members={[{ sender: "test", deviceId: "123" }] as CallMembership[]}
+    />,
+  );
+  expect(container).toBeTruthy();
+  expect(queryAllByTestId("audio")).toHaveLength(1);
+});
+it("should not render without member", () => {
+  const { container, queryAllByTestId } = render(
+    <MatrixAudioRenderer
+      members={[{ sender: "othermember", deviceId: "123" }] as CallMembership[]}
+    />,
+  );
+  expect(container).toBeTruthy();
+  expect(queryAllByTestId("audio")).toHaveLength(0);
+});
+
+it("should not setup audioContext gain and pan if there is no need to.", () => {
+  render(
+    <MatrixAudioRenderer
+      members={[{ sender: "test", deviceId: "123" }] as CallMembership[]}
+    />,
+  );
+  const audioTrack = tracks[0].publication.track! as RemoteAudioTrack;
+
+  expect(audioTrack.setAudioContext).toHaveBeenCalledTimes(1);
+  expect(audioTrack.setAudioContext).toHaveBeenCalledWith(undefined);
+  expect(audioTrack.setWebAudioPlugins).toHaveBeenCalledTimes(1);
+  expect(audioTrack.setWebAudioPlugins).toHaveBeenCalledWith([]);
+
+  expect(testAudioContext.gain.gain.value).toEqual(1);
+  expect(testAudioContext.pan.pan.value).toEqual(0);
+});
+it("should setup audioContext gain and pan", () => {
+  vi.spyOn(MediaDevicesContext, "useEarpieceAudioConfig").mockReturnValue({
+    pan: 1,
+    volume: 0.1,
+  });
+  render(
+    <MatrixAudioRenderer
+      members={[{ sender: "test", deviceId: "123" }] as CallMembership[]}
+    />,
+  );
+
+  const audioTrack = tracks[0].publication.track! as RemoteAudioTrack;
+  expect(audioTrack.setAudioContext).toHaveBeenCalled();
+  expect(audioTrack.setWebAudioPlugins).toHaveBeenCalled();
+
+  expect(testAudioContext.gain.gain.value).toEqual(0.1);
+  expect(testAudioContext.pan.pan.value).toEqual(1);
+});
--- a/src/livekit/MatrixAudioRenderer.test.tsx
+++ b/src/livekit/MatrixAudioRenderer.test.tsx
@@ -34,7 +34,7 @@ afterEach(() => {
 
 vi.mock("@livekit/components-react", async (importOriginal) => {
   return {
-    ...(await importOriginal()), // this will only affect "foo" outside of the original module
+    ...(await importOriginal()),
     AudioTrack: (props: { trackRef: TrackReference }): ReactNode => {
       return (
         <audio data-testid={"audio"}>
--- a/src/livekit/MatrixAudioRenderer.tsx
+++ b/src/livekit/MatrixAudioRenderer.tsx
@@ -7,7 +7,7 @@ Please see LICENSE in the repository root for full details.
 
 import { getTrackReferenceId } from "@livekit/components-core";
 import { type RemoteAudioTrack, Track } from "livekit-client";
-import { useEffect, useMemo, useRef, type ReactNode } from "react";
+import { useEffect, useMemo, useRef, useState, type ReactNode } from "react";
 import {
   useTracks,
   AudioTrack,
@@ -88,7 +88,8 @@ export function MatrixAudioRenderer({
     },
   ).filter((ref) => {
     const isValid = validIdentities?.has(ref.participant.identity);
-    if (!isValid) logInvalid(ref.participant.identity, validIdentities);
+    if (!isValid && !ref.participant.isLocal)
+      logInvalid(ref.participant.identity, validIdentities);
     return (
       !ref.participant.isLocal &&
       ref.publication.kind === Track.Kind.Audio &&
@@ -97,11 +98,11 @@ export function MatrixAudioRenderer({
   });
 
   // This component is also (in addition to the "only play audio for connected members" logic above)
-  // to mimic earpice audio on iphones.
-  // The safari audio devices enumeration does not expose an earpice audio device.
+  // responsible for mimicking earpiece audio on iPhones.
+  // The Safari audio devices enumeration does not expose an earpiece audio device.
   // We alternatively use the audioContext pan node to only use one of the stereo channels.
 
-  // This component does get additionally complicated because of a safari bug.
+  // This component does get additionally complicated because of a Safari bug.
   // (see: https://bugs.webkit.org/show_bug.cgi?id=251532
   // and the related issues: https://bugs.webkit.org/show_bug.cgi?id=237878
   // and https://bugs.webkit.org/show_bug.cgi?id=231105)
@@ -109,31 +110,38 @@ export function MatrixAudioRenderer({
   // AudioContext gets stopped if the webview gets moved into the background.
   // Once the phone is in standby audio playback will stop.
   // So we can only use the pan trick only works is the phone is not in standby.
-  // If earpice mode is not used we do not use audioContext to allow standby playback.
+  // If earpiece mode is not used we do not use audioContext to allow standby playback.
   // shouldUseAudioContext is set to false if stereoPan === 0 to allow standby bluetooth playback.
 
   const { pan: stereoPan, volume: volumeFactor } = useEarpieceAudioConfig();
   const shouldUseAudioContext = stereoPan !== 0;
 
   // initialize the potentially used audio context.
-  const audioContext = useMemo(() => new AudioContext(), []);
+  const [audioContext, setAudioContext] = useState<AudioContext | undefined>(
+    undefined,
+  );
+  useEffect(() => {
+    const ctx = new AudioContext();
+    setAudioContext(ctx);
+    return (): void => {
+      void ctx.close();
+    };
+  }, []);
   const audioNodes = useMemo(
     () => ({
-      gain: audioContext.createGain(),
-      pan: audioContext.createStereoPanner(),
+      gain: audioContext?.createGain(),
+      pan: audioContext?.createStereoPanner(),
     }),
     [audioContext],
   );
 
   // Simple effects to update the gain and pan node based on the props
   useEffect(() => {
-    audioNodes.pan.pan.value = stereoPan;
-  }, [audioNodes.pan.pan, stereoPan]);
+    if (audioNodes.pan) audioNodes.pan.pan.value = stereoPan;
+  }, [audioNodes.pan, stereoPan]);
   useEffect(() => {
-    // *4 to balance the transition from audio context to normal audio playback.
-    // probably needed due to gain behaving differently than el.volume
-    audioNodes.gain.gain.value = volumeFactor;
-  }, [audioNodes.gain.gain, volumeFactor]);
+    if (audioNodes.gain) audioNodes.gain.gain.value = volumeFactor;
+  }, [audioNodes.gain, volumeFactor]);
 
   return (
     // We add all audio elements into one <div> for the browser developer tool experience/tidyness.
@@ -155,8 +163,8 @@ interface StereoPanAudioTrackProps {
   muted?: boolean;
   audioContext?: AudioContext;
   audioNodes: {
-    gain: GainNode;
-    pan: StereoPannerNode;
+    gain?: GainNode;
+    pan?: StereoPannerNode;
   };
 }
 
@@ -182,15 +190,18 @@ function AudioTrackWithAudioNodes({
   // (adding the audio context when already mounted did not work outside strict mode)
   const [trackReady, setTrackReady] = useReactiveState(
     () => false,
-    [audioContext || audioNodes],
+    // We only want the track to reset once both (audioNodes and audioContext) are set.
+    // for unsetting the audioContext its enough if one of the the is undefined.
+    [audioContext && audioNodes],
   );
 
   useEffect(() => {
     if (!trackRef || trackReady) return;
     const track = trackRef.publication.track as RemoteAudioTrack;
-    track.setAudioContext(audioContext);
+    const useContext = audioContext && audioNodes.gain && audioNodes.pan;
+    track.setAudioContext(useContext ? audioContext : undefined);
     track.setWebAudioPlugins(
-      audioContext ? [audioNodes.gain, audioNodes.pan] : [],
+      useContext ? [audioNodes.gain!, audioNodes.pan!] : [],
     );
     setTrackReady(true);
   }, [audioContext, audioNodes, setTrackReady, trackReady, trackRef]);
--- a/src/livekit/MatrixAudioRenderer.tsx
+++ b/src/livekit/MatrixAudioRenderer.tsx
@@ -1,0 +1,201 @@
+/*
+Copyright 2025 New Vector Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+Please see LICENSE in the repository root for full details.
+*/
+
+import { getTrackReferenceId } from "@livekit/components-core";
+import { type RemoteAudioTrack, Track } from "livekit-client";
+import { useEffect, useMemo, useRef, type ReactNode } from "react";
+import {
+  useTracks,
+  AudioTrack,
+  type AudioTrackProps,
+} from "@livekit/components-react";
+import { type CallMembership } from "matrix-js-sdk/lib/matrixrtc";
+import { logger } from "matrix-js-sdk/lib/logger";
+
+import { useEarpieceAudioConfig } from "./MediaDevicesContext";
+import { useReactiveState } from "../useReactiveState";
+
+export interface MatrixAudioRendererProps {
+  /**
+   * The list of participants to render audio for.
+   * This list needs to be composed based on the matrixRTC members so that we do not play audio from users
+   * that are not expected to be in the rtc session.
+   */
+  members: CallMembership[];
+  /**
+   * If set to `true`, mutes all audio tracks rendered by the component.
+   * @remarks
+   * If set to `true`, the server will stop sending audio track data to the client.
+   */
+  muted?: boolean;
+}
+
+/**
+ * The `MatrixAudioRenderer` component is a drop-in solution for adding audio to your LiveKit app.
+ * It takes care of handling remote participants’ audio tracks and makes sure that microphones and screen share are audible.
+ *
+ * It also takes care of the earpiece audio configuration for iOS devices.
+ * This is done by using the WebAudio API to create a stereo pan effect that mimics the earpiece audio.
+ * @example
+ * ```tsx
+ * <LiveKitRoom>
+ *   <MatrixAudioRenderer />
+ * </LiveKitRoom>
+ * ```
+ * @public
+ */
+export function MatrixAudioRenderer({
+  members,
+  muted,
+}: MatrixAudioRendererProps): ReactNode {
+  const validIdentities = useMemo(
+    () =>
+      new Set(members?.map((member) => `${member.sender}:${member.deviceId}`)),
+    [members],
+  );
+
+  const loggedInvalidIdentities = useRef(new Set<string>());
+  /**
+   * Log an invalid livekit track identity.
+   * A invalid identity is one that does not match any of the matrix rtc members.
+   *
+   * @param identity The identity of the track that is invalid
+   * @param validIdentities The list of valid identities
+   */
+  const logInvalid = (identity: string, validIdentities: Set<string>): void => {
+    if (loggedInvalidIdentities.current.has(identity)) return;
+    logger.warn(
+      `Audio track ${identity} has no matching matrix call member`,
+      `current members: ${Array.from(validIdentities.values())}`,
+      `track will not get rendered`,
+    );
+    loggedInvalidIdentities.current.add(identity);
+  };
+
+  const tracks = useTracks(
+    [
+      Track.Source.Microphone,
+      Track.Source.ScreenShareAudio,
+      Track.Source.Unknown,
+    ],
+    {
+      updateOnlyOn: [],
+      onlySubscribed: true,
+    },
+  ).filter((ref) => {
+    const isValid = validIdentities?.has(ref.participant.identity);
+    if (!isValid) logInvalid(ref.participant.identity, validIdentities);
+    return (
+      !ref.participant.isLocal &&
+      ref.publication.kind === Track.Kind.Audio &&
+      isValid
+    );
+  });
+
+  // This component is also (in addition to the "only play audio for connected members" logic above)
+  // to mimic earpice audio on iphones.
+  // The safari audio devices enumeration does not expose an earpice audio device.
+  // We alternatively use the audioContext pan node to only use one of the stereo channels.
+
+  // This component does get additionally complicated because of a safari bug.
+  // (see: https://bugs.webkit.org/show_bug.cgi?id=251532
+  // and the related issues: https://bugs.webkit.org/show_bug.cgi?id=237878
+  // and https://bugs.webkit.org/show_bug.cgi?id=231105)
+  //
+  // AudioContext gets stopped if the webview gets moved into the background.
+  // Once the phone is in standby audio playback will stop.
+  // So we can only use the pan trick only works is the phone is not in standby.
+  // If earpice mode is not used we do not use audioContext to allow standby playback.
+  // shouldUseAudioContext is set to false if stereoPan === 0 to allow standby bluetooth playback.
+
+  const { pan: stereoPan, volume: volumeFactor } = useEarpieceAudioConfig();
+  const shouldUseAudioContext = stereoPan !== 0;
+
+  // initialize the potentially used audio context.
+  const audioContext = useMemo(() => new AudioContext(), []);
+  const audioNodes = useMemo(
+    () => ({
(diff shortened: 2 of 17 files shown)


src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,144 @@
...
+  channels: number;
+  timestamp: number;
+}
+
+/** @public */
+export interface RoomAudioRendererProps {
+  /** Does an additional gain control on all audio tracks rendered by this component. This does not interfere with the
+   * `setVolume` of the track itself.
+   */
+  volumeFactor?: number;
```
Half-Shot: Bit confused by this. Is the volumeFactor cumulative or just the default?

I.e. is it

1. `volumeFactor * trackVolumeFactor` for each track?
2. `trackVolumeFactor ||  volumeFactor` for each track?

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,144 @@
...
+  members?: CallMembership[];
+  /**
+   * The balance of the audio tracks.
+   * This is a value between -1 and 1, where -1 is full left, 0 is center, and 1 is full right.
+   */
+  balance?: number;
+}
+
+/**
+ * The `RoomAudioRenderer` component is a drop-in solution for adding audio to your LiveKit app.
```
Half-Shot: This seems like the description of the parent component?

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,144 @@
...
+        gainNode.gain.value = volumeFactor ?? 1;
+        track.setWebAudioPlugins([panNode, gainNode]);
+        trackNodes.current.set(trackRef.participant.identity, {
+          pan: panNode,
+          gain: gainNode,
+        });
+      }
+
+      // Remove nodes from trackNodes map.
+      trackNodes.current.forEach((_, identity) => {
```
Half-Shot: Do we need to do any cleanup of the audioContext when removing tracks? Can they end up added twice or something?

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,144 @@
...
+        if (!tracks.some((ref) => ref.participant.identity === identity)) {
+          trackNodes.current.delete(identity);
+        }
+      });
+    });
+  }, [balance, tracks, volumeFactor]);
+
+  return (
+    <div style={{ display: "none" }}>
```
Half-Shot: ooi, do the AudoTracks' render anything? We need to display: none them?

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
toger5: its
`track.connect(gainNode).connect(panNode).connect(gainNode2)`
where:
`gainNode`: is the configured by the volumeFactor and `gainNode2` is configured by setVolume on the RemoteAudioTrack
So it think 1. fits better but then gain is not linear i think. so its similar but different.

src/settings/SettingsModal.tsx
```diff
@@ -121,6 +122,16 @@ export const SettingsModal: FC<Props> = ({
             title={t("settings.devices.speaker")}
             numberedLabel={(n) => t("settings.devices.speaker_numbered", { n })}
           />
+          <FieldRow>
+            <InputField
```
Half-Shot: This is more of a product nit, but I'm surprised we don't just expose a pan setting like most volume controls do.

Half-Shot (dismissed):
Just had some questions. Implementation looks sane

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
Half-Shot: alright, that sounds like it would match what I'd expect the buttons to do :)

src/settings/SettingsModal.tsx
toger5: Yes we definitly should. I am not sure what solution I should use however. android does expose this autmatically so we dont need the stereo pan for it.
So we could make this a url flag. Or maybe there is an obvious mobile safari browser hint we can consume and expose the option then.

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
toger5: I think this was just to group them into one collapsable section?

src/useAudioContext.tsx
```diff
@@ -105,23 +111,30 @@ export function useAudioContext<S extends string>(
     if (audioContext && "setSinkId" in audioContext) {
       // https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/setSinkId
       // @ts-expect-error - setSinkId doesn't exist yet in types, maybe because it's not supported everywhere.
-      audioContext.setSinkId(devices.audioOutput.selectedId).catch((ex) => {
+      audioContext.setSinkId(audioOutput.selectedId).catch((ex) => {
         logger.warn("Unable to change sink for audio context", ex);
       });
     }
-  }, [audioContext, devices]);
+  }, [audioContext, audioOutput]);
```
robintown:
```suggestion
  }, [audioContext, audioOutput.selectedId]);
```

Hopefully React is cool with this?

src/useAudioContext.test.tsx
```diff
@@ -39,94 +39,156 @@ const TestComponent: FC = () => {
...
+    if (MockAudioContext.testContext) {
+      this.gain = MockAudioContext.testContext.gain;
+      this.pan = MockAudioContext.testContext.pan;
+      this.setSinkId = MockAudioContext.testContext.setSinkId;
+      this.decodeAudioData = MockAudioContext.testContext.decodeAudioData;
+      this.createBufferSource = MockAudioContext.testContext.createBufferSource;
+      this.createGain = MockAudioContext.testContext.createGain;
+      this.createStereoPanner = MockAudioContext.testContext.createStereoPanner;
+      this.close = MockAudioContext.testContext.close;
+    }
```
robintown: This is a bit awkward, I take it you're trying to have repeated calls to `new MockAudioContext` return "the same value" essentially. I think you could achieve this more easily by using a simple object + function instead of the ES6 `class` syntax:

```typescript
const testContext = { gain: ..., pan: ..., /* et cetera */ }
function MockAudioContext() { return testContext }

const ctx1 = new MockAudioContext()
const ctx2 = new MockAudioContext()
expect(ctx1).toBe(ctx2) // this now works
```

src/settings/DeveloperSettingsTab.tsx
```diff
@@ -175,6 +179,20 @@ export const DeveloperSettingsTab: FC<Props> = ({ client, livekitRoom }) => {
           )}
         />
       </FieldRow>
+      <FieldRow>
+        <InputField
+          id="alwaysShowIphoneEarpiece"
+          type="checkbox"
+          label={t("developer_mode.always_show_iphone_earpiece")}
+          checked={!!alwaysShowIphoneEarpiece}
```
robintown:
```suggestion
          checked={alwaysShowIphoneEarpiece}
```

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
robintown, line 1: Why has this ended up in the MediaPipe folder?

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+   * The list of participants to render audio for.
+   * This list needs to be composed based on the matrixRTC members so that we do not play audio from users
+   * that are not expected to be in the rtc session.
+   */
+  members: CallMembership[];
+  /**
+   * If set to `true`, mutes all audio tracks rendered by the component.
+   * @remarks
+   * If set to `true`, the server will stop sending audio track data to the client.
+   * @alpha
```
robintown: Why `@alpha`? Is this expected to be unstable?

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+ * It takes care of handling remote participants’ audio tracks and makes sure that microphones and screen share are audible.
+ *
+ * @example
+ * ```tsx
+ * <LiveKitRoom>
+ *   <MatrixAudioRenderer />
+ * </LiveKitRoom>
+ * ```
+ * @public
+ */
```
robintown: This doc comment looks copied & pasted / inaccurate now that this component is more custom

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+export function MatrixAudioRenderer({
+  members,
+  muted,
+}: MatrixAudioRendererProps): ReactNode {
+  const validIdentities = useMemo(
+    () => members?.map((member) => member.sender + ":" + member.deviceId),
```
robintown:
```suggestion
    () => members?.map((member) => `${member.sender}:${member.deviceId}`),
```

Templates are more performant whenever there are more than 2 parts to the string, IIRC

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+      Track.Source.Microphone,
+      Track.Source.ScreenShareAudio,
+      Track.Source.Unknown,
+    ],
+    {
+      updateOnlyOn: [],
+      onlySubscribed: true,
+    },
+  ).filter((ref) => {
+    const isValid = validIdentities?.includes(ref.participant.identity);
```
robintown: Perhaps it's appropriate for `validIdentities` to be a `Set`, if this is what we're using it for

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+    if (!isValid) logInvalid(ref.participant.identity, validIdentities);
```
robintown: This looks like it could cause a lot of repeated logs for the same users, as it runs on every update to the list of tracks. Is there an easy mitigation here to only log once per participant?

It also looks like this line could end up warning that we're not rendering the local user's audio, since it comes before the `!ref.participant.isLocal` branch, even though that'd be a non-issue.

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
robintown: A comment that this `<div>` exists for dev experience wouldn't hurt

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+    </div>
+  );
+}
+
+interface StereoPanAudioTrackProps {
+  /**
+   * This additional volume factor is applied to all audio tracks from livekit.
+   *
+   * @remark It does not interfere with the `setVolume` of the track itself.
+   * The seperta `volume` property will override/conflict with what is set using `track.setVolume`
```
robintown:
```suggestion
   * The seperate `volume` property will override/conflict with what is set using `track.setVolume`
```

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+   */
+  volumeFactor: number;
+  /**
+   * The left right stereo pan of the audio tracks.
```
robintown:
```suggestion
   * The left-right stereo pan of the audio tracks.
```

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+  // So we can only use the pan trick in the following conditions
+  // - webview is visible
+  // - stereoPan is not 0
+  //
+  // The architecture of the component is as follows:
+  // the `audioContext` state is used to determine if we should use the audio context or not.
+  // the actual `audioContext` is stored in `audioContextRef` to minimize reallocations.
+  // whenever we notice the audioContext is active but should not: `!!audioContext !== shouldUseAudioContext`
+  // we remount the audio tracks using a true -> false -> true transition for the `trackReady` state.
+  const audioContextRef = useRef(new AudioContext());
```
robintown: This will create a new AudioContext on each render, which is dangerous: you need to at least call the `close()` method.

```suggestion
  const audioContext = useMemo(() => new AudioContext(), []);
  useEffect(() => () => audioContext.close(), [audioContext]);
```

But in general we've seen this cause big resource leaks in the past, back when walkie-talkie mode was a thing. Is there some way that we could just create a single AudioContext in the global scope and use that across the entire app? (that is, use the same context in `useAudioContext` too?) Given that an accumulation of AudioContexts can cause stuttering audio, I think we should consider them expensive.

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+  const gainRef = useRef(audioContextRef.current.createGain());
+  const panRef = useRef(audioContextRef.current.createStereoPanner());
+
+  // Simple effects to update the gain and pan node based on the props
+  useEffect(() => {
+    panRef.current.pan.value = stereoPan;
+  }, [stereoPan]);
+  useEffect(() => {
+    // *4 to balance the transition from audio context to normal audio playback.
+    // probably needed due to gain bahving differently than el.volume
```
robintown:
```suggestion
    // probably needed due to gain behaving differently than el.volume
```

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+  //
+  // The architecture of the component is as follows:
+  // the `audioContext` state is used to determine if we should use the audio context or not.
+  // the actual `audioContext` is stored in `audioContextRef` to minimize reallocations.
+  // whenever we notice the audioContext is active but should not: `!!audioContext !== shouldUseAudioContext`
+  // we remount the audio tracks using a true -> false -> true transition for the `trackReady` state.
+  const audioContextRef = useRef(new AudioContext());
+
+  const gainRef = useRef(audioContextRef.current.createGain());
+  const panRef = useRef(audioContextRef.current.createStereoPanner());
```
robintown: These should use `useMemo` too

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+  // This is used to unmount/remount the AudioTrack component.
+  // Mounting needs to happen after the audioContext is set.
+  // (adding the audio context when already mounted did not work outside strict mode)
+  const [trackReady, setTrackReady] = useState(false);
+
+  const [isVisible, setIsVisible] = useState(true);
+  useEventTarget(document, "visibilitychange", () =>
+    setIsVisible(!document.hidden),
+  );
```
robintown:
```suggestion
  useEventTarget(document, "visibilitychange", useCallback(() =>
    setIsVisible(!document.hidden),
  []));
```

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+    gainRef.current.gain.value = Math.min(volumeFactor * 4, 1);
+  }, [volumeFactor]);
+  const audioContextEnabled = useRef(false);
+
+  // This is used to unmount/remount the AudioTrack component.
+  // Mounting needs to happen after the audioContext is set.
+  // (adding the audio context when already mounted did not work outside strict mode)
+  const [trackReady, setTrackReady] = useState(false);
+
+  const [isVisible, setIsVisible] = useState(true);
```
robintown:
```suggestion
  const [isVisible, setIsVisible] = useState(!document.hidden);
```

to be safe?

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+    track.setWebAudioPlugins(
+      audioContextEnabled.current ? [panRef.current, gainRef.current] : [],
+    );
+    setTrackReady(true);
+  }, [audioContextRef, trackReady, trackRef]);
+
+  const volume = useMemo(
+    () => (audioContextEnabled.current ? props.volume : volumeFactor),
+    [audioContextEnabled, props.volume, volumeFactor],
+  );
```
robintown: This value is not guaranteed to be reactive with respect to `audioContextEnabled.current`, since that's a ref.

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,215 @@
...
+    if (!trackRef || trackReady) return;
+    const track = trackRef.publication.track as RemoteAudioTrack;
+    track.setAudioContext(
+      audioContextEnabled.current ? audioContextRef.current : undefined,
+    );
+    track.setWebAudioPlugins(
+      audioContextEnabled.current ? [panRef.current, gainRef.current] : [],
+    );
+    setTrackReady(true);
+  }, [audioContextRef, trackReady, trackRef]);
```
robintown: I think you could get rid of the `audioContextEnabled` ref and instead react to changes in `shouldUseAudioContext`, like:

```typescript
  const shouldUseAudioContext = isVisible && stereoPan !== 0;
  useEffect(() => {
    // Once sth changes (stereoPan or isVisible) so that we change if we use the audioContext.
    // We will disable the audio track.
    // this will unmount the AudioTrack component.
    // it will get remounted once the track is updated with the new audioContext configurations.
    setTrackReady(false);
  }, [shouldUseAudioContext]);

  useEffect(() => {
    if (!trackRef || trackReady) return;
    const track = trackRef.publication.track as RemoteAudioTrack;
    track.setAudioContext(
      shouldUseAudioContext ? audioContextRef.current : undefined,
    );
    track.setWebAudioPlugins(
      shouldUseAudioContext ? [panRef.current, gainRef.current] : [],
    );
    setTrackReady(true);
  }, [audioContextRef, trackReady, trackRef, shouldUseAudioContext]);
```

or perhaps even:

```typescript
  const shouldUseAudioContext = isVisible && stereoPan !== 0;
  const [trackReady, setTrackReady] = useReactiveState(() => false, [shouldUseAudioContext]);

  useEffect(() => {
    if (!trackRef || trackReady) return;
    const track = trackRef.publication.track as RemoteAudioTrack;
    track.setAudioContext(
      shouldUseAudioContext ? audioContextRef.current : undefined,
    );
    track.setWebAudioPlugins(
      shouldUseAudioContext ? [panRef.current, gainRef.current] : [],
    );
    setTrackReady(true);
  }, [audioContextRef, trackReady, trackRef, shouldUseAudioContext]);
```

This could help with the previous comment about reactivity too.

src/livekit/MediaDevicesContext.tsx
```diff
@@ -102,27 +111,39 @@ function useMediaDevice(
             // Create a virtual default audio output for browsers that don't have one.
             // Its device ID must be the empty string because that's what setSinkId
             // recognizes.
+            // we also create this if we do not have any available devices, so that
+            // we can use the default or the earpiece.
```
robintown: I'd be interested to know what platforms/conditions can cause there to be no available devices, because it was useful to have this `available.size` branch to hide the device list in settings during the loading state (while we're still awaiting the `enumerateDevices` call).

src/livekit/MediaDevicesContext.tsx
```diff
@@ -102,27 +111,39 @@ function useMediaDevice(
...
               ]);
+            if (
+              kind === "audiooutput" &&
+              (navigator.userAgent.match("iPhone") || alwaysShowIphoneEarpice)
+            )
+              // On IPhones we have to create a virtual earpiece device, because
+              // the earpiece is not available as a device ID.
+              available = new Map([
+                ...available,
+                [EARPIECE_CONFIG_ID, { type: "name", name: "Earpiece" }],
```
robintown: This `"Earpiece"` string needs translation. Perhaps instead of using `type: "name"`, we could introduce a new `type: "earpiece"` variant to `DeviceLabel`.

robintown (commented):
Not fully through reviewing the last bits of the MediaDevicesContext changes but here's what I've got:

src/useAudioContext.test.tsx
toger5: I ended up to keep testContext a class instance instead of a struct/object.
Reason being that we need the types in the test to `expect` the mocked values.
Those types otherwise wuold have need to been defined with an additional interface.
This way we get it automatically.

src/livekit/MediaDevicesContext.tsx
toger5: Thanks for the explanation.
The new change then seems to be the correct solution to me.

With the shortcoming, that we can already see default + earpice while we are still enumerating devices.
But it will remove the "fake default" once we are done enumerating.

To answer your question: I only tested android and it lists not output on enumerate devices.

src/livekit/MediaDevicesContext.tsx
toger5: That makes sense. Was thinking i could save the work and just reuse name but this ends up to be clearer.

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
toger5: Our local participation also should not be invalid rigth?
Even though it would not break the call the log still might be helpful, that we connected to lk before our memberhsip event was set.
Since we do not have local echos this might be a fairly commen case however...

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
toger5: Are you sure it would create one each render.
Or do you mean a new one on each creation of the compoenent?

But that will be the same for the memo version.

I am not sure we can have the same context with different destinations?
Now they are all just connected to one html element.
Lk also provides a way to pass a single mixed audio context to the lk room.
But this would be an entirely differnt architecture and I am not sure we would be able to enforece the correct remounting ourder as easily when switching from audio context to not using audio context.

I would postpone a central audioContext refactor once we have a decision if locking ios should mute audio or switch back to native html audio?

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
toger5: This would not make a difference except allowing us to call `.disconnect()`?

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
toger5: Why is that needed? It seems like bette hygene but what would happen wihtou use callback?

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
toger5: Dont we also need to put shouldUseAudioContext into a memo?

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
toger5: It seemed to be most related.

toger5: I move it to the livekit folder.

toger5:
@robintown This has changed a bit since we cannot do the fallback to HTML5 audio because HTML5 audio volume control is broken on safari...
I rebased and squashed the changes to cleanup the PR and work on the playwright tests (thats why i wanted/needed to rebase)
I hope this is okay.
I only touched the `MatrixAudioRenderer` after your review. and it has changed enough so it makes sense to review the whole file again (it got much simpler)

src/livekit/MediaDevicesContext.tsx
robintown: Nice, thanks for adding `available.size` back

src/room/InCallView.test.tsx
```diff
@@ -48,6 +44,8 @@ import {
 } from "../settings/settings";
 import { ReactionsSenderProvider } from "../reactions/useReactionsSender";
 import { useRoomEncryptionSystem } from "../e2ee/sharedKeyManagement";
+// import { testAudioContext } from "../useAudioContext.test";
```
robintown:
```suggestion
```

src/settings/DeviceSelection.tsx
```diff
@@ -22,30 +22,34 @@ import {
...
-      devices.select(e.target.value);
+      device.select(e.target.value);
     },
-    [devices],
+    [device],
   );
 
-  if (devices.available.size == 0) return null;
+  // There is no need to show the menu if there is no choice that can be made.
+  if (device.available.size == 1) return null;
```
robintown:
```suggestion
  if (device.available.size <= 1) return null;
```

I suspect this will currently show a heading with no contents in the case of 0 devices (visible in Firefox while awaiting the `enumerateDevices` call)

src/livekit/MatrixAudioRenderer.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+});
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+  vi.clearAllMocks();
+});
+
+vi.mock("@livekit/components-react", async (importOriginal) => {
+  return {
+    ...(await importOriginal()), // this will only affect "foo" outside of the original module
```
robintown: Was "foo" meant to say something else?

src/livekit/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,201 @@
...
+  // This component does get additionally complicated because of a safari bug.
+  // (see: https://bugs.webkit.org/show_bug.cgi?id=251532
+  // and the related issues: https://bugs.webkit.org/show_bug.cgi?id=237878
+  // and https://bugs.webkit.org/show_bug.cgi?id=231105)
+  //
+  // AudioContext gets stopped if the webview gets moved into the background.
+  // Once the phone is in standby audio playback will stop.
+  // So we can only use the pan trick only works is the phone is not in standby.
+  // If earpice mode is not used we do not use audioContext to allow standby playback.
+  // shouldUseAudioContext is set to false if stereoPan === 0 to allow standby bluetooth playback.
```
robintown:
```suggestion
  // This component is also (in addition to the "only play audio for connected members" logic above)
  // responsible for mimicking earpiece audio on iPhones.
  // The Safari audio devices enumeration does not expose an earpiece audio device.
  // We alternatively use the audioContext pan node to only use one of the stereo channels.

  // This component does get additionally complicated because of a Safari bug.
  // (see: https://bugs.webkit.org/show_bug.cgi?id=251532
  // and the related issues: https://bugs.webkit.org/show_bug.cgi?id=237878
  // and https://bugs.webkit.org/show_bug.cgi?id=231105)
  //
  // AudioContext gets stopped if the webview gets moved into the background.
  // Once the phone is in standby audio playback will stop.
  // So we can only use the pan trick only works is the phone is not in standby.
  // If earpiece mode is not used we do not use audioContext to allow standby playback.
  // shouldUseAudioContext is set to false if stereoPan === 0 to allow standby bluetooth playback.
```

src/livekit/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+  const { pan: stereoPan, volume: volumeFactor } = useEarpieceAudioConfig();
+  const shouldUseAudioContext = stereoPan !== 0;
+
+  // initialize the potentially used audio context.
+  const audioContext = useMemo(() => new AudioContext(), []);
```
robintown: This will still result in an growing number of un-closed AudioContexts if you join and leave calls repeatedly. I think we do need to either move the AudioContext to a global/module level `const`, or else use the trick we discussed where you create the AudioContext in an effect, after all.

```typescript
const [audioContext, setAudioContext] = useState<AudioContext | undefined>(undefined)
useEffect(() => {
  const ctx = new AudioContext()
  setAudioContext(ctx)
  return () => ctx.close()
}, [])
```

(that one, just for reference)

src/livekit/MatrixAudioRenderer.tsx
```diff
@@ -0,0 +1,201 @@
...
+  );
+
+  // Simple effects to update the gain and pan node based on the props
+  useEffect(() => {
+    audioNodes.pan.pan.value = stereoPan;
+  }, [audioNodes.pan.pan, stereoPan]);
+  useEffect(() => {
+    // *4 to balance the transition from audio context to normal audio playback.
+    // probably needed due to gain behaving differently than el.volume
+    audioNodes.gain.gain.value = volumeFactor;
```
robintown: I don't see a `* 4` here

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
robintown: I'm thinking of the case where my client fails to refresh the delayed event due to poor homeserver performance while I remain connected to the SFU, in the middle of the call. That happens to me every few days in our stand-ups.

> Even though it would not break the call the log still might be helpful, that we connected to lk before our memberhsip event was set.

Potentially, though warning that we're not rendering our own audio is still a misleading thing to say in this case

src/settings/DeviceSelection.tsx
toger5: I think you are right...

src/livekit/MatrixAudioRenderer.tsx
toger5: I think I tried that and it did not work. But let me play around with that again.

src/livekit/MatrixAudioRenderer.tsx
toger5: yes the *4 could get removed. Thanks the comment needs updating.

src/mediapipe/imageSegmenter/audioNativeRenderer/MatrixAudioRenderer.tsx
toger5: I added a not is local check.

robintown (approved):

PR #3255, opened 2025-05-09
Fix connection leaks: Ensure that any pending connection open are cancelled/undo when ActiveCall is unmounted

### LINKED ISSUE #3268: Potential race causing a re-connection to livekit after the component is unmounted.
### Steps to reproduce

Some Rageshake stating that Oher members could still hear a user after he left.
https://github.com/element-hq/element-call-rageshakes/issues/10192

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

Yes

BillCarsonFr:
Fixes #3268


src/livekit/useECConnectionState.ts
```diff
@@ -59,7 +59,8 @@ async function doConnect(
   livekitRoom: Room,
   sfuConfig: SFUConfig,
   audioEnabled: boolean,
-  audioOptions: AudioCaptureOptions,
+  initialDeviceId: string | undefined,
+  cancellable: Cancellable,
```
toger5:
```suggestion
  cancelable: Cancelable,
```

toger5: Is `abortHandle` a good name for this. `cancelable` sounds more like a bool flag?

src/livekit/useECConnectionState.ts
```diff
@@ -82,19 +83,40 @@ async function doConnect(
...
     if (audioTracks.length < 1) {
       logger.info("Tried to pre-create local audio track but got no tracks");
     } else {
       preCreatedAudioTrack = audioTracks[0];
     }
+    // There was a yield point previously (awaiting for the track to be created) so we need to check
+    // if the operation was cancelled and stop connecting if needed.
+    if (cancellable.isCancelled()) {
+      logger.info(
+        "[Lifecycle] Signal Aborted: Pre-created audio track but connection aborted",
```
toger5, line 98: Can we use a logger child instead of a manual `[Lifecycle]` prefix?

src/livekit/useECConnectionState.ts
```diff
@@ -247,6 +281,22 @@ export function useECConnectionState(
 
   const currentSFUConfig = useRef(Object.assign({}, sfuConfig));
 
+  // Protection against potential leaks, where the component to be unmounted and there is
```
toger5, line 284: "is about to be"?

src/livekit/useECConnectionState.ts
```diff
@@ -247,6 +281,22 @@ export function useECConnectionState(
...
+  // still a pending doConnect promise. This would lead the user to still be in the call even
+  // if the component is unmounted.
+  const cancelBag = useRef(new Set<Cancellable>());
```
toger5: `abortHandles`?

src/livekit/useECConnectionState.ts
```diff
@@ -273,11 +323,14 @@ export function useECConnectionState(
       // always capturing audio: it helps keep bluetooth headsets in the right mode and
       // mobile browsers to know we're doing a call.
       setIsInDoConnect(true);
+      const cancellable = new Cancellable();
+      cancelBag.current.add(cancellable);
```
toger5: Is there an assumption we can do about having only one doConnect running?
Can we enforce this somehow? This would allow us to not use a Set for the `Cancellable`s (`abortHandles`)

src/livekit/useLivekit.ts
```diff
@@ -155,9 +155,7 @@ export function useLivekit(
   );
 
   const connectionState = useECConnectionState(
-    {
-      deviceId: initialDevices.current.audioInput.selectedId,
-    },
+    initialDevices.current.audioInput.selectedId,
```
toger5, line 158: Really good that you detected and fixed this!

src/utils/cancellable.ts
```diff
@@ -0,0 +1,18 @@
...
+  public constructor(private cancelled = false) {}
+
+  public cancel(): void {
+    this.cancelled = true;
+  }
+
+  public isCancelled(): boolean {
+    return this.cancelled;
+  }
+}
```
toger5:
```suggestion
export class AbortHandle {
  public constructor(private aborted = false) {}

  public abort(): void {
    this.aborted = true;
  }

  public isAborted(): boolean {
    return this.aborted;
  }
}
```

src/livekit/useECConnectionState.ts
```diff
@@ -247,6 +281,22 @@ export function useECConnectionState(
...
+  // This is a cleanup function that will be called when the component is unmounted.
+  // It will cancel all cancellables in the bag
+  useEffect(() => {
+    const bag = cancelBag.current;
+    return (): void => {
+      bag.forEach((cancellable) => {
+        cancellable.cancel();
+      });
+    };
+  }, []);
```
toger5:
```suggestion
  useEffect(() => () => {
      cancelBag.current.forEach((cancellable) => cancellable.cancel());
    };
  }, []);
```

toger5 (changes requested):
Looking good.
I think renaming would make sense. (If cancellable is from "able to cancel" its spelled wrong?)

src/livekit/useECConnectionState.ts
BillCarsonFr: It is on purpose to avoid a warning. When using `cancelBag.current` , technically the `current`could have changed by the time the cleanup function is called. Not really relevant in our case given it is only set once. (TS30010: [react-hooks/exhaustive-deps]

src/livekit/useECConnectionState.ts
```diff
@@ -284,15 +284,15 @@ export function useECConnectionState(
...
 
-  // This is a cleanup function that will be called when the component is unmounted.
-  // It will cancel all cancellables in the bag
+  // This is a cleanup function that will be called when the component is about to be unmounted.
+  // It will cancel all abortHandles in the bag
   useEffect(() => {
-    const bag = cancelBag.current;
+    const bag = abortHandlesBag.current;
     return (): void => {
       bag.forEach((cancellable) => {
```
toger5: can we also rename the variable and property to `abortHandle`

src/livekit/useECConnectionState.ts
toger5: this seems to not have worked

toger5 (approved):

PR #3261, opened 2025-05-12
Docker-compose: do not access ports <443

toger5:
This does not allow podman (also docker) to start the containers without altering linux default security measurements.

The playwright tests also run with those the alternative ports: 4443, 8080, 8081

### DIFF
--- a/backend/dev_nginx.conf
+++ b/backend/dev_nginx.conf
@@ -145,7 +145,7 @@ server {
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
       proxy_set_header X-Forwarded-Proto $scheme;
 
-      proxy_pass http://element-web:81;
+      proxy_pass http://element-web:8081;
       proxy_ssl_verify              off;
 
     }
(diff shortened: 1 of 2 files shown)


fkwp:
this PR is missing the required changes for `dev-backend-docker-compose.yml` to work properly as removing port 443 requires a bunch of config files to be adapted https://hostname becomes then https://hostname:port

fkwp:
As a workaround for podman

You can modify the `net.ipv4.ip_unprivileged_port_start` sysctl to change the lowest port. For example `sysctl net.ipv4.ip_unprivileged_port_start=443` allows rootless Podman containers to bind to ports >= 443.

toger5:
I updated the PR to still include the `:Z` flag for the file copy and allow to set the min port to 443. (instead of 80)

fkwp:
The .well-known URI can only be reached via HTTPS on the default 443 port. Due to this limitation the highest possible port is bound to 443

fkwp (approved):

toger5:
I actually did run into issues using the dev env in firefox (not playwright) not sure if that is an issue. I think some routing is done through port 80.
So eventually we need to update the nginx conf or also allow port 80?

toger5:
Lets merge this as is. In case there is another issue with the local lk server we can fix this in another PR.

PR #3266, opened 2025-05-13
Add delayed events to docs and playwridgt/dev server configs.

toger5:


fkwp:
msc4140 is enabled by setting `max_event_delay_duration` eg to `24h`

PR #3267, opened 2025-05-13
Update RTCSession configuration to non deprecated names.

toger5:
As introduced in: https://github.com/matrix-org/matrix-js-sdk/pull/4714

~~Is blocked until we bump the js-sdk to a version supporting the new names.~~

This pr now also updates the js-sdk version

### DIFF
--- a/src/e2ee/matrixKeyProvider.ts
+++ b/src/e2ee/matrixKeyProvider.ts
@@ -44,7 +44,7 @@ export class MatrixKeyProvider extends BaseKeyProvider {
     encryptionKeyIndex: number,
     participantId: string,
   ): void => {
-    createKeyMaterialFromBuffer(encryptionKey).then(
+    createKeyMaterialFromBuffer(encryptionKey.buffer as ArrayBuffer).then(
       (keyMaterial) => {
         this.onSetEncryptionKey(keyMaterial, participantId, encryptionKeyIndex);
 
--- a/src/e2ee/matrixKeyProvider.ts
+++ b/src/e2ee/matrixKeyProvider.ts
@@ -5,7 +5,7 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 Please see LICENSE in the repository root for full details.
 */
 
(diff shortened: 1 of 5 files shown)


src/e2ee/matrixKeyProvider.ts
```diff
@@ -44,7 +44,7 @@ export class MatrixKeyProvider extends BaseKeyProvider {
     encryptionKeyIndex: number,
     participantId: string,
   ): void => {
-    createKeyMaterialFromBuffer(encryptionKey).then(
+    createKeyMaterialFromBuffer(encryptionKey.buffer as ArrayBuffer).then(
```
BillCarsonFr: This warning is a bit annoying, a `.buffer` is not necessarly an ArrayBuffer? 
I wonder why we are having this warning only now... TBH I don't understand

At the end it is calling subtle crypto `importKey` that accepts An ArrayBuffer, a TypedArray, a DataView...

I don't know what is the best way to fix that.
We change our API to use ArrayBuffer? We call directly import key? or we request a change to `createKeyMaterialFromBuffer`

BillCarsonFr: So could be that:

```
   crypto.subtle.importKey('raw', encryptionKey, 'HKDF', false, [
      'deriveBits',
      'deriveKey',
    ]).then(
  ```
  
  If not what do we do if `encryptionKey.buffer` is not an ArrayBuffer? Maybe we could throw an error ?

BillCarsonFr (changes requested):
I am not sure what is the correct way to fix the ArrayBuffer thing... I put some suggestions

src/e2ee/matrixKeyProvider.ts
toger5: This seems like a good idea to me. I just added a commit patching this.

fkwp (approved):

toger5:
I just tried a to-device call with this and .dev and they seem to understand their keys (so we dont break the key format in a unexpected way. Which is the thing i was most concerned about)

PR #3270, opened 2025-05-14
Audio device controls for mobile native audio device selection

### LINKED ISSUE #3262: Implement Native Audio Switching for Element Call Widget on iOS/Android
### Background

Element Call is implemented on our mobile platform via a one-fits-all widget approach. Web views on mobile platforms (Android, iOS) do not support properly selecting an audio output device from within the web view. Those mobile platforms "try" to find a "smart" auto selection. 

### Plan

In order to workaround the above constraints, on mobile we do not want to change the audio output/input via the web view, but instead do this all natively through the widget hosting native messenger app.

We still want that the control/change action of input/outbut audio lives in the widget / web view. In order to allow this we use js<->swift/cotlin web ffi calls.

Note the operating system can also change the audio input/output devices which is also signaled to the widget.

See FFI calls defined in: [controlledOutput](https://github.com/element-hq/element-call/blob/f69c75322f0e247fd42a9d2aa3e24d4b6bb959be/docs/controls.md)

The implementation wrt. Element Call widget is carried out in: https://github.com/element-hq/element-call/pull/3270

Current state:
 - [x] Element Call https://github.com/element-hq/element-call/pull/3270 
 - [x] iOS
   - [x] set the url flag to use controlled audio
   - [x] audio output works (in general. the call audio is audible)
   - [x] changing the output in the os (wither by connecting bt headphones or by using a native dialog) also updates the ui in EC
   - [x] (ios specific) Earpiece mode is only delectable when on default audio output device
   - [x] (ios specific) Earpiece mode works as good as it can.
   - [x] changing the output in app also changes the actual audible output. (on ios this means that the button in the ui forwards us to the native picker)
   - [x]  initial list of devices is correct (on ios this is only about showing the earpice when starting with speaker mode)
   - [x] all related PRs merged
       - https://github.com/element-hq/element-x-ios/pull/4136
 - [x] android
   - [x] set the url flag to use controlled audio
   - [x] audio output works (in general. the call audio is audible)
   - [x] changing the output in the os (wither by connecting bt headphones or by using a native dialog) also updates the ui in EC
   - [x] changing the output in app also changes the actual audible output. 
   - [x] initial list of devices is correct (we had issues with the timing of sending over the initial devices)
   - [x] all related PRs merged
       - [x] https://github.com/element-hq/element-x-android/pull/4663

toger5:
Fixes: https://github.com/element-hq/element-call/issues/3262
See: [controlledOutput](https://github.com/element-hq/element-call/blob/f69c75322f0e247fd42a9d2aa3e24d4b6bb959be/docs/controls.md)


docs/controls.md
```diff
@@ -1,7 +1,18 @@
...
 - `controls.canEnterPip(): boolean` Determines whether it's possible to enter picture-in-picture mode.
 - `controls.enablePip(): void` Puts the call interface into picture-in-picture mode. Throws if not in a call.
 - `controls.disablePip(): void` Takes the call interface out of picture-in-picture mode, restoring it to its natural display mode. Throws if not in a call.
+
+## Audio output devices
+
+These functions must be used in conjunction with the `controlledOutput` URL parameter in order to have any effect.
+
+- `controls.setOutputDevices(devices: { id: string, name: string, forEarpiece?: boolean }[]): void` Sets the list of available audio outputs. `forEarpiece` is used on ios only.
+  It flags the device that should be used if the user selects earpice mode. This should be the main (stereo loudspeaker) of the device.
```
robintown:
```suggestion
  It flags the device that should be used if the user selects earpiece mode. This should be the main stereo loudspeaker of the device.
```

src/UrlParams.ts
```diff
@@ -281,6 +284,8 @@ export const getUrlParams = (
     fontScale: Number.isNaN(fontScale) ? null : fontScale,
     allowIceFallback: parser.getFlagParam("allowIceFallback"),
     perParticipantE2EE: parser.getFlagParam("perParticipantE2EE"),
+    // TODO this should not default to true!
+    controlledOutput: parser.getFlagParam("controlledMediaDevices", true),
```
robintown: Leaving this comment here so we can't forget the TODO

src/App.tsx
```diff
@@ -72,26 +91,15 @@ export const App: FC = () => {
...
-                  </MediaDevicesProvider>
+                  <ProcessorProvider>
+                    {controlledOutput ? (
+                      <ControlledOutputMediaDevicesProvider>
+                        {inner}
+                      </ControlledOutputMediaDevicesProvider>
+                    ) : (
+                      <MediaDevicesProvider>{inner}</MediaDevicesProvider>
+                    )}
+                  </ProcessorProvider>
```
robintown: As I hinted at the other week I think we can do

```diff
@@ -67,19 +67,7 @@ export const App: FC = () => {
       .catch(logger.error);
   });
 
-  const inner = (
-    <Sentry.ErrorBoundary
-      fallback={(error) => <ErrorPage error={error} widget={widget} />}
-    >
-      <DisconnectedBanner />
-      <Routes>
-        <SentryRoute path="/" element={<HomePage />} />
-        <SentryRoute path="/login" element={<LoginPage />} />
-        <SentryRoute path="/register" element={<RegisterPage />} />
-        <SentryRoute path="*" element={<RoomPage />} />
-      </Routes>
-    </Sentry.ErrorBoundary>
-  );
+  const MediaDevicesProviderComponent = controlledOutput ? ControlledOutputMediaDevicesProvider : MediaDevicesProvider
 
   return (
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
@@ -92,13 +80,19 @@ export const App: FC = () => {
               <Suspense fallback={null}>
                 <ClientProvider>
                   <ProcessorProvider>
-                    {controlledOutput ? (
-                      <ControlledOutputMediaDevicesProvider>
-                        {inner}
-                      </ControlledOutputMediaDevicesProvider>
-                    ) : (
-                      <MediaDevicesProvider>{inner}</MediaDevicesProvider>
-                    )}
+                    <MediaDevicesProviderComponent>
+                      <Sentry.ErrorBoundary
+                        fallback={(error) => <ErrorPage error={error} widget={widget} />}
+                      >
+                        <DisconnectedBanner />
+                        <Routes>
+                          <SentryRoute path="/" element={<HomePage />} />
+                          <SentryRoute path="/login" element={<LoginPage />} />
+                          <SentryRoute path="/register" element={<RegisterPage />} />
+                          <SentryRoute path="*" element={<RoomPage />} />
+                        </Routes>
+                      </Sentry.ErrorBoundary>
+                    </MediaDevicesProviderComponent>
                   </ProcessorProvider>
                 </ClientProvider>
               </Suspense>
```

and then it becomes a bit cleaner as `inner` remains inline.

src/room/InCallView.tsx
```diff
@@ -222,7 +223,9 @@ export const InCallView: FC<InCallViewProps> = ({
     room: livekitRoom,
   });
 
-  const [muteAllAudio] = useSetting(muteAllAudioSetting);
+  const muteAllAudioControlled = useObservableEagerState(setOutputEnabled$);
+  const [muteAllAudioFromSetting] = useSetting(muteAllAudioSetting);
+  const muteAllAudio = muteAllAudioControlled || muteAllAudioFromSetting;
```
robintown: I would expect this `setOutputEnabled$` subject to be consumed by the CallViewModel, piped through `startWith` to give it an initial value, combined with the setting there, and exposed as a single state observable. (I'm actually surprised if this isn't crashing the InCallView component currently because `setOutputEnabled$`, being a subject, doesn't provide an initial value.)

Tangentially, I think you forgot a `!`:

```suggestion
  const muteAllAudioControlled = !useObservableEagerState(setOutputEnabled$);
  const [muteAllAudioFromSetting] = useSetting(muteAllAudioSetting);
  const muteAllAudio = muteAllAudioControlled || muteAllAudioFromSetting;
```

src/room/GroupCallView.tsx
```diff
@@ -104,8 +106,9 @@ export const GroupCallView: FC<Props> = ({
   const [externalError, setExternalError] = useState<ElementCallError | null>(
     null,
   );
-
-  const [muteAllAudio] = useSetting(muteAllAudioSetting);
+  const muteAllAudioControlled = useObservableEagerState(setOutputEnabled$);
+  const [muteAllAudioFromSetting] = useSetting(muteAllAudioSetting);
+  const muteAllAudio = muteAllAudioControlled || muteAllAudioFromSetting;
```
robintown: Oh gosh, I guess this usage of `setOutputEnabled$` in the GroupCallView shows that, if we turn this into a state observable, that state observable needs to live not on the CallViewModel, but rather something a bit more global than that. Since we don't yet have anything outside the InCallView refactored to use MVVM, I suppose accepting a bit of duplication or putting the `muteAllAudio` observable somewhere in the global/module scope would be fine with me.

src/livekit/MediaDevicesContext.tsx
```diff
@@ -59,12 +60,26 @@ export interface MediaDevice {
...
   videoInput: MediaDevice;
   startUsingDeviceNames: () => void;
   stopUsingDeviceNames: () => void;
+  usingNames: boolean;
+}
+
+export interface MediaDevices extends Omit<InputDevices, "usingNames"> {
+  audioOutput: MediaDevice;
+}
+function useShowEarpiece(): boolean {
```
robintown:
```suggestion
}

function useShowEarpiece(): boolean {
```

src/livekit/MediaDevicesContext.tsx
```diff
@@ -59,12 +60,26 @@ export interface MediaDevice {
...
+  const [alwaysShowIphoneEarpice] = useSetting(alwaysShowIphoneEarpieceSetting);
+  const m = useMemo(
+    () =>
+      (navigator.userAgent.match("iPhone")?.length ?? 0) > 0 ||
```
robintown: Is there a reason we can't do this here?

```suggestion
      navigator.userAgent.includes("iPhone") ||
```

src/livekit/MediaDevicesContext.tsx
```diff
@@ -74,7 +89,7 @@ function useMediaDevice(
 ): MediaDevice {
   // Make sure we don't needlessly reset to a device observer without names,
   // once permissions are already given
```
robintown: This comment belongs with the `hasRequestedPermissions` line

src/livekit/MediaDevicesContext.tsx
```diff
@@ -249,6 +253,144 @@ export const MediaDevicesProvider: FC<Props> = ({ children }) => {
...
+        map((devices) => {
+          const devicesMap = new Map<string, DeviceLabel>(
+            devices.map(({ id, name }) => [id, { type: "name", name }]),
+          );
+          if (showEarpiece)
+            devicesMap.set(EARPIECE_CONFIG_ID, { type: "earpiece" });
+          return devicesMap;
+        }),
+      ),
+    ),
```
robintown: It's not safe to reference `showEarpiece` here inside `useObservable` due to https://observable-hooks.js.org/guide/gotchas.html#epic-closure

Instead you could pass `showEarpiece` to the inputs array of `useObservable` like so:

```suggestion
    useObservable((showEarpiece$) =>
      combineLatest([setOutputDevices$.pipe(startWith<OutputDevice[]>([])), showEarpiece$], (devices, showEarpiece) => {
        const devicesMap = new Map<string, DeviceLabel>(
          devices.map(({ id, name }) => [id, { type: "name", name }]),
        );
        if (showEarpiece)
          devicesMap.set(EARPIECE_CONFIG_ID, { type: "earpiece" });
        return devicesMap;
      }),
      [showEarpiece],
    ),
```

Or you could even convert `useShowEarpiece` from a React hook into an Observable, if that feels nicer.

Oh, or I suppose replacing `useObservable` with `useMemo` would work too. Lots of options :)

src/livekit/MediaDevicesContext.tsx
```diff
@@ -249,6 +253,144 @@ export const MediaDevicesProvider: FC<Props> = ({ children }) => {
...
+  );
+  const earpiceDevice = useObservableEagerState(
```
robintown:
```suggestion
  const earpieceDevice = useObservableEagerState(
```

src/livekit/MediaDevicesContext.tsx
```diff
@@ -249,6 +253,144 @@ export const MediaDevicesProvider: FC<Props> = ({ children }) => {
...
+        window.controls.onOutputDeviceSelect?.(selectedId);
+  }, [selectedId]);
+
+  const [asEarpice, setAsEarpiece] = useState(false);
+
+  const select = useCallback(
+    (id: string) => {
+      if (id === EARPIECE_CONFIG_ID) {
+        setAsEarpiece(true);
+        if (earpiceDevice) setPreferredId(earpiceDevice.id);
```
robintown: I feel like we should log a warning or something if there is no `earpieceDevice`

src/livekit/MediaDevicesContext.tsx
```diff
@@ -249,6 +253,144 @@ export const MediaDevicesProvider: FC<Props> = ({ children }) => {
...
+      // Worst case it is not available and the browser will gracefully fall
+      // back to some other device for us when requesting the media stream.
+      // Otherwise, select the first available device.
+      return (preferredId !== undefined && available.has(preferredId)) ||
+        (available.size === 1 && available.has(""))
+        ? preferredId
+        : available.keys().next().value;
+    }
+    return undefined;
+  }, [available, preferredId]);
```
robintown: This duplicates some code earlier in this file - could be turned into a reusable function

src/livekit/MediaDevicesContext.tsx
```diff
@@ -249,6 +253,144 @@ export const MediaDevicesProvider: FC<Props> = ({ children }) => {
...
+
+  useEffect(() => {
+    if (selectedId === EARPIECE_CONFIG_ID)
+      if (selectedId !== undefined)
+        window.controls.onOutputDeviceSelect?.(selectedId);
```
robintown: Why does this only happen if `selectedId === EARPIECE_CONFIG_ID`?

src/livekit/MediaDevicesContext.tsx
toger5: I ended up witht he combine latest approach

src/livekit/MediaDevicesContext.tsx
toger5: good idea.
I also removed all the earpiece logic from the non controlled logic branch (It will only be used on ios where we also use controlled outputs)

src/livekit/MediaDevicesContext.tsx
toger5: yes needs to be done in any case. Updated it so that it will be done with the correct id but in dependent of the `selectedId === EARPIECE_CONFIG_ID` condition

src/App.tsx
toger5: Do you think the solution with just one context is good as well?

src/room/InCallView.tsx
toger5: I am not sure this is possible since we need the mute info also in the groupCallView

src/room/GroupCallView.tsx
toger5: oh I see.

toger5: I moved this into a file called `MuteAllAudioModel.ts` With a comment that this might converge into a `GroupCallViewModel.ts`.
If that is what we think is appropritate. otherwise we can move it into a place where it fits once we have such a place.

src/livekit/MediaDevicesContext.tsx
toger5: With the recent developments in how we need to solve it on ios we actually need the case that there is no earpiece device.

toger5: If bt headphones are selected the user first has to use the native picker to go back to speaker and then the earpice can be selected.
Modeling it with earpice being available in the list or not is the most elegant we can do I think .

src/room/InCallView.tsx
toger5: moved it into its own file and fixed the `!` mistake.

src/UrlParams.ts
toger5: We have: https://github.com/matrix-org/matrix-rust-sdk/pull/5049 now se I remove the wrong default value.

docs/controls.md
```diff
@@ -1,7 +1,21 @@
...
 
 - `controls.canEnterPip(): boolean` Determines whether it's possible to enter picture-in-picture mode.
 - `controls.enablePip(): void` Puts the call interface into picture-in-picture mode. Throws if not in a call.
 - `controls.disablePip(): void` Takes the call interface out of picture-in-picture mode, restoring it to its natural display mode. Throws if not in a call.
+
+## Audio output devices
+
+These functions must be used in conjunction with the `controlledMediaDevices` URL parameter in order to have any effect.
+
+- `controls.setAvailableOutputDevices(devices: { id: string, name: string, forEarpiece?: boolean, isEarpiece?: boolean isSpeaker?: boolean, isExternalHeadset?, boolean;}[]): void` Sets the list of available audio outputs. `forEarpiece` is used on ios only.
```
robintown:
```suggestion
- `controls.setAvailableOutputDevices(devices: { id: string, name: string, forEarpiece?: boolean, isEarpiece?: boolean isSpeaker?: boolean, isExternalHeadset?, boolean; }[]): void` Sets the list of available audio outputs. `forEarpiece` is used on iOS only.
```

docs/controls.md
```diff
@@ -1,7 +1,21 @@
...
+  It flags the device that should be used if the user selects earpiece mode. This should be the main stereo loudspeaker of the device.
+- `controls.onOutputDeviceSelect: ((id: string) => void) | undefined` Callback called whenever the user or application selects a new audio output.
+- `controls.setOutputDevice(id: string): void` Sets the selected audio device in EC menu. This should be used if the os decides to automatically switch to bluetooth.
```
robintown:
```suggestion
- `controls.setOutputDevice(id: string): void` Sets the selected audio device in Element Call's menu. This should be used if the OS decides to automatically switch to Bluetooth, for example.
```

Generally I'm trying to keep these docs in EC descriptive of how the APIs *could* be used rather than how downstream projects do happen to use them currently.

docs/controls.md
```diff
@@ -1,7 +1,21 @@
...
+- `controls.setOutputEnabled(enabled: boolean)` Enables/disables all audio output from the application. This can be useful for temporarily pausing audio while the controlling application is switching output devices. Output is enabled by default.
```
robintown: Just curious: do we still need this? I thought I remembered you saying that disabling audio output didn't actually help, in testing - not sure if I'm recalling that correctly

docs/controls.md
```diff
@@ -1,7 +1,21 @@
...
+- `showNativeOutputDevicePicker: () => void`. This callback will be code by the webview if the user presses the output button in the settings menu.
+  This button is only shown on ios. (`userAgent.includes("IPhone")`)
```
robintown:
```suggestion
- `showNativeOutputDevicePicker: (() => void) | undefined`. Callback called whenever the user presses the output button in the settings menu.
  This button is only shown on iOS. (`userAgent.includes("iPhone")`)
```

To align with the other callback docs

src/controls.ts
```diff
@@ -5,15 +5,49 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
+  isSpeaker?: boolean;
+  isExternalHeadset?: boolean;
+}
+
+/**
+ * If pipMode is enabled, EC will render a adapted call view layout.
+ */
 export const setPipEnabled$ = new Subject<boolean>();
+// BehaviorSubject since the client might set this before we have subscribed (GroupCallView still in "loading" state)
+// We want the that has been set during loading to be be available immediately once loaded.
```
robintown:
```suggestion
// We want the devices that have been set during loading to be available immediately once loaded.
```

src/controls.ts
```diff
@@ -5,15 +5,49 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
+export const setAvailableOutputDevices$ = new BehaviorSubject<OutputDevice[]>(
+  [],
+);
+// BehaviorSubject since the client might set this before we have subscribed (GroupCallView still in "loading" state)
+// We want the that has been set during loading to be be available immediately once loaded.
```
robintown:
```suggestion
// We want the device that has been set during loading to be available immediately once loaded.
```

src/controls.ts
```diff
@@ -5,15 +5,49 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
+  isExternalHeadset?: boolean;
+}
+
+/**
+ * If pipMode is enabled, EC will render a adapted call view layout.
+ */
 export const setPipEnabled$ = new Subject<boolean>();
+// BehaviorSubject since the client might set this before we have subscribed (GroupCallView still in "loading" state)
+// We want the that has been set during loading to be be available immediately once loaded.
+export const setAvailableOutputDevices$ = new BehaviorSubject<OutputDevice[]>(
```
robintown: As long as these are BehaviorSubjects (representing stateful values rather than events), I think it makes sense to name these like values:

```suggestion
export const availableOutputDevices$ = new BehaviorSubject<OutputDevice[]>(
```

src/controls.ts
```diff
@@ -5,15 +5,49 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
+  [],
+);
+// BehaviorSubject since the client might set this before we have subscribed (GroupCallView still in "loading" state)
+// We want the that has been set during loading to be be available immediately once loaded.
+export const setOutputDevice$ = new BehaviorSubject<string | undefined>(
```
robintown:
```suggestion
export const outputDevice$ = new BehaviorSubject<string | undefined>(
```

src/controls.ts
```diff
@@ -5,15 +5,49 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
+  undefined,
+);
+/**
+ * This is currently unused. It might be possible to allow the os to mute the call this way if the user
+ * presses the volume down button when it is at the minimum volume.
+ *
+ * This should also be used to display a darkened overlay screen letting the user know that audio is muted.
+ */
+export const setOutputEnabled$ = new Subject<boolean>();
```
robintown: This comment seems inaccurate - isn't this implemented in the MuteAllAudioModel thing?

src/controls.ts
```diff
@@ -27,4 +61,17 @@ window.controls = {
...
+  },
+  setOutputDevice(id: string): void {
+    setOutputDevice$.next(id);
+  },
+  setOutputEnabled(enabled: boolean): void {
+    if (!setOutputEnabled$.observed)
+      throw new Error(
+        "Output controls are disabled. No setOutputEnabled$ observer",
+      );
+    setOutputEnabled$.next(!enabled);
```
robintown:
```suggestion
    setOutputEnabled$.next(enabled);
```

src/livekit/MediaDevicesContext.tsx
```diff
@@ -59,24 +63,73 @@ export interface MediaDevice {
...
+ *  - hide any input devices (they do not work anyhow on ios)
+ *  - Show a button to show the native output picker instead.
+ *  - Only show the earpiece toggle option if the earpiece is available:
+ *   `setAvailableOutputDevices$.includes((d)=>d.forEarpiece)`
+ */
+export const iosDeviceMenu$ = alwaysShowIphoneEarpieceSetting.value$.pipe(
+  startWith(
+    alwaysShowIphoneEarpieceSetting.getValue() ||
+      navigator.userAgent.includes("iPhone"),
+  ),
```
robintown: This `startWith` should be redundant since `Setting["value$"]` refers to a behavior-like observable (an observable that emits immediately upon subscription)

src/livekit/MediaDevicesContext.tsx
```diff
@@ -59,24 +63,73 @@ export interface MediaDevice {
...
+    }
+    return undefined;
+  }, [available, preferredId]);
+}
+
+/**
+ * Hook to get access to a mediaDevice handle for a kind. This allows to list
+ * the available devices, read and set the selected device.
+ * @param kind audio input, output or video output.
+ * @param setting The setting this handles selection should be synced with.
```
robintown:
```suggestion
 * @param kind Audio input, output or video output.
 * @param setting The setting this handle's selection should be synced with.
```

src/livekit/MediaDevicesContext.tsx
```diff
@@ -273,6 +317,73 @@ export const MediaDevicesProvider: FC<Props> = ({ children }) => {
...
+          );
+          return {
+            devicesMap: new Map<string, DeviceLabel>(deviceMapTuple),
+            deviceForEarpiece,
+          };
+        }),
+      );
+
+      return combineLatest([outputDeviceData$, iosDeviceMenu$]).pipe(
+        map(([{ devicesMap, deviceForEarpiece }, iosShowEarpiece]) => {
```
robintown:
```suggestion
      return combineLatest([outputDeviceData$, iosDeviceMenu$], ({ devicesMap, deviceForEarpiece }, iosShowEarpiece) => {
```

src/livekit/MediaDevicesContext.tsx
```diff
@@ -273,6 +317,73 @@ export const MediaDevicesProvider: FC<Props> = ({ children }) => {
...
+        }),
+      );
+    }),
+  );
+  const [preferredId, setPreferredId] = useSetting(audioOutputSetting);
+  useEffect(() => {
+    setOutputDevice$.subscribe((id) => {
+      if (id) setPreferredId(id);
+    });
+  }, [setPreferredId]);
```
robintown: This effect needs a clean-up function which unsubscribes

src/livekit/MediaDevicesContext.tsx
```diff
@@ -273,6 +317,73 @@ export const MediaDevicesProvider: FC<Props> = ({ children }) => {
...
+  const selectedId = useSelectedId(available, preferredId);
+
+  const [asEarpiece, setAsEarpiece] = useState(false);
+
+  useEffect(() => {
+    // In earpiece mode we just sent the EARPIECE_CONFIG_ID to the native code
+    // This only happens on ios where we use the native picker.
+    // So this only is needed so that ios can know if the proximity sensor should be used or not.
+    if (selectedId) window.controls.onOutputDeviceSelect?.(selectedId);
+    setAsEarpiece(selectedId === EARPIECE_CONFIG_ID);
```
robintown: I'm finding these comments hard to follow since they're mixing a description of what EC does with a description of what downstream applications do. Is something like this still accurate?

```suggestion
    // Let the hosting application know which output device has been selected.
    // This information is probably only of interest if the earpiece mode has been
    // selected - for example, Element X iOS listens to this to determine whether it
    // should enable the proximity sensor.
    if (selectedId) window.controls.onOutputDeviceSelect?.(selectedId);
    setAsEarpiece(selectedId === EARPIECE_CONFIG_ID);
```

src/settings/SettingsModal.tsx
```diff
@@ -101,20 +104,36 @@ export const SettingsModal: FC<Props> = ({
   const [showDeveloperSettingsTab] = useSetting(developerMode);
 
   const { available: isRageshakeAvailable } = useSubmitRageshake();
+  // If we are on ios we will show a button to open the native picker.
```
robintown:
```suggestion
  // If we are on iOS we will show a button to open the native audio device picker.
```

src/settings/SettingsModal.tsx
```diff
@@ -101,20 +104,36 @@ export const SettingsModal: FC<Props> = ({
...
+  const iosDeviceMenu = useObservableEagerState(iosDeviceMenu$);
+  // In controlled devices we will not show the input section
```
robintown: How come? Wouldn't the `iosDeviceMenu` condition be more appropriate here?

src/state/MuteAllAudioModel.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * This can transition into sth more complete: `GroupCallViewModel.ts`
+ */
+export const muteAllAudio$ = combineLatest([
+  setOutputEnabled$,
+  muteAllAudioSetting.value$,
+]).pipe(
+  startWith([true, muteAllAudioSetting.getValue()]),
+  map(([outputEnabled, settingsMute]) => !outputEnabled || settingsMute),
+);
```
robintown:
```suggestion
export const muteAllAudio$ = combineLatest([
  setOutputEnabled$.pipe(startWith(true)),
  muteAllAudioSetting.value$,
], (outputEnabled, settingsMute) => !outputEnabled || settingsMute);
```

Since the setting is already behavior-like

src/state/MuteAllAudioModel.test.ts
```diff
@@ -0,0 +1,36 @@
...
+*/
+
+import { test, vi } from "vitest";
+import { expect } from "vitest";
+
+import { setOutputEnabled$ } from "../controls";
+import { muteAllAudio as muteAllAudioSetting } from "../settings/settings";
+import { muteAllAudio$ } from "./MuteAllAudioModel";
+
+test("muteAllAudio$", () => {
```
robintown, line 15: Not a blocker, but this would be more idiomatic as a marble test

robintown (commented):
Sorry, still lots of comments, though it's mostly docs stuff

docs/controls.md
toger5: Yes that is correct. We do have another possible usecase however.
Since we might want the audio mute button, there is a chance we want to enable the audio again if the user presses the hardware volume up button. Also we might want a "hardware down + lowest volume X times" to mute the audio.
This is a simple enough interface control this.
Once could still argue that this can be dropped from this pr.

src/livekit/MediaDevicesContext.tsx
toger5: I got tricked by it being stored in an Observable. Which I now notice is just a trick to hide the `.next` "write" api and making this a readonly behaviour subject basically.
But since it references to the BahaviourSubject its still inheriting the "emit on subscription" property.

src/livekit/MediaDevicesContext.tsx
toger5: very accurate and undestandable to me!

src/settings/SettingsModal.tsx
toger5: There are 3 cases:
 - normal (input:yes + output:yes)
 - ios (input:no, output: menu button + earpice picker vs speaker in output)
 - android (input: no ouput:yes)

 we model them with the two bools iosDeviceMenu, controlledMediaDevices. where one is responsible for the input: yes/no and the other on for output: default/custom button ios thing

docs/controls.md
robintown: Makes sense to me, I suggest we leave it in and just drop the sentence "This can be useful for temporarily pausing audio while the controlling application is switching output devices" then.

src/settings/SettingsModal.tsx
robintown: Well, `controlledMediaDevices` relates to the output devices rather than the input devices, right? So something seems off here.

> - android (input: no ouput:yes)

Why do we want hide the input devices on Android, out of interest?

src/UrlParams.ts
```diff
@@ -124,9 +124,15 @@ export interface UrlParams {
...
-   * Whether we the app should use per participant keys for E2EE.
+   * Whether the app should use per participant keys for E2EE.
    */
   perParticipantE2EE: boolean;
+  /**
+   * Whether the global JS controls for audio output devices should be enabled,
+   * allowing the list of output devices to be controlled by the app hosting
+   * Element Call.
+   */
+  controlledMediaDevices: boolean;
```
robintown: By the way, maybe this should instead be named `controlledOutputDevices` to reflect its behavior? I think I'm responsible for that mis-naming.

src/settings/SettingsModal.tsx
toger5: The controlled media devices bool value is responsible to control whether the input is shown. If it is true we do not show the input if it is false we do show the input selection.

So yes in controller media devices input is never shown and that is how/why they define the input visibilty

toger5: On Android we hide the input because it also uses controlled media devices and the os does not expose input selection. It combines a mic and a speaker into one media device. So we change both at once and output is easier to understand. (Main speaker vs bluetooth device in input is confusing. But in output it makes a lot of sense and that the input is switched at the same time is hopefully expected and hence not too confusing for the user.)

robintown: I think I get it now: for the use cases in which `mediaDeviceControls` would be active, we actually expect both the audio input devices and audio output devices to reflect the very same device list? If so, I would suggest renaming the URL parameter to `audioDeviceControls` and likewise changing the `controls` API names & docs to reference "audio input/output devices" (or maybe just "audio devices") as a unified concept, rather than specifically "audio output devices".

I also suggest you add a code comment with the explanation you just gave for why the devices are easier to understand as output devices, in the UI - I found that helpful.

src/UrlParams.ts
robintown: Different renaming suggested at https://github.com/element-hq/element-call/pull/3270#discussion_r2101824073

src/state/MuteAllAudioModel.test.ts
toger5: I agree. Since it s not a blocker I prefer postponing this to a later PR.

src/settings/SettingsModal.tsx
toger5: The rename also makes sense but I dont see to much value in the output vs media choice. Media is very vague so it could also contain input (and even video tbh) so I think it is acceptable.
We also have rust and mobile code that already uses the naming: `controlledMediaDevices`.
I also think `controlled` is a good name for the url parameter since that explains the best what is actually happing, that the devices are controlled. `mediaDeviceControls` = `false` could be read as: There wont be any ui to control the media. (which is not true. The media devices will just not be controlled from the outside but will interact with the webview directly)

robintown: Ah right, `controlledAudioDevices` was what I meant to suggest.

> Media is very vague so it could also contain input (and even video tbh)

That last bit, that it could apply to video devices, is how I see `controlledAudioDevices` could be an improvement. What do you think / how set-in-stone is the Rust and mobile code?

With the other names, I kind of just find them misleading if they only mention outputs. IMO we should make their usage expectations (that input + output are coupled) be reflected in their names while we have the chance:

```typescript
setAvailableAudioDevices(devices: AudioDevice[]): void;
setAudioDevice(id: string): void;
onAudioDeviceSelect?: (id: string) => void;
setAudioEnabled(enabled: boolean): void;
showNativeAudioDevicePicker?: () => void;
```

docs/controls.md
```diff
@@ -1,7 +1,21 @@
...
+
+## Picture-in-picture
 
 - `controls.canEnterPip(): boolean` Determines whether it's possible to enter picture-in-picture mode.
 - `controls.enablePip(): void` Puts the call interface into picture-in-picture mode. Throws if not in a call.
 - `controls.disablePip(): void` Takes the call interface out of picture-in-picture mode, restoring it to its natural display mode. Throws if not in a call.
+
+## Audio output devices
+
+These functions must be used in conjunction with the `controlledMediaDevices` URL parameter in order to have any effect.
```
robintown: Florian suggests this say:

> On mobile platforms (iOS, Android), web views do not reliably support selecting audio output devices such as the main speaker, earpiece, or headset. To work around this limitation, the following functions allow the hosting application (e.g., Element Web, Element X) to take control of audio device management via exposed JavaScript interfaces. These functions must be enabled by the URL parameter `controlledAudioDevices` in order to have any effect.

robintown (approved):

docs/controls.md
```diff
@@ -1,7 +1,21 @@
...
+
+- `controls.setAvailableAudioDevices(devices: { id: string, name: string, forEarpiece?: boolean, isEarpiece?: boolean isSpeaker?: boolean, isExternalHeadset?, boolean; }[]): void` Sets the list of available audio outputs. `forEarpiece` is used on iOS only.
+  It flags the device that should be used if the user selects earpiece mode. This should be the main stereo loudspeaker of the device.
+- `controls.onAudioDeviceSelect: ((id: string) => void) | undefined` Callback called whenever the user or application selects a new audio output.
+- `controls.setAudioDevice(id: string): void` Sets the selected audio device in Element Call's menu. This should be used if the OS decides to automatically switch to Bluetooth, for example.
+- `controls.setAudioEnabled(enabled: boolean)` Enables/disables all audio output from the application. This can be useful for temporarily pausing audio while the controlling application is switching output devices. Output is enabled by default.
```
robintown:
```suggestion
- `controls.setAudioEnabled(enabled: boolean)` Enables/disables all audio output from the application. Output is enabled by default.
```

I still suggest we drop this (since it won't be used that way after all)

toger5: okay cool.

docs/url-params.md
```diff
@@ -63,6 +63,7 @@ These parameters are relevant to both [widget](./embedded-standalone.md) and [st
...
+| `controlledAudioDevices`                                | `true` or `false`                                                                                    | No, defaults to `false`                                                                           | No, defaults to `false`      | Whether the [global JS controls for audio output devices](./controls.md#audio-devices) should be enabled, allowing the list of output devices to be controlled by the app hosting Element Call.                                                                                                                                                                                                     |
```
robintown: This should reference "audio devices" or "audio input/output devices" rather than "audio output devices"

docs/controls.md
```diff
@@ -1,7 +1,21 @@
...
-A few aspects of Element Call's interface can be controlled through a global API on the `window`:
+A few aspects of Element Call's interface can be controlled through a global API on the `window`.
+
+## Picture-in-picture
 
 - `controls.canEnterPip(): boolean` Determines whether it's possible to enter picture-in-picture mode.
 - `controls.enablePip(): void` Puts the call interface into picture-in-picture mode. Throws if not in a call.
 - `controls.disablePip(): void` Takes the call interface out of picture-in-picture mode, restoring it to its natural display mode. Throws if not in a call.
+
+## Audio output devices
```
robintown: Same here

robintown (changes requested):

docs/url-params.md
```diff
@@ -63,6 +63,7 @@ These parameters are relevant to both [widget](./embedded-standalone.md) and [st
...
+| `controlledAudioDevices`                                | `true` or `false`                                                                                    | No, defaults to `false`                                                                           | No, defaults to `false`      | Whether the [global JS controls for audio output devices](./controls.md#audio-devices) should be enabled, allowing the list of audio devices to be controlled by the app hosting Element Call.                                                                                                                                                                                                     |
```
robintown:
```suggestion
| `controlledAudioDevices`                                | `true` or `false`                                                                                    | No, defaults to `false`                                                                           | No, defaults to `false`      | Whether the [global JS controls for audio devices](./controls.md#audio-devices) should be enabled, allowing the list of audio devices to be controlled by the app hosting Element Call.                                                                                                                                                                                                     |
```

robintown (approved):

PR #3283, opened 2025-05-22
Add nginx upstream group of servers for JWT Service

fkwp:
Adding nginx upstream servers for the JWT service including the docker-compose provided JWT-Service as well as a locally running one outside of docker-compose. This helps with developing the JWT Service while having the whole MatrixRTC stack running via docker-compose.

In order to start the MatrixRTC stack with out the JWT-Service you can run:
`docker-compose -f ./dev-backend-docker-compose.yml up nginx synapse redis livekit`


toger5 (approved):
jwt on 60xx portspace makes sense.

PR #3284, opened 2025-05-22
Disable device switching when in controlled audio devices mode

robintown:


toger5:
Am I right in assuming, that the only reason for this being a draft, is that we have the custom js-sdk branch as a dep?

fkwp:
> Am I right in assuming, that the only reason for this being a draft, is that we have the custom js-sdk branch as a dep?

yup -- I think that is the only reason after a test with a real device confirmed its working.

toger5:
Blocked until the underlying PR is merged. (we theoretically could merge with the PR branch but lets try getting the underlying js-sdk pr merged first.)

fkwp (approved):

PR #3293, opened 2025-05-27
Fix creating two lk rooms if there is no local store setup (fixes a resulting disconnect bug)

toger5:


src/main.tsx
```diff
@@ -29,7 +29,7 @@ window.setLKLogLevel = setLKLogLevel;
 initRageshake().catch((e) => {
   logger.error("Failed to initialize rageshake", e);
 });
-setLKLogLevel("warn");
+setLKLogLevel("info");
```
BillCarsonFr, line 32: Agree, info is engouh to get the livekit room logs we want

BillCarsonFr (approved):

PR #3295, opened 2025-05-27
Only use ios platform logic when in controlled media device condition

toger5:


robintown (approved):

toger5:
~~I am not exactly sure anymore we want/need this. It does make sense but on the other hand it is also useful to have the native button in the SPA on iphone. At least as long as we still have the SPA.~~
We do since the button relies on controlled devices. So even though it would be nice to have this in spa it just is not possible -> we should not show the button.

PR #3299, opened 2025-05-28
Fix some correctness issues in useLocalStorage

robintown:
See the commits for the full details. I discovered these issues while looking back over https://github.com/element-hq/element-call/pull/3293 with Timo:

- It seems 6258bcec549c1b743601dff5e2c4869f97de8e61 *does* represent a race condition that we're hitting in practice with the `useRoomEncryptionSystem` shared secret logic. https://github.com/element-hq/element-call/pull/3293, as I understand it, causes us to depend on the existence of the race. But really we ought to fix the race (which this commit does), and also fix `useRoomEncryptionSystem` to not depend on it (PR from Timo is https://github.com/element-hq/element-call/pull/3302; this PR will be blocked until then).
- We haven't yet triggered the failure mode addressed by ccf168cadd58c3047e1a6ce178d415337cbd140b in practice, I think; this is more of a theoretical correctness thing.

### DIFF
--- a/src/useLocalStorage.test.tsx
+++ b/src/useLocalStorage.test.tsx
@@ -7,7 +7,8 @@ Please see LICENSE in the repository root for full details.
 
 import { test } from "vitest";
 import { render, screen } from "@testing-library/react";
-import { type FC, useEffect } from "react";
+import { type FC, useEffect, useState } from "react";
+import userEvent from "@testing-library/user-event";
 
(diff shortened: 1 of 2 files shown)


robintown:
I've learned that *lots* of other hooks in this project are susceptible to the exact same failure modes, so I've generalized the fix: https://github.com/element-hq/element-call/pull/3316

PR #3302, opened 2025-05-29
Leave issue refactor

toger5:
A follow up refactor after the leave issue fix.

It contains two parts:
 - simplifying the local storage key management
 - refactoring useLivekit

This change tries to make the code more explicit so that we only do the things we really need to do and rely less on react updating everything correctly.

It also surfaces, that we are currently implementing useLivekit in a way, so that we can change the encryption system on the fly and recreate the room. I am not sure this is a case we need to support?

### DIFF
--- a/src/e2ee/sharedKeyManagement.ts
+++ b/src/e2ee/sharedKeyManagement.ts
@@ -7,16 +7,27 @@ Please see LICENSE in the repository root for full details.
 
 import { useEffect, useMemo } from "react";
 
-import { setLocalStorageItem, useLocalStorage } from "../useLocalStorage";
+import {
+  setLocalStorageItemReactive,
+  useLocalStorage,
+} from "../useLocalStorage";
 import { getUrlParams } from "../UrlParams";
 import { E2eeType } from "./e2eeType";
 import { useClient } from "../ClientContext";
+import { logger } from "matrix-js-sdk/lib/logger";
 
 /**
  * This setter will update the state for all `useRoomSharedKey` hooks
+ * if the password is different from the one in local storage or if its not yet in the local storage.
  */
 export function saveKeyForRoom(roomId: string, password: string): void {
-  setLocalStorageItem(getRoomSharedKeyLocalStorageKey(roomId), password);
+  if (
+    localStorage.getItem(getRoomSharedKeyLocalStorageKey(roomId)) !== password
+  )
+    setLocalStorageItemReactive(
+      getRoomSharedKeyLocalStorageKey(roomId),
+      password,
+    );
 }
 
 const getRoomSharedKeyLocalStorageKey = (roomId: string): string =>
@@ -48,7 +59,12 @@ const useRoomSharedKey = (
 
 export function getKeyForRoom(roomId: string): string | null {
   const { roomId: urlRoomId, password } = getUrlParams();
-  if (roomId !== urlRoomId) return null;
+  if (roomId !== urlRoomId)
+    logger.warn(
+      "requested key for a roomId which is not the current call room id (from the URL)",
+      roomId,
+      urlRoomId,
+    );
   return (
     password ?? localStorage.getItem(getRoomSharedKeyLocalStorageKey(roomId))
   );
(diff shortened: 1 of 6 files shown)


src/useLocalStorage.ts
```diff
@@ -13,7 +13,11 @@ type LocalStorageItem = ReturnType<typeof localStorage.getItem>;
 // Bus to notify other useLocalStorage consumers when an item is changed
 export const localStorageBus = new EventEmitter();
 
-// Like useState, but reads from and persists the value to localStorage
+/**
+ * Like useState, but reads from and persists the value to localStorage
+ * this hook will not update when we write to localStorage.setItem(key, value) directly.
+ * For the hook to react either use the returned setter or `saveKeyForRoom`.
```
robintown:
```suggestion
 * This hook will not update when we write to localStorage.setItem(key, value) directly.
 * For the hook to react either use the returned setter or `setLocalStorageItem`.
```

src/e2ee/sharedKeyManagement.ts
```diff
@@ -8,56 +8,52 @@ Please see LICENSE in the repository root for full details.
...
+  // make sure we never return the initial null value from `useLocalStorage`
+  return [setInitialValue ?? roomSharedKey, setRoomSharedKey];
 };
 
 export function getKeyForRoom(roomId: string): string | null {
-  saveKeyFromUrlParams(getUrlParams());
-  const key = getRoomSharedKeyLocalStorageKey(roomId);
-  return localStorage.getItem(key);
+  const { roomId: urlRoomId, password } = getUrlParams();
+  if (roomId !== urlRoomId) return null;
```
robintown: I think in this case we should still return the localStorage item, if it exists

src/home/useGroupCallRooms.ts
```diff
@@ -81,11 +81,13 @@ function sortRooms(client: MatrixClient, rooms: Room[]): Room[] {
...
 const roomIsJoinable = (room: Room): boolean => {
-  if (!room.hasEncryptionStateEvent() && !getKeyForRoom(room.roomId)) {
-    // if we have an non encrypted room (no encryption state event) we need a locally stored shared key.
+  const password = getKeyForRoom(room.roomId);
+  if (!room.hasEncryptionStateEvent() && !password) {
+    // if we have a non encrypted room (no encryption state event) we need a locally stored shared key.
     // in case this key also does not exists we cannot join the room.
     return false;
   }
+  if (password) saveKeyForRoom(room.roomId, password);
```
robintown: I remember we were talking about how we were both confused by the `saveKeyForRoom` side effect being executed in functions whose names sounded like they wouldn't perform any such effects - and I think this change, making `roomIsJoinable` save the key, kind of doubles down on that issue. So, why do we need to save the password here, and is there a higher-level function that we could perform the saving of the key in instead?

robintown:
By the way, I'm wondering if the source of our room disconnect troubles wouldn't go away entirely if we fixed this effect to actually depend on `livekitRoom`, like the linter thinks it should:

https://github.com/element-hq/element-call/blob/5f8a8d4c5be1e1a619e1dc1ed410e9cca56a40f5/src/room/InCallView.tsx#L151

It totally seems like it should - I can't come up with any explanation why we would want to ignore the effect dependency there. What do you think?

toger5:
> By the way, I'm wondering if the source of our room disconnect troubles wouldn't go away entirely if we fixed this effect to actually depend on `livekitRoom`, like the linter thinks it should:
> 
> https://github.com/element-hq/element-call/blob/5f8a8d4c5be1e1a619e1dc1ed410e9cca56a40f5/src/room/InCallView.tsx#L151
> 
> It totally seems like it should - I can't come up with any explanation why we would want to ignore the effect dependency there. What do you think?

I think we absolutely should. But since we now only create the room once and then never set the state again, i think we would still have the room disconnect under control?

src/e2ee/sharedKeyManagement.ts
toger5: That makes sense. I will log it as a warning. I am really sure if that case happens things will be very broken but the log line should help identifying the issue.

src/home/useGroupCallRooms.ts
toger5: I moved this into an effect in the group call view. So we try storing it explicitly

src/room/InCallView.tsx
```diff
@@ -148,7 +148,7 @@ export const ActiveCall: FC<ActiveCallProps> = (props) => {
         });
     };
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
```suggestion
```

No longer necessary

robintown (approved):

PR #3309, opened 2025-06-03
add onAudioPlaybackStarted callback for controls

toger5:


src/controls.ts
```diff
@@ -54,7 +55,13 @@ export const outputDevice$ = new BehaviorSubject<string | undefined>(undefined);
  * This should also be used to display a darkened overlay screen letting the user know that audio is muted.
  */
 export const setAudioEnabled$ = new Subject<boolean>();
-
+let trackReadyEmitted = false;
+export const setTrackReady = (): void => {
+  trackReadyEmitted = true;
+  if (!trackReadyEmitted) {
```
jmartinesp: Something tells me this should be reversed 😅 .

jmartinesp:
```suggestion
  if (!trackReadyEmitted) {
    trackReadyEmitted = true;
```

toger5: Oh dear

jmartinesp:
I think I finally fixed the bug we discussed on EXA thanks to this callback, thanks!

src/controls.ts
toger5: thanks for fixing this!

jmartinesp (approved):

src/controls.ts
```diff
@@ -20,6 +20,7 @@ export interface Controls {
   /** @deprecated use  onAudioDeviceSelect instead*/
   onOutputDeviceSelect?: (id: string) => void;
   onAudioDeviceSelect?: (id: string) => void;
+  onAudioTrackReady?: () => void;
```
robintown: I see that this is being called the first time we connect a remote audio track to the audio context, but the name sounds equally like something you might call when you've acquired a local microphone audio stream. Needs docs in docs/controls.md, and would a name like `onAudioPlaybackStarted` be more appropriate?

src/livekit/MatrixAudioRenderer.tsx
```diff
@@ -204,6 +204,7 @@ function AudioTrackWithAudioNodes({
       useContext ? [audioNodes.gain!, audioNodes.pan!] : [],
     );
     setTrackReady(true);
+    controls.setTrackReady();
```
robintown: If the callback is meant to be called the first time we try to play some audio, as I understand it, then I would expect it to be called from the reaction audio renderer and the join/leave audio renderer as well

toger5:
@robintown really good points. I will convert it into a draft for now since a proper PR needs your changes. It was done as a "maybe this helps unblock android" experiment.

robintown (approved):
Still could be nice to have a little more context as to what issue EXA was facing without this change, but the API looks right for what it does

PR #3311, opened 2025-06-03
Break loop in acquiring media on Safari

robintown:
As of this week we've started seeing an issue in mobile Safari where Element Call will fail to acquire media, calling `getUserMedia` over and over in a loop. How this happened:

- It seems that, for whatever reason (a recent Safari update?) the browser will emit a MediaDevices 'devicechange' event whenever you call `getUserMedia`, even though there is no change in the devices returned by `enumerateDevices`.
- Then, because Element Call's reactivity happened to be a little too coarse-grained in a couple of spots (specifically, the LobbyView and MuteStates files), every 'devicechange' event would have the unintentional side effect of triggering a new call to `getUserMedia`. And so, the loop begins.

While it is pretty annoying that Safari has started emitting spurious 'devicechange' events, we can easily work around this situation by making our reactivity more fine-grained, breaking the loop. And this may result in performance improvements for all browsers, so, why not.

Here's the specifics of the loop:

- LobbyView calls `usePreviewTracks`
- `usePreviewTracks` has an effect which calls `getUserMedia`
- The browser emits a MediaDevices 'devicechange' event
- Media devices observer emits a new devices map
- MediaDevicesContext invalidates its available devices memo and returns a new devices map
- Devices map is passed to MuteStates
- MuteStates (unnecessarily) invalidates its mute states memos and returns new ones
- Mute states are passed to the LobbyView
- LobbyView (unnecessarily) invalidates its `onError` callback and passes a new one to `usePreviewTracks`
- `usePreviewTracks` uses the `onError` callback as one of the effect dependencies, so the effect runs again, and we call `getUserMedia` again

### DIFF
--- a/src/livekit/MediaDevicesContext.tsx
+++ b/src/livekit/MediaDevicesContext.tsx
@@ -17,9 +17,10 @@ import {
   type JSX,
 } from "react";
 import { createMediaDeviceObserver } from "@livekit/components-core";
-import { combineLatest, map, startWith } from "rxjs";
+import { combineLatest, distinctUntilChanged, map, startWith } from "rxjs";
 import { useObservable, useObservableEagerState } from "observable-hooks";
 import { logger } from "matrix-js-sdk/lib/logger";
+import { isEqual } from "lodash-es";
 
 import {
   useSetting,
@@ -140,7 +141,13 @@ function useMediaDeviceHandle(
         kind,
         () => logger.error("Error creating MediaDeviceObserver"),
         requestPermissions,
-      ).pipe(startWith([])),
+        // This Observable emits new values whenever the browser fires a
+        // MediaDevices 'devicechange' event. One would think, innocently, that
+        // a 'devicechange' event means the devices have changed. But as of the
+        // time of writing, we are seeing mobile Safari firing spurious
+        // 'devicechange' events (where no change has actually occurred) when
+        // we call MediaDevices.getUserMedia. So, filter by deep equality.
+      ).pipe(startWith([]), distinctUntilChanged<MediaDeviceInfo[]>(isEqual)),
     [kind, requestPermissions],
   );
   const available = useObservableEagerState(
--- a/src/room/MuteStates.ts
+++ b/src/room/MuteStates.ts
@@ -60,7 +60,7 @@ function useMuteState(
     // Determine the default value once devices are actually connected
     (prev) =>
       prev ?? (device.available.size > 0 ? enabledByDefault() : undefined),
-    [device],
+    [device.available.size],
   );
   return useMemo(
     () =>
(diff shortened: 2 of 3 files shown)


robintown:
Up to the reviewer whether this needs a test, but… the failure mode is pretty idiosyncratic, and testing the lobby component is going to be painful, so I'd prefer to not!

src/room/MuteStates.ts
```diff
@@ -60,7 +60,7 @@ function useMuteState(
     // Determine the default value once devices are actually connected
     (prev) =>
       prev ?? (device.available.size > 0 ? enabledByDefault() : undefined),
-    [device],
+    [device.available.size],
```
BillCarsonFr, line 63: So if I get it correctly, previously this effect was replayed because `device` was changed. But it was actually the "same" inner map of devices? And by changing that dependency to `device.available.size` it will not be re-triggered because the map have the same size?

Or maybe we just don't care about the content of the map, we jsut want to know if there is more than one choice.

So it is mostly a cleanup that was causing an un-wanted chain of effect replay?

As per the PR description, if we fixe that, then i won't `invalidates its mute states memos` then the `onError callback` will not be invalidated? So we touch it? is it done for additional safety?

Honestly I am a bit confused by this change and how it fixes the problem..
I was more expecting a change regarding `devicechange` so that it is stable when the same list is emited again

BillCarsonFr (approved):
I approve but I am a bit confused by the change, and I wonder if it is not working around the real problem?

src/room/MuteStates.ts
toger5: What if we make sure that: 
> MediaDevicesContext invalidates its available devices memo and returns a new devices map

Gets fixed to: 
MediaDevicesContext only ever updates its device map if there is an actual change in the device map.

Would that also fix the loop? I think as long as we have the media device context to be a rapid firing react state hook we will run into trouble sooner or later again?

robintown: I've implemented the suggestion of comparing the devices list by deep equality, so now we get 3x safety from this bug.

src/livekit/MediaDevicesContext.tsx
```diff
@@ -140,7 +141,13 @@ function useMediaDeviceHandle(
...
         () => logger.error("Error creating MediaDeviceObserver"),
         requestPermissions,
-      ).pipe(startWith([])),
+        // This Observable emits new values whenever the browser fires a
+        // MediaDevices 'devicechange' event. One would think, innocently, that
+        // a 'devicechange' event means the devices have changed. But as of the
+        // time of writing, we are seeing mobile Safari firing spurious
+        // 'devicechange' events (where no change has actually occurred) when
+        // we call MediaDevices.getUserMedia. So, filter by deep equality.
+      ).pipe(startWith([]), distinctUntilChanged<MediaDeviceInfo[]>(isEqual)),
```
toger5, line 150: Nice that distinctUntilChanged can get a comparator.

PR #3313, opened 2025-06-04
Upgrade Android publishing to use Maven Central Portal

jmartinesp:
- Upgrade maven publishing plugin to `v0.32.0`.
- Change the `SonatypeHost` to point to `CENTRAL_PORTAL`.


robintown (approved):

jmartinesp:
Ah sorry, we talked with @fkwp before and we agreed to not merge this yet to not block releasing the next library version, but I forgot to leave a message here. 

Can we revert the PR and I can submit it again as a draft later?

jmartinesp:
The alternative is doing the migration in all `io.element` libraries, which is feasible, but maybe not as soon as we'd need it.

PR #3314, opened 2025-06-04
Fix react errors in settings tab

toger5:
Fixes:
Screenshot From 2025-06-04 12-27-38
And:
Screenshot From 2025-06-04 12-27-41

### DIFF
--- a/src/settings/FeedbackSettingsTab.tsx
+++ b/src/settings/FeedbackSettingsTab.tsx
@@ -57,14 +57,12 @@ export const FeedbackSettingsTab: FC<Props> = ({ roomId }) => {
 
   const [optInAnalytics, setOptInAnalytics] = useOptInAnalytics();
   const optInDescription = (
-    <Text size="sm">
-      <Trans i18nKey="settings.opt_in_description">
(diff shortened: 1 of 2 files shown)


src/settings/FeedbackSettingsTab.tsx
```diff
@@ -57,14 +57,12 @@ export const FeedbackSettingsTab: FC<Props> = ({ roomId }) => {
...
-        You may withdraw consent by unchecking this box. If you are currently in
-        a call, this setting will take effect at the end of the call.
-      </Trans>
-    </Text>
+    <Trans i18nKey="settings.opt_in_description">
+      <AnalyticsNotice />
+      <br />
+      You may withdraw consent by unchecking this box. If you are currently in a
+      call, this setting will take effect at the end of the call.
+    </Trans>
```
robintown: This changes the font size of the description, instead you could set `as="span"` on the Text component to avoid the warning

robintown (changes requested):

robintown (approved):

toger5:
good we finally got rid of those...

PR #3315, opened 2025-06-04
Media device settings modal

toger5:
This tries fixing the ios external device settings modal issue.

It is WIP and mostly to have a ci build to test some changes.


toger5:
A large part of this is solved more elegantly in: https://github.com/element-hq/element-call/pull/3311

The other part is sth we want to address differntly @robintown will try to not bluntly rip out all device name permission logic but fix it instead.

PR #3316, opened 2025-06-04
Avoid reactivity bugs in how we track external state

robintown:
Many of our hooks which attempt to bridge external state from an EventEmitter or EventTarget into React had subtle bugs which could cause them to fail to react to certain updates. The conditions necessary for triggering these bugs are explained by the tests that I've included.

In the majority of cases, I don't think we were triggering these bugs in practice. They could've become problems if we refactored our components in certain ways. The one concrete case I'm aware of in which we actually triggered such a bug was the race condition with the `useRoomEncryptionSystem` shared secret logic (addressed by a1110af6d549e2e3b5cd33912e0a6389d7bdf326).

But, particularly with all the weird reactivity issues we're debugging this week, I think we need to eliminate the possibility that any of the bugs in these hooks are the cause of our current headaches.

### DIFF
--- a/src/room/useActiveFocus.ts
+++ b/src/room/useActiveFocus.ts
@@ -9,10 +9,12 @@ import {
   type MatrixRTCSession,
   MatrixRTCSessionEvent,
 } from "matrix-js-sdk/lib/matrixrtc";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useRef } from "react";
 import { deepCompare } from "matrix-js-sdk/lib/utils";
 import { logger } from "matrix-js-sdk/lib/logger";
 import { type LivekitFocus, isLivekitFocus } from "matrix-js-sdk/lib/matrixrtc";
+
+import { useTypedEventEmitterState } from "../useEvents";
 
 /**
  * Gets the currently active (livekit) focus for a MatrixRTC session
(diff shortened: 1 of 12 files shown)


src/room/useActiveFocus.ts
```diff
@@ -22,38 +24,27 @@ import { type LivekitFocus, isLivekitFocus } from "matrix-js-sdk/lib/matrixrtc";
...
+    }, [rtcSession]),
+  );
 
-  const onMembershipsChanged = useCallback(() => {
-    const newActiveFocus = rtcSession.getActiveFocus();
-    if (!!newActiveFocus && !isLivekitFocus(newActiveFocus)) return;
-    if (!deepCompare(activeFocus, newActiveFocus)) {
+  const prevActiveFocus = useRef(activeFocus);
+  useEffect(() => {
+    if (!deepCompare(activeFocus, prevActiveFocus.current)) {
```
toger5, line 39: why do we use deepCompare here and the lodash isEqual for the rxjs pipe.

src/room/useRoomState.ts
```diff
@@ -5,25 +5,33 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
     room,
+    RoomEvent.CurrentStateUpdated,
+    useCallback(() => room.currentState, [room]),
+  );
+  return useTypedEventEmitterState(
+    currentState,
     RoomStateEvent.Update,
-    useCallback(() => setNumUpdates((n) => n + 1), [setNumUpdates]),
+    useCallback(() => f(currentState), [f, currentState]),
   );
```
toger5, line 36: why id is it not enough, to listen to `RoomEvent.CurrentStateUpdated`?
Shouldnt that also update the currentState if currentState emits RoomStateUpdate.Update

toger5 (commented):
I think we can just do `getLiveTimeline().getState(EventTimeline.FORWARDS)`

src/useLocalStorage.ts
```diff
@@ -18,32 +18,28 @@ export const localStorageBus = new EventEmitter();
...
-    ),
-  ];
-};
+  const value = useSyncExternalStore(subscribe, getValue);
+  const setValue = useCallback(
+    (newValue: string) => setLocalStorageItemReactive(key, newValue),
+    [key],
+  );
+
+  return [value, setValue];
```
toger5: ah nice, its of course the same thing.
Can this use the other hook?
```
const getValue = useCallback(() => localStorage.getItem(key), [key]);
const value = useTypedEventEmitterState(localStoreBus, key, getValue);
 const setValue = useCallback(
    (newValue: string) => setLocalStorageItemReactive(key, newValue),
    [key],
  );
return [value, setValue];
```

src/useMediaQuery.ts
```diff
@@ -1,28 +1,25 @@
...
+      return (): void => mediaQuery.removeEventListener("change", onChange);
+    },
+    [mediaQuery],
   );
-
-  // We want any change to the update counter to trigger an update here
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  return useMemo(() => mediaQuery.matches, [mediaQuery, numChanges]);
+  const getState = useCallback(() => mediaQuery.matches, [mediaQuery]);
+  return useSyncExternalStore(subscribe, getState);
```
toger5, line 24: this should also work with the event emitter state hook?

src/room/useActiveFocus.ts
robintown: https://github.com/element-hq/element-call/pull/3318

src/room/useRoomState.ts
robintown: No, it's only called when the `currentState` object is replaced with something else: https://github.com/matrix-org/matrix-js-sdk/blob/44399f601733c8f4daf5a6c891a355cf425192a9/src/models/room.ts#L1319

src/useLocalStorage.ts
robintown: With a little bit of type-casting, sure

src/useMediaQuery.ts
robintown: This one is an EventTarget rather than an EventEmitter, so it has different methods (`addEventListener` vs. `on`)

robintown:
> I think we can just do `getLiveTimeline().getState(EventTimeline.FORWARDS)`

The way I see it, it's currently impossible to use this correctly in a reactive manner. This chain of methods depends on the current values of:

1. `Room.timelineSets`
2. `EventTimelineSet.liveTimeline`
3. `EventTimeline.endState`

And I have no idea what events you're supposed to listen to to stay up to date on all three links in the chain. All three of these seem to have cases where matrix-js-sdk can modify them without emitting anything.

`currentState` on the other hand gives you the `CurrentStateUpdated` event, which provides the right contract for reacting to the latest room state. (I kinda just disagree with the decision to deprecate `currentState` in matrix-js-sdk. It's so simple, and serves common use cases well!)

src/useLocalStorage.ts
toger5: Oh the bus was an EventEmitter instead of a TypedEventEmitter... But making it typed is a good thing anyways imo.
Thx

src/useMediaQuery.ts
toger5: okay I see.

toger5:
I think on the `currentState` topic we stick to the deprecated solution for now and bring this up with the web team eventually

toger5 (approved):

PR #3317, opened 2025-06-04
Always render audio from the current set of participants

robintown:
We forgot to tell React that we need the audio renderer to react to changes in the set of MatrixRTC participants; instead we had it referencing `rtcSession.memberships` non-reactively.
    
Now, I'm not 100% confident that this is going to fix the "speaking from the void" issues observed in the wild, because I can't reproduce them and, in my testing, the InCallView component always seemed to be rendered redundantly when the MatrixRTC participants change, even though we hadn't explicitly stated that it needs to react. (This makes sense as we haven't memoized the component.) But it's worth a shot.

Maybe fixes https://github.com/element-hq/voip-internal/issues/344

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -104,6 +104,7 @@ import { ConnectionLostError } from "../utils/errors.ts";
 import { useTypedEventEmitter } from "../useEvents.ts";
 import { MatrixAudioRenderer } from "../livekit/MatrixAudioRenderer.tsx";
 import { muteAllAudio$ } from "../state/MuteAllAudioModel.ts";
+import { useMatrixRTCSessionMemberships } from "../useMatrixRTCSessionMemberships.ts";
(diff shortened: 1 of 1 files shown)


toger5 (approved):
While debugging on Tuesday I also really wanted to do this. But concluded that this is really bad but should work since we will always re-render. But maybe there is a condition where we do not?

Also I was not aware of the already existing `useMatrixRTCSessionMemberships` so this is much more low handing than i anticipated.

Definitely worth a shot!

toger5:
why is this dependent on: https://github.com/element-hq/element-call/pull/3316
It seems entirely de-tangled since only the impl of `useMatrixRTCSessionMemberships` changes but not the signature.

robintown:
Sorry, yeah, I realized it's independent just now as well. I've rebased to exclude the other changes.

PR #3321, opened 2025-06-05
Use html audio element to call setSinkId for reactions and call sounds.

toger5:
setSinkId on the audio context is not supported on safari ios. So we try using an html audio element and `audioElement.setSinkId` instead with this PR.


toger5:
With more testing on the subject of audio devices I suspect there is hope that this is not needed. lets keep this in draft until I know for sure.

toger5:
As of the most recent tests with ios 26 this seems to not be required. So we can close this.

PR #3322, opened 2025-06-05
html-audio + always-request-device-permissions

toger5:


toger5:
Closed because we will use a different approach which will not ask for all permissions on startup.
@robintown

PR #3325, opened 2025-06-10
fix: Force the docker image for lk-jwt-service to be pulled everytime

fkwp:


dev-backend-docker-compose.yml
```diff
@@ -3,7 +3,7 @@ networks:
 
 services:
   auth-service:
-    image: ghcr.io/element-hq/lk-jwt-service:latest-ci
+    image: ghcr.io/element-hq/lk-jwt-service:0.2.3
```
BillCarsonFr: I wonder if instead we should put `pull_policy: always` for the `dev` compose file. I though there was a way to specify a minumum version but lools like not

fkwp: I think it's fine from the EC dev PoV to rely on the latest stable lk-jwt-service. If you need a dev version of the lk-jwt-service you can do so by:
- running your dev version of lk-jwt-service locally with `LIVEKIT_INSECURE_SKIP_VERIFY_TLS="YES_I_KNOW_WHAT_I_AM_DOING" LIVEKIT_URL="wss://matrix-rtc.m.localhost/livekit/sfu" LIVEKIT_KEY=devkey LIVEKIT_SECRET=secret LIVEKIT_JWT_PORT=6080 LIVEKIT_LOCAL_HOMESERVERS=synapse.localhost  ./lk-jwt-service`
- `docker-compose -f ./dev-backend-docker-compose.yml up nginx synapse redis livekit`, i.e., starting all services BUT the lk-jwt service

Note the `backend/dev_nginx.conf` is automatically using `host.docker.internal:6080`, i.e., running the local dev version.
```
   43  # MatrixRTC reverse proxy
   44  # - MatrixRTC Authorization Service
   45  # - LiveKit SFU websocket signaling connection
   46  upstream jwt-auth-services {
   47      server auth-server:6080;
   48      server host.docker.internal:6080;
   49  }
```

fkwp: addressed by https://github.com/element-hq/element-call/pull/3325/commits/80ebfcae452153a10cbac1f0b0d8bd2be77f60c2

BillCarsonFr (approved):
Maybe we coudl do it for livekit and synapse too?

PR #3303, opened 2025-05-30
Expose keep alive event configurarion in ec config and update defaults.

toger5:

PR #3332, opened 2025-06-11
Apply new timing constants to MatrixRTC session

### LINKED ISSUE #3331: Adjust timing parameters: reset-period to 4s, reset-timeout to 18s, abort-timeout to 4s

As a follow-up from rageshakes we can improve our timings wrt. the delayed leave event.

Apply the following timings to your default and sample config

- reset-period 5s -> 4s
- reset-timeout 15s -> 18s
- abort-timeout 5s -> 4s

After https://github.com/element-hq/element-call/pull/3303 being merged update various config files

Config files PRs:
- https://github.com/element-hq/element-call-webapp-deployments/pull/13
- https://github.com/element-hq/element-call/pull/3332

fkwp:
Depends on:
- https://github.com/element-hq/element-call/pull/3303

fixes https://github.com/element-hq/element-call/issues/3331

### DIFF
--- a/config/config.devenv.json
+++ b/config/config.devenv.json
@@ -8,5 +8,12 @@
   "features": {
     "feature_use_device_session_member_events": true
   },
-  "ssla": "https://static.element.io/legal/element-software-and-services-license-agreement-uk-1.pdf"
+  "ssla": "https://static.element.io/legal/element-software-and-services-license-agreement-uk-1.pdf",
+  "matrix_rtc_session": {
+    "wait_for_key_rotation_ms": 3000,
+    "membership_event_expiry_ms": 6000000,
+    "delayed_leave_event_delay_ms": 18000,
+    "delayed_leave_event_restart_ms": 4000,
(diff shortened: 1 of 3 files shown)


config/config.devenv.json
```diff
@@ -8,5 +8,12 @@
   "features": {
     "feature_use_device_session_member_events": true
   },
-  "ssla": "https://static.element.io/legal/element-software-and-services-license-agreement-uk-1.pdf"
+  "ssla": "https://static.element.io/legal/element-software-and-services-license-agreement-uk-1.pdf",
+  "matrix_rtc_session": {
+    "wait_for_key_rotation_ms": 3000,
+    "membership_event_expiry_ms": 6000000,
```
toger5: What do you think about making this 180 000 000 (Which is 5*60*60*1000 so 5h)
So we reduce the chance of a resend as much as possible.

toger5 (approved):

PR #3334, opened 2025-06-13
Refactor media devices to live outside React as Observables

robintown:
This moves the media devices state out of React to further our transition to a MVVM architecture in which we can more easily model and store complex application state. I have created an AppViewModel to act as the overarching state holder for any future non-React state we end up creating, and the MediaDevices reside within this. We should move more application logic (including the CallViewModel itself) there in the future.


src/main.tsx
```diff
@@ -60,7 +61,7 @@ Initializer.initBeforeReact()
   .then(() => {
     root.render(
       <StrictMode>
-        <App />
+        <App vm={new AppViewModel()} />
```
toger5, line 64: This is a very significant change that almost deserves its own PR. But since it so obvious that this would be coming eventually and to reduce unnecassary friction me for my part is happy to merge it with this PR.

src/utils/observable.ts
```diff
@@ -38,3 +38,18 @@ export function accumulate<State, Event>(
...
+ * This function exists to help with certain cases of bridging Observables into
+ * React, where an initial value is needed. You should never use it to create an
+ * Observable derived from another Observable; use reactive operators instead.
+ */
+export function getValue<T>(state$: Observable<T>): T {
+  let value: T | typeof nothing = nothing;
+  state$.subscribe((x) => (value = x)).unsubscribe();
+  if (value === nothing) throw new Error("Not a state Observable");
+  return value;
+}
```
toger5, line 55: Sth feels off ab out this helper method. We only use it for `selected$` right now which is an Observable because the interface wants it to be an observable. But with the `Not a state Observable` error it really feels like the typing is not correct. If we have a stateful observable we will have a behavior subject like observable right? So would it make sense to just enforce a behavior subject on the interface level to then just call `selected$.getValue()`? (Of course a readonly behavior subject)

I am not sure if one can easily convert a scope.state() observable into a BehaviourSubject type...

src/useAudioContext.tsx
```diff
@@ -73,7 +71,10 @@ export function useAudioContext<S extends string>(
   props: Props<S>,
 ): UseAudioContext<S> | null {
   const [soundEffectVolume] = useSetting(soundEffectVolumeSetting);
-  const { audioOutput } = useMediaDevices();
+  const devices = useMediaDevices();
+  const audioOutputId = useObservableEagerState(
+    devices.audioOutput.selected$,
+  )?.id;
```
toger5: This would also work right?
I have seen the pattern, that we wrap hooks into hooks more often recently so this would not be a code style inconsistency.
```suggestion
  const audioOutputId = useObservableEagerState(
    useMediaDevices().audioOutput.selected$,
  )?.id;
```

src/useAudioContext.tsx
```diff
@@ -73,7 +71,10 @@ export function useAudioContext<S extends string>(
   props: Props<S>,
 ): UseAudioContext<S> | null {
   const [soundEffectVolume] = useSetting(soundEffectVolumeSetting);
-  const { audioOutput } = useMediaDevices();
+  const devices = useMediaDevices();
+  const audioOutputId = useObservableEagerState(
```
toger5: Should we move this right next to the effect?
```
useEffect(() => {
    if (
      audioContext &&
      "setSinkId" in audioContext &&
      !controlledAudioDevices
    ) {
      ```
Its only used in the effect? I guess this is a "all state before all effects" vs "self contained codeblocks" tradeoff so both works but in this case I prefer it being next to the effect.

File: src/utils/observable.ts (hunk 2 above)
Comment (toger5) [reply src/utils/observable.ts:55]: What is the benefit of scope.state() vs a conversion to a behaviour subject like this. Then we do not need the `getValue` method but can use behavior subjects.
```
function toStateObservable<T>(observable: Observable<T>, initialValue: T): BehaviorSubject<T> {
  const subject = new BehaviorSubject<T>(initialValue);
  observable.subscribe(subject); // Push values from observable into subject
  return subject;
}
```

File: src/controls.ts [hunk 5]
```diff
@@ -42,12 +42,11 @@ export interface OutputDevice {
...
  */
 export const setPipEnabled$ = new Subject<boolean>();
-// BehaviorSubject since the client might set this before we have subscribed (GroupCallView still in "loading" state)
-// We want the devices that have been set during loading to be available immediately once loaded.
-export const availableOutputDevices$ = new BehaviorSubject<OutputDevice[]>([]);
-// BehaviorSubject since the client might set this before we have subscribed (GroupCallView still in "loading" state)
-// We want the device that has been set during loading to be available immediately once loaded.
-export const outputDevice$ = new BehaviorSubject<string | undefined>(undefined);
+
+export const availableOutputDevices$ = new Subject<OutputDevice[]>();
```
toger5, line 49: Do we use the subject here to have a simple non statful api?
Or is there another reason for this? I do think it is reasonable to have a global Behaviour subject state here?

src/state/MediaDevices.ts
```diff
@@ -0,0 +1,369 @@
...
+
+  // Start using device names as soon as requested. This will cause LiveKit to
+  // briefly request device permissions and acquire media streams for each
+  // device type while calling `enumerateDevices`, which is what browsers want
+  // you to do to receive device names in lieu of a more explicit permissions
+  // API. This flag never resets to false, because once permissions are granted
+  // the first time, the user won't be prompted again until reload of the page.
+  private readonly usingNames$ = this.deviceNamesRequest$.pipe(
+    map(() => true),
+    startWith(false),
```
toger5, line 346: I played around with this a bit more to fully understand why we start with false.
On chrome this all feels sane. On FF on the other hand the rerequesting of permissions just restarts the camera but does not prompt me with another permissions request. I also dont see any difference in setting this to start with `true` vs `false`
In the room list I do not get a request, in the lobby i do get a request for both (even when setting start with to `false`). Once I open the settings I do not get a request prompt in both cases but in the startWith `false` I do see a camera stream rerequest.

src/state/MediaDevices.ts
```diff
@@ -0,0 +1,369 @@
...
+    SelectedAudioInputDevice
+  > = new AudioInput(this.usingNames$, this.scope);
+
+  public readonly audioOutput: MediaDevice<
+    AudioOutputDeviceLabel,
+    SelectedAudioOutputDevice
+  > = getUrlParams().controlledAudioDevices
+    ? new ControlledAudioOutput(this.scope)
+    : new AudioOutput(this.usingNames$, this.scope);
+
```
toger5, line 361: Observables have the really great property, that they allow us to conditionally call them (compared to hooks) I like that a lot more than what it was before!

toger5 (commented):
This looks good.
I have a couple of comments and the follow up PR (which as expected is now super small) here: https://github.com/element-hq/element-call/pull/3342

src/main.tsx
robintown: Right well, as the AppViewModel holds nothing besides the media devices at this moment, extracting this to a separate PR would mean I'd first introduce an empty class which is effectively dead code.

src/utils/observable.ts
robintown: Yep, so we're back to the pretty fundamental flaw of RxJS that it makes no type-level distinction between [behaviors and streams](https://monoid.dk/post/behaviors-and-streams-why-both/). I agree that now-ish with the introduction of this weird function could be the right time to do something about this.

I think we probably want to introduce our own `Behavior` type so we can enforce that these things are read-only. That might be as simple as writing `type Behavior<T> = Omit<Observable<T>, 'next'>`, actually.

The other challenge is that a `toStateObservable` function like you propose would not be usable with the `Observable.pipe` method, so it would create an extra layer of indentation wherever it's used. I suggest that to make this as convenient as possible, we could instead patch an `Observable.behavior` method into the Observable prototype, then we could start using it with the familiar method chaining syntax of JS. I also want to be careful when designing such a function that it limits the subscription to an ObservableScope (to avoid resource leaks) and pulls the initial value directly from the input Observable to avoid asking the caller to calculate it with redundant code.

Stepping back, I just wish we had a better solution to functional reactive programming in the JS ecosystem, really—[Hareactive](https://github.com/Funkia/hareactive) seemed really thoughtfully designed but the approach of functional purity may appear a bit extreme to some, and also the repo is inactive :cry: Without broader excitement for FRP in the ecosystem, a sustainable solution might come from inlining a simple & home-made reactivity library, ultimately.

src/useAudioContext.tsx
robintown: Sure, why not

src/controls.ts
robintown: My reasoning is that I originally wanted the controls.ts file to be as minimal a wrapper as possible around the `windows.controls` methods, channeling the data from any calls to these methods off to the files that are actually meant to be responsible for application logic. Then we're never tempted to "mutate" BehaviorSubjects from outside the file, bouncing data through controls.ts that didn't really belong there.

But I get that the BehaviorSubject approach is closer to the imperative approach of just having some variables for your state and mutating them. Would you like me to steer this back in the BehaviorSubject direction?

src/state/MediaDevices.ts
robintown: Wow, you're right, what the heck is going on here…

robintown: Okay, turns out that we start our state observables lazily (only once the first subscriber arrives, in this case once the settings menu is opened if you're sitting on the room list). This is why starting with `true` appears to have such minimal effects.

I think there's a path forward where we start all our state observables *eagerly*, stick with the initial value of `false` (to postpone the permissions request until it's actually needed), and avoid the re-request of the camera stream in Firefox. I'll add the proposed fix here in a bit.

src/controls.ts
toger5: I think for now I think we should focus on this file being consistent. I saw your "Behavior" PR where I think we can then do a consistent update to this file if we prefer it.

toger5 (approved):
Approving this.
We still have the issue with firefox video stream restarting, but since this is not a regression we can work on this in another PR. @robintown I think you have an idea and I also started a take here: https://github.com/element-hq/element-call/pull/3353

Issue #3271, opened 2025-05-15
UX implementation of fast audio output switching between Main Speaker and Earpiece Speaker (for In-App Calling)


fkwp:
#### What would you like to do?

Depending on the environment users do want to route the audio output of Element Call to different speakers in particular the Earpiece speaker which preserves privacy of the video call.

#### How would you like to achieve it?

Implementation of the designs as agreed in https://github.com/element-hq/voip-internal/issues/338

Additional notes:
- We want to only show the button if the application is *not* using a bluetooth headset, but rather one of the speaker or earpiece
- We want to have the icon reflect the current state (rather than the state it switches you to - needs follow-up with Aaron)

PR #3335, opened 2025-06-13
Add a global control for toggling earpiece mode

robintown:
:warning: **Based on https://github.com/element-hq/element-call/pull/3334**, please only review the new commits

This method will be used by Element X to show an earpiece toggle button in the header. I have not yet implemented the screen-dimming overlay; that will come in a separate PR.

For https://github.com/element-hq/element-call/issues/3271


toger5:
How do we plan to proceed with this? With the updated timeline and the plan to aim for in web view controls directly, do we still want to merge this as a stop gap?

robintown:
No, we no longer want this

Issue #3272, opened 2025-05-15
Make the reactions button available on small mobile screen width layouts


fkwp:
#### What would you like to do?

Have access to the reactions button also in case of small mobile screen width layouts.

#### Why would you like to do it?

Because reactions are a very essential part of video conferences :-)

robintown:
I intend to fix this by moving the 'switch camera' button out of the footer and onto the local user's video tile, to match the most recent designs. That will free up space for the reactions button.

PR #3336, opened 2025-06-13
Move the switch camera button to the local user's tile

robintown:
For https://github.com/element-hq/element-call/issues/3272

Screenshot 2025-07-10 at 16-14-22 Element Call test


src/state/MediaViewModel.ts
```diff
@@ -433,20 +437,35 @@ abstract class BaseUserMediaViewModel extends BaseMediaViewModel {
  * The local participant's user media.
  */
 export class LocalUserMediaViewModel extends BaseUserMediaViewModel {
+  /**
+   * The local video track as an observable that emits whenever the track
+   * changes, the camera is switched, or the track is muted.
+   */
+  private readonly videoTrack$: Observable<LocalVideoTrack | null> =
```
toger5, line 453: is this a behavior? If so we might (just for the transition) want to add comments so we can easily find the leftover bits we still should convert to behaviors?

src/state/MediaViewModel.ts
```diff
@@ -433,20 +437,35 @@ abstract class BaseUserMediaViewModel extends BaseMediaViewModel {
...
+      switchMap((v) => {
+        const track = v?.publication?.track;
+        if (!(track instanceof LocalVideoTrack)) return of(null);
+        return merge(
+          // Watch for track restarts because they indicate a camera switch
+          fromEvent(track, TrackEvent.Restarted).pipe(
+            startWith(null),
+            map(() => track),
+          ),
+          fromEvent(track, TrackEvent.Muted).pipe(map(() => null)),
```
toger5, line 466:
```suggestion
          // Emits only when Muted instead of emitting on mute state changes. (on un-mute it will not emit)
          // We set the track to null here. On un-mute, the `fromEvent(..., TrackEvent.Restarted)` will make sure that `videoTrack$` is set to `track` again.
          fromEvent(track, TrackEvent.Muted).pipe(map(() => null)),
```

Do I understand this correctly?

src/state/MediaViewModel.ts
```diff
@@ -457,12 +476,46 @@ export class LocalUserMediaViewModel extends BaseUserMediaViewModel {
   public readonly alwaysShow$ = alwaysShowSelf.value$;
   public readonly setAlwaysShow = alwaysShowSelf.setValue;
 
+  /**
+   * Callback for switching between the front and back cameras.
+   */
+  public readonly switchCamera$: Observable<(() => void) | null> =
```
toger5: Also a behavior right?

Is there any variable name rule to indicate that this is a callback? We dont have this for normal (non observable) vars. So probably this is not needed.

toger5: This method/observable is really good to read!

src/state/MediaViewModel.test.ts
```diff
@@ -79,21 +105,100 @@ test("toggle fit/contain for a participant's video", async () => {
...
+        available$: of(new Map()),
+        selected$: of(undefined),
+        select: selectVideoInput,
+      },
+    }),
+    async (vm) => {
+      // Switch to back camera
+      getValue(vm.switchCamera$)!();
+      expect(restartTrack).toHaveBeenCalledTimes(1);
+      expect(restartTrack).toHaveBeenCalledWith({ facingMode: "environment" });
```
toger5:
```suggestion
      expect(restartTrack).toHaveBeenCalledExactlyOnceWith({ facingMode: "environment" });
```
Do you prefer using the dedicated matcher for this case?

toger5:
What was the reason for the block? Just sequencing or did we wait for sth specific to get merged first?

src/state/MediaViewModel.ts
toger5: with the comment it reads, like it might actually be an observable...

robintown: It is an Observable, since in this case we need to notify about changes to the mutable LiveKit track object (Behaviors on the other hand have immutable data as their values)

src/state/MediaViewModel.ts
robintown: Yes - commented.

src/state/MediaViewModel.ts
robintown: Changed this in the rebase to become a behavior.

> Is there any variable name rule to indicate that this is a callback?

Kind of - I just name it as I would a method, with the verb first. My little naming convention for Observables that emit "events" is to put the verb last, as in `toggleSpotlightExpanded$` / `spotlightExpandedToggle$`. But the types are always the best source of information.

toger5 (approved):

PR #3342, opened 2025-06-17
Toger5/robin/mvvm-media-devices

toger5:
- dont use preferred vs selected concept in controlled media. Its not needed since we dont use the id for actual browser media devices (the id's are not even actual browser media devices)
  - add more logging
  - add more conditions to not accidently set a deviceId that is not a browser deviceId but one provided via controlled.

### DIFF
--- a/src/controls.ts
+++ b/src/controls.ts
@@ -6,6 +6,9 @@ Please see LICENSE in the repository root for full details.
 */
 
 import { Subject } from "rxjs";
+import { logger as rootLogger } from "matrix-js-sdk/lib/logger";
+
+const logger = rootLogger.getChild("[controlled-output]");
 
 export interface Controls {
   canEnterPip(): boolean;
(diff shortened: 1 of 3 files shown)


toger5:
@robintown 
I am wondering if instead of doing this conditional logic our media devices should expose:
`selectedId /** The id of the available devices list, that needs to be used to show where the selection in the ui should be*/`
`deviceId /** the actual browser deviceId which currently is selected. On controlled media this is always undefined. */`

We might even be able to use the media stream creation promises to use selectedId=deviceId as a condition if a switch is ongoing and give the user instant feedback with some kind of secondary ui that shows when the device has actually switched.

robintown (approved):
Feel free to merge this to my branch

robintown:
@toger5 Oops, missed your question. Something like that sounds fine to me. If you wanna go that route, I suggest we add a new boolean field the `SelectedAudioOutputDevice` interface indicating whether the web view is responsible for setting the output, rather than create a separate Observable, as that keeps this complexity from spreading to the other device types.

toger5:
> Feel free to merge this to my branch

Cool. I kept it as is without any refactor to the selectedId logic.

Issue #2579, opened 2024-08-24
All iPhones/iPads getting burning hot, high battery drain due to unsupported VP8 codec


jacotec:
### Steps to reproduce

Users are refusing to use Element for video calls after migrating to Element-X / Element-Call for video conferences / calls as their iPhones are getting burning hot during video calls, battery drain of 10%(+) per 15 min.

We never saw this in legacy element with 1:1 calls or Jitsi, nor with BigBlueButton or Nextcloud Talk.

Investigating further the root cause is the use of VP8 which is not supported in any way by Apple hardware at all. There are plenty of reports in the web which are documenting the same issue when using VP8, as VP8 needs to be decoded/encoded fully in software and can't use hardware acceleration.

#### What did you expect?

Also Apple participants should be able to use Element Call without suffering extreme temperature dissipation and battery loss.

Widespread supported standard codecs like H.264 or H.265 should be at least offered as an alternative to all clients to support hardware acceleration

#### What happened instead?

Apple devices getting hot, losing battery quickly in Element calls. People refusing to use Element-X for calls due to this issue.

### Operating system

IOS, iPadOS, MacOS

### Browser information

Element-X, Element-Desktop

### URL for webapp

Element call 0.6.0

### Will you send logs?

No

jacotec:
Workaround for self-hosted Element-Call deployments (Docker):

In `element-call/volumes/elementcall/backend/livekit.yaml` add this section:

```
room:
  enabled_codecs:
    - mime: video/h264
    - mime: audio/opus
```

This will force livekit to use H.264 and all is back to normal in the Apple world, iPhones not getting hot anymore, battery use for 15 minutes video call on an iPhone 14 is down to 3% now.

fkwp:
@jacotec thx for the research :-) very good datapoints. We already have this topic on our radar. Currently we are finishing getting-the-basics-right; in that order:
- Stable and reliable room state
- Sane E2EE via individual sender keys based on to-device messages in order to provide Perfect Fordward Secrecy and Post Compromise Security
- Homeserver presence (or similar) to get MatrixRTC also working like Matrix in partitioned scenarios

We are aware of the various codes out there and its complicated dependencies. Where h264 makes a lot of sense in the Apple eco system (energy performance) its looks different on Android. Here https://livekit.io/webrtc/codecs-guide is a good summary with all the pros and cons 

[...]

Issue #2373, opened 2024-05-11
Consider H264 as default codec for video


ndren:
#### What would you like to do?
Replace VP8 with H264 as the default video codec.
#### Why would you like to do it?
This will reduce the CPU/power consumption and bitrate of video transmission in element call, especially on devices without hardware VP8 encoding/decoding, but also other devices as H264 is well-standardised and optimised as an older standard.
#### How would you like to achieve it?
Replace this in https://github.com/element-hq/element-call/blob/livekit/src/livekit/options.ts

### Have you considered any alternatives?

Keep VP8 or use some other codec.

### Additional context

Sending video is the highest difference in CPU usage (2x total usage, from 30% to over 60%) when using element-call for me, receiving video is negligible.

VP8 is not hardware decoded on my laptop according to about:support. I'm almost certain it's not hardware encoded either.

toger5:
Looking at:
https://livekit.io/webrtc/bitrate-guide
-|H.264|	VP8	|VP9	|AV1
--|---|---|--|--
1920x1080	|2.70Mbps	|2.00Mbps|	1.20Mbps	|900kbps
1280x720	|1.25Mbps	|1.00Mbps	|700kbps	|550kbps
960x540|	700kbps|	600kbps|	450kbps|	350kbps
640x360|	400kbps|	400kbps|	270kbps	|190kbps
384x216|	170kbps|	180kbps|	120kbps|	90kbps
320x180|	140kbps|	160kbps|	90kbps	|70kbps
160x90|	50kbps|	90kbps	|35kbps|	30kbps

At the resolutions we are using vp8 is better bandwidth wise then H.264. For lower resolutions it is not however.

Dependent on the hardware h264 might be less resource intensive. Thanks for the ticket so we have a place to discuss this.

Quackdoc:
What bitrate is being? I have found that x264 is far better pretty much across the board then libvpx is, which I presume the apps are using when no hardware acceleration is being used. Different encoders can differ greatly but even hwenc h264 are probably better then libvpx is. 

This guide is fairly lackluster in general, but more often then not, h264 will be far better quality at a given bitrate, even really low ones.

as a related sidenote, Can we get a toggle in the web app/clients to swap between codecs? If we can swap between h264, vp8, vp9, and av1 that would be greatly appreciated. Maybe as an "advanced room" this would also help users provide feedback as to what codec is the best for default.

[...]

PR #3345, opened 2025-06-18
Changing the video codec from VP8 to H.264

### LINKED ISSUE #3166: Selected video resolution often not that great (EC defaults lower than livekit defaults)
### Steps to reproduce

I start by creating Element Calls between myself. The wifi or LAN connection are good, the SFU is not saturated at all. This is current EXA (25.3.4) and Element web (1.11.96), both using the EC widget v0.9 (either integrated or at call.element.io). My SFU is version 1.8.4.

### Outcome

The resolution that EXA chose was 180x320 and it appeared quite blocky in Element-Web, in other calls I have seen even way worse resolutions though. I have submitted a rageshake (send feedback within Element Call) just some time ago with the following text:

```
Good quality network connection, mobile chose 180x320 pixel video (EXA EC v0.9). SFU 1.8.4
I have never switched video sources so far.
For Florian.
```

Turning the camera on and off seemed to often make the situation a bit better (it started out low and switched to a **somewhat** higher resolution shortly after), switching to front-side camera and back-sid did not help at all (but it also did not make matters worse).

The Element-Web webcam on the other hand did send a 720x1280 resolution. (switch camera off and on  again in EC will lead not always to the same resolutions being picked though). So, it is possible to have higher-resolution calls.

Is livekit erring on the low-quality side of things here? Are there livekit knobs that I can turn. It is a stock installation basically.

### Operating system

EXA and EW (Firefox v137/Linux)

### Browser information

_No response_

### URL for webapp

integrated v0.9 and call.element.io (v0.9)

### Will you send logs?

Yes

fkwp:
Recent research has confirmed that the number of devices lacking H.264 decoding capabilities is negligible.

In contrast, many modern devices come equipped with native H.264 encoders and decoders. By switching to the H.264 codec, we can significantly reduce CPU usage. Especially mobile devices will benefit, where it can lead to better battery life.

Furthermore, we add a 540p simulcast video layer.

fixes https://github.com/element-hq/element-call/issues/3166 https://github.com/element-hq/element-call/issues/2579 https://github.com/element-hq/element-call/issues/2373

### Current state
H264 video is still broken on iOS — other participants won’t receive video from an iOS device. This happens when the iOS device is the first to join a LiveKit room and starts streaming H264.

However, it's working if a non-iOS client is already streaming H264 before the iOS device joins. Note Using VP8 on the other client does not fix the issue — it must be H264.


src/livekit/options.ts
```diff
@@ -23,12 +23,16 @@ const defaultLiveKitPublishOptions: TrackPublishDefaults = {
   red: false,
   forceStereo: false,
   simulcast: true,
-  videoSimulcastLayers: [VideoPresets.h180, VideoPresets.h360] as VideoPreset[],
+  videoSimulcastLayers: [
+    VideoPresets.h180,
+    VideoPresets.h360,
+    VideoPresets.h540,
+  ] as VideoPreset[],
```
spaetz: I might be misunderstanding the livekit docs. But can we actually hand it three presets here? Livekit [docs](https://docs.livekit.io/reference/client-sdk-js/interfaces/trackpublishdefaults.html) stated up to 2 custom layers. Or are those standard preset no custom layers?
```
videoSimulcastLayers: [VideoPreset]

custom video simulcast layers for camera tracks, defaults to h180, h360, h540.
You can specify**up to two custom layers** that will be used instead of the LiveKit default layers.
```

fkwp: nope its limited to three in total including the base layer.

src/main.tsx
```diff
@@ -29,7 +29,7 @@ window.setLKLogLevel = setLKLogLevel;
 initRageshake().catch((e) => {
   logger.error("Failed to initialize rageshake", e);
 });
-setLKLogLevel("info");
+setLKLogLevel("debug");
```
BillCarsonFr, line 34: Is this on purpose? or a left over of a dev setup?

BillCarsonFr (approved):
Just a small comment regarding the loglevel

src/main.tsx
fkwp: this is part of debugging why h264 with simulcast is not working on iOS devices.

src/livekit/options.ts
```diff
@@ -11,10 +11,20 @@ import {
   type RoomOptions,
   ScreenSharePresets,
   type TrackPublishDefaults,
-  type VideoPreset,
+  VideoPreset,
   VideoPresets,
 } from "livekit-client";
 
+const VideoPresetsH264 = {
```
toger5: what makes this specific H264 presets? Aren't those just regular presets that could also be used with other encordings?

src/livekit/options.ts
```diff
@@ -23,10 +33,13 @@ const defaultLiveKitPublishOptions: TrackPublishDefaults = {
   red: false,
   forceStereo: false,
   simulcast: true,
-  videoSimulcastLayers: [VideoPresets.h180, VideoPresets.h360] as VideoPreset[],
+  videoSimulcastLayers: [
+    VideoPresetsH264.h144,
+    VideoPresets.h360,
```
toger5: why the mixture of `VideoPresetsH264` and `VideoPresets`?

src/livekit/options.ts
fkwp: this PR is in draft -- this is part of debugging why h264 with simulcast is not working on iOS devices.

using h264 specific presets help finding out if the h264 macro-block constraint is the problem here.

toger5:
What is the status on this one? Can the OP get a description what this is blocked by. (I do remember some device specific issues where some resolution just were not send at all?)

fkwp:
> What is the status on this one? Can we OP get a description what this is blocked by. (I do remember some device specific issues where some resolution just were not send at all?)

H264 video is still broken on iOS — other participants won’t receive video from an iOS device. This happens when the iOS device is the first to join a LiveKit room and starts streaming H264.

However, it's working if a non-iOS client is already streaming H264 before the iOS device joins. Note Using VP8 on the other client does not fix the issue — it must be H264.

toger5:
Test on safari on mac os works.
(to make sure its ios and not safari in general

src/livekit/options.ts
```diff
@@ -23,10 +33,13 @@ const defaultLiveKitPublishOptions: TrackPublishDefaults = {
...
+  ] as VideoPreset[],
   screenShareEncoding: ScreenSharePresets.h1080fps30.encoding,
   stopMicTrackOnMute: false,
-  videoCodec: "vp8",
+  videoCodec: "h264",
```
toger5, line 29:
```suggestion
  videoCodec: platform === "ios" || !useH264Setting ? "vp8" : "h264",
```

Adding a dev setting for using h264 would allow merging this and would provide us more data on how this behaves in real live.

Let me know if I should update this PR.

toger5: @fkwp any thoughts on this? This would also be nice to have in the 15 release ?

shootie22:
bump, any progress on this? what is blocking it? how about iOS 26? 👀

fkwp:
> bump, any progress on this? what is blocking it? how about iOS 26? 👀

still blocked on Safari as h264 with e2ee as a first joiner with Safari does not work.

thorw4ld:
so this is not fixable until safari itself changes? which might never happen

toger5:
## Update
Things have changed after rebasing it on a large refactor including numerous livekit udpates. Those are the observations compared to previous ones:

>H264 video is still broken on iOS — other participants won’t receive video from an iOS device. This happens when the iOS device is the first to join a LiveKit room and starts streaming H264.

This is not the case anymore. Its the opposite now: starting with iOS H264 (first to join the lk room) is realiably resulting in the other device receiving media. _I do have slow network and matrix.org accounts (and matrix.org is also slow atm) so things might have started to work if i would have been a bit more patient_

> However, it's working if a non-iOS client is already streaming H264 before the iOS device joins. Note Using VP8 on the other client does not fix the issue — it must be H264.

If another client streams VP8 before it OCCASIONALLY does not work (not work = the vp8 client cannot see the ios stream, ios can see both. the local echo stream and the vp8 stream)


Both on vp8 seemed to always work. (independent on who joined first)

**Tested on:**
Iphone + macOs (firefox)

@shootie22 @thorw4ld If you want to help out on testing this proceed as following:
on EW type in `/devtools` when sending the message a modal will appear with a custom EC url field.
on EX go to the settings tab and press the version number multiple times (i think 7) a devtool option will appear with a custom EC url filed

Use this URL: `https://pr3345--element-call.netlify.app/room` to test with the h264 build. (try all combinations of EW and EX using/not using it with all devices you have and report the observations here)

shootie22:
@toger5 thats awesome, thank you so much! I will try to test it out tonight after I get home and report back.

shootie22:
alright, I tested all combinations - result: as long as I'm using the EC version you provided on EX iOS, my phone will not transmit any video.  

I tested using:
My phone - iPhone 11 Pro Max on iOS 26.1, EX 25.11.3
Linux desktop - Fedora 43, Element Nightly 2025111401

the combination of EC version or which device joined first didn't seem to matter, as long as my phone was using the PR version of EC, it would send no video. :( I will note that my phone does indeed get hot. I don't think I noticed until now, but it gets hot on either version, stable or the one from this PR.

is there anything specific you'd like me to test or look for, perhaps in the developer menu?

vkovtash:
Very nice initiative. But why h264 instead of h265?
H264 will only be supported by safari, and maybe some implementations of webrtc that are not from Google. Chrome can't use it due to licensing concerns and will resort to software implementation (openh264) which is slower than vp8.

h265 on the other hand, has no licensing concerns as long hardware codec is used. A lot of modern chips (2018 and onwards) have both hardware encoding and decoding. Chrome supports it since the beginning of 2025, safari as well.

Using h265 with a fallback to vp8 would improve both quality and performance for many users as it's hardware accelerated and provides better compression.

Additionally it can unlock 1080p for many devices that were constrained by cpu before (its target bitrate is 3Mbps at 1080p, the same as vp8 at 720p).

toger5:
Related: https://github.com/livekit/livekit/issues/2792

There are known compatibility issues. Worth investigating if they are resolved nowadays?

vkovtash:
Yeah, I suggested it prematurely. It definitely has issues on Windows with Intel encoder. 
Issues listed in https://github.com/livekit/livekit/issues/2792 would not be blocking:
 - Neither vp8 nor h264 support svc. It would be nice to have in h265, but even without it it's better in quality, compression and hardware support.
 - Nvenc not supporting some resolutions. Not nice but element call is opinionated about supported resolutions anyway. Just need to keep that in mind, configuring it.

In my tests with element call. It works well between Mac/iOS/Andriod. Even at 1080p cpu usage is low.
Windows has issues. It can decode h265 but produces broken encodings. I've spent some time on it, looks like a live kit bug.
Linux I didn't test. Decoding would most like work. And encoding is likely challenging to enable.

I think it worth trying to at least allow to opt in for it on Mac/iOS/Android, and keep serving vp8 on Windows and linux for some time until the bugs are fixed.

Egoistically:
Could a setting be added to let the user decide the codec? I think keeping a widely-supported codec the default is the right choice, but for certain use-cases something like h265 or av1 would be great.

RealCrystalNight:
Any new news on this? I believe https://github.com/fluxerapp/fluxer has good LiveKit implementation for screensharing. It would be nice to screenshare on Linux without straining my CPU. It's a super annoying limitation.

drewwells:
Do you need testers for this? I see it approved already but not with the dev flag merged to flip on/off h264. 

It would be great to just focus on that one codec rather than getting too worried about h265 which could be added on pretty easily later as another dev toggle.

BillCarsonFr:
FWIW this PR is adding a settings drop down to select the codec (via advanced settings) https://github.com/element-hq/element-call/pull/3736

robintown:
We're currently too worried about breakages with devices that don't fully support H264 (I know it's a widespread codec, but the risks due to its previous status as a patented codec are real, as evidenced by field reports), so we would prefer to add a setting that allows the user or deployment to choose H264 instead. Therefore we need to get back to https://github.com/element-hq/element-call/pull/3736.

PR #3346, opened 2025-06-18
Create a type-level distinction between raw Observables and Behaviors

robintown:
https://monoid.dk/post/behaviors-and-streams-why-both/
See https://github.com/element-hq/element-call/pull/3334#discussion_r2152640104 for more context

Reviewing will be easiest if you hide whitespace changes from the diff.

### DIFF
--- a/src/state/Behavior.ts
+++ b/src/state/Behavior.ts
@@ -5,9 +5,7 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 Please see LICENSE in the repository root for full details.
 */
 
-import { BehaviorSubject, distinctUntilChanged, Observable } from "rxjs";
-
-import { type ObservableScope } from "./ObservableScope";
+import { BehaviorSubject } from "rxjs";
 
 /**
  * A stateful, read-only reactive value. As an Observable, it is "hot" and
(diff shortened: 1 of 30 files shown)


robintown:
If anyone wants to look into the test failures while I'm away, be my guest

src/state/Behavior.ts
```diff
@@ -0,0 +1,52 @@
...
+
+declare module "rxjs" {
+  interface Observable<T> {
+    /**
+     * Converts this Observable into a Behavior. This requires the Observable to
+     * synchronously emit an initial value.
+     */
+    behavior(scope: ObservableScope): Behavior<T>;
+  }
+}
```
toger5: I am wondering if we want two api's one where the observable provides the current/initial value and one where we can pass in an initial value manually so that also non state observables can be used as behaviors?

src/state/Behavior.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const nothing = Symbol("nothing");
+
+Observable.prototype.behavior = function <T>(
+  this: Observable<T>,
+  scope: ObservableScope,
+): Behavior<T> {
+  const subject$ = new BehaviorSubject<T | typeof nothing>(nothing);
+  // Push values from the Observable into the BehaviorSubject
+  this.pipe(scope.bind()).subscribe(subject$);
```
toger5: is the `scope.bind()` mandatory? It seems we can make it optional since it is only needed to make the observable complete when the scope ends?

src/state/MediaViewModel.ts
```diff
@@ -387,8 +391,8 @@ abstract class BaseUserMediaViewModel extends BaseMediaViewModel {
     encryptionSystem: EncryptionSystem,
     livekitRoom: LivekitRoom,
     displayname$: Observable<string>,
```
toger5: Why is this not becoming a Behavior as well?

src/state/Behavior.ts
robintown: Yeah, having a shortcut for `observable.pipe(startWith(initialValue)).behavior(scope)` wouldn't be a bad idea. I could see it making sense to either overload the `Observable.behavior` method, add a separate `behaviorFrom(scope, initialValue)` method, or call this thing `stepper` like other FRP libraries do.

toger5: maybe another helpful thing would be:
`behavior.pipeBehavior(Opertaors...) = behavior.pipe(Operators...).behavior(scope)` 

But I am not sure if we can infer the scope from the behavior.

src/state/Behavior.ts
robintown: We do really need to make sure that subscriptions end at some point so we don't create resource leaks, and the ObservableScope approach was the most straightforward way I could come up with of encouraging this practice on the view model level

src/state/MediaViewModel.ts
robintown: It did become a Behavior, maybe you're looking at the wrong commit?

src/state/Behavior.ts
robintown: There wouldn't be a way to infer the scope as far as I'm aware. But a way to avoid the extra indentation created by `.behavior(scope)` could be nice to have.

src/state/MediaViewModel.ts
toger5: very likely!

toger5 (approved):
I like the latest updates!
I also did the merge with current livekit. (I hope that is fine)

PR #3347, opened 2025-06-18
Earpiece switcher and overlay

### LINKED ISSUE #3271: UX implementation of fast audio output switching between Main Speaker and Earpiece Speaker (for In-App Calling)

robintown:
A couple things still need doing here:
- [x] Write tests
  - move mediaDevice test to callViewModel
- [x] Remove the controls API for toggling the earpiece switcher, which we no longer need
- [x] ~~Split the header out into another PR, maybe?~~
- [x] Modernize the new context
- [x] Force dark theme for the overlay (and also the header and footer, when it is active)
   - The light theme looks very correct to me when the phone is in light theme mode (this is tbd)
   - Let's ask Aaron
   - Result: We do the same then what we do with emoji and settings overlays: dark overlay in light and dark theme
     This requires us to also swap the icon colors for the header bar icons
   - Result V2: we are fine with the inconsistency between modal background darkening and earpiece overlay lightening
- [x] Fix android designs (video goes to top of the bar)
- [x] Fix hover border being clipped
- [x] Fix header being shown in PiP
- [x] Only present the fast toggling UX in case either the Earpiece speaker or the Main speaker are selected
    - The fast toggling UX is only toggling between Earpiece and Main speaker
    - We also consider always showing the toggles but making them actions (shows what will happen when pressed instead of the current state) This is currently implemented in the pr as of: https://github.com/element-hq/element-call/pull/3347/commits/df3f962f4707bccc6100001ad1b9bffee95d5d0c
    - This might be not enough: we also need to make sure to always show the button (even if in bluetooth mode)
- [x] landscape should hide the header for now
- [x] Update Element Web version used for end-to-end tests to a version that includes https://github.com/element-hq/element-web/pull/30213

Potential candidates for a follow up PR
    
- [x] fix android designs OneOnOneLayout (so that the app_bar overlays the video and but the floating tile does still respect the bounds)
- [x] fix OneOnOneLayout Landscape (currently the app bar is not visible in this mode.

Closes https://github.com/element-hq/element-call/issues/3271


docs/controls.md
```diff
@@ -23,3 +23,4 @@ On mobile platforms (iOS, Android), web views do not reliably support selecting
   playing audio in the webview. It can be helpful to do device setup on the native app when the webviews audio is ready.
   In particular android is using it to setup the output channel so that the call volume can
   be controlled by the hardware volume rocker.
+- `controls.toggleEarpieceMode(): void` Switches audio between the earpiece device and an arbitrary non-earpiece device.
```
toger5: Do we really need this. It is for the native OS to enforce EC to switch to earpiece mode.
Do we have any UX branch that requires this change to get initiated by the native os?
On android we can just do so by setting the device right?
On iOS the only scenario I can imagine is using the proximity sensor but that is not active if we are not yet in earpiece mode.

robintown: We don't need this, it is left over from when I thought the button was going to have to be in the native header

robintown:
@fkwp This is now ready for testing! Expect to see two headers on mobile until we get to the EX tasks listed in https://github.com/element-hq/voip-internal/issues/360

toger5 (approved):

PR #3353, opened 2025-06-20
Skip unnecassary media devices permissions requests (video feed flicker when opening settings)

toger5:
This is based on the discussion:
https://github.com/element-hq/element-call/pull/3334#discussion_r2155369391

Which I think we should not forget to fix.
I merged the initial PR so this PR is also a tracker to not forget about this issue.

~~This does not work in the lobby but does work in call. The lobby is planned to become part of the call anyways so we do not have the camera restart when switching from lobby to call.~~
A way was found to also fix this in the lobby.

### DIFF
--- a/src/state/MediaDevices.ts
+++ b/src/state/MediaDevices.ts
@@ -8,6 +8,7 @@ Please see LICENSE in the repository root for full details.
 import {
   combineLatest,
   filter,
+  identity,
   map,
   merge,
   of,
@@ -18,7 +19,7 @@ import {
   type Observable,
 } from "rxjs";
 import { createMediaDeviceObserver } from "@livekit/components-core";
-import { logger } from "matrix-js-sdk/lib/logger";
+import { logger as rootLogger } from "matrix-js-sdk/lib/logger";
 
 import {
   audioInput as audioInputSetting,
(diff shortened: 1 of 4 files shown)


src/state/MediaDevices.ts
```diff
@@ -93,15 +94,16 @@ export const iosDeviceMenu$ = navigator.userAgent.includes("iPhone")
...
   kind: MediaDeviceKind,
-  usingNames$: Observable<boolean>,
+  updateAvailableDeviceRequests$: Observable<boolean>,
   scope: ObservableScope,
 ): Observable<MediaDeviceInfo[]> {
-  return usingNames$.pipe(
-    switchMap((usingNames) =>
+  return updateAvailableDeviceRequests$.pipe(
+    startWith(false),
+    switchMap((withPermissions) =>
```
robintown: Instead of having updateAvailableDeviceRequests$, we could stick with usingNames$ and just expand the logic here to do some "trial and error" by creating a device observer without permissions and then re-creating it if we need names which aren't present.

Or, since this is getting kind of hairy, maybe we should just submit a fix upstream to LiveKit Components :thinking:

robintown (approved):

PR #3358, opened 2025-06-24
Remove usages of forwardRef

robintown:
It has been deprecated in React 19, which allows functional components to receive refs just like any other prop.

To review, hide whitespace changes from the diff.

### DIFF
--- a/src/useMergedRefs.ts
+++ b/src/useMergedRefs.ts
@@ -5,21 +5,21 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 Please see LICENSE in the repository root for full details.
 */
 
-import { type MutableRefObject, type RefCallback, useCallback } from "react";
+import { type RefCallback, type RefObject, useCallback } from "react";
 
 /**
  * Combines multiple refs into one, useful for attaching multiple refs to the
  * same DOM node.
  */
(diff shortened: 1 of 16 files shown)


src/useMergedRefs.ts
```diff
@@ -5,21 +5,21 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
-  ...refs: (MutableRefObject<T | null> | RefCallback<T | null> | null)[]
+  ...refs: (RefObject<T | null> | RefCallback<T | null> | null | undefined)[]
 ): RefCallback<T | null> =>
   useCallback(
     (value) =>
       refs.forEach((ref) => {
         if (typeof ref === "function") {
           ref(value);
-        } else if (ref !== null) {
+        } else if (ref != null) {
```
toger5: Doesnt this just become `if(ref)`, if we also treat `ref=undefined` as false in this condition?

toger5 (approved):
This looks mostly like noops. I hope I managed to check all non-changes that git tracks as changes due to indentations correctly...

src/useMergedRefs.ts
robintown: Yep, I've replaced it with that simplified condition

toger5:
This ended up not being a squash merge. Was this intentional?

PR #3359, opened 2025-06-24
Modernize how we use React contexts

robintown:
Here I'm just applying a couple of quick [codemods](https://github.com/reactjs/react-codemod) that are available for React 19.


toger5 (approved):
Cool. I was always wondering if `.Provider` really is the only way or if they could find a way to obsolete it.

PR #3360, opened 2025-06-24
Use the platform property for ios detection

toger5:


fkwp:
Check on a real device -> works

fkwp (approved):
LGTM and verified with real devices (iPhone, iPad)

PR #3361, opened 2025-06-24
bugfix: #3344 Reconnecting to the same SFU on membership change

### LINKED ISSUE #3344: Reconnecting to the same SFU on membership change
### Steps to reproduce

- Starting a call on https://call.element.dev
  -  Join with participant A
  -  Join with participant B
  -  Join with participant C

- if participant A (the oldest one) leaves the call, all remaining participants will re-connect to the same SFU

loglines include

`2025-06-16T15:47:24.984Z 2 matrix SFU config changed! URL was wss://matrix-org.livekit.cloud now wss://matrix-org.livekit.cloud
`

#### What did you expect?

Not reconnected to the very same SFU, i.e., stay connected to the SFU

#### What happened instead?

Re-connecting to the same SFU

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

Yes

BillCarsonFr:
Fixes #3344

~~This a an ugly workaround, the root cause is that we have a problem with react hooks and a re-render is causing effects with non properly working dependencies to be ran again?~~
This has changed to actually address the root cause: the `activeFocus` state returned by the `useActiveLivekitFocus` hook was updating with new objects that had the same values. (so we switched to the same focus)
Now `useActiveLivekitFocus` behaves as expected, only returning new objects when there is a change.

### DIFF
--- a/src/livekit/openIDSFU.ts
+++ b/src/livekit/openIDSFU.ts
@@ -8,7 +8,7 @@ Please see LICENSE in the repository root for full details.
 import { type IOpenIDToken, type MatrixClient } from "matrix-js-sdk";
 import { logger } from "matrix-js-sdk/lib/logger";
 import { type MatrixRTCSession } from "matrix-js-sdk/lib/matrixrtc";
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { type LivekitFocus } from "matrix-js-sdk/lib/matrixrtc";
 
 import { useActiveLivekitFocus } from "../room/useActiveFocus";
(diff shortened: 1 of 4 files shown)


src/livekit/openIDSFU.ts
```diff
@@ -55,13 +61,28 @@ export function useOpenIDSFU(
         },
       );
     } else {
+      currentFocus.current = {} as LivekitFocus;
       setSFUConfig(undefined);
     }
   }, [client, activeFocus, showErrorBoundary]);
```
toger5: Using the primitive types in the dependecy array, should also make this work:
```suggestion
  }, [client, activeFocus?.type, activeFocus?.livekit_service_url, activeFocus?.livekit_alias, showErrorBoundary]);
```
We might need to then recompose the active focus to make eslint happy:
```
getSFUConfigWithOpenID(client,
    {
      type: activeFocus.type,
      livekit_service_url: activeFocus.livekit_service_url,
      livekit_alias: activeFocus.livekit_alias
      }
     ).then(...
```
or:
```
// inside the effect
const activeFocus = {
      type: activeFocus.type,
      livekit_service_url: activeFocus.livekit_service_url,
      livekit_alias: activeFocus.livekit_alias
}
```
But we dont need the ref and the comparison function anymore.

robintown (approved):

PR #3372, opened 2025-06-27
Hangup when last person in call (based on url params)

toger5:
If the call widget is started with the URL parameter: `telephoneAutoLeave` we want the call to automatically hangup when the other participants leaves. This PR implements this behavior.

### DIFF
--- a/src/UrlParams.ts
+++ b/src/UrlParams.ts
@@ -210,6 +210,12 @@ export interface UrlConfiguration {
    * Whether and what type of notification EC should send, when the user joins the call.
    */
   sendNotificationType?: RTCNotificationType;
+  /**
+   * Whether the app should automatically leave the call when there
+   * is no one left in the call.
+   * This is one part to make the call matrixRTC session behave like a telephone call.
+   */
+  telephoneAutoLeave: boolean;
 }
 
 // If you need to add a new flag to this interface, prefer a name that describes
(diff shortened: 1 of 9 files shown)


BillCarsonFr (changes requested):
It would have been easier to review if there was not a mix of refactoring to add an option object to the view model and mode the encryptionSystem in it as well as adding the new param.

This feature doesn't seem trivial for me and I see zero test to check what it is doing.
It looks to me that this as the potential to auto hangup on you, and I expect some test for the happy path as well as the edge cases. Maybe it will auto hangup on join at this point..

toger5:
I put it out of review and back into draft since I also wanted the tests.

BillCarsonFr:
> I put it out of review and back into draft since I also wanted the tests.

👍 I followed a notification to review it. Good luck with the testing.

BTW is it compatible with the Lobby? or it will only work when no lobby (asking because of the `telephone` prefix in the option name)

toger5:
> BTW is it compatible with the Lobby? or it will only work when no lobby (asking because of the telephone prefix in the option name)

It will only happen in the call. Since we only have a running call view model once we join a call.

I am not sure the `telephone` prefix makes sense however

toger5:
This would be nice to be built on top of the [UrlParam refactor pr](https://github.com/element-hq/element-call/pull/3376) but its still open to review since it does not really make a difference if we update this one with the changes of the refactor or if we update the rafactor to include the new params introduced here.

toger5:
> This would be nice to be built on top of the https://github.com/element-hq/element-call/pull/3376 but its still open to review since it does not really make a difference if we update this one with the changes of the refactor or if we update the rafactor to include the new params introduced here.

It is now rebased on top of the url refactor.

fkwp (approved):
lgtm

PR #3376, opened 2025-06-30
UrlParams: Intent system update, split into configuration and propreties

toger5:
Needed for implementing https://github.com/element-hq/element-call/pull/3372 more cleanly

This PR updates the intend system. We want to solve this before starting on the telephone usecase implementation to prohibit us needing to make too many rust-sdk PR's and to converge to a solution we think brings the best benefits to maintain the url paramters.

This solution tries to get us the most features from this list:
 * Authority in EC (change all platforms at once)  
 * Flexible usage of EC in other clients (secondary)  
 * Very simple/fast setup on EX (one prop preferably)  
 * As much transparency as possible (intent \-\> changes things the dev did not know about anymore) *No time should be wasted on: “was this handled by an intent, do i need to look into the intent description, or do i need to look through feature list”*  
 * No artificial entanglement (flexibility) If the intent “telephoneOneOnOne” controls “autoleave” (for example) we will always get “startUnmuted” if we want autoleave. Sometimes only one of them is desired.  
 * (off topic but related) We want EC to feel like a custom react component with properties. It can execute functions (send call member event, encryption keys) but things like “sendNotifyEvent” might be configurable as params

By implementing intends with a "preset" architecture:
 - The intend is optional and it is possible to tweak each individual configuration
 - Each intend is a list of presets that also adapt based on `platform`
 - Web + EX configurations should only set the intend and keep all the other options blank. So everything can get controlled in the EC codebase.
 - It is still possible to provide an intend but overwrite a subset of the configuration properties via individual query parametrs.

This PR changes:
 - remove all deprecated fields
 - split the url parameters into `UrlProperties` and `UrlConfiguration` (configuration are the ones where you can skip with one intend, properties are always required)


src/UrlParams.ts
```diff
@@ -31,12 +33,12 @@ export enum HeaderStyle {
...
 }
 
-// If you need to add a new flag to this interface, prefer a name that describes
-// a specific behavior (such as 'confineToRoom'), rather than one that describes
-// the situations that call for this behavior ('isEmbedded'). This makes it
-// clearer what each flag means, and helps us avoid coupling Element Call's
-// behavior to the needs of specific consumers.
-export interface UrlParams {
+/**
+ * This UrlProperties contains as the properties that are used by to pass required data to the widget.
```
robintown:
```suggestion
 * This UrlProperties contains the properties that are used by applications to pass required data to the widget.
```

src/UrlParams.ts
```diff
@@ -193,6 +128,99 @@ export interface UrlParams {
...
+   * can be "light", "dark", "light-high-contrast" or "dark-high-contrast".
+   */
+  theme: string | null;
+}
+
+/**
+ * The configuration for the app, which can be set via URL parameters.
+ * Those property are different to the UrlProperties, since they are all optional
+ * and configure the behavior of the app. There value is the same if EC is used in
+ * the same context but different accoutns/users.
```
robintown:
```suggestion
 * and configure the behavior of the app. Their value is the same if EC is used in
 * the same context but with different accounts/users.
```

src/UrlParams.ts
```diff
@@ -235,6 +263,17 @@ class ParamParser {
     return this.fragmentParams.get(name) ?? this.queryParams.get(name);
   }
 
+  public getEnumParam<T extends string>(
+    name: string,
+    type: { [s: string]: T } | ArrayLike<T>,
+  ): T | undefined {
+    const value = this.getParam(name);
+    if (value && Object.values(type).includes(value as T)) {
```
robintown:
```suggestion
    if (value !== null && Object.values(type).includes(value as T)) {
```

To avoid surprises with falsy strings

src/UrlParams.ts
```diff
@@ -262,73 +305,114 @@ export const getUrlParams = (
...
+    // In SPA mode the user should always exit to the home screen when hanging
+    // up, rather than being sent back to the lobby
+    returnToLobby: isWidget ? parser.getFlag("returnToLobby") : false,
+  };
+
+  return {
+    ...properties,
+    ...intentPreset,
+    ...pickBy(configuration, (v) => v !== undefined),
+    intent,
```
robintown: What do you think about the idea of not including `intent` in the returned object here? That would enforce that we never predicate any behavior on the intent alone, and always provide some other URL parameter to override the behavior with.

src/UrlParams.ts
```diff
@@ -262,73 +305,114 @@ export const getUrlParams = (
...
+        controlledAudioDevices: platform === "desktop" ? false : true,
+        skipLobby: false,
+        returnToLobby: false,
+      };
+      break;
+    // Non widget usecase defaults
+    default:
+      intentPreset = {
+        confineToRoom: false,
+        appPrompt: false,
```
robintown: I think this should still be `true`.

src/UrlParams.ts
toger5: let me check it was used for. But your explanation for why it is not needed (should not be needed) sounds very sane to me.

src/UrlParams.ts
```diff
@@ -267,79 +300,131 @@ export const getUrlParams = (
...
+        preload: true,
+        header: platform === "desktop" ? HeaderStyle.None : HeaderStyle.AppBar,
+        showControls: true,
+        hideScreensharing: false,
+        allowIceFallback: true,
+        perParticipantE2EE: true,
+        controlledAudioDevices: platform === "desktop" ? false : true,
+        skipLobby: false,
+        returnToLobby: false,
+        sendNotificationType: undefined,
```
robintown: This should be `"notification"`, because it's possible to start a call under this intent too if the previous call ends while you're on the lobby screen.

src/UrlParams.ts
```diff
@@ -267,79 +300,131 @@ export const getUrlParams = (
...
+      };
```
robintown: Non-blocking: Since the difference between these cases is pretty minute (only `skipLobby` changes), it'd be nice to collapse these two branches into one, so it's clearer what changes

src/UrlParams.ts
toger5: And the logic for checking this would run in EC right?

toger5: `sendNotificationType` is slightly wrong. More correct would be `requestSendingNotificationtype`

(I remember we discussed that we do not have a "should do sth if checks pass" var prefix)

robintown (approved):

PR #3382, opened 2025-07-02
Shorter timeout for delayed event restart https requests

fkwp:
bump js-sdk to delayed event reset branch

See js-sdk changes here: https://github.com/matrix-org/matrix-js-sdk/pull/4896

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "livekit-client": "^2.13.0",
     "lodash-es": "^4.17.21",
     "loglevel": "^1.9.1",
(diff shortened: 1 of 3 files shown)


toger5:
This is controversial since we also need this in the rust-sdk.

The question is if this should be
 - configured per SDK (the the rust sdk will send delayed event restarts with a custom timeout and the js-sdk as well)
 - configured through the widget api. The widget api action passes over a timeout values which is then used by the widget driver.

toger5:
I think we can treat this as resolved by: https://github.com/element-hq/element-call/commit/7dd45acd2985ec05d07a8e301ed5eb075f54af61

PR #3388, opened 2025-07-03
Fix bluetooth iOS issue (device auto switches from Bluetooth to speaker)

### LINKED ISSUE #3389: When un-muting on EXI the audio switches to the phone speaker instead of staying on bluetooth headset
### Steps to reproduce

1. Starting a call on my iPhone with EXI
2. The call starts as muted, and I can hear the sound through my bluetooth headset
3. Click on unmute

## Actual

The sound switches to the iPhone speaker!
The `Change audio device` pop-up still says that I am on the bluetooth headset and no way to switch back

## Expected

Should still be using the bluetooth headset

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

Yes

toger5:
Fixes https://github.com/element-hq/element-call/issues/3389

This fixes the ios bluetooth device selection on the `livekit` branch.

The main trick in this PR is:
 - we still sync devices in controlled output audio mode
 - we subscribe to `createMediaObserver` with `kind=audiooutput` in controlled media
 - we call `devices.requestDeviceNames` when the audio mute state changes.

The issue is that the os will switch back to the speaker on unmute and subscribing to `createMediaObserver` with `kind=audiooutput`will trigger a sync device call taht syncs us back to the state before.

TODO: this does need a follow up task to find out if it is enough to only call `devices.requestDeviceNames` without the subscribe to `createMediaObserver` step.

### DIFF
--- a/src/livekit/useLivekit.ts
+++ b/src/livekit/useLivekit.ts
@@ -47,6 +47,7 @@ import { useUrlParams } from "../UrlParams";
 import { useInitial } from "../useInitial";
 import { getValue } from "../utils/observable";
 import { type SelectedDevice } from "../state/MediaDevices";
+import { platform } from "../Platform";
 
 interface UseLivekitResult {
   livekitRoom?: Room;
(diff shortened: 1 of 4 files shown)


src/livekit/useLivekit.ts
```diff
@@ -320,16 +320,18 @@ export function useLivekit(
...
-      connectionState === ConnectionState.Connected &&
-      !controlledAudioDevices
-    ) {
+    if (room !== undefined && connectionState === ConnectionState.Connected) {
       const syncDevice = (
         kind: MediaDeviceKind,
         selected$: Observable<SelectedDevice | undefined>,
       ): Subscription =>
         selected$.subscribe((device) => {
+          logger.info(
```
Half-Shot, line 329: leftover debug line? if it's being kept it might need more words :)

BillCarsonFr (approved):
Reviewed in pair sessions !
Maybe you can add more details on the actual bug it is fixing? Do we have an existing issue? do you want me to create one?

BillCarsonFr:
Fixes https://github.com/element-hq/element-call/issues/3389

src/livekit/useLivekit.ts
toger5: We would like to keep this yes.

PR #3400, opened 2025-07-10
Prefer showing the reaction button at small screen widths

### LINKED ISSUE #3272: Make the reactions button available on small mobile screen width layouts

robintown:
Closes https://github.com/element-hq/element-call/issues/3272


BillCarsonFr:
> ⚠️ Based on https://github.com/element-hq/element-call/pull/3336, review just the last commit

I usually let github handles that by selecting the base branch:
image

Like that the review is only the last commit. And when the dependency is merged this will change the base branch here automatically.

Any reason why you are doing differently?

BillCarsonFr (changes requested):
I think we could have a test for that?
I think we can test that with playwright `await browser.newContext({ reducedMotion: "reduce", screen: {width: 340, height: 720} })`?

robintown:
Test added.

BillCarsonFr (approved):

PR #3412, opened 2025-07-21
Trim extra characters from roomId parameter in URL

Half-Shot:
Some users get bitten by their client pasting random extra crap onto the end of a URL, which mucks up parsing. I think this has a low risk since we're only trimming unprintable characters.

### DIFF
--- a/src/UrlParams.ts
+++ b/src/UrlParams.ts
@@ -387,10 +387,14 @@ export function getRoomIdentifierFromUrl(
 
   // Make sure roomId is valid
   let roomId: string | null = parser.getParam("roomId");
-  if (!roomId?.startsWith("!")) {
-    roomId = null;
-  } else if (!roomId.includes("")) {
-    roomId = null;
+  if (roomId !== null) {
+    // Replace any non-printable characters that another client may have inserted.
+    roomId = roomId.replaceAll(/^[^ -~]+|[^ -~]+$/g, "");
+    if (!roomId.startsWith("!")) {
+      roomId = null;
+    } else if (!roomId.includes("")) {
+      roomId = null;
+    }
   }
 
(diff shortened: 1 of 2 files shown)


src/UrlParams.ts
```diff
@@ -258,6 +260,8 @@ export const getUrlParams = (
   search = window.location.search,
   hash = window.location.hash,
 ): UrlParams => {
+  // Replace any non-printable characters that another client may have inserted.
+  search = search.replaceAll(/^[^ -~]+|[^ -~]+$/g, "");
```
robintown: Looks like you're doing this replacement twice as ParamParser is invoked just after this.

Also, we do want to allow non-ASCII characters in the URL parameters in general. For example `displayName` or `homeserver` could contain all kinds of Unicode characters and it's valid for them to appear at the end of the search string.

robintown (commented):
I'm curious how this happens. Would it not be preferable to fix the client?

Half-Shot:
>Would it not be preferable to fix the client?

Ideally yes, but it seems in some cases these are OS-level bugs. Arguably you could fix it at any layer, but pragmatically speaking a roomId has a expected grammar and there isn't really any danger to doing some extra work here I think.

src/UrlParams.ts
Half-Shot: Yeah, good point. Maybe just limit this to roomId for now as it has a strict grammar.

robintown (commented):
This is due to a buggy browser then? It would be perfect if this included a comment about which version of which browser or system web view is responsible here, so we can understand when we might consider it unsupported, and in what cases we might need to expand this treatment to other search params.

robintown (approved):

PR #3418, opened 2025-07-24
Request capability to send call notifications

robintown:
0e0fba657558964376e6f2375271630d9cf52b84 added the ability to send call notification events when starting a call, but I forgot to give the widget the right capabilities to do this. The effect was that notifications just wouldn't send in widget mode.

Auto-approval changes:
- https://github.com/element-hq/element-web/pull/30404
- https://github.com/matrix-org/matrix-rust-sdk/pull/5446

I will backport this once reviewed.

### DIFF
--- a/src/widget.ts
+++ b/src/widget.ts
@@ -106,6 +106,10 @@ export const widget = ((): WidgetHelpers | null => {
       if (!baseUrl) throw new Error("Base URL must be supplied");
 
       // These are all the event types the app uses
+      const sendEvent = [
+        EventType.CallNotify, // Sent as a deprecated fallback
+        EventType.RTCNotification,
+      ];
       const sendRecvEvent = [
         "org.matrix.rageshake_request",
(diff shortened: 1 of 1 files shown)


toger5 (commented):
I think we want the notification to be sendrecv as well. A decline probably will also do sth inside EC.

You could argue that this is over eager since we currently dont work on it. Its very easy to do here and saves us a step if we eventually end up working on it.

robintown:
Didn't we agree that decline should be a different event type?

toger5:
> Didn't we agree that decline should be a different event type?

We discussed it but I am not sure there was a conclusion yet. At least nothing that was captured in the MSC.

But the arguments were really strong. (Most of the fields are different so a new type make quiet a bit of sense)

Since the reading needs is not guaranteed based on whatever the outcome of this discussion when it happens would be. It does make sense to merge it like it is.

toger5 (approved):

PR #3427, opened 2025-07-29
Backport: Set available devices to empty map on safari.

toger5:
Backport of: https://github.com/element-hq/element-call/pull/3426


toger5:
My assumption is, that It's enough to back port this PR to fix it in version v0.14.0. The lk bump would be more direct way to prohibit calls to setSinkId but this should already skip all possible codepaths we have to call setSinkId.

robintown (approved):

toger5:
@robintown I added: https://github.com/element-hq/element-call/pull/3427/commits/961697ab88912099831c750ea25de273c72bc334

I will still merge this without a rereview but raising it for awareness in case there are concerns/comments

PR #3434, opened 2025-08-01
Fix: missing logs in rageshake

toger5:
Dont construct logger before rageshake initialization


BillCarsonFr (approved):
Discussed live.

PR #3436, opened 2025-08-01
Reset LiveKit connection timeouts to their defaults

robintown:
I believe that the issue we were originally investigating using these increased timeouts was the fault of my earlier ISP in SW Virginia. I can't recall reproducing the exact issue on other networks.


fkwp (approved):
lgtm

PR #3444, opened 2025-08-05
Ensure that the jwt service is called before starting a call

BillCarsonFr:


src/rtcSessionHelpers.ts
```diff
@@ -77,6 +86,10 @@ async function makePreferredLivekitFoci(
       livekit_service_url: urlFromConf,
       livekit_alias: livekitAlias,
     };
+    if (shouldWarmup) {
+      // this will call the jwt/sfu/get endpoint to pre create the livekit room.
+      await getSFUConfigWithOpenID(rtcSession.room.client, focusFormConf);
+    }
```
toger5: I think the overall logic should be more like:
 - do we end up with "focus in use" or a new one (config or well_known)
 - at the end of the function `makePreferredLivekitFoci` check the condition from step one and based on that call: `await getSFUConfigWithOpenID`

We will always call `await getSFUConfigWithOpenID` only once but the way it is currenlty implemented it takes a bit to confirm that.

src/rtcSessionHelpers.ts
```diff
@@ -77,10 +80,17 @@ async function makePreferredLivekitFoci(
       livekit_service_url: urlFromConf,
       livekit_alias: livekitAlias,
     };
+    if (!toWarmUp) {
+      toWarmUp = focusFormConf;
+    }
```
toger5:
```suggestion
    toWarmUp = toWarmUp ?? focusFormConf;
```

toger5 (approved):
Looks good. I added a small style suggestion (not sure if you think that is more confusing than your current version? Feel free to leave it as is)

PR #3453, opened 2025-08-15
Pause media tracks and show a message when reconnecting to MatrixRTC

### LINKED ISSUE #3329: Stop sending media while Element Call is in reconnecting state
### Your use case

While Element Call is in reconnection state (e.g., due to a delayed leave event being emitted) stop sending media.

_No response_

### Additional context

_No response_

robintown:
[Screencast From 2025-08-15 16-03-34-trimmed.webm](https://github.com/user-attachments/assets/e22bc01d-5099-4081-b710-c2e7b42e08ce)

Closes https://github.com/element-hq/voip-internal/issues/361
Closes https://github.com/element-hq/element-call/issues/3329

- [x] Implement the local membership manager timeout mentioned in the linked issue in order to see the 'reconnecting' state appear while you can't receive any sync responses.
- [x] Avoid showing the sync loop disconnection banner in combination with the 'Reconnecting' message - they're a bit redundant.
- [x] Stop all audio and video playback while the 'Reconnecting' message is shown to avoid making people worry that the app's state is splitbrained.


src/room/InCallView.tsx
```diff
@@ -272,7 +277,7 @@ export const InCallView: FC<InCallViewProps> = ({
   const [useExperimentalToDeviceTransport] = useSetting(
     useExperimentalToDeviceTransportSetting,
   );
-  const encryptionSystem = useRoomEncryptionSystem(rtcSession.room.roomId);
```
toger5, line 273: Thank you soo much that you added the matrix room as a prop. Really fought with myself to not add it to the auto disconnect PR (to keep it simple) but i very much agree with this change!

src/state/CallViewModel.ts
```diff
@@ -481,12 +488,35 @@ export class CallViewModel extends ViewModel {
...
     fromEvent(this.matrixRTCSession, MatrixRTCSessionEvent.MembershipsChanged),
     // Handle room membership changes (and displayname updates)
-    fromEvent(this.matrixRTCSession.room, RoomStateEvent.Members),
+    fromEvent(this.matrixRoom, RoomStateEvent.Members),
   ).pipe(
-    startWith(this.matrixRTCSession.memberships),
-    map(() => {
-      return this.matrixRTCSession.memberships;
-    }),
+    startWith(null),
```
toger5: why do we start with null. If there are memberships already we would need to awiat for the first Members event from the session or the room? This which is possible to never happen?

src/state/CallViewModel.ts
```diff
@@ -481,12 +488,35 @@ export class CallViewModel extends ViewModel {
...
+      // connection but are now disconnected
+      scan(
+        ({ connectedPreviously, reconnecting }, connectedNow) => ({
+          connectedPreviously: connectedPreviously || connectedNow,
+          reconnecting: connectedPreviously && !connectedNow,
+        }),
+        { connectedPreviously: false, reconnecting: false },
+      ),
+      map(({ reconnecting }) => reconnecting),
+    ),
```
toger5: I am not sure if this is the best way to do this.
To make sure I understand correctly:
 - We implicitly track membershipsa based on the `RoomStateEvent.Members` and `MatrixRTCSessionEvent.MembershipsChanged` events
 - We use this to check our own state on each change and update the reconnecting state based on that.

In the membership manager we do a very similar logic whenever we recalculate the memebrs list. If we are missing we go into reconnection state.

Having this logic twise (once here and once in the js-sdk) feels dangerus since we might have different values based on the order of exacution which is tricky to predict with the observables + event emitter setup.

Can we read this valie from the rtc session memebership manager state?

src/state/CallViewModel.ts
```diff
@@ -1497,5 +1529,44 @@ export class CallViewModel extends ViewModel {
...
+
+    // Pause all media tracks when we're disconnected from MatrixRTC, because it
+    // can be an unpleasant surprise for the app to say 'reconnecting' and yet
+    // still be transmitting your media to others.
+    this.matrixRTCConnected$.pipe(this.scope.bind()).subscribe((connected) => {
+      const publications =
+        this.livekitRoom.localParticipant.trackPublications.values();
+      if (connected) {
+        for (const p of publications) {
+          if (p.track?.isUpstreamPaused === true) {
```
toger5, line 1604: in the vm filter + forEach might fit better since it is more aligned with the observable syntax? wdyt? (`for` `if` also is nice because its so simple however...)

```suggestion
        publications
            .filter((p) => p.track?.isUpstreamPaused)
            .forEach((p) => {
            ...
            });
```

toger5 (changes requested):
Looks very nice. The only part that needs additional consideration is the bit on how we aquire the reconnection state.

As mentioned in the comment, I feel like we should use the one already existing value from the membership manager.

src/state/CallViewModel.ts
robintown: You'll notice that the `map` operation on the line just after this doesn't depend on the value of the input Observable at all - the `null` here is just a dummy value used to tell the `map` to emit an initial value, like a Behavior would. (So the refactor makes no changes in behavior to `memberships$` - it was just to clarify that the `startWith` value is unused.)

src/state/CallViewModel.ts
robintown: It's a micro-optimization for sure, but I like avoiding the overhead of the filtered array in this case. I don't feel strongly about it though.

src/state/CallViewModel.ts
robintown: Your understanding is correct. Do you consider this solved by the introduction of `probablyLeft`, or would you like the membership manager's `Status` to also be re-emitted by MatrixRTCSession and then we depend on that?

I do think it's important to check the current membership status in some way or another in addition to `probablyLeft`, because we should react to cases where another member overwrites our state event due to the lack of state protection, for example.

src/state/CallViewModel.ts
toger5: oh of course.

src/state/CallViewModel.ts
toger5: Yes you are right. The probably left is an alternative to this.
But checking for both also sounds sensible, we might even want to have different error screens (one is more like a lost connection and the other one a reconnecting state)

And for the reconnecting the best source of truth indeed is the membership managers status.

toger5: I will [create another PR](https://github.com/element-hq/element-call/pull/3453#discussion_r2288398352) to expose the status event through the session.

This allow us to do `status === Status.Reconnecting` and `probablyLeft === true` to compose the error message observable.
`(reconncting ? "reconnecting" : (probablyLeft ? "connection lost" : undefined))`

src/state/CallViewModel.ts
toger5: lets keep it as is.

src/room/CallEventAudioRenderer.test.tsx
```diff
@@ -66,64 +67,64 @@ beforeEach(() => {
...
     mockRtcMemberships,
   );
 
   render(<CallEventAudioRenderer vm={vm} />);
   expect(playSound).not.toBeCalled();
   act(() => {
-    remoteRtcMemberships$.next(
-      mockRtcMemberships.slice(0, MAX_PARTICIPANT_COUNT_FOR_SOUND - 1),
+    rtcMemberships$.next(
+      mockRtcMemberships.slice(0, MAX_PARTICIPANT_COUNT_FOR_SOUND),
```
toger5, line 129: this change seems to not be tracked in the PR description/title?

src/room/InCallView.tsx
```diff
@@ -777,8 +790,15 @@ export const InCallView: FC<InCallViewProps> = ({
...
+      <Toast
+        onDismiss={onDismissReconnectingToast}
+        open={reconnecting}
+        portal={false}
+      >
+        {t("common.reconnecting")}
+      </Toast>
       <EarpieceOverlay
-        show={earpieceMode}
+        show={earpieceMode && !reconnecting}
```
toger5, line 801: good catch. Thats an easy one to miss.

src/state/CallViewModel.ts
```diff
@@ -403,7 +414,7 @@ export class CallViewModel extends ViewModel {
   private readonly remoteParticipantHolds$ = this.scope.behavior<
```
toger5, line 494: Irrelevant for this PR but a general question:
 - do we still need to concept of holds?
    - Tiles are based on the matrix room state memberhsips -> sfu switches should be trivial (we still show the tile in connection state until we find the media on the new sfu)
    - they will be hold on screen anyways even if lk connection is lost due to being based on member events.

src/state/CallViewModel.ts
```diff
@@ -468,12 +479,73 @@ export class CallViewModel extends ViewModel {
...
-    fromEvent(this.matrixRTCSession.room, RoomStateEvent.Members),
+    fromEvent(this.matrixRoom, RoomStateEvent.Members),
   ).pipe(
-    startWith(this.matrixRTCSession.memberships),
-    map(() => {
-      return this.matrixRTCSession.memberships;
-    }),
+    startWith(null),
+    map(() => this.matrixRTCSession.memberships),
+  );
```
toger5: Based on the codes snippet below (taken from the matrixRTCSession file) I think we dont need the room members event listener:

```suggestion
fromEvent(this.matrixRTCSession, MatrixRTCSessionEvent.MembershipsChanged),
).pipe(
    startWith(null),
    map(() => this.matrixRTCSession.memberships),
  );
```

reference:
```

    /**
     * Call this when the Matrix room members have changed.
     */
    public onRoomMemberUpdate = (): void => {
        this.recalculateSessionMembers();
    };

    /**
     * Call this when something changed that may impacts the current MatrixRTC members in this session.
     */
    public onRTCSessionMemberUpdate = (): void => {
        this.recalculateSessionMembers();
    };

    /**
     * Call this when anything that could impact rtc memberships has changed: Room Members or RTC members.
     *
     * Examines the latest call memberships and handles any encryption key sending or rotation that is needed.
     *
     * This function should be called when the room members or call memberships might have changed.
     */
    private recalculateSessionMembers = (): void => {
        const oldMemberships = this.memberships;
        this.memberships = MatrixRTCSession.sessionMembershipsForRoom(this.room, this.sessionDescription);

        this._callId = this._callId ?? this.memberships[0]?.sessionDescription.id;

        const changed =
            oldMemberships.length != this.memberships.length ||
            oldMemberships.some((m, i) => !CallMembership.equal(m, this.memberships[i]));

        if (changed) {
            this.logger.info(
                `Memberships for call in room ${this.roomSubset.roomId} have changed: emitting (${this.memberships.length} members)`,
            );
            logDurationSync(this.logger, "emit MatrixRTCSessionEvent.MembershipsChanged", () => {
                this.emit(MatrixRTCSessionEvent.MembershipsChanged, oldMemberships, this.memberships);
            });

```

src/state/CallViewModel.ts
```diff
@@ -468,12 +479,73 @@ export class CallViewModel extends ViewModel {
...
+
+  private readonly matrixConnected$ = this.scope.behavior(
```
toger5: This now just becomes 
```
this.scope.behavior(
fromEvent(session, MembershipManagerEvent.StatusChanged).pipe(
  startWith([undefined, State.Unknown]),
  map((prev,current)=>current)
)
```

src/state/CallViewModel.ts
```diff
@@ -468,12 +479,73 @@ export class CallViewModel extends ViewModel {
...
+        map(([state]) => state === SyncState.Syncing),
+      ),
+      // We can see our own call membership
+      this.memberships$.pipe(
+        map((ms) =>
+          ms.some(
+            (m) => m.sender === this.userId && m.deviceId === this.deviceId,
+          ),
+        ),
+      ),
```
toger5: This now just becomes:
this.scope.behavior(
fromEvent(session, MembershipManagerEvent.StatusChanged).pipe(
  startWith(null),
  map(()=>session.membershipStatus===Status.Reconnecting)
)

is this correct?

toger5: This also would be the better condisiont since we otherwiese might end up with a connection overlay on initial join.

src/state/CallViewModel.ts
```diff
@@ -468,12 +479,73 @@ export class CallViewModel extends ViewModel {
...
+      // connection but are now disconnected
+      scan(
+        ({ connectedPreviously, reconnecting }, connectedNow) => ({
+          connectedPreviously: connectedPreviously || connectedNow,
+          reconnecting: connectedPreviously && !connectedNow,
+        }),
+        { connectedPreviously: false, reconnecting: false },
+      ),
+      map(({ reconnecting }) => reconnecting),
+    ),
```
toger5: This `scan` of course allows us to just use the connection condition without the specific reconnection logic.
But maybe we can just skip this observable and instead make `matrixConnected` -> `matrixReconnecting` by inverting the logic and using the `membershipStatus`.

It does not work that well with the sync bool observable...

Maybe keeping this and changing the above to:
```
session.mambershipStatus !== connected
```
Makes more sense?

src/state/CallViewModel.ts
```diff
@@ -1486,5 +1560,44 @@ export class CallViewModel extends ViewModel {
     >,
   ) {
     super();
+
+    // Pause all media tracks when we're disconnected from MatrixRTC, because it
+    // can be an unpleasant surprise for the app to say 'reconnecting' and yet
+    // still be transmitting your media to others.
+    this.matrixConnected$.pipe(this.scope.bind()).subscribe((connected) => {
```
toger5, line 1599: shouldnt this be `reconnecting$`?

src/state/MediaViewModel.ts
```diff
@@ -611,6 +626,23 @@ export class RemoteUserMediaViewModel extends BaseUserMediaViewModel {
     ),
   );
 
+  /**
+   * The local volume, taking into account whether we're supposed to pretend
+   * that the audio stream is disconnected (since we don't necessarily want that
+   * to modify the UI state).
+   */
+  private readonly actualLocalVolume$ = this.scope.behavior(
+    this.pretendToBeDisconnected$.pipe(
```
robintown: Let's move this inline to where we actually subscribe to the thing

src/state/CallViewModel.ts
robintown: Still need to update the PR to use the status events

src/state/CallViewModel.ts
robintown: We can consider this again once the Status.Reconnecting state is actually used in matrix-js-sdk

src/state/CallViewModel.ts
robintown: Turns out we do still need to react to this event because we need to observe display name changes (and those are not part of the call membership data itself)

src/room/CallEventAudioRenderer.test.tsx
robintown: Added a comment. This was about db59679ad4872f4f282e96309212e80f77ca3649 which causes the local RTC member to be included in all these test membership arrays and therefore the length of this array went up by 1.

toger5 (approved):
wonderful

PR #3464, opened 2025-08-22
Test sync loop status and membership status in reconnection test as well

robintown:
We skipped testing these things ahead of the RC due to time constraints, so I'm going back and adding them. First commit is a refactor of the test file.


toger5 (approved):
This looks very great. Absolutly liking the refactor!! Not only the optional part but also that we have the field names as part of the function now. It was hard to remember the meaning of first, second... field of the.

PR #3466, opened 2025-08-25
Add dialing/ringing state to CallViewModel (`callPickupState$`)

toger5:
This adds a new public behavior to the call view model that represents the current waiting state when this client initiated the call. This state is described in the doc comment:
```
/**
   * The current call pickup state of the call.
   *  - "ringing": The call is ringing on other devices in this room (This client should give audiovisual feedback that this is happening).
   *  - "unknown": The client has not yet sent the notification event. We don't know if it will because it first needs to send its own membership.
   *     Then we can conclude if we were the first one to join or not.
   *  - "timeout": No-one picked up in the defined time this call should be ringing on others devices.
   *     The call failed. If desired this can be used as a trigger to exit the call.
   *  - "success": Someone else joined. The call is in a normal state. Stop audiovisual feedback.
   *  - null: EC is configured to never show any waiting for answer state.
   */
```

It will be used by the view to render a "waiting" screen and play a "dialing/waiting sound" (not part of this PR)

It is missing any decline state which will be added in a seperate PR.


src/UrlParams.ts
```diff
@@ -216,6 +216,17 @@ export interface UrlConfiguration {
...
+  /**
+   * If the client should behave like it is awaiting an answer if a notification was sent (wait for call pick up).
+   * This is a no-op if not combined with sendNotificationType.
+   *
+   * This entails:
+   *  - show ui that it is awaiting an answer
+   *  - play a sound that indicates that it is awaiting an answer
+   *  - auto-dismiss the call widget once the notification lifetime expires on the receivers side.
+   */
+  shouldWaitForCallPickup: boolean;
```
BillCarsonFr: This is hard to find a correct naming for that :/
If I get it right it is when I start a call that will ring on the other side. And we want like a phone call experience, it should show the other user avatar/name, should give a ringing feedback, and gave up after a timeput

src/room/__snapshots__/InCallView.test.tsx.snap
```diff
@@ -49,6 +49,34 @@ exports[`InCallView > rendering > renders 1`] = `
               </svg>
             </span>
           </div>
+          <div
```
BillCarsonFr, line 52: Why is there this change?

src/UrlParams.ts
```diff
@@ -216,6 +216,17 @@ export interface UrlConfiguration {
...
   autoLeaveWhenOthersLeft: boolean;
+
+  /**
+   * If the client should behave like it is awaiting an answer if a notification was sent (wait for call pick up).
+   * This is a no-op if not combined with sendNotificationType.
+   *
+   * This entails:
+   *  - show ui that it is awaiting an answer
+   *  - play a sound that indicates that it is awaiting an answer
+   *  - auto-dismiss the call widget once the notification lifetime expires on the receivers side.
```
BillCarsonFr, line 227: Not sure to understand that, what does this mean "expires on the recevievers side?" Like the receiver should send some expiration signal?? Or is it a local timeout for when to stop trying to call?

src/room/__snapshots__/InCallView.test.tsx.snap
robintown: I believe it's because, now that the participant count has been moved to the view model, this test is properly simulating the participant count rather than always passing `participantCount={0}` as a prop.

src/UrlParams.ts
robintown: My understanding is that it's a local timeout - we get to specify a duration when sending the calling event, so if we assume that the receiver respects it, we can use a local timer to approximate when their ringing stops and we should give up.

src/UrlParams.ts
robintown: Right. A nitpick: we could simplify the name to `waitForCallPickup`. Or to clarify that the ringing is happening on the other side, perhaps `waitForRemoteAnswer`.

src/state/CallViewModel.ts
```diff
@@ -427,7 +435,14 @@ export class CallViewModel extends ViewModel {
...
         startWith(null),
-        map(() => this.matrixRTCSession.membershipStatus === Status.Connected),
+        map(
+          () =>
+            (
+              this.matrixRTCSession as unknown as {
+                membershipStatus?: Status;
+              }
+            ).membershipStatus === Status.Connected,
+        ),
```
robintown: Why the casts? (also on line 456)

src/state/CallViewModel.ts
```diff
@@ -825,49 +844,68 @@ export class CallViewModel extends ViewModel {
...
+        },
+        { ids: [], joined: [], left: [] },
+      ),
     ),
   );
 
+  /**
+   * The number of participants currently in the call.
+   *
+   *  - Each participant has one livekit connection
```
robintown: Not sure what this means

src/state/CallViewModel.ts
```diff
@@ -825,49 +844,68 @@ export class CallViewModel extends ViewModel {
...
+   *  - Each participant has a corresponding MatrixRTC membership state event
+   *  - There can be multiple participants for one matrix user.
```
robintown:
```suggestion
   *  - There can be multiple participants for one Matrix user if they join from multiple devices.
```

robintown (commented):
Here's my initial review - still need to look at the newly added behaviors in more detail.

src/state/CallViewModel.test.ts
```diff
@@ -1228,6 +1231,234 @@ test("autoLeaveWhenOthersLeft$ doesn't emits when autoLeaveWhenOthersLeft option
...
 });
 
+describe("shouldWaitForCallPickup$", () => {
+  test("unknown -> ringing -> timeout when notified and nobody joins", () => {
+    withTestScheduler(({ hot, schedule, expectObservable, scope }) => {
+      // No one ever joins (only local user)
+      withCallViewModel(
+        {
+          remoteParticipants$: scope.behavior(hot("a", { a: [] }), []),
+          rtcMembers$: scope.behavior(hot("a", { a: [localRtcMember] }), []),
```
robintown: There's a `behavior` test helper that we can use here instead of `scope.behavior` + `hot`.

src/state/CallViewModel.test.ts
```diff
@@ -1228,6 +1231,234 @@ test("autoLeaveWhenOthersLeft$ doesn't emits when autoLeaveWhenOthersLeft option
...
+          connectionState$: of(ConnectionState.Connected),
+          speaking: new Map(),
+          mediaDevices: mockMediaDevices({}),
```
robintown: A lot of these arguments can be omitted since they're the same as the default values

src/state/CallViewModel.test.ts
```diff
@@ -1228,6 +1231,234 @@ test("autoLeaveWhenOthersLeft$ doesn't emits when autoLeaveWhenOthersLeft option
...
+        },
+      );
+    });
+  });
+
+  test("ringing -> success if someone joins before timeout", () => {
+    withTestScheduler(({ hot, schedule, expectObservable, scope }) => {
+      // Someone joins at 20ms (both LiveKit participant and MatrixRTC member)
+      const remote$ = scope.behavior(
+        hot("a--b", { a: [], b: [aliceParticipant] }),
```
robintown: Let's use numerical durations in the marble diagrams if we're referencing such a duration in the comment

src/state/CallViewModel.test.ts
```diff
@@ -1228,6 +1231,234 @@ test("autoLeaveWhenOthersLeft$ doesn't emits when autoLeaveWhenOthersLeft option
...
+                  event_id: string;
+                } & ICallNotifyContent,
+              );
+            },
+          });
+
+          expectObservable(vm.callPickupState$).toBe("a 2ms c", {
+            a: "unknown",
+            c: "success",
+          });
```
robintown: This result looks wrong, there's no ringing state

src/state/CallViewModel.test.ts
```diff
@@ -1228,6 +1231,234 @@ test("autoLeaveWhenOthersLeft$ doesn't emits when autoLeaveWhenOthersLeft option
...
+                {} as unknown as {
+                  event_id: string;
+                } & ICallNotifyContent,
+              );
+            },
+          });
+          expectObservable(vm.callPickupState$).toBe("a 1ms b", {
+            a: "unknown",
+            b: "success",
+          });
```
robintown: Also doesn't match the description

src/state/CallViewModel.test.ts
```diff
@@ -1228,6 +1231,234 @@ test("autoLeaveWhenOthersLeft$ doesn't emits when autoLeaveWhenOthersLeft option
...
+          mediaDevices: mockMediaDevices({}),
+        },
+        (vm, rtcSession) => {
+          schedule("          10ms r", {
+            r: () => {
+              rtcSession.emit(
+                MatrixRTCSessionEvent.DidSendCallNotification,
+                { lifetime: 0 } as unknown as {
+                  event_id: string;
+                } & IRTCNotificationContent, // no lifetime
```
robintown: Should we actually fully omit the lifetime field here to test this?

robintown (commented):
The new behaviors look correct, though I'd like to try my hand at rewriting them into a slightly more idiomatic RxJS style. Some notes on the tests as well.

src/UrlParams.ts
robintown: I pushed a commit changing this to `waitForCallPickup` for now. If anyone has a preference for something else, let me know.

robintown (approved):
Since Timo is out, I fixed up my own comments. @BillCarsonFr Would you like to review again, or do we just consider this done?

src/room/CallEventAudioRenderer.test.tsx
```diff
@@ -172,7 +170,7 @@ test("should not play a sound when a hand raise is retracted", () => {
       },
     });
   });
-  expect(playSound).toHaveBeenCalledTimes(2);
+  expect(playSound).toHaveBeenCalledTimes(1);
```
BillCarsonFr: What is triggering this behavior change?  looks unrelated to the current change?
Or maybe this test should not depend on that magic number? maybe reset the mock before the calling the `next` and ensure that it was not called

BillCarsonFr (approved):
If you can clarify the behavior change on the raise hand test

src/room/CallEventAudioRenderer.test.tsx
robintown: It seems that previously, both a join sound and a hand raise sound would be triggered in this test. Since I refactored the join sound effect behaviors, I changed them so that they no longer consider the initial value of the participant list to be a change in participants, which I think is more correct. Therefore it's not playing the join sound here anymore.

I did double check that we still hear a join sound in practice when joining calls - that still happens because the local participant only appears after we have successfully sent the session membership event.

But yeah, I agree that this test is better off if the number only reflects any sounds that are due to the `act` block.

PR #3467, opened 2025-08-25
View model for decline logic

toger5:
This extends the `callPickupState` with a `"decline"` state that occurs once a decline event is received.


src/state/CallViewModel.ts
```diff
@@ -935,26 +937,35 @@ export class CallViewModel extends ViewModel {
    * "ringing": The notification event was sent.
    * "ringEnded": The notification events lifetime has timed out -> ringing stopped on all receiving clients.
    */
-  private readonly notificationEventIsRingingOthers$: Observable<
-    "unknown" | "ringing" | "ringEnded" | null
-  > = fromEvent<[IRTCNotificationContent, ICallNotifyContent]>(
-    this.matrixRTCSession,
-    MatrixRTCSessionEvent.DidSendCallNotification,
-  ).pipe(
+  private readonly rtcNotificationEventState$: Observable<
```
fkwp: what happens if the notification type is notify and not ringing? Is that covered by unknown or do we need a none state?

src/state/CallViewModel.test.ts
```diff
@@ -1249,10 +1256,12 @@ describe("shouldWaitForCallPickup$", () => {
             r: () => {
               rtcSession.emit(
                 MatrixRTCSessionEvent.DidSendCallNotification,
-                { lifetime: 30 } as unknown as {
+                { event_id: "$notif1", lifetime: 30 } as unknown as {
```
fkwp: what is the time unit ms or seconds?

src/state/CallViewModel.ts
```diff
@@ -935,26 +937,35 @@ export class CallViewModel extends ViewModel {
...
-          of<"ringing" | null>("ringing"),
-          timer(lifetime).pipe(map((): "ringEnded" | null => "ringEnded")),
+          of({
+            state: "ringing",
+            event_id: notificationEvent.event_id,
+          } as {
+            state: "ringing";
+            event_id: string;
+          }),
+          timer(lifetime).pipe(
```
fkwp: should we add the time unit to lifetime? lifetime_ms

src/state/CallViewModel.test.ts
robintown: Milliseconds. There's a lot of precedent in the Matrix spec for keeping the time units of duration fields implicit (for example see [m.call.invite](https://spec.matrix.org/unstable/client-server-api/#mcallinvite)) and only elaborating in the docs. So I personally feel fine naming the event content field this way.

src/state/CallViewModel.ts
robintown: Added insurance against this in 2541f810

robintown (approved):
I've reviewed this PR and pushed the changes that I would've suggested, since Timo is away (it's pretty much the same stuff as on the last pickup state PR, continued).

src/state/CallViewModel.ts
```diff
@@ -887,17 +891,59 @@ export class CallViewModel extends ViewModel {
...
+        NEVER,
+      ).pipe(
+        takeUntil(
+          (
+            fromEvent(this.matrixRoom, RoomEvent.Timeline) as Observable<
+              Parameters<EventTimelineSetHandlerMap[RoomEvent.Timeline]>
+            >
+          ).pipe(
+            filter(
+              ([event]) =>
```
BillCarsonFr, line 936: What is exactly the expected logic here?
- Are we trying to check if the call was declined by another of our device? and then stop ringing here
- Or is it that we are the caller and trying to see if one/all of the callee have declined?

So should we check the decline event sender? if we are in a DM or not?

BillCarsonFr (commented):
I have some questions. Not sure to see what case we are handling? declining from another of our sessions? or we are calling and the person we call did decline? (in that case we should stop the call?)

src/state/CallViewModel.ts
robintown: It's the latter - This change is about making it so that, when you start a call in a DM and send a ringing notification to the other user, they can decline your notification and EC will stop playing the dial tone.

> (in that case we should stop the call?)

Good question, I checked the design shown [here](https://www.figma.com/design/NKjLzSlbD8JnldYQ5sFpce/Element-Call-Updates--New----Q1-2025) and it seems to have persistent views which say "declined" or "no answer", leaving it up to the user to dismiss them manually once they've read the message. So I believe not.

> So should we check the decline event sender? if we are in a DM or not?

For now I'd like to assume that this behavior is always being used for calls in DMs. I just don't know what a group call dial + "wait for pickup" experience would look like, or whether such a feature would ever end up on our roadmap yet.

A simple check that the decline event sender is, in fact, another user doesn't sound like a bad idea. I guess it would protect us from mis-attributing a stray decline event sent by another one of the user's own buggy clients to the to receiving party. (Like, it would protect us from telling Alice "Bob declined the call" if in fact it was a weird Matrix client on Alice's other laptop that sent the decline event.)

toger5:
@robintown thanks for working on this and pushing it over the line. I checked the changes and they all make a lot of sense to me.

PR #3471, opened 2025-08-29
OAuth 2.0 login / MAS for Element Call SPA

AndrewFerr:
TODO:
- [ ] support for guest accounts in MAS (https://github.com/element-hq/matrix-authentication-service/issues/1445)
- [ ] fix the post-login redirect page to load properly (currently requires a manual refresh after a few seconds)
- [ ] use the refresh token (if returned) to handle expiring access tokens (otherwise requests will fail once the access token expires)
- [ ] support logging out
- [ ] support account registration

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,7 @@ A docker compose file `dev-backend-docker-compose.yml` is provided to start the
 whole stack of components which is required for a local development environment:
 
 - Minimum Synapse Setup (servername: `synapse.m.localhost`)
+- Matrix Authentication Service Setup (issuer: `mas.m.localhost`)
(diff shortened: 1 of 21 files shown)


robintown:
We're unfortunately not sure at this point in time that MAS is the right fit for the Element Call SPA. One of the issues is that MAS will of course prompt users to log in with existing accounts if they have them available, and we do not have interest in adding cross signing and verification flows to the Element Call SPA. We really would prefer for the accounts used to be "burner" accounts linked to a single device.

This PR may still be useful in the future if we reevaluate this, of course.

PR #3474, opened 2025-09-03
Use a PNG for the background gradient for better performance

sandhose:
I had the issue in another project with this exact same background. The SVG is really expensive to render, as it's a gaussian blur of 200px. This is noticeable on loading, but especially when resizing the window. This is quite bad on safari, but especially on firefox.

Here's a recording before/after on safari:

https://github.com/user-attachments/assets/50c8757d-30aa-40e3-989d-90fdfd926053


toger5:
Thanks @sandhose. This looks way better indeed. It does look very different to the svg version so this needs a design review to see if the png version is desired.

sandhose:
Is it really that different? It's closer to the Figma designs, and on safari the rendering differences are negligible. I think the biggest difference is in Firefox, but I would say that the SVG version in Firefox is the one currently most far off the designs?

toger5:
Yes I was testing on FF.
The PNG also seems a bit grainy. Is that intentional?

sandhose:
> The PNG also seems a bit grainy. Is that intentional?

It is very low res; I can probably double the resolution if it's too noticeable

bmarty:
My 2 cents: I prefer SVG as it's a vector format. Would it be possible, to mitigate the white stripes when increasing the browser window size to apply another plain dark color background? Or maybe render the background to 120% (for instance) of the window size and centered so that we have room to increase the browser size without viewing the white stripes?

toger5:
I am wondering what we do with this.
I like the perf benefit. But I think the rendering only happens on resize so its not that big of a deal from a resource consumption point of view. (its not happening all the time)

I think what would have a bigger impact on how it feels is making the background behind the picture also black?
The png is also larger (1kb vs 60kb) so both very small...

Johennes:
> I think what would have a bigger impact on how it feels is making the background behind the picture also black?

I had a very brief look at this but it seems like the white background is part of the browser chrome? I couldn't find a way to change it.

sandhose:
> > I think what would have a bigger impact on how it feels is making the background behind the picture also black?
> 
> I had a very brief look at this but it seems like the white background is part of the browser chrome? I couldn't find a way to change it.

I also don't think it's really a 'do we see the white background' problem but rather a general paint performance problem. The SVG is really expensive to paint, the PNG isn't

sandhose:
> The png is also larger (1kb vs 60kb) so both very small...

Since AVIF is quite well supported, I've included an AVIF version as well now (with PNG fallback), which is 3.2KiB

reivilibre:
> > I think what would have a bigger impact on how it feels is making the background behind the picture also black?
> 
> I had a very brief look at this but it seems like the white background is part of the browser chrome? I couldn't find a way to change it.

FWIW (as agreed with Q that the perf problem is the main problem), my experience is that the white background can be changed using `:root` or `html` CSS selector, setting a `background-color` onto it, can't remember which. This is my experience with the 'elastic overscrolling' effect that Firefox recently gained and that you see a lot on mobile; to be honest I don't know if it would apply to a laggy relayout.

Johennes:
> I also don't think it's really a 'do we see the white background' problem but rather a general paint performance problem. The SVG is really expensive to paint, the PNG isn't

I wonder if there is a way to quantify the impact? I have rather poor vision but FWIW I don't really _feel_ any difference when loading the frontend in Firefox on this branch vs. the `livekit` branch. I briefly tried profiling the page load but the time from `FirstContentfulPaint` to `LargestContentfulPaint` appeared to be roughly 300ms in all cases.

robintown:
Picked this back up, but there were merge conflicts and I also found out that we can get the least visual artifacts by using even smaller PNGs than what this PR started out with. Therefore I've opened https://github.com/element-hq/element-call/pull/4180 instead.

PR #3486, opened 2025-09-10
Fix the interactivity of buttons while reconnecting or in earpiece mode

### LINKED ISSUE #3480: We do not consistently block interaction with tiles in background in earpiece mode
### Steps to reproduce

1. Enter earpiece mode

#### What did you expect?

Since the video tiles are now masked by the translucent overlay, I should not be able to interact with them.

#### What happened instead?

I can't interact with them with a pointing device, but I still can interact with them via keyboard navigation. (And also their contents will still be entirely visible to accessibility technologies.) We should at least choose a consistent behavior.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

### LINKED ISSUE #3481: 'Reconnecting' toast blocks interaction with header and footer buttons
### Steps to reproduce

1. Join a call
2. Get disconnected from the Internet
3. Wait for the 'reconnecting' toast to appear
4. Try to click the hangup button

#### What did you expect?

I can click the hangup button.

#### What happened instead?

It doesn't respond to input because the toast has captured focus and set `pointer-events: none;` on all other elements in the application.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

Yes

robintown:
When we're in one of these modes, we need to ensure that everything above the overlay (the header and footer buttons) is interactive, while everything obscured by the overlay (the media tiles) is non-interactive and removed from the accessibility tree. It's not a very easy task to trap focus *outside* an element, so the best solution I could come up with is to set `tabindex="-1"` manually on all interactive elements belonging to the media tiles.

Closes https://github.com/element-hq/element-call/issues/3481
Closes https://github.com/element-hq/element-call/issues/3480


robintown:
I need to do more debugging of the Playwright test to understand what error it's getting stuck on, but this is otherwise ready to be reviewed.

fkwp (approved):
lgtm

toger5 (approved):

PR #3488, opened 2025-09-11
Do not use preload mode by default in embedded mode

### LINKED ISSUE #3487: Unhelpful white screen of death when using a DM intent (e.g. `start_call_dm`) but preload is `false`
### Steps to reproduce

1. Start a call with the URL parameters `preload=false` but `intent=start_call_dm`

#### What did you expect?

An error, as this is an invalid configuration. Or, the lobby to be shown anyway.

#### What happened instead?

A white screen

### Operating system

N/A

### Browser information

N/A

### URL for webapp

local / develop

### Will you send logs?

No

Half-Shot:
Fixes #3487 

This changes `preload` to false, which means that the widget will *not* wait for the host application to indicate it's ready (via widget action) before trying to join the call. Myself and @toger5 think it's no longer required for our applications. This can still be enabled by clients that need it.

I've also cleaned up some of the boolean logic to help understand what's going on, as well as pulling through some types and adding more comments.


src/room/GroupCallView.tsx
```diff
@@ -83,8 +83,8 @@ interface Props {
   client: MatrixClient;
   isPasswordlessUser: boolean;
   confineToRoom: boolean;
-  preload: boolean;
-  skipLobby: boolean;
+  preload: UrlParams["preload"];
+  skipLobby: UrlParams["skipLobby"];
```
Half-Shot, line 87: Probably a preference thing but this is a nice way (imo) to indicate where this value comes from, rather than having to duplicate docs.

robintown (approved):

PR #3489, opened 2025-09-11
Add "wait for pickup" overlay with sound. Leave on decline/timeout

toger5:
Sounds are reusing existing audio as a stop-gap.

- [x] show dialing animation
- [x] play dialing sound
- [ ] show timeout screen _just close EC without interactive message in this iteration_
- [ ] play timeout sound _only play sound when the other user declines_
- [ ] show decline screen (and auto quit EC) _only doing the auto quit of EC_

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -117,7 +117,10 @@ import waitingStyles from "./WaitingForJoin.module.css";
 import { prefetchSounds } from "../soundUtils";
 import { useAudioContext } from "../useAudioContext";
 // TODO: Dont use this!!!  use the correct sound
-import { GenericReaction } from "../reactions";
+import genericSoundOgg from "../sound/reactions/generic.ogg?url";
+import genericSoundMp3 from "../sound/reactions/generic.mp3?url";
+import leftCallSoundMp3 from "../sound/left_call.mp3";
+import leftCallSoundOgg from "../sound/left_call.ogg";
 
 const canScreenshare = "getDisplayMedia" in (navigator.mediaDevices ?? {});
 
@@ -273,18 +276,18 @@ export const InCallView: FC<InCallViewProps> = ({
   const muteAllAudio = useBehavior(muteAllAudio$);
   // Call pickup state and display names are needed for waiting overlay/sounds
   const callPickupState = useBehavior(vm.callPickupState$);
-  const displaynames = useBehavior(vm.memberDisplaynames$);
-  // Preload a subtle waiting/ringing tone
-  const [waitingSoundCache, setWaitingSoundCache] = useState<ReturnType<
-    typeof prefetchSounds
-  > | null>(null);
-  useEffect(() => {
-    if (!waitingSoundCache && callPickupState === "ringing") {
-      setWaitingSoundCache(prefetchSounds({ waiting: GenericReaction.sound! }));
-    }
-  }, [waitingSoundCache, callPickupState]);
-  const waitingAudio = useAudioContext({
-    sounds: waitingSoundCache,
+
+  // Preload a waiting and decline sounds
+  const pickupPhaseSoundCache = useInitial(async () => {
+    return prefetchSounds({
+      waiting: { mp3: genericSoundMp3, ogg: genericSoundOgg },
+      decline: { mp3: leftCallSoundMp3, ogg: leftCallSoundOgg },
+      // do we want a timeout sound?
+    });
+  });
+
+  const pickupPhaseAudio = useAudioContext({
+    sounds: pickupPhaseSoundCache,
     latencyHint: "interactive",
     muted: muteAllAudio,
   });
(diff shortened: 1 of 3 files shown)


src/room/InCallView.tsx
```diff
@@ -328,6 +354,76 @@ export const InCallView: FC<InCallViewProps> = ({
   const audioOutputSwitcher = useBehavior(vm.audioOutputSwitcher$);
   useSubscription(vm.autoLeave$, onLeave);
 
+  // When we enter timeout or decline we will leave the call.
+  useEffect((): void | (() => void) => {
+    if (callPickupState === "timeout") {
+      onLeave();
+    }
+    if (callPickupState === "decline") {
+      void pickupPhaseAudio
```
Half-Shot, line 368: I'm undecided on this but, a comment might be nice to say the sound needs awaiting before onLeave can be called or the iframe will unload and the sound will cut.

src/room/InCallView.tsx
```diff
@@ -328,6 +354,76 @@ export const InCallView: FC<InCallViewProps> = ({
...
+        ?.playSound("decline")
+        .then(() => {
```
Half-Shot: Presumably, also this needs moving after catch. 
```suggestion
        .finally(() => {
```

Otherwise a failed sound will prevent a leave.

src/room/InCallView.tsx
```diff
@@ -328,6 +354,76 @@ export const InCallView: FC<InCallViewProps> = ({
...
+  });
+
+  // When waiting for pickup, loop a waiting sound
+  useEffect((): void | (() => void) => {
+    if (callPickupState !== "ringing") return;
+    // play immediately and then every ~2.5s while in ringing
+    void pickupPhaseAudio?.playSound("waiting");
+    const id = window.setInterval(
+      () => void pickupPhaseAudio?.playSound("waiting"),
+      2500,
```
Half-Shot: So, couple of things from me:

 - We should `const` this so we don't need to fish for the value if the sound effect were to change.
 - A *better* way to define this would be the gap between tones, and only schedule the next sound *after* the previous sound finishes. This means we could change the sound (or offer users a way to customise the sound) without having to adjust the delay value here.

src/room/InCallView.tsx
```diff
@@ -328,6 +354,76 @@ export const InCallView: FC<InCallViewProps> = ({
...
+    const roomOthers = [
+      ...matrixRoom.getMembersWithMembership("join"),
+      ...matrixRoom.getMembersWithMembership("invite"),
+    ].filter((m) => m.userId !== client.getUserId());
+    // Yield if there are not other members in the room.
+    if (roomOthers.length === 0) return null;
+
+    const otherMember = roomOthers.length > 0 ? roomOthers[0] : undefined;
+    const isOneOnOne = roomOthers.length === 1 && otherMember;
+    const text = isOneOnOne
```
Half-Shot, line 403: I am slightly lost here, are we showing this for non-1to1 chats?

Half-Shot (changes requested):
Looks smashing, I've got a variety of thoughts!

src/room/InCallView.tsx
toger5: No but from a EC point of view we dont know. It is possible for EC to ring in groups as well. It just was the natural way to handle the case that there are more members. Not checking for `roomOthers.length` felt wrong too.

Treating EC like a react component which also does send the notification event it also works by sending a notification event in a group room so we should support the waiting screen. (even though the defaults are configured to not do this. But we cannot trust EX/EW to set the correct intent)

src/room/InCallView.tsx
toger5: I updated this bit.

> We should const this so we don't need to fish for the value if the sound effect were to change.

this i dont understand i am afraid.

src/room/InCallView.tsx
```diff
@@ -328,6 +354,83 @@ export const InCallView: FC<InCallViewProps> = ({
...
+  // When waiting for pickup, loop a waiting sound
+  const [playing, setPlaying] = useState(false);
+  useEffect((): void | (() => void) => {
+    if (callPickupState !== "ringing") return;
+    setPlaying(true);
+    // play and then wait for 1.5s
+    const playSoundAndWait = async (): Promise<void> => {
+      await pickupPhaseAudio?.playSound("waiting");
+      await new Promise((res) => setTimeout(res, 1500));
+    };
```
Half-Shot: Right, this is what I wanted! The `playSound` promise only resolves when the sound has finished playing which means the gap between tones will always be 1.5s, no matter how long the "waiting" sound is.

Sorry, this might have been the case previously but is more clear to me in this diff.

src/room/InCallView.tsx
Half-Shot: Continued in https://github.com/element-hq/element-call/pull/3489#discussion_r2343860557

src/room/InCallView.tsx
Half-Shot: Right, got it. So this is probably an edge case in reality but it's better to support something than error.

Half-Shot (approved):
This is satisfying to me :+1:

toger5:
I will try to give this one real world test with: https://github.com/element-hq/element-web/pull/30730 and then merge it

toger5:
I found a couple of small things. Ready to merge now.

fkwp (approved):
lgtm

PR #3490, opened 2025-09-12
Add sounds for ringing

Half-Shot:
Should be merged with #3489


toger5 (changes requested):
Those also need to be used in the `InCallView.ts`

Additinaly we need to implement a sound on timeout. right now there is none.

And on top of this we still play the user leave sound. once left. So we migth need to silence the leave member sound if its our own member that leaves?

src/state/CallViewModel.ts
```diff
@@ -983,6 +975,23 @@ export class CallViewModel extends ViewModel {
       )
     : constant(null);
 
+  public readonly leaveSoundEffect$ = combineLatest([
+    this.callPickupState$,
+    this.userMedia$,
+  ]).pipe(
+    // Until the call is successful, do not play a leave sound.
```
toger5, line 982: Can you add the comment on "null = not configured" here as well?

toger5 (approved):

toger5 (approved):
@Half-Shot and @toger5 discussed the 2* logic again. It was challenged due to its complexity. But the value we get from it:
 - perfectly synced animation
 - very easy to exchange audio files without custom silence

justify the complexity.

PR #3491, opened 2025-09-16
Prevent showing calling view when disconnected from Livekit.

Half-Shot:
Hopefully fixes the problem where we still show 1:1 calling UI while disconnected from livekit. This also refactors InCallView to use CallViewModel for disconnection state.

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -25,7 +25,11 @@ import useMeasure from "react-use-measure";
 import { type MatrixRTCSession } from "matrix-js-sdk/lib/matrixrtc";
 import classNames from "classnames";
 import { BehaviorSubject, map } from "rxjs";
-import { useObservable, useSubscription } from "observable-hooks";
+import {
+  useObservable,
+  useObservableEagerState,
+  useSubscription,
+} from "observable-hooks";
 import { logger } from "matrix-js-sdk/lib/logger";
 import { RoomAndToDeviceEvents } from "matrix-js-sdk/lib/matrixrtc/RoomAndToDeviceKeyTransport";
 import {
@@ -249,7 +253,7 @@ export const InCallView: FC<InCallViewProps> = ({
     useReactionsSender();
 
   useWakeLock();
-  const isDisconnected = useBehavior(vm.livekitDisconnected$);
+  const isDisconnected = useObservableEagerState(vm.livekitConnectionState$);
 
   // annoyingly we don't get the disconnection reason this way,
   // only by listening for the emitted event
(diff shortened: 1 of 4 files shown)


src/room/InCallView.tsx
```diff
@@ -249,19 +242,18 @@ export const InCallView: FC<InCallViewProps> = ({
...
     useReactionsSender();
 
   useWakeLock();
+  const isDisconnected = useBehavior(vm.livekitDisconnected$);
 
   // annoyingly we don't get the disconnection reason this way,
   // only by listening for the emitted event
-  if (connState === ConnectionState.Disconnected)
-    throw new ConnectionLostError();
+  if (isDisconnected) throw new ConnectionLostError();
```
toger5: I think we dont need to be so verbal in the vm.
Making `livekitConnectionState$` public and running the logic:
`if (livekitConnectionState === Disconnected) throw new ConnectionLostError();`
Is an acceptable amount of logic in the View.

src/state/CallViewModel.ts
```diff
@@ -465,6 +465,14 @@ export class CallViewModel extends ViewModel {
...
   );
 
+  public readonly livekitDisconnected$ = this.scope.behavior(
+    and$(
+      this.livekitConnectionState$.pipe(
+        map((state) => state === ConnectionState.Disconnected),
+      ),
+    ),
+  );
+
```
toger5: This can than get removed.

src/state/CallViewModel.ts
```diff
@@ -959,13 +968,20 @@ export class CallViewModel extends ViewModel {
...
+          switchMap(([disconnected, someoneElseJoined]) => {
+            if (disconnected) {
+              // Do not ring until we're connected.
+              return of("unknown" as const);
+            } else if (someoneElseJoined) {
+              return of("success" as const);
+            }
+            // Show the ringing state of the most recent ringing attempt.
+            return this.ring$.pipe(switchAll());
+          }),
```
toger5: also works with livekit connection state (and without livektiDisconnected). might need a distinct until changed.

toger5 (changes requested):
Cool. This looks like an overall good change. And fixing the double overlay is of course a good thing too.

src/room/InCallView.tsx
```diff
@@ -249,19 +246,18 @@ export const InCallView: FC<InCallViewProps> = ({
...
   header: headerStyle,
-  connState,
   onShareClick,
 }) => {
   const { t } = useTranslation();
   const { supportsReactions, sendReaction, toggleRaisedHand } =
     useReactionsSender();
 
   useWakeLock();
+  const isDisconnected = useObservableEagerState(vm.livekitConnectionState$);
```
toger5: This does not look right?
It probably wants to be `...useLivekitConnectionState$) !== Disconnected`

Half-Shot: oh duh!

toger5 (approved):

PR #3492, opened 2025-09-16
[WIP] Multi-SFU

toger5:
This is a draft to get a build from the multi sfu branch.

### DIFF
--- a/src/livekit/livekitSubscriptionRoom.ts
+++ b/src/livekit/livekitSubscriptionRoom.ts
@@ -1,0 +1,123 @@
+/*
+Copyright 2023, 2024 New Vector Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+Please see LICENSE in the repository root for full details.
+*/
+
+import {
+  ConnectionState,
+  type E2EEManagerOptions,
+  ExternalE2EEKeyProvider,
+  LocalVideoTrack,
+  Room,
+  type RoomOptions,
+} from "livekit-client";
+import { useEffect, useRef } from "react";
+import E2EEWorker from "livekit-client/e2ee-worker?worker";
+import { logger } from "matrix-js-sdk/lib/logger";
+import { type MatrixRTCSession } from "matrix-js-sdk/lib/matrixrtc";
+
+import { defaultLiveKitOptions } from "./options";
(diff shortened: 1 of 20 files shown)


src/livekit/livekitSubscriptionRoom.ts
```diff
@@ -0,0 +1,123 @@
+/*
+Copyright 2023, 2024 New Vector Ltd.
```
BillCarsonFr, line 2: 2025 ^^

src/main.tsx
```diff
@@ -60,9 +60,9 @@ if (fatalError !== null) {
 Initializer.initBeforeReact()
   .then(() => {
     root.render(
-      <StrictMode>
-        <App vm={new AppViewModel()} />
-      </StrictMode>,
+      // <StrictMode>
```
BillCarsonFr, line 63: why?

src/room/GroupCallView.tsx
```diff
@@ -73,6 +73,12 @@ import { muteAllAudio$ } from "../state/MuteAllAudioModel.ts";
 import { useAppBarTitle } from "../AppBar.tsx";
 import { useBehavior } from "../useBehavior.ts";
 
+/**
+ * If there already are this many participants in the call, we automatically mute
+ * the user.
+ */
+export const MUTE_PARTICIPANT_COUNT = 8;
```
BillCarsonFr, line 80: nit: Could have a better name? `AUTO_MUTE_THRESHOLD`

BillCarsonFr (commented):
I did a static review, but there is a lot going on here, including refactoring and moving things to viewmodels. It makes it hard to review given all the moving parts.

I can still see  multiple `TODO-MULTI-SFU`, I suppose that these are all pending items that must be resolved before thinking about merging ?

I am wondering if this PR is also invalidating most of the docs regarding SFU selection/election?

I don't think I see a feature flag to enable multi SFU? is there a plan to do that?

toger5:
> I can still see multiple TODO-MULTI-SFU, I suppose that these are all pending items that must be resolved before thinking about merging ?

This is still manly a fragment from the hackathon and not really a PR that is planned to get merged like this.
We would need to get the branch into a valid good shape. For actually merging this, we could consider packaging this into multiple PR's with a smaller more specific scope.

toger5:
Closing this in favor of creating a new one based on the rebase branch.

PR #3494, opened 2025-09-17
default mute states (unmuted!) in widget mode (embedded + intent)

toger5:
fixes: https://github.com/element-hq/voip-internal/issues/394

### DIFF
--- a/src/UrlParams.ts
+++ b/src/UrlParams.ts
@@ -228,13 +228,20 @@ export interface UrlConfiguration {
    */
   waitForCallPickup: boolean;
 }
+interface IntentDerivedConfiguration {
+  defaultAudioEnabled?: boolean;
+  defaultVideoEnabled?: boolean;
+}
 
 // If you need to add a new flag to this interface, prefer a name that describes
 // a specific behavior (such as 'confineToRoom'), rather than one that describes
 // the situations that call for this behavior ('isEmbedded'). This makes it
 // clearer what each flag means, and helps us avoid coupling Element Call's
 // behavior to the needs of specific consumers.
-export interface UrlParams extends UrlProperties, UrlConfiguration {}
+export interface UrlParams
+  extends UrlProperties,
+    UrlConfiguration,
+    IntentDerivedConfiguration {}
 
 // This is here as a stopgap, but what would be far nicer is a function that
 // takes a UrlParams and returns a query string. That would enable us to
@@ -343,8 +350,7 @@ export const getUrlParams = (
     ? UserIntent.Unknown
     : (parser.getEnumParam("intent", UserIntent) ?? UserIntent.Unknown);
   // Here we only use constants and `platform` to determine the intent preset.
-  let intentPreset: UrlConfiguration;
-  const inAppDefault = {
+  let intentPreset: UrlConfiguration = {
     confineToRoom: true,
     appPrompt: false,
     preload: false,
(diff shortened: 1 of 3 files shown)


src/UrlParams.ts
```diff
@@ -408,6 +405,22 @@ export const getUrlParams = (
       };
   }
 
+  const intentDerivedConfiguration: IntentDerivedConfiguration = {};
+  const desktopMobile = platform === "desktop" ? "desktop" : "mobile";
```
Half-Shot, line 427: Desktop was confusing me because on web we have separate platforms for "web" and "desktop". I assume this is effectively "mobile" or "other".

src/UrlParams.ts
```diff
@@ -362,31 +368,22 @@ export const getUrlParams = (
   };
   switch (intent) {
     case UserIntent.StartNewCall:
-      intentPreset = {
-        ...inAppDefault,
-        skipLobby: true,
-      };
+      intentPreset.skipLobby = true;
```
Half-Shot, line 387: ooi, struggling to see the difference aside from syntax, what prompted the changes here?

src/UrlParams.ts
```diff
@@ -408,6 +405,22 @@ export const getUrlParams = (
...
+    case "desktop":
+    case "mobile":
+      switch (intent) {
+        case UserIntent.StartNewCall:
+        case UserIntent.JoinExistingCall:
+        case UserIntent.StartNewCallDM:
+        case UserIntent.JoinExistingCallDM:
+          intentDerivedConfiguration.defaultAudioEnabled = true;
+          intentDerivedConfiguration.defaultVideoEnabled = true;
+          break;
```
Half-Shot: Slight cautious noises here that the url params are now being altered by the platform you are on, which feels a bit unexpected and may make things harder to debug. I'd sort of rather the mobile clients injected their own parameters for tracing reasons.

In lieu of that, can we at least log that extra configuration is being enabled here based on platform so that debugging is easier?

src/room/MuteStates.test.tsx
```diff
@@ -110,9 +110,10 @@ function mockMediaDevices(
   return new MediaDevices(scope);
 }
 
-describe("useMuteStates", () => {
+describe("useMuteStates in full mode", () => {
```
Half-Shot: is "full" == SPA?

Half-Shot (commented):
I have lots of questions. I don't think it's necessarily a bad thing but I think I lack context to understand the changes.

src/UrlParams.ts
Half-Shot: This has been confirmed oob that this is how it works.

src/UrlParams.ts
```diff
@@ -408,6 +405,22 @@ export const getUrlParams = (
       };
   }
 
+  const intentDerivedConfiguration: IntentDerivedConfiguration = {};
```
Half-Shot: If I was being pedantic, this is more like intentAndPlatformDerivedConfiguration. The intent derived configuration is what you are doing above :)

src/UrlParams.ts
Half-Shot: >I'd sort of rather the mobile clients injected their own parameters for tracing reasons.

It has been explained to me this is to reduce friction with making changes to mobile platforms on behalf of product, and instead just make changes on ECall.

I would still like some logging though :)

Half-Shot (approved):
Re-reviewed with context. Nothing seems bad here, but I'd like a small bit of logging if possible.

src/UrlParams.ts
toger5: added a comment.

src/UrlParams.ts
toger5: Its shorter and more straight forward due to not needing to wrap you head around the spread syntax.

src/UrlParams.ts
toger5: Logging this is a really good idea. We also do have a url param section in the debug tools.
But logging is a good stop gap until we have a clear: url vs final parameters view in the debugging tools.

src/room/MuteStates.test.tsx
toger5: yes its the package name ("embedded" vs "full") this totally deserves a better description

src/UrlParams.ts
toger5: I like this and i dont see any harm in having a really long variable name here.

BillCarsonFr (approved):
Nice to see the tests here!
Looks to me that some of the existing param overrides was not tested? This would have allow to ensure that the change of syntax does not introduce behavior change.
Maybe a task for later to add more tests?

src/UrlParams.ts
```diff
@@ -331,12 +331,15 @@ let urlParamCache: {
 export const getUrlParams = (
   search = window.location.search,
   hash = window.location.hash,
+  /** Skipping the cache might be needed in tests, to allow recomputing based on mocked platform changes. */
```
Half-Shot, line 334: I think this will be needed in every test, for test environment hygiene reasons. Hm, I wonder how we can do this without having to hack around for every test.

toger5:
@BillCarsonFr
>  Maybe a task for later to add more tests?


Tracking this todo here:
https://github.com/element-hq/element-call/issues/3497

src/UrlParams.ts
toger5: another option would be to export the cache and reset it in beforeEach (or export `resetUrlParamCache`)

toger5: Or we do sth ridiculus like using urls that contain a `Date.now()` so that we will never be able to use the cache... Or maybe not...

Half-Shot: Exporting a func to reset sounds like a good soln to me.

toger5: For this PR. should we just keep it like that. Once we acutally need it a better solution might appear?

toger5: probably compute + get might be nice.
`computeUrlParams` is uncached and `getUrlParams` is using the cache. so we mostly write test for compute and the get method just gets a single test if it loads from the cache.

Issue #3477, opened 2025-09-04
Screenshare stops sharing after some minutes


Half-Shot:
### Steps to reproduce

1. Start a screenshare in a group call (on SPA)
2. Screenshare starts as normal and works for a few minutes
3. Notice after some time that the screenshare stops (silently)

#### What did you expect?

Screenshare runs until the user explicitly stops sharing.

#### What happened instead?

Screenshare stops on it's own, no feedback as to why.

### Operating system

Not sure (will check)

### Browser information

Not sure (will check)

### URL for webapp

call.element.io

### Will you send logs?

Yes

Half-Shot:
I had a look through the logs but I could find no information as to why the stream was stopping.

fkwp:
We should log when a track is published or unpublished.

toger5:
closed as we merged https://github.com/element-hq/element-call/pull/3495

Half-Shot:
no no no, it's still a bug :) #3495 gives me logs to debug it

robintown:
Blocked until we can capture some logs.

PR #3495, opened 2025-09-18
Log when a track is unpublished or runs into an error

Half-Shot:
Improves logging for #3477

### DIFF
--- a/src/livekit/useLivekit.ts
+++ b/src/livekit/useLivekit.ts
@@ -9,6 +9,7 @@ import {
   ConnectionState,
   type E2EEManagerOptions,
   ExternalE2EEKeyProvider,
+  type LocalTrackPublication,
   LocalVideoTrack,
   Room,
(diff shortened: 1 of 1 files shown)


toger5 (approved):
We can totally add this to the RC but this whole file will get removed with the multi sfu branch. Any idea how we could make sure we do not loose the work when moving to the multi sfu branch? Maybe its fine if this is just part of this release for debugging?
Otherwise you could try porting this to the multi sfu branch into `Connection.ts`

Half-Shot:
> We can totally add this to the RC but this whole file will get removed with the multi sfu branch. Any idea how we could make sure we do not loose the work when moving to the multi sfu branch? Maybe its fine if this is just part of this release for debugging? Otherwise you could try porting this to the multi sfu branch into `Connection.ts`

Happy to do the port job after merge into multi-sfu.

PR #3499, opened 2025-09-18
Add ring notification to UserIntent.StartNewCallDM

toger5:


jmartinesp:
~~Is using the deployed site as the custom call URL enough to test this on EX? If so, it doesn't seem to work, I still only see the `.member` event being sent.~~

---

It's actually working fine and this PR actually makes it ring as expected. It was the 'unverified sessions' issue striking again, when I was sure I didn't have any...

fkwp (approved):
lgtm

PR #3502, opened 2025-09-19
Fix: never stop ring feedback on the sender side.

toger5:
Make ring$ a behavior and add code comments to justify/explain the change.


Half-Shot (approved):
tick as per the other PR

BillCarsonFr (approved):
kudos for the test

PR #3503, opened 2025-09-19
Backport: Fix: never stop ring feedback on the sender side

toger5:
backport of: https://github.com/element-hq/element-call/pull/3502


Half-Shot (approved):
(we paired on this, so I'm happy)

PR #3504, opened 2025-09-19
Refactor ring$ observable

toger5:

### DIFF
--- a/src/state/CallViewModel.test.ts
+++ b/src/state/CallViewModel.test.ts
@@ -1274,7 +1274,7 @@ describe("waitForCallPickup$", () => {
             },
           });
 
-          expectObservable(vm.callPickupState$).toBe("a 9ms b 29ms c", {
+          expectObservable(vm.callPickupState$).toBe("a 9ms b 19ms c", {
             a: "unknown",
             b: "ringing",
             c: "timeout",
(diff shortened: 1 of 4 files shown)


src/state/CallViewModel.test.ts
```diff
@@ -1272,7 +1274,7 @@ describe("waitForCallPickup$", () => {
             },
           });
 
-          expectObservable(vm.callPickupState$).toBe("a 9ms b 29ms c", {
+          expectObservable(vm.callPickupState$).toBe("a 9ms b 19ms c", {
```
toger5, line 1277: This was actually wrong before the refactor.

Before the timer was only ever started once connected. But it might have sent the notification event a couple of seconds before (if the connection to the sfu is very poor)
The test has passed previously because we only were running `ring$.pipe(...)` once we were connected. Now we subscribe to `ring$` right in the `combineLatest` so we start the timer instantly.
Basically even though the observable with the timer was already created it was only subscribed later. (the inner observable at least. the other one was subscribed by the behavior scope immediately that is why the `didSendCallNotification$` was still caught.

I hope this makes sense. The more direct justification for this change is:
```
 // Fire a call notification IMMEDIATELY (its important for this test, that this happens before the livekitConnectionState$ emits)
          schedule("n", {
            n: () => {
              rtcSession.emit(
                MatrixRTCSessionEvent.DidSendCallNotification,
                mockRingEvent("$notif1", 30),
                mockLegacyRingEvent,
              );
            },
          });
```
Says we fire the event immediately and it rings for 30ms -> it should stop after 30ms -> c=timeout should happen after 30ms
before: `"a 9ms b 29ms c"` 1ms+9ms+1ms+29ms = 40ms is just wrong
now: `"a 9ms b 19ms c"` 1ms+9ms+1ms+19ms = 30ms is what we would expect with a 30ms notificaiton event

Half-Shot: That makes sense to me :+1: , good catch.

Half-Shot (approved):

fkwp (approved):

PR #3511, opened 2025-09-22
DevX: Add another Matrix site to docker-compose

fkwp:
This allows to develop and test MatrixRTC over federation including a Multi-SFU scenario.

Note due to the new wildcard domain *.othersite.m.localhost you need to reimport [./backend/dev_tls_local-ca.crt](https://github.com/element-hq/element-call/blob/livekit/backend/dev_tls_local-ca.crt) (once merged) to your local browser


toger5 (approved):
I am wondering if using a `dev2` or `devB` prefix is simpler? (also for ctrl+P file opening)
a `-otherside` post fix is also very clear however.

PR #3512, opened 2025-09-23
A new world: "Multi-SFU" media transport

toger5:
This PR is a big step up in how we treat the ownership of the media connection in a call. Instead of using a unsuited "centralised" solution to decide on what SFU will be used for the call this PR allows everyone to own their connection and still have a call with all participants without a system to choose one specific SFU. Apart from this being the much better suited appraoch fro a decentralised system there are the following real word benefits:
 - NO sudden reconnections mid call if someone disconnects.
 - NO unpredictable performance in the same room (because someone else started a call)
 - Better performance since calls can be spread over multiple SFU's
 - Clear Connection data ownership and clean up of the related code, allowing for simpler adoption of other SFU concepts (full mesh, cascading)


fkwp (approved):
lgtm

PR #3513, opened 2025-09-24
Add sticky event support

toger5:
This uses the newly merged sticky event work. Largely this just adds a new setting to enable it when *starting* a call. Sticky events will *always* work if someone else on the call enables it.

#### Netilify

- use the linked netlify build https://pr3513--element-call.netlify.app/
- open the netlify url WITH a `homeserver` url parameter to link your custom homeserver (example: `https://localhost:3000?homeserver=https%3A%2F%2Ftest.com` encoding is of course needed)
- start a call with this link (using the homeserver url param) (in multiple browser envs (private, non private OR firefox container extension)

#### Locally

pull/download this PR/branch
start the docker dev env via `yarn backend` (synapse, livekit sfu, nginx, jwt authorizationService...)
(in a new terminal) start EC `yarn dev`

Now it should "just work" to do calls on https://localhost:3000

### DIFF
--- a/src/home/useGroupCallRooms.ts
+++ b/src/home/useGroupCallRooms.ts
@@ -143,7 +143,7 @@ const roomHasCallMembershipEvents = (room: Room): boolean => {
 
   // Check for *active* calls using sticky events.
   for (const sticky of room._unstable_getStickyEvents()) {
-    if (sticky.getType() === EventType.GroupCallMemberPrefix) {
+    if (sticky.getType() === EventType.RTCMembership) {
       return true;
     }
   }
--- a/src/home/useGroupCallRooms.ts
+++ b/src/home/useGroupCallRooms.ts
@@ -20,7 +20,6 @@ import {
   MatrixRTCSessionManagerEvents,
   type MatrixRTCSession,
 } from "matrix-js-sdk/lib/matrixrtc";
-import { logger } from "matrix-js-sdk/lib/logger";
 
 import { getKeyForRoom } from "../e2ee/sharedKeyManagement";
 
@@ -121,7 +120,8 @@ const roomIsJoinable = (room: Room): boolean => {
  * @returns `true` if the room has call events.
  */
 const roomHasCallMembershipEvents = (room: Room): boolean => {
-  // Legacy events.
+  // Check our room membership first, to rule out any rooms
+  // we can't have a call in.
   const myMembership = room.getMyMembership();
   if (myMembership === KnownMembership.Knock) {
     // Assume that a room you've knocked on is able to hold calls
@@ -131,9 +131,8 @@ const roomHasCallMembershipEvents = (room: Room): boolean => {
     return false;
   }
 
+  // Legacy member state checks (cheaper to check.)
   const timeline = room.getLiveTimeline();
-
-  // Check legacy events first, because it's cheaper.
   if (
     timeline
       .getState(EventTimeline.FORWARDS)
@@ -142,7 +141,15 @@ const roomHasCallMembershipEvents = (room: Room): boolean => {
     return true;
   }
 
-  // There was call membership events at some point in the timeline.
+  // Check for *active* calls using sticky events.
+  for (const sticky of room._unstable_getStickyEvents()) {
+    if (sticky.getType() === EventType.GroupCallMemberPrefix) {
+      return true;
+    }
+  }
+
+  // Otherwise, check recent event history to see if anyone had
+  // sent a call membership in here.
   return timeline.getEvents().some(
     (e) =>
       // Membership events only count if both of these are true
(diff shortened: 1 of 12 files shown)


src/home/useGroupCallRooms.ts
```diff
@@ -113,19 +113,41 @@ const roomIsJoinable = (room: Room): boolean => {
...
+  if (
+    timeline
+      .getState(EventTimeline.FORWARDS)
+      ?.events?.has(EventType.GroupCallMemberPrefix)
+  ) {
+    return true;
+  }
+
+  // There was call membership events at some point in the timeline.
+  return timeline.getEvents().some(
```
BillCarsonFr, line 153: There is no API to get the sticky events? Having the sticky key doesn't make the event sticky. 
I am a bit lost with what this method is expecting to do

BillCarsonFr, line 153: What is this method exactly? Try some heuristic to see if a room is a call room? like `type": "org.matrix.msc3417.call"`?

Or if there might be an active call?
And now it is checking for possible past sticky events (but potentially not anymore sticky)?
I don't get what this doing and what for

BillCarsonFr (commented):
So this is added a new lab option? 
It is just for Netlify? so why a review?

Half-Shot:
> It is just for Netlify? so why a review?

Sorry failed to change the title :facepalm:

toger5 (commented):
I think the following might make sense:
- since sticky events imply the new event format it also implies multi-sfu (new event format does not have a oldest member transport)

- rename the dev option to: use sticky m.rtc.member` events
- Always use multi sfu when stikcy evnets are activated (gray out the multi sfu option) 
Currently turning multi-sfu off and sticky on results in not sending any transport at all

toger5:
I think this might be ready to merge @robintown

src/rtcSessionHelpers.ts
```diff
@@ -118,18 +119,12 @@ export async function enterRTCSession(
...
 ): Promise<void> {
-  const {
-    encryptMedia,
-    useNewMembershipManager = true,
-    useExperimentalToDeviceTransport = false,
-    useMultiSfu = true,
-  } = options;
-
+  const { encryptMedia, useExperimentalToDeviceTransport = false } = options;
+  const useMultiSfu = preferStickyEvents.getValue() ?? options.useMultiSfu;
```
robintown:
```suggestion
  const useMultiSfu = preferStickyEvents.getValue() || options.useMultiSfu;
``

File: src/rtcSessionHelpers.ts [hunk 3]
```diff
@@ -164,6 +158,7 @@ export async function enterRTCSession(
       membershipEventExpiryMs:
         matrixRtcSessionConfig?.membership_event_expiry_ms,
       useExperimentalToDeviceTransport,
+      unstableSendStickyEvents: preferStickyEvents.getValue(),
```
robintown: Could be nice to have `preferStickyEvents` be an explicit parameter of `enterRtcSession`, like `multiSfu` is, so we can easily turn on sticky events in the middle of a call when dogfooding

src/home/useGroupCallRooms.ts
```diff
@@ -114,19 +113,49 @@ const roomIsJoinable = (room: Room): boolean => {
...
+    timeline
+      .getState(EventTimeline.FORWARDS)
+      ?.events?.has(EventType.GroupCallMemberPrefix)
+  ) {
+    return true;
+  }
+
+  // Check for *active* calls using sticky events.
+  for (const sticky of room._unstable_getStickyEvents()) {
+    if (sticky.getType() === EventType.GroupCallMemberPrefix) {
```
robintown: Presumably we should refer to `RTCMembership` here rather than `GroupCallMemberPrefix` (though TBH we could probably replace this entire function body with `return true;` and it would be fine… we're definitely not aiming to support using the SPA in cases where the homeserver is also used by messenger apps… well, maybe we are with the guest access use case, but I don't think we'll care about giving guest access users a home page at all.)

src/rtcSessionHelpers.ts
toger5: we would still need to rejoin and setup a new membership manager. I am not sure its worth the effort?

src/home/useGroupCallRooms.ts
toger5: 100%.
Lets keep the function in for now. There might be some use for it in the guest case and it is very detangled from the rest of element-call -> not complexifying the codebase. (just makes it larger)

src/rtcSessionHelpers.ts
robintown: Sure but with the `scope.reconcile` method already used for starting and stopping the membership managers as needed, this ends up being pretty trivial.

Incidentally I see it as necessary to make the CallViewModel also aware of the relationship between `multiSfu` and `preferStickyEvents` (the fact that one force-enables the other), otherwise it would pass the wrong transport to `enterRtcSession`. So I just committed e313cf04a6fe2c4c953ec6e86ad70e41f412ad61 which handles both these things. Does it look reasonable to you?

robintown (approved):
I think this looks pretty much ready now. But just one thing I want to get clarity on (for my future reference as we eventually roll the Matrix 2.0 changes out to the public):

> Largely this just adds a new setting to enable it when starting a call. Sticky events will always work if someone else on the call enables it.

From what I can tell by looking at the matrix-js-sdk diff, it's not the case that it switches membership manager implementations when another participant sending sticky events is detected. Do you mean that the setting enables *sending* sticky events, but *receiving* sticky events will always work?

toger5:
I added another todo comment so we can dont forget discussion the `transport$` naming. I am not exactly sure transport makes so much sense considering it now includes the sticky option.

I will merge this into the main multi sfu branch as it is now however so we have less branches to handle.

PR #3521, opened 2025-10-01
Connection State: Error Handling + Cleaning / refactoring / 

BillCarsonFr:
Better reviewed commit per commit


src/livekit/MatrixAudioRenderer.tsx
```diff
@@ -32,8 +32,11 @@ export interface MatrixAudioRendererProps {
    * This list needs to be composed based on the matrixRTC members so that we do not play audio from users
    * that are not expected to be in the rtc session.
    */
+  // TODO: Why do we have this structure? looks like we only need the valid/active participants (not the room member or id)?
```
robintown: Just laziness, really. The CallViewModel needs the ID and room member internally to create the media tiles, so I took the same data and made it public for the purpose of audio rendering without making the public type any stricter. I've remedied this in: c96e81bfd375262be9c992cbd1885c7d878f5b8d

src/main.tsx
```diff
@@ -11,7 +11,6 @@ Please see LICENSE in the repository root for full details.
 // dependency references.
 import "matrix-js-sdk/lib/browser-index";
 
-import { StrictMode } from "react";
```
robintown, line 14: Oops, I didn't mean to commit my temporary disabling of strict mode. e346c8c148bd6aad302598d57f16f02dc91cf468

src/state/MuteStates.ts
```diff
@@ -132,6 +138,7 @@ class MuteState<Label, Selected> {
   ) {}
 }
 
+// TODO there is another MuteStates in src/room/MuteStates.tsx ?? why
```
robintown: Dead code, we can remove it. 5da780ed301bc6e259ba9fb52c7b92b3aef88525

src/state/Async.ts
```diff
@@ -5,16 +5,10 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
-  map,
-  Observable,
-  of,
-  startWith,
-  switchMap,
-} from "rxjs";
+import { catchError, from, map, type Observable, of, startWith } from "rxjs";
 
+// TODO where are all the comments? ::cry::
+// There used to be an unitialized state!, a state might not start in loading
```
robintown: IMO the "uninitialized" state doesn't belong here. In cases where it might be tempting to use "uninitialized", I'd like it if we explicitly add such a case into our types as needed (`Async<T>` → `Async<T | undefined>`) or keep an an eye out for better ways to model the problem. With just these three states, there's a very clear mapping between async values and Promises.

I added hopefully a decent amount of comments: b1d143720aa322d9f263c4e3736a0dd5ebc0d468

src/state/Connection.ts
```diff
@@ -10,293 +10,246 @@ import {
...
-    protected readonly scope: ObservableScope,
-    protected readonly remoteTransports$: Behavior<
-      { membership: CallMembership; transport: LivekitTransport }[]
-    >,
-    e2eeLivekitOptions: E2EEOptions | undefined,
-    livekitRoom: LivekitRoom | undefined = undefined,
+  /**
+   * The focus server to connect to.
+   */
+  public readonly localTransport: LivekitTransport;
```
robintown, line 160: Why is this called `localTransport` and not `transport`? It's often remote.

src/state/CallViewModel.ts
```diff
@@ -558,48 +563,51 @@ export class CallViewModel extends ViewModel {
...
    * The transport over which we should be actively publishing our media.
    */
-  private readonly localTransport$: Behavior<Async<LivekitTransport> | null> =
+  private readonly localTransport$: Behavior<Async<LivekitTransport>> =
     this.scope.behavior(
       this.transports$.pipe(
         map((transports) => transports?.local ?? null),
         distinctUntilChanged(deepCompare),
       ),
     );
```
robintown, line 577: For the record this can be null, it's just that `deepCompare` defaults its type parameter to `any` which ended up hiding the type error here. Restored in: 3691e7120d572082af5560aa5539aa02a5eb6906

src/utils/test.ts
```diff
@@ -416,13 +418,15 @@ export const deviceStub = {
...
+export function mockMediaDevices(
+  data: Partial<MediaDevices>,
+): MockedObject<MediaDevices> {
+  return vi.mocked<MediaDevices>({
     audioInput: deviceStub,
     audioOutput: deviceStub,
     videoInput: deviceStub,
     ...data,
-  } as MediaDevices;
+  } as MediaDevices);
```
robintown: This object isn't a `Vitest.MockedObject` in the sense that functions on it don't have the `.mockClear()`, `.mockReturnValue()` etc. methods available, so we shouldn't assert that it is. dcc3ab641f59ae986edca267d9b7787ed9c546fa

src/livekit/MatrixAudioRenderer.test.tsx
```diff
@@ -48,42 +57,148 @@ vi.mock("@livekit/components-react", async (importOriginal) => {
...
+      { userId: "@alice", deviceId: "DEV1" },
+      { userId: "@bob", deviceId: "DEV0" },
+    ],
+    livekitParticipantIdentities: [
+      { id: "@alice:DEV0", isLocal: true },
+      "@bob:DEV0",
+      "@alice:DEV1",
+    ],
+    expectedAudioTracks: 2,
+  },
```
robintown: For the record I'm removing this test case because I managed to make the prop types of MatrixAudioRenderer strict enough so that it only accepts RemoteParticipants. 00daf834b655140d402515beb52505946e79b7d0

src/state/Connection.ts
```diff
@@ -10,293 +10,246 @@ import {
...
+  /** Optional factory to create the Livekit room, mainly for testing purposes. */
+  livekitRoomFactory?: (options?: RoomOptions) => LivekitRoom;
+}
+
+export type FocusConnectionState =
+  | { state: "Initialized" }
+  | { state: "FetchingConfig"; focus: LivekitTransport }
+  | { state: "ConnectingToLkRoom"; focus: LivekitTransport }
+  | { state: "PublishingTracks"; focus: LivekitTransport }
+  | { state: "FailedToStart"; error: Error; focus: LivekitTransport }
```
robintown, line 54: I made an attempt in my code to stick to the new term 'transport' and phase out usage of 'focus'. Do you think it makes sense to continue that here?

src/state/Connection.ts
```diff
@@ -10,293 +10,246 @@ import {
...
     this.stopped = false;
-    const { url, jwt } = await this.sfuConfig;
-    if (!this.stopped) await this.livekitRoom.connect(url, jwt);
+    try {
+      this._focusedConnectionState$.next({
+        state: "FetchingConfig",
+        focus: this.localTransport,
+      });
+      // TODO could this be loaded earlier to save time?
+      const { url, jwt } = await this.getSFUConfigWithOpenID();
```
robintown, line 100: Oh yeah, good idea. I think it's only safe for our users' privacy to do so for the *local* connection though. (We shouldn't leak to untrusted remote servers whether you've left your client sitting on the lobby screen of a video room, or something. Only once there is a clear intent to join the call should we fetch the JWT.)

src/state/Connection.ts
```diff
@@ -10,293 +10,246 @@ import {
...
             )
             // Pair with their associated LiveKit participant (if any)
-            .map((membership) => {
+            // Uses flatMap to filter out memberships with no associated rtc participant ([])
+            .flatMap((membership) => {
               const id = `${membership.sender}:${membership.deviceId}`;
               const participant = participants.find((p) => p.identity === id);
-              return { participant, membership };
+              return participant ? [{ participant, membership }] : [];
             }),
```
robintown, line 202: I think something went wrong here with the merge. If you take a look at 1820cac3f66e9b9801209990d5ad5469ad21e97a I was trying to make sure that even RTC memberships with no associated LiveKit participant are included in `publishingParticipants$`. (One could add them back in CallViewModel, but it seemed way easier and more efficient to just avoid filtering them out here.)

The type that I was aiming for was:

```typescript
Connection.publishingParticipants$: Behavior<{
  participant: RemoteParticipant | undefined;
  membership: CallMembership;
}[]>
```

Once we get the CallViewModel tests back up and running we'll be back to a state where this can't regress without a red flag.

src/state/Connection.ts
```diff
@@ -10,293 +10,246 @@ import {
...
+      .behavior<ConnectionState>(connectionStateObserver(this.livekitRoom))
+      .subscribe((connectionState) => {
+        const current = this._focusedConnectionState$.value;
+        // Only update the state if we are already connected to the LiveKit room.
+        if (current.state === "ConnectedToLkRoom") {
+          this._focusedConnectionState$.next({
+            state: "ConnectedToLkRoom",
+            connectionState,
+            focus: current.focus,
+          });
```
robintown, line 217: Perhaps a more elegant way to do this would be to change the type to:

```diff
 {
   state: "ConnectedToLkRoom";
-  connectionState: ConnectionState;
+  connectionState: Observable<ConnectionState>;
   focus: LivekitTransport;
 }
```

Then (on line 111) you can just write:

```typescript
this._focusedConnectionState$.next({
  state: "ConnectedToLkRoom",
  focus: this.localTransport,
  connectionState: connectionStateObserver(this.livekitRoom),
});
```

and remove this line 212 code.

src/state/PublishConnection.ts
```diff
@@ -0,0 +1,277 @@
...
+    super(room, args);
+
+    // Setup track processor syncing (blur)
+    this.observeTrackProcessors(scope, room, trackerProcessorState$);
+    // Observe mute state changes and update LiveKit microphone/camera states accordingly
+    this.observeMuteStates(scope);
+    // Observe media device changes and update LiveKit active devices accordingly
+    this.observeMediaDevices(scope, devices, controlledAudioDevices);
+
+    this.workaroundRestartAudioInputTrackChrome(devices, scope);
```
robintown, line 86: Nice, this cleaner constructor is also a welcome change

src/state/PublishConnection.ts
```diff
@@ -0,0 +1,277 @@
...
+        await this.livekitRoom.localParticipant.setCameraEnabled(desired);
+      } catch (e) {
+        logger.error("Failed to update LiveKit video input mute state", e);
+      }
+      return this.livekitRoom.localParticipant.isCameraEnabled;
+    });
+    scope.onEnd(() => {
+      this.muteStates.audio.unsetHandler();
+      this.muteStates.video.unsetHandler();
+    });
```
robintown, line 229: In 86fb026be86d7ea0fa904babd920b6e1d0b9f1c0 I moved these calls to `unsetHandler` to the `stop` method as `scope.onEnd` happens much later than what we need for SFU-hopping. I believe they should stay there in `stop`.

src/state/PublishConnection.ts
```diff
@@ -0,0 +1,277 @@
...
+      video: this.muteStates.video.enabled$.value,
+    });
+    if (this.stopped) return;
+    for (const track of tracks) {
+      // TODO: handle errors? Needs the signaling connection to be up, but it has some retries internally
+      // with a timeout.
+      await this.livekitRoom.localParticipant.publishTrack(track);
+      if (this.stopped) return;
+      // TODO: check if the connection is still active? and break the loop if not?
+    }
```
robintown, line 117: We should restore this diff from 86fb026be86d7ea0fa904babd920b6e1d0b9f1c0:

```diff
@@ -130,23 +131,35 @@ export class PublishConnection extends Connection {
     if (!this.stopped) await this.livekitRoom.connect(url, jwt);
 
     if (!this.stopped) {
-      const tracks = await this.livekitRoom.localParticipant.createTracks({
-        audio: this.muteStates.audio.enabled$.value,
-        video: this.muteStates.video.enabled$.value,
-      });
-      for (const track of tracks) {
-        await this.livekitRoom.localParticipant.publishTrack(track);
+      // TODO-MULTI-SFU: Prepublish a microphone track
+      const audio = this.muteStates.audio.enabled$.value;
+      const video = this.muteStates.video.enabled$.value;
+      // createTracks throws if called with audio=false and video=false
+      if (audio || video) {
+        const tracks = await this.livekitRoom.localParticipant.createTracks({
+          audio,
+          video,
+        });
+        for (const track of tracks) {
+          await this.livekitRoom.localParticipant.publishTrack(track);
+        }
       }
     }
   }
```

(by which I mean just the `TODO` and the `audio || video` guard.)

robintown (commented):
Very nice work with fixing things up as you go. For the sake of getting our two branches merged into one as soon as possible, I went ahead and applied some changes directly for things which seemed hopefully uncontrovertial - if you like you can have a look through them. The rest of my comments are mostly about things from my branch that seem to have been lost in the previous merge conflicts.

I did also confirm that the app runs and connects to an SFU successfully.

src/state/PublishConnection.ts
robintown: Actually, scratch that, we should just give the Connection class a proper ObservableScope of its own that ends at the right time (when CallViewModel is about to drop the Connection object).

robintown: I can work on that.

src/state/Connection.ts
robintown: a6c4fb41483c005f839c2c91b726b62dea68fe85

src/state/PublishConnection.ts
robintown: 95069fd3fafd3cd40fb2b48edb061da50796e8db

src/state/Connection.ts
robintown: ea17ed7253cb47adb29475732bace0486aff0cb7

src/state/Connection.ts
robintown: 2dc61346061b5078c46a345784b62f456bd43bbb

src/state/Connection.ts
robintown: 9f4e99310bd6ae5f0e4ff86450f2ab4757103cf0

src/state/Connection.ts
robintown: b0eb566a4f959503c8be8d4c2a03c227553041fd

PR #3531, opened 2025-10-06
Disable call button while the call is connecting.

### LINKED ISSUE #3530: The 'join' button appears to do nothing if the JWT check is delayed.
### Steps to reproduce

1. Start a video call in a room (embedded in EW/ED)
2. Hit the join button from the lobby
3. EC tries to authorise livekit to participate in the room via JWT stuff, which may take a while (e.g. if the matrixrtc auth service can't query synapse via the federation API, or takes a while for DNS to resolve, or whatever)
4. Nothing happens.
5. You keep hitting the join button.
6. Eventually it suddenly comes to life.

#### What did you expect?

The join button should be greyed out or show a spinner or something after you start the process of joining.

#### What happened instead?

There's no evidence anything is happening despite the delay; the app feels flakey

### Operating system

macOS

### Browser information

ED

### URL for webapp

Nightly

### Will you send logs?

No

Half-Shot:
Fixes https://github.com/element-hq/element-call/issues/3530

### DIFF
--- a/src/room/LobbyView.tsx
+++ b/src/room/LobbyView.tsx
@@ -193,12 +193,12 @@ export const LobbyView: FC<Props> = ({
 
   useTrackProcessorSync(videoTrack);
 
+  const [waitingToEnter, setWaitingToEnter] = useState(false);
   const onEnterCall = useCallback(() => {
-    setWaiting(true);
-    void onEnter().finally(() => setWaiting(false));
+    setWaitingToEnter(true);
+    void onEnter().finally(() => setWaitingToEnter(false));
   }, [onEnter]);
 
-  const [waitingToEnter, setWaitingToEnter] = useState(false);
   const waiting = waitingForInvite || waitingToEnter;
 
   // TODO: Unify this component with InCallView, so we can get slick joining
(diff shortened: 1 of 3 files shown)


toger5 (approved):
Thanks. This one seems super simple.

Does it make sense to port this one to the multi-sfu branch as well?

src/room/LobbyView.tsx
```diff
@@ -193,6 +193,13 @@ export const LobbyView: FC<Props> = ({
 
   useTrackProcessorSync(videoTrack);
 
+  const onEnterCall = useCallback(() => {
+    setWaiting(true);
+    void onEnter().finally(() => setWaiting(false));
+  }, [onEnter]);
+
+  const [waiting, setWaiting] = useState(waitingForInvite);
```
robintown: I know that `waitingForInvite` never actually gets updated in practice without fully unmounting and remounting the lobby view, but just to avoid a future surprise: could we have this component react to its updates?

```suggestion
  const [waitingToEnter, setWaitingToEnter] = useState(false);
  const waiting = waitingForInvite || waitingToEnter;
```

robintown (changes requested):

robintown:
> Does it make sense to port this one to the multi-sfu branch as well?

Sure, I'll continue merging livekit → multi-sfu every now and then.

robintown (approved):

PR #3535, opened 2025-10-15
Revert transport state to avoid having an embedded observable in one of the variant

BillCarsonFr:
Proposal to revert that change https://github.com/element-hq/element-call/pull/3512/commits/b0eb566a4f959503c8be8d4c2a03c227553041fd and keep only static variant of state instead of having one of the observable variant beeing him self another observable.
```
 | { state: "FailedToStart"; error: Error; transport: LivekitTransport }
  | {
      state: "ConnectedToLkRoom";
      connectionState$: Observable<ConnectionState>;
      transport: LivekitTransport;
    }
```

I think it makes it easier to use. I added back the test that was deleted
https://github.com/element-hq/element-call/pull/3512/commits/2c66e11a0a1f496e251c4139d506f05c220b1852
image.png

If this proposal is not accepted, then at least the deleted test should be added back using the new API

### DIFF
--- a/src/state/Connection.ts
+++ b/src/state/Connection.ts
@@ -21,7 +21,7 @@ import {
   type CallMembership,
   type LivekitTransport,
 } from "matrix-js-sdk/lib/matrixrtc";
-import { BehaviorSubject, combineLatest, type Observable } from "rxjs";
+import { BehaviorSubject, combineLatest } from "rxjs";
 
 import {
   getSFUConfigWithOpenID,
@@ -60,7 +60,7 @@ export type TransportState =
   | { state: "FailedToStart"; error: Error; transport: LivekitTransport }
   | {
       state: "ConnectedToLkRoom";
-      connectionState$: Observable<ConnectionState>;
+      livekitState: ConnectionState;
       transport: LivekitTransport;
     }
   | { state: "Stopped"; transport: LivekitTransport };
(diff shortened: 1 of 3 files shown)


src/state/Connection.ts
```diff
@@ -250,6 +250,20 @@ export class Connection {
       [],
     );
 
+    scope
```
BillCarsonFr, line 253: maybe we could have a switch map from a new startLifecycle observable if you prefer

src/state/Connection.ts
```diff
@@ -250,6 +250,20 @@ export class Connection {
...
+      .behavior<ConnectionState>(connectionStateObserver(livekitRoom))
+      .subscribe((livekitState) => {
```
robintown, line 255: This `subscribe` call would need a `.pipe(this.scope.bind())` to avoid risk of leaking, for the record

robintown (commented):
I don't agree that this is necessarily easier to use. As the diff shows it's basically a matter of `map` versus `switchMap`, and meanwhile what was a clean declarative specification with almost nothing to test becomes imperative + with poorer locality of behavior.

There are also some things that become easier to do when state is not flattened in this way, such as `map`ing with ideal performance when you don't care about the `ConnectionState`, or accumulating a state that resets whenever you're outside the `ConnectedToLkRoom` phase.

This is not high-temperature feedback and you're welcome to merge this if you still like it better, but I just wanted to give context as to why, more generally, I think we shouldn't shy away from dynamic values that contain other dynamic values when they're also a valid way of modelling a problem.

PR #3536, opened 2025-10-17
Avoid gratuitous animations when the speaker changes

robintown:
The purpose of the TileStore is to help keep call layouts more visually stable by only creating, destroying, or reordering tiles when absolutely necessary. One of the goals with this implementation was always to have the TileStore "swap" the media of the spotlight tile with the media of another grid tile whenever the spotlight speaker changes, so we don't have a constant visual distraction of animations during back-and-forth conversations.

Example of intended behavior (no animations):

[Screencast From 2025-10-17 11-49-45.webm](https://github.com/user-attachments/assets/ba8339f4-9cb0-4c03-ab65-adfd4c44ba79)

Unfortunately, this regressed in 79c40f198cb662481df012594f8fdbbb67be63bd because of the overlooked renaming of the `speaking` field to `speaking$`, meaning the TileStore wasn't accurately detecting when things were user media. I have added a new assertion to a relevant test to prevent this regressing again.

Also, I thought I remembered that the implementation would likewise swap the media out on PiP tiles in the expanded spotlight layout without constantly destroying & recreating the tile. However, this was apparently left unimplemented during the first iteration of the TileStore. It's a welcome UI optimization I think, and we can reliably test for it, so I added that here too.

Example of new PiP behavior (no animations):

[Screencast From 2025-10-17 11-54-42.webm](https://github.com/user-attachments/assets/44221fef-0012-4e1a-9612-8bfcb2895cb0)

### DIFF
--- a/src/state/TileStore.ts
+++ b/src/state/TileStore.ts
@@ -262,6 +262,33 @@ export class TileStoreBuilder {
   }
 
   /**
+   * Sets up a PiP tile for the given media. This is a special kind of grid tile
+   * that is expected to stand on its own and switch between speakers, so this
+   * method will more eagerly try to reuse an existing tile, replacing its
+   * media, than registerGridTile would.
+   */
+  public registerPipTile(media: UserMediaViewModel): void {
+    if (DEBUG_ENABLED)
+      logger.debug(
+        `[TileStore, ${this.generation}] register PiP tile: ${media.member?.rawDisplayName ?? "[👻]"}`,
+      );
+
+    // If there is a single grid tile that we can reuse
+    if (this.prevGrid.length === 1) {
+      const entry = this.prevGrid[0];
+      this.stationaryGridEntries[0] = entry;
+      // Do the media swap
+      entry.media = media;
+      this.prevGridByMedia.delete(entry.media);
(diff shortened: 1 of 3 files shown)


src/state/TileStore.ts
```diff
@@ -201,7 +202,7 @@ export class TileStoreBuilder {
       if (
         media === this.prevSpotlightSpeaker &&
         this.spotlight.media.length === 1 &&
-        "speaking" in this.spotlight.media[0] &&
```
toger5, line 204: would a `this.spotlight.media[0]?.speaking$ !== undefined` have helped here? Would ts let us know that `speaking` is not available anyhow...

toger5 (approved):

src/state/TileStore.ts
robintown: TypeScript wouldn't allow us to access that property since it doesn't exist on `ScreenShareViewModel`. But what I can do is add a `satisfies` assertion to make it clear what's happened here: 1a26a85a7888aeb0546ca3e4c37d0ebb4251b1bd

robintown:
Bypassing known end-to-end test failures (caused by downstream Element Web changes)

PR #3537, opened 2025-10-17
Fix resource leaks when we stop using a connection

robintown:
The execution of certain Observables related to a local or remote connection would continue even after we stopped caring about said connection because we were failing to give these state holders a proper ObservableScope of their own, separate from the CallViewModel's longer-lived scope. With this commit they now have scopes managed by a new helper function `generateKeyed$`.

Please take a look at the individual commits when reviewing. The first commit introduces a refactor that makes scope management less error-prone and makes `generateKeyed$` possible to implement, the second one actually fixes the resource leaks, and the third is just an additional bit of test clean-up that the refactor enabled.


src/state/CallViewModel.ts
```diff
@@ -755,80 +746,76 @@ export class CallViewModel extends ViewModel {
...
+    generateKeyed$<
+      [typeof this.participantsByRoom$.value, number],
+      MediaItem,
+      MediaItem[]
+    >(
+      combineLatest([this.participantsByRoom$, duplicateTiles.value$]),
+      ([participantsByRoom, duplicateTiles], createOrGet) => {
+        const items: MediaItem[] = [];
+
+        for (const { livekitRoom, participants, url } of participantsByRoom) {
```
toger5, line 771: Do we have some kind of naming guide?
participant: livekit participant
roomMember: matrix room member
member: rtc memebership member

(If we follow the above strictly it might be worth mentioning it explicitly somewhere?)

src/state/CallViewModel.ts
```diff
@@ -755,80 +746,76 @@ export class CallViewModel extends ViewModel {
...
-                  if (prevMedia instanceof UserMedia)
-                    prevMedia.updateParticipant(participant);
-
-                  yield [
+    generateKeyed$<
+      [typeof this.participantsByRoom$.value, number],
+      MediaItem,
+      MediaItem[]
+    >(
+      combineLatest([this.participantsByRoom$, duplicateTiles.value$]),
```
toger5, line 767: Can this get a commented?
`generateKeyed` is introduced in our own codebase and more help on how it has to be read in this usecase might be helpful.
Overall i like the way `mediaItems$` work now (if I understood it correctly which i think i do)

src/state/CallViewModel.ts
robintown: 27f24ca5659e309a8ddfb883340effb1b3fdde5b

src/state/CallViewModel.ts
robintown: 1b3a56427f76a6e44539c25735afdd3b9552d2ed

src/utils/observable.ts
```diff
@@ -117,3 +118,71 @@ export function pauseWhen<T>(pause$: Behavior<boolean>) {
...
 }
+
+/**
+ * Maps a changing input value to an output value consisting of items that have
+ * automatically generated ObservableScopes tied to a key. Items will be
+ * automatically created when their key is requested for the first time, reused
+ * when the same key is requested at a later time, and destroyed (have their
+ * scope ended) when the key is no longer requested.
+ *
+ * @param input$ The input value to be mapped.
```
toger5, line 129: This is not easy to understand for me.
I think I would be able to get the concept of this better if there would be a very simple usecase (the test in the test file also is not immediately obvious to me.)

Maybe its also just missing a motivation why this method exists in the docstring?

toger5 (approved):
This looks good.

The `generateKeyed$` also seems like an obvious candidate for the CallMembership static function on the RTCSession. (sadly this is js-sdk. But once we build our own rtc sdk that might be interesting again)

PR #3546, opened 2025-10-27
Combine preferStickyEvents with useMultiSFU and add a url parameter

Half-Shot:
Needed so that Element Web and friends can specify they want to use sticky events (to match up with slot permissions behaviour).


BillCarsonFr (approved):
Approved. But I think we need to make some clean-up on the different options.
We can probably merge `multiSfu` and `preferStickyEvent` to a single `userMatrix2_0` option

src/settings/DeveloperSettingsTab.tsx
```diff
@@ -180,10 +159,10 @@ export const DeveloperSettingsTab: FC<Props> = ({ client, livekitRooms }) => {
...
           id="multiSfu"
           type="checkbox"
-          label={t("developer_mode.multi_sfu")}
-          // If using sticky events we implicitly prefer use multi-sfu
-          checked={multiSfu || preferStickyEvents}
-          disabled={preferStickyEvents}
+          label={t("developer_mode.multi_sfu.label")}
+          description={t("developer_mode.multi_sfu.description")}
+          checked={multiSfu}
+          disabled={!stickyEventsSupported}
```
Half-Shot, line 165: Note, we should have been checking this. MSFU can't be enabled without sticky event support by the homeserver.

toger5:
So this implies, that we will disable the state event based multi sfu case?

toger5:
In that case we really should rename multi sfu to sth more general like `Matrix2.0 mode (use sticky events, multi sfu, and m.rtc.member instead of m.call.member)`

Half-Shot:
> So this implies, that we will disable the state event based multi sfu case?

I think we were already only running multisfu on sticky events based on the code, I've just removed the single SFU sticky event use case.

>In that case we really should rename multi sfu to sth more general like Matrix2.0 mode (use sticky events, multi sfu, and m.rtc.member instead of m.call.member)

I've been trying to come up with a name, but I don't really know how to describe it to power users.

toger5:
Heads up, that this will collide with the refactor of the call vm.

The call vm refactor does already include sth very similar to this.

(But it does not drop the state event + multi sfu case)

toger5:
@Half-Shot I am mostly sure, that we don't need this change anymore. The current lk branch should have a good solution for this?
Can you check livekit and close this eventually?

Half-Shot:
Yeah I agree.

PR #3547, opened 2025-10-27
fix: Send close widget action on auto-leave

BillCarsonFr:
Fixes https://github.com/element-hq/voip-internal/issues/407

Widget was not closing automatically when you leave the call when all other user leave the call. Leaving you stuck with a blank widget screen.

Looks like a regression since the addition of new leave reason. 
The widget should not auto-close only if the call was left due to an error (this is to allow to display the error screen)


Half-Shot (approved):
Looks okay to me

toger5:
Why is this skipping one of the new tests?

PR #3567, opened 2025-11-17
Move one-on-one layout into CallViewModel

toger5:
_as for our goal to now work with actual PR's on this branch_

This moves another bit of logic in the `view` into the `CallViewModel`.

It also fixes a bug where we featched all members but are only interested about joined members. (and invited members for the sake of determining if its a one-on-one call.

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -262,6 +262,8 @@ export const InCallView: FC<InCallViewProps> = ({
   const earpieceMode = useBehavior(vm.earpieceMode$);
   const audioOutputSwitcher = useBehavior(vm.audioOutputSwitcher$);
   const sharingScreen = useBehavior(vm.sharingScreen$);
+  const localUserIsAlone = useBehavior(vm.localUserIsAlone$);
+  const oneOnOneMember = useBehavior(vm.isOneOnOneWith$);
 
   const fatalCallError = useBehavior(vm.configError$);
   // Stop the rendering and throw for the error boundary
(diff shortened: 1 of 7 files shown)


src/room/InCallView.tsx
```diff
@@ -262,6 +262,8 @@ export const InCallView: FC<InCallViewProps> = ({
   const earpieceMode = useBehavior(vm.earpieceMode$);
   const audioOutputSwitcher = useBehavior(vm.audioOutputSwitcher$);
   const sharingScreen = useBehavior(vm.sharingScreen$);
+  const localUserIsAlone = useBehavior(vm.localUserIsAlone$);
+  const oneOnOneMember = useBehavior(vm.isOneOnOneWith$);
```
toger5: Better naming or comment
 - oneOnOneMember -> "dm"sth.
 - local user Alone -> make sure its clear its session state (not room state)

src/room/InCallView.tsx
```diff
@@ -300,24 +302,15 @@ export const InCallView: FC<InCallViewProps> = ({
...
-      ...matrixRoom.getMembersWithMembership("invite"),
-    ].filter((m) => m.userId !== client.getUserId());
-    // Yield if there are not other members in the room.
-    if (roomOthers.length === 0) return null;
-
-    const otherMember = roomOthers.length > 0 ? roomOthers[0] : undefined;
-    const isOneOnOne = roomOthers.length === 1 && otherMember;
-    const text = isOneOnOne
-      ? `Waiting for ${otherMember.name ?? otherMember.userId} to join…`
+    if (callPickupState !== "ringing" || localUserIsAlone) return null;
```
toger5: Add a comment that this still is not high level enough for the view?

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -327,10 +327,35 @@ export class CallViewModel {
...
     this.scope,
     this.scope.behavior(this.memberships$.pipe(map((mems) => mems.value))),
-    createRoomMembers$(this.scope, this.matrixRoom),
+    this.matrixRoomMembers$,
+  );
+
+  /**
+   * Returns the Member {userId, getMxcAvatarUrl, rawDisplayName} of the other user in the call, if it's a one-on-one call.
+   */
+  public isOneOnOneWith$ = this.scope.behavior(
```
toger5: Change this whole logic to be beased on isDM

BillCarsonFr (approved):
Reviewed in watercooler.
could add some comments on the new isOneOne/IsAlone. This is not yet a snapshot as there is still some logic, but it's an improvement

PR #3568, opened 2025-11-17
refactor to createCallViewModel$

toger5:
This is a "experiement" to review how things would feel like if we move to a function for the creation of the call view model.

It has the following advantages:
 - No cluttering/misuse of `private readonly`
 - clear return type (public interface) with and a central point for documenting them
 - very clear input overview. Seeing what needs to injected for tests.
 - normal context for the whole function body (no class toplevel context where some things are awkward like adding intermediate fields)
 
Disadvantages:
 - It being a function might feel odd since a view model is a struct/class in the end.
 - A big change to the already big reafactor
 - It is unclear how it compares to a class appraoch where with strict rules
    - Do not initialize fields inline
    - declare all public types at the top of the class
    - add all impls in the constrcutor and set all public fields at the end of the constrcutor
    
    
^ the class approach seems to be very similar but might provide even better field documentation?

## Final decision
We opt for the class based appraoch since this is the smaller change since no imports and types need to be changed. (even though a class with just a constructor wants to be a function)

But we keep a similar structure a function would have had
```
// all public fields with types and docs
public myProp$: Behavior<bool>
...
constructor(
// all inputs with types and docs
...){
   // all the temp fields with impl
   const myTemp$ = ...
   
   // return = set all public fields
   this.myProp$ = myPorp
}

```


BillCarsonFr (approved):

toger5:
I added a bunch more comments.

PR #3571, opened 2025-11-17
`createCallViewModel$` function refactor

toger5:


src/state/CallViewModel/CallViewModel.ts
```diff
@@ -172,63 +172,64 @@ type AudioLivekitItem = {
 };
 
 /**
- * A view model providing all the application logic needed to show the in-call
- * UI (may eventually be expanded to cover the lobby and feedback screens in the
- * future).
+ * The return of createCallViewModel$
+ * This interface represents the root snapshot for the call view. Snapshots in EC with rxjs behave like snapshot trees.
+ * They are a list of observables and objects containing observables to allow for a very granular update mechanism.
```
robintown: I think I wouldn't call it a snapshot if it's not literally a static snapshot of plain data. I would say "this interface represents the root source of data for the call view", perhaps.

robintown (approved):
This does look very simple in the end.

PR #3572, opened 2025-11-18
Correct Caddy reverse proxy config example

alutarius:
remote_addr in the configuration example for using Caddy as reverse proxy is not a valid placeholder - see https://caddyserver.com/docs/caddyfile/concepts#placeholders .
Due to the mistake Caddy will pass the string "{remote_addr}" instead of the remote IP address. I guess this was missed while rewriting the nginx example.
The correct placeholder is remote_host


fkwp (approved):
thx for improving the caddy reverse proxy config example

PR #3575, opened 2025-11-20
Restrict importing src/lib in most cases for matrix- libs

Half-Shot:
Based off @BillCarsonFr 's struggles, we can use the linter to prevent us importing /paths, which is usually a bad smell.


BillCarsonFr (approved):
Tested and works perfectly, thx

toger5 (approved):
sweet

PR #3576, opened 2025-11-20
Refactor local transport testing and local membership initialization

toger5:
A test in the callViewModel uncovered that there is a much better way to split test for the call setup.
This splits the test into localTransport tests that check for the different error cases and a test for the local Membership that checks that on error we will end up in the correct state for the callViewModel. (based on this state we set `fatalError$`

Also splits homeserverConnected into its own file and behavior so that it can be tested seperatly and mocked in a dedicated way.

### DIFF
--- a/src/livekit/openIDSFU.ts
+++ b/src/livekit/openIDSFU.ts
@@ -22,11 +22,13 @@ export type OpenIDClientParts = Pick<
   "getOpenIdToken" | "getDeviceId"
 >;
 /**
- *
+ * Gets a bearer token from the homeserver and then use it to authenticate
+ * to the matrix RTC backend in order to get acces to the SFU.
+ * It has built-in retry for calls to the homeserver with a backoff policy.
  * @param client
  * @param serviceUrl
  * @param matrixRoomId
- * @returns
+ * @returns Object containing the token information
  * @throws FailToGetOpenIdToken
  */
 export async function getSFUConfigWithOpenID(
--- a/src/livekit/openIDSFU.ts
+++ b/src/livekit/openIDSFU.ts
@@ -21,7 +21,14 @@ export type OpenIDClientParts = Pick<
   MatrixClient,
   "getOpenIdToken" | "getDeviceId"
 >;
-
+/**
+ *
+ * @param client
+ * @param serviceUrl
+ * @param matrixRoomId
+ * @returns
+ * @throws FailToGetOpenIdToken
+ */
 export async function getSFUConfigWithOpenID(
   client: OpenIDClientParts,
   serviceUrl: string,
--- a/src/room/GroupCallErrorBoundary.tsx
+++ b/src/room/GroupCallErrorBoundary.tsx
@@ -22,6 +22,7 @@ import {
   WebBrowserIcon,
 } from "@vector-im/compound-design-tokens/assets/web/icons";
 import { Button } from "@vector-im/compound-web";
+import { logger } from "matrix-js-sdk/lib/logger";
 
 import {
   ConnectionLostError,
@@ -33,7 +34,6 @@ import {
 import { FullScreenView } from "../FullScreenView.tsx";
 import { ErrorView } from "../ErrorView.tsx";
 import { type WidgetHelpers } from "../widget.ts";
-import { logger } from "matrix-js-sdk/lib/logger";
 
 export type CallErrorRecoveryAction = "reconnect"; // | "retry" ;
 
--- a/src/room/GroupCallErrorBoundary.tsx
+++ b/src/room/GroupCallErrorBoundary.tsx
@@ -33,6 +33,7 @@ import {
 import { FullScreenView } from "../FullScreenView.tsx";
 import { ErrorView } from "../ErrorView.tsx";
 import { type WidgetHelpers } from "../widget.ts";
+import { logger } from "matrix-js-sdk/lib/logger";
 
 export type CallErrorRecoveryAction = "reconnect"; // | "retry" ;
 
@@ -53,7 +54,7 @@ const ErrorPage: FC<ErrorPageProps> = ({
   widget,
 }: ErrorPageProps): ReactElement => {
   const { t } = useTranslation();
-
+  logger.log("Error boundary caught:", error);
   let icon: ComponentType<SVGAttributes<SVGElement>>;
   switch (error.category) {
     case ErrorCategory.CONFIGURATION_ISSUE:
--- a/src/room/GroupCallErrorBoundary.tsx
+++ b/src/room/GroupCallErrorBoundary.tsx
@@ -54,7 +54,7 @@ const ErrorPage: FC<ErrorPageProps> = ({
   widget,
 }: ErrorPageProps): ReactElement => {
   const { t } = useTranslation();
-  logger.log("Error boundary caught:", error);
+  logger.error("Error boundary caught:", error);
   let icon: ComponentType<SVGAttributes<SVGElement>>;
   switch (error.category) {
     case ErrorCategory.CONFIGURATION_ISSUE:
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -266,7 +266,7 @@ export const InCallView: FC<InCallViewProps> = ({
   const sharingScreen = useBehavior(vm.sharingScreen$);
 
   const ringOverlay = useBehavior(vm.ringOverlay$);
-  const fatalCallError = useBehavior(vm.configError$);
+  const fatalCallError = useBehavior(vm.fatalError$);
   // Stop the rendering and throw for the error boundary
   if (fatalCallError) throw fatalCallError;
 
--- a/src/state/CallViewModel/CallViewModel.ts
+++ b/src/state/CallViewModel/CallViewModel.ts
@@ -97,8 +97,8 @@ import {
 } from "../layout-types.ts";
 import { type ElementCallError } from "../../utils/errors.ts";
 import { type ObservableScope } from "../ObservableScope.ts";
+import { createHomeserverConnected$ } from "./localMember/HomeserverConnected.ts";
 import {
-  createHomeserverConnected$,
   createLocalMembership$,
   enterRTCSession,
   LivekitState,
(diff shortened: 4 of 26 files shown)


src/livekit/openIDSFU.ts
```diff
@@ -21,7 +21,14 @@ export type OpenIDClientParts = Pick<
   MatrixClient,
   "getOpenIdToken" | "getDeviceId"
 >;
-
+/**
+ *
```
BillCarsonFr: Suggestion?

Gets a bearer token from the homeserver and then use it to authenticate to the matrix RTC backend in order to get acces to the SFU.
It has built-in retry for calls to the homeserver with a backoff policy. Will bubble errors from the matrix RTC backend :(.

src/room/GroupCallErrorBoundary.tsx
```diff
@@ -53,7 +54,7 @@ const ErrorPage: FC<ErrorPageProps> = ({
   widget,
 }: ErrorPageProps): ReactElement => {
   const { t } = useTranslation();
-
+  logger.log("Error boundary caught:", error);
```
BillCarsonFr: log.error?

src/room/InCallView.tsx
```diff
@@ -266,7 +266,7 @@ export const InCallView: FC<InCallViewProps> = ({
   const sharingScreen = useBehavior(vm.sharingScreen$);
 
   const ringOverlay = useBehavior(vm.ringOverlay$);
-  const fatalCallError = useBehavior(vm.configError$);
+  const fatalCallError = useBehavior(vm.fatalError$);
```
BillCarsonFr, line 269: 👍 far better name

src/state/CallViewModel/localMember/LocalMembership.test.ts
```diff
@@ -6,154 +6,234 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
-  },
-}));
-
-test("It joins the correct Session", async () => {
-  const focusFromOlderMembership = {
-    type: "livekit",
-    livekit_service_url: "http://my-oldest-member-service-url.com",
-    livekit_alias: "my-oldest-member-service-alias",
-  };
+// vi.mock("../../../widget", async (importOriginal) => ({
```
BillCarsonFr: nit: delete?

src/state/CallViewModel/localMember/LocalMembership.ts
```diff
@@ -102,17 +98,21 @@ export interface LocalMemberConnectionState {
...
-  options: Behavior<EnterRTCSessionOptions>;
+  // TODO add a comment into some code style readme or file header callviewmodel
+  // that the inputs for those createSomething$() functions should NOT contain any js-sdk objectes
   scope: ObservableScope;
-  mediaDevices: MediaDevices;
   muteStates: MuteStates;
   connectionManager: IConnectionManager;
-  matrixRTCSession: MatrixRTCSession;
-  matrixRoom: MatrixRoom;
+  createPublisherFactory: (connection: Connection) => Publisher;
```
BillCarsonFr, line 103: nice

src/state/CallViewModel/localMember/LocalMembership.ts
```diff
@@ -102,17 +98,21 @@ export interface LocalMemberConnectionState {
...
+  joinMatrixRTC: (trasnport: LivekitTransport) => Promise<void>;
```
BillCarsonFr, line 104: nice

src/state/CallViewModel/localMember/LocalMembership.ts
```diff
@@ -188,18 +181,32 @@ export const createLocalMembership$ = ({
...
   const tracks$ = new BehaviorSubject<LocalTrack[]>([]);
 
+  // unwrap the local transport and set the state of the LocalMembership to error in case the transport is an error.
+  const localTransport$ = scope.behavior(
+    localTransportCanThrow$.pipe(
+      catchError((e: unknown) => {
+        if (e instanceof ElementCallError) {
+          state.livekit$.next({ state: LivekitState.Error, error: e });
+        } else {
+          logger.error("Unknown error from localTransport$", e);
```
BillCarsonFr: Could we create an `UnknownCallError` instead?
`new UnknownCallError(error instanceof Error ? error : new Error("Unknown error from localTransport"))`

src/state/CallViewModel/localMember/LocalMembership.ts
```diff
@@ -265,18 +238,15 @@ export const createLocalMembership$ = ({
...
-      publisher$.next(
-        new Publisher(
-          scope,
-          connection,
-          mediaDevices,
-          muteStates,
-          trackProcessorState$,
-        ),
-      );
+      // @valere will take care of this!
```
BillCarsonFr: 😇

src/state/CallViewModel/localMember/LocalMembership.ts
```diff
@@ -509,9 +492,9 @@ export const createLocalMembership$ = ({
   const sharingScreen$ = scope.behavior(
     localConnection$.pipe(
       switchMap((c) =>
-        c === null
-          ? of(false)
-          : observeSharingScreen$(c.livekitRoom.localParticipant),
+        c !== null && c.livekitRoom
```
BillCarsonFr: livekitRoom can be null? 
In that case will we get notified when it is set on the connection? I am not convinced this switchmap works?

src/state/CallViewModel/localMember/LocalMembership.ts
```diff
@@ -631,3 +615,44 @@ export async function enterRTCSession(
     await widget.api.transport.send(ElementWidgetActions.JoinCall, {});
   }
 }
+
+/**
+ * Whether we are connected to the MatrixRTC session.
+ */
+export function createHomeserverConnected$(
```
BillCarsonFr: I think we talked about extracting that to its own file? with some tests. Maybe next PR

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -49,6 +46,8 @@ interface Props {
  *
  * @prop useOldestMember Whether to use the same transport as the oldest member.
  * This will only update once the first oldest member appears. Will not recompute if the oldest member leaves.
+ *
+ * @throws MatrixRTCTransportMissingError | FailToGetOpenIdToken
```
BillCarsonFr, line 60: 👍 I wonder if we can have a linter rule that uses that information to ensure the caller handle the errors

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -124,7 +133,7 @@ async function makeTransportInternal(
       "Using LiveKit transport from local storage: ",
       transportFromStorage,
     );
-    return transportFromStorage;
+    transport = transportFromStorage;
```
BillCarsonFr, line 146: Can we have a test for that? (checking that is trying local storage)

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -154,26 +162,15 @@ async function makeTransportInternal(
       livekit_alias: livekitAlias,
     };
     logger.log("Using LiveKit transport from config: ", transportFromConf);
-    return transportFromConf;
+    transport = transportFromConf;
```
BillCarsonFr, line 176: is it also returning transport?

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -446,15 +454,31 @@ export function createCallViewModel$(
...
     muteStates: muteStates,
-    mediaDevices: mediaDevices,
+    joinMatrixRTC: async (transport: LivekitTransport) => {
+      return enterRTCSession(
+        matrixRTCSession,
+        transport,
+        connectOptions$.value,
+      );
+    },
+    createPublisherFactory: (connection: Connection) => {
```
BillCarsonFr, line 470: nit: maybe we can extract it as a const just before instead of inlining? WDYT

BillCarsonFr (changes requested):
Great refactor 👍 
Just a few comments

src/state/CallViewModel/localMember/LocalMembership.ts
toger5: Good catch. It cannot be null
The story:
In a test it was not undefined
so this fell through and crashed. I did not check the types and assumend that this might be a possible scenario and adding a "undefined" check seemed like a non breaking okay thing todo. But just now it looked very confusing to me. Will fix this in the tests if it still is an issue!

src/state/CallViewModel/localMember/LocalTransport.ts
toger5: I searched a bit. not possible.

src/state/CallViewModel/localMember/LocalTransport.ts
toger5: this needed some update (the priority should have been the other way around) updated it.

src/state/CallViewModel/CallViewModel.ts
toger5: This was consciously done inline to make clear that this is "one Package" of input and not entangled with the rest of the call view model.
So that its easier to understand why callviewmodel test just mock the while `createLocalMembership$` return value.

src/state/CallViewModel/localMember/LocalMembership.ts
```diff
@@ -188,18 +176,38 @@ export const createLocalMembership$ = ({
   // This should be used in a combineLatest with publisher$ to connect.
   const tracks$ = new BehaviorSubject<LocalTrack[]>([]);
 
+  // unwrap the local transport and set the state of the LocalMembership to error in case the transport is an error.
+  const localTransport$ = scope.behavior(
+    localTransportCanThrow$.pipe(
+      catchError((e: unknown) => {
+        let error: ElementCallError;
+        if (e instanceof ElementCallError) {
```
BillCarsonFr, line 184: I think we have that at several place in the code, we might want some utility at some point

BillCarsonFr (approved):
Ok for me 👍

PR #3585, opened 2025-11-21
Add log to screenshare

toger5:
We need this to debug spurious screen share stops.

### DIFF
--- a/src/state/CallViewModel/localMember/LocalMembership.ts
+++ b/src/state/CallViewModel/localMember/LocalMembership.ts
@@ -10,6 +10,7 @@ import {
   type Participant,
   ParticipantEvent,
   type LocalParticipant,
+  type ScreenShareCaptureOptions,
 } from "livekit-client";
 import { observeParticipantEvents } from "@livekit/components-core";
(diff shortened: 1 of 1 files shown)


robintown (approved):

src/state/CallViewModel/localMember/LocalMembership.ts
```diff
@@ -516,27 +517,37 @@ export const createLocalMembership$ = ({
...
-              audio: true,
-              selfBrowserSurface: "include",
-              surfaceSwitching: "include",
-              systemAudio: "include",
-            })
-            .catch(logger.error)
-      : null;
+  ) {
+    toggleScreenSharing = (): void => {
+      const screenshareSettings: ScreenShareCaptureOptions = {
```
BillCarsonFr, line 526: could probably be a more global const SCREEN_SHARE_CAPTURE_OPTIONS ?
I wonder what is audio true

BillCarsonFr (approved):
could add some description on the PR that it is to investigate the unexpected stopping of screen shares

PR #3586, opened 2025-11-24
Remove all `OpenTelementry` related code

toger5:
Currently we do not use it in a good way. Its just a bunch of untested logic and unused code.

We still have this remove commit once merged which can be used to add it back if we need it in the way it was done until now.


toger5:
Screenshot 2025-11-24 at 12 48 20
Screenshot 2025-11-24 at 12 49 06

Half-Shot (approved):
For some reason for the longest time I read "otel" as a British person saying "Hotel"  and entirely missed this was OpenTelemetry.

Anyway 2026 is the year of clearing out dead code! Looks good to me.

PR #3590, opened 2025-11-25
Config: UrlParams to control noiseSuppression and echoCancellation

BillCarsonFr:
Add optional url params `echoCancellation` and `noiseSuppression` to setup the default audio capture info.
Both default to true (current behavior)


toger5:
Can this link the contribution that adds this via the settings modal?

src/state/CallViewModel/remoteMembers/ConnectionFactory.ts
```diff
@@ -41,8 +41,10 @@ export class ECConnectionFactory implements ConnectionFactory {
    * @param client - The OpenID client parts for authentication, needed to get openID and JWT tokens.
    * @param devices - Used for video/audio out/in capture options.
    * @param processorState$ - Effects like background blur (only for publishing connection?)
-   * @param e2eeLivekitOptions - The E2EE options to use for the LiveKit Room.
+   * @param livekitKeyProvider
```
toger5, line 45: good catch!

src/state/CallViewModel/remoteMembers/ConnectionFactory.ts
toger5: this function got to a point, where It would benefit from a `generateRoomOption({})` pattern.

Not required however. Its only used here.

src/state/CallViewModel/remoteMembers/ECConnectionFactory.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+import { ObservableScope } from "../../ObservableScope.ts";
+import { ECConnectionFactory } from "./ConnectionFactory.ts";
+import type { OpenIDClientParts } from "../../../livekit/openIDSFU.ts";
+import { exampleTransport, mockMediaDevices } from "../../../utils/test.ts";
+import type { ProcessorState } from "../../../livekit/TrackProcessorContext.tsx";
+import { constant } from "../../Behavior";
+
+// At the top of your test file, after imports
+vi.mock("livekit-client", async () => {
```
toger5: the first param of the factory should be `importOriginal`
```suggestion
vi.mock("livekit-client", async (actual) => {
```
so you can skip `  const actual = await vi.importActual("livekit-client");` (I think)

src/state/CallViewModel/remoteMembers/ECConnectionFactory.test.ts
toger5, line 1: very nice to have tests on this.

toger5 (approved):
looks very complete.

Just the comment on using the object based function signature & `const actual = await vi.importActual("livekit-client");` which is not blocking.

src/state/CallViewModel/remoteMembers/ECConnectionFactory.test.ts
BillCarsonFr: cannot really skip, but I can use the param as the import function

src/state/CallViewModel/remoteMembers/ConnectionFactory.ts
BillCarsonFr: done

PR #3591, opened 2025-11-25
Fix reactivity based on localTransport updates (Cleanup for LocalMembership.ts)

toger5:
This PR refactors LocalMembership.ts to use reconcile for cleaning up tracks. It also enables track and publisher creation to be reactive to the local transport and the connection.


src/state/CallViewModel/localMember/LocalMembership.ts
```diff
@@ -22,58 +22,77 @@ import {
...
-  type ConnectionState,
-} from "../remoteMembers/Connection.ts";
+import { type Connection } from "../remoteMembers/Connection.ts";
 
 export enum LivekitState {
-  Uninitialized = "uninitialized",
-  Connecting = "connecting",
-  Connected = "connected",
   Error = "error",
+  /** Not even a transport is available to the LocalMembership */
```
BillCarsonFr, line 58: nit: maybe we could name this RTCBackendState? rather than LivekitState? Because it could be confused to the actual Livekit Room connection state?
It would also better match the name we use in the EC readme main schema?

src/state/CallViewModel/localMember/LocalMembership.ts
```diff
@@ -136,48 +155,27 @@ export const createLocalMembership$ = ({
   muteStates,
   matrixRTCSession,
 }: Props): {
-  requestConnect: () => LocalMemberConnectionState;
+  requestConnect: () => void;
```
BillCarsonFr: nit: Maybe add some quick comments on what all these does?

src/state/CallViewModel/localMember/LocalMembership.ts
```diff
@@ -218,52 +216,23 @@ export const createLocalMembership$ = ({
...
     and$(
       homeserverConnected$,
-      localConnection$.pipe(
-        switchMap((c) =>
-          c
-            ? c.state$.pipe(map((state) => state.state === "ConnectedToLkRoom"))
-            : of(false),
-        ),
+      localConnectionState$.pipe(
+        map((state) => (state ? state.state === "ConnectedToLkRoom" : false)),
```
BillCarsonFr: When `state.state === "ConnectedToLkRoom"` should we then look at the inner `livekitConnectionState`? to look for `Disconnected` / `SignalReconnecting`?
```javascript
  | {
      state: "ConnectedToLkRoom";
      livekitConnectionState$: Observable<LivekitConenctionState>;
    }

File: src/state/CallViewModel/localMember/LocalMembership.ts [hunk 4]
```diff
@@ -286,103 +255,231 @@ export const createLocalMembership$ = ({
...
+  const publisher$ = new BehaviorSubject<Publisher | null>(null);
+  /**
+   * Extract the tracks from the published. Also reacts to changing publishers.
+   */
+  const tracks$ = scope.behavior(
+    publisher$.pipe(switchMap((p) => (p?.tracks$ ? p.tracks$ : constant([])))),
+  );
+  const publishing$ = scope.behavior(
+    publisher$.pipe(
+      switchMap((p) => (p?.publishing$ ? p.publishing$ : constant(false))),
```
BillCarsonFr: nit: I am not a fan of the condensed/ternary notation. Can I at least get a comment of what is doing and why?

src/state/ObservableScope.ts
```diff
@@ -123,32 +123,38 @@ export class ObservableScope {
...
   ): void {
     let latestValue: T | typeof nothing = nothing;
-    let reconciledValue: T | typeof nothing = nothing;
+    let reconcilePromise: Promise<void> | undefined = undefined;
     let cleanUp: (() => Promise<void>) | void = undefined;
+    let prevVal: T | typeof nothing = nothing;
+
+    // While this loop runs it will process the latest from `value$` until it caught up with the updates.
+    // It might skip updates from `value$` and only process the newest value after callback has resolved.
+    const reconcileLoop = async (): Promise<void> => {
```
BillCarsonFr, line 132: we might want to merge develop in to have the reconcile test https://github.com/element-hq/element-call/pull/3597

BillCarsonFr (changes requested):
Great work thanks. The quality gate is 🤩
We did some live pair session on it already.
I'd like just that we add more debug logs on the LocalMembership/publisher. things about the lifecycle:
`start tracks requested`
`Creating traks`
`Request disconnect`

Looks like we only have logging on error. I'd like that we add a lot more logs, and eventually revert later if too verbose or impacting perfs

src/state/ObservableScope.ts
toger5: good idea. Done

src/state/CallViewModel/localMember/LocalMembership.ts
toger5: this can become `    publisher$.pipe(switchMap((p) => p?.publishing$ ?? constant(false))),` which is much nicer imo

src/state/CallViewModel/localMember/LocalMembership.ts
toger5: and flatten the whole thing?
This is a good idea. But a bit of work. I will give it a try and see how it goes.

toger5: I think the best would actually be to have `state.state` to unwrap the inner state.
But for reducing the risk i kept similar to how it is now but added the check on`livekitConnectionState`

BillCarsonFr (approved):
Fine for me but playwright not happy

PR #3593, opened 2025-11-26
test: Enable quality gate and touch a file

BillCarsonFr:
Enabled the quality code for new code. 
Also touching a file and add a test, to check impact

### DIFF
--- a/src/state/ObservableScope.test.ts
+++ b/src/state/ObservableScope.test.ts
@@ -5,7 +5,7 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 Please see LICENSE in the repository root for full details.
 */
 
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import { BehaviorSubject, combineLatest, Subject } from "rxjs";
 import { logger } from "matrix-js-sdk/lib/logger";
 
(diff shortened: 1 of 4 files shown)


BillCarsonFr:
I added a non tested method to check that it is working

image
image

Will revert it now

toger5:
Is it possible to also add back the codecov report? So we can see by what value the quality gate was passed?

src/state/ObservableScope.test.ts
```diff
@@ -102,3 +102,52 @@ describe("Epoch", () => {
...
+          });
+        }
+        return Promise.resolve();
+      },
+    );
+
+    // behavior$.next(1);
+    await vi.advanceTimersByTimeAsync(200);
+    behavior$.next(2);
+    await vi.advanceTimersByTimeAsync(300);
```
toger5: The comment can be removed and we do not need the advance timers right. there never is a time check?

It could make the whole thing simpler to read however, if you just make it wait 100ms for setup and 100ms for cleanup. Than you can just use `behavior$.next` and `vi.advanceTimersByTimeAsync(100);` to control your reconsile.

src/state/ObservableScope.test.ts
```diff
@@ -102,3 +102,52 @@ describe("Epoch", () => {
...
+          return Promise.resolve(async (): Promise<void> => cleanUp1.promise);
+        } else if (value === 2) {
+          return Promise.resolve(async (): Promise<void> => {
+            cleanUp2();
+            return Promise.resolve(undefined);
+          });
+        }
+        return Promise.resolve();
+      },
+    );
```
toger5:
```suggestion
    const setup = vi.fn().mockImplementation(async ()=>await sleep(100)));
    const cleanup = vi.fn().mockImplementation(async ()=>await sleep(100)))
    scope.reconcile(
      behavior$,
      async () => {
       await setup();
       return cleanup
      },
    );
```
Could be enough for the `reconcile` call. Then the test can be fully controlled by: `behavior$.next` and `vi.advanceTimersByTimeAsync(100);` as outlined below.

BillCarsonFr:
> Is it possible to also add back the codecov report? So we can see by what value the quality gate was passed?

I have no idea why it didn't show up after reverting the commit

src/state/ObservableScope.test.ts
BillCarsonFr: I removed this test, was mostly to check how the quality gate respond to it. Will create a new PR for them

BillCarsonFr:
> > Is it possible to also add back the codecov report? So we can see by what value the quality gate was passed?
> 
> I have no idea why it didn't show up after reverting the commit

Looks like it is because the PR currently has no code changes, only some doc and new tests? report visible here https://app.codecov.io/gh/element-hq/element-call/pull/3593

toger5 (approved):

src/state/ObservableScope.test.ts
BillCarsonFr: Please see here https://github.com/element-hq/element-call/pull/3597

PR #3595, opened 2025-11-27
Reset overwrite url if it is invalid (does fail to reach sfu)

toger5:
The overwrite option currently can brick your element call and you need to update the local store manually. This PR automatically resets the overwrite url if it errored once while trying to get the livekit focus. (auto fallback to nen dev overwrite)

Screenshot 2025-11-27 at 15 08 25
Screenshot 2025-11-27 at 15 08 29
Screenshot 2025-11-27 at 15 08 34


src/settings/DeveloperSettingsTab.tsx
```diff
@@ -92,9 +93,8 @@ export const DeveloperSettingsTab: FC<Props> = ({
     alwaysShowIphoneEarpieceSetting,
   );
 
-  const [customLivekitUrl, setCustomLivekitUrl] = useSetting(
-    customLivekitUrlSetting,
-  );
+  const [customLivekitUrl, setCustomLivekitUrl, customLivekitUrlUpdateReason] =
```
Half-Shot: Can we test it here somehow, even just a basic attempt to auth? We can do nicer UX here and set the input to busy, let it run to test whether the URL works and that way the URL doesn't change unexpectedly. We can also show a nice error string :)

Half-Shot (changes requested):
We can do better UX here.

src/settings/DeveloperSettingsTab.tsx
toger5: Do you think this is worthit for a dev tool? I almost though i am doing too nice ux for a devtool

Half-Shot: Well, if it was a days effort then maybe but this should a relatively trivial bit of code. I think even devs should have a nice experience :)

toger5:
@Half-Shot I updated this PR so it now checks before setting the value.

src/settings/DeveloperSettingsTab.tsx
```diff
@@ -229,14 +234,36 @@ export const DeveloperSettingsTab: FC<Props> = ({
...
+            if (
+              customLivekitUrlTextBuffer === "" ||
+              customLivekitUrlTextBuffer === null
+            ) {
+              setCustomLivekitUrl(null);
+              return Promise.resolve();
+            }
+
+            try {
+              logger.debug("try setting");
```
Half-Shot: Assume all these debug logs will go before merge?

src/settings/DeveloperSettingsTab.tsx
```diff
@@ -229,14 +234,36 @@ export const DeveloperSettingsTab: FC<Props> = ({
...
-                ? null
-                : customLivekitUrlTextBuffer,
-            );
+          async (e: React.FormEvent<HTMLFormElement>): Promise<void> => {
+            if (
+              customLivekitUrlTextBuffer === "" ||
+              customLivekitUrlTextBuffer === null
+            ) {
+              setCustomLivekitUrl(null);
+              return Promise.resolve();
```
Half-Shot: Do we need to return an explicit resolve in an async function?

src/settings/DeveloperSettingsTab.tsx
```diff
@@ -229,14 +234,36 @@ export const DeveloperSettingsTab: FC<Props> = ({
...
+            }
+
+            try {
+              logger.debug("try setting");
+              await getSFUConfigWithOpenID(
+                client,
+                customLivekitUrlTextBuffer,
+                "Test-room-alias-" + Date.now().toString() + client.getUserId(),
```
Half-Shot:
```suggestion
                `Test-room-alias-${Date.now()}${client.getUserId()}`,
```

Half-Shot: What's the room alias thing about, btw?

src/settings/DeveloperSettingsTab.tsx
```diff
@@ -229,14 +234,36 @@ export const DeveloperSettingsTab: FC<Props> = ({
...
+              );
+              logger.debug("done setting! Success");
+              setCustomLivekitUrlUpdateError(null);
+              setCustomLivekitUrl(customLivekitUrlTextBuffer);
+            } catch (e) {
+              logger.error("failed setting", e);
+              setCustomLivekitUrlUpdateError("invalid URL (did not update)");
+              // automatically unset the error after 4 seconds (2 seconds will be for the save label)
```
Half-Shot: This seems like an unusual thing to do? Usually inputs change the error state when the value changes, 4 seconds isn't even enough time to understand the error.

src/settings/DeveloperSettingsTab.tsx
```diff
@@ -251,7 +278,13 @@ export const DeveloperSettingsTab: FC<Props> = ({
           },
           [setCustomLivekitUrl],
         )}
-      />
+      >
+        {customLivekitUrlUpdateError !== null && (
+          <Text size="sm" priority="low">
```
Half-Shot: This is not how you error an EditInPlace :)

https://compound.element.io/?path=/docs/compound-web_form-controls-editinplace--docs#with-error

uses a proper semantic error component with `serverInvalid` to ensure the field is in error state (so accessibility is informed etc)

src/settings/DeveloperSettingsTab.tsx
Half-Shot: Also, since you're not cancelling the timeout, you'll have these stack up if you try to save within that 2 second window.

Half-Shot (changes requested):

src/settings/DeveloperSettingsTab.tsx
toger5: This can be any id lk should use. But I am not sure the jwt service will use it for checking if we are part of the room. its saver to set this to the actual roomId.

src/settings/DeveloperSettingsTab.tsx
toger5: think it should be fine to just show the error until the user has actually entered sth reasonable.
 I was worried it might be confusing what is actually set right now if the control shows an error. But probably users will change it until they have set it to sth that works.

src/settings/DeveloperSettingsTab.tsx
toger5: At the time I did not have the async call below this was needed i suspect ;) thanks for catching it.

src/settings/DeveloperSettingsTab.test.tsx
```diff
@@ -5,13 +5,17 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
-import { render, waitFor } from "@testing-library/react";
+import { afterEach, describe, expect, it, type Mock, vi } from "vitest";
+import { render, waitFor, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { TooltipProvider } from "@vector-im/compound-web";
 
 import type { MatrixClient } from "matrix-js-sdk";
 import type { Room as LivekitRoom } from "livekit-client";
+// import { DeveloperSettingsTab } from "./DeveloperSettingsTab";
 import { DeveloperSettingsTab } from "./DeveloperSettingsTab";
```
Half-Shot: You seem to have a commented out import :)

src/settings/DeveloperSettingsTab.test.tsx
```diff
@@ -20,6 +24,14 @@ vi.mock("../UrlParams", () => ({
   }),
 }));
 
+// IMPORTANT: mock the same specifier used by DeveloperSettingsTab
+vi.mock("../livekit/openIDSFU", () => ({
+  getSFUConfigWithOpenID: vi.fn().mockResolvedValue({
+    url: "mock-url",
+    jwt: "mock-jwt",
```
Half-Shot, line 30: I think this is okay, but just beware there are bits of code now that expect a valid JWT so in the future this might fail. There is a fixture that exports a valid test JWT.

src/settings/DeveloperSettingsTab.test.tsx
```diff
@@ -52,33 +64,34 @@ function createMockMatrixClient(): MatrixClient {
...
+    url: string;
+    isLocal?: boolean;
+  }[] = [
+    createMockLivekitRoom(
+      "wss://local-sfu.example.org",
+      { region: "local", version: "1.2.3" },
+      "local-metadata",
+    ),
+    {
+      isLocal: false,
```
Half-Shot: could this also use `createMockLivekitRoom` (and thus avoid the `as unknown` stuff) if we expose `isLocal` as a param?

Half-Shot (approved):
Code looks good, I've added a few ideas but looks otherwise okay.

Issue #3587, opened 2025-11-24
We do NOT mute the camera in earpiece mode


toger5:
### Steps to reproduce

1. Where are you starting? What can you see?
2. What do you click?
3. More steps…

### Will you send logs?

Yes

PR #3596, opened 2025-11-27
Fix: Camera is not muted when the earpiece mode is enabled

BillCarsonFr:
See https://github.com/element-hq/element-call/issues/3587

Looks like there was a merge problem? These changes have been lost https://github.com/element-hq/element-call/pull/3351

I integrated it a bit differently and added the test.
We should get rid of MuteStates it is way too complex to understand, and since the refactoring it looks like part of what it is doing should be under the responsability of the Publisher/LocalMembership

### DIFF
--- a/src/state/MuteStates.ts
+++ b/src/state/MuteStates.ts
@@ -61,7 +61,7 @@ export class MuteState<Label, Selected> {
     this.handler$.next(defaultHandler);
   }
 
-  private readonly devicesConnected$ = combineLatest([
+  private readonly canControlDevices$ = combineLatest([
     this.device.available$,
     this.forceMute$,
   ]).pipe(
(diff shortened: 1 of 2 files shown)


src/state/MuteStates.ts
```diff
@@ -135,21 +159,45 @@ class MuteState<Label, Selected> {
     private readonly device: MediaDevice<Label, Selected>,
     private readonly joined$: Observable<boolean>,
     private readonly enabledByConfig: boolean,
+    /**
+     * An optional observable which, when it emits `true`, will force the mute.
+     * Used for video to stop camera when earpiece mode is on.
+     * @private
+     */
+    private readonly forceMute$: Observable<boolean>,
```
toger5, line 170: Maybe this is also what we should use if we run have an error for any of the devices. Add the force mute flag and then render the button as in disabled state.

BillCarsonFr: Yes we have to check for these errors and when they occur. But maybe now it will just make the `available` map empty, that is already muting. What error are you thinking about?

src/state/MuteStates.ts
```diff
@@ -38,31 +38,58 @@ interface MuteStateData {
...
     if (this.handler$.value !== defaultHandler)
       throw new Error("Multiple mute state handlers are not supported");
     this.handler$.next(handler);
   }
+
   public unsetHandler(): void {
     this.handler$.next(defaultHandler);
   }
 
+  private readonly devicesConnected$ = combineLatest([
```
toger5: I am not sure I like this name. `devicesConnected` now is implied by force mute?
I would expect the other way around, that no connected device implies force muting.
But force mute = true -> no devices connected feels like it messes with causality.

src/state/MuteStates.ts
toger5: An error while muting/unmuting. But I think I mixed things up since if we use the lower level api `mute` `unmute` this will never throw.

src/state/MuteStates.ts
BillCarsonFr: renamed to `canControlDevices` https://github.com/element-hq/element-call/pull/3596/commits/44980a2744e5f5b198c83e5de91f7d03aba0ab93

toger5 (approved):
Sorry for letting you wait on this for so long.

PR #3600, opened 2025-12-01
Element Call SDK target

toger5:
This adds another build target to the EC codebase to make it possible to build a library from the EC source, that allows using EC just as a rtc sdk (for real time data) inside a widget.


sdk/helper.ts
```diff
@@ -0,0 +1,55 @@
+/*
+Copyright 2025 New Vector Ltd.
```
BillCarsonFr: Copyright 2025 Element Creations Ltd.

sdk/main.ts
```diff
@@ -0,0 +1,308 @@
+/*
+Copyright 2025 New Vector Ltd.
```
BillCarsonFr: Copyright

vite-sdk.config.ts
```diff
@@ -0,0 +1,28 @@
+/*
+Copyright 2025 New Vector Ltd.
```
BillCarsonFr: copyright

BillCarsonFr (approved):
Ok for me but fix the copyrights

PR #3601, opened 2025-12-01
fix: Automatically switch to spotlight if remote screenshare

### LINKED ISSUE #3588: Starting to screenshare does not always switch to spotlight
### Steps to reproduce

2. What do you click?
3. More steps…

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

Yes

BillCarsonFr:
fixes #3588

Not sure when it did regress, the main change is to make `gridModeUserSelection$` emit an initial value (changed it from Subject, to BehaviorSubject`).

If it is no initial value, `gridMode$` is directly set to `grid` without doing the `switchmap` in the chain that is checking the current screenShare state in order to override the userChoise to `spotlight` when there is a screen share.

The test for that behavior was skipped, it is now enabled again

### DIFF
--- a/src/state/CallViewModel/CallViewModel.test.ts
+++ b/src/state/CallViewModel/CallViewModel.test.ts
@@ -267,7 +267,7 @@ describe("CallViewModel", () => {
     });
   });
 
-  it.skip("screen sharing activates spotlight layout", () => {
+  test("screen sharing activates spotlight layout", () => {
     withTestScheduler(({ behavior, schedule, expectObservable }) => {
       // Start with no screen shares, then have Alice and Bob share their screens,
       // then return to no screen shares, then have just Alice share for a bit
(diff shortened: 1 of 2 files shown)


src/state/CallViewModel/CallViewModel.test.ts
```diff
@@ -267,7 +267,7 @@ describe("CallViewModel", () => {
     });
   });
 
-  it.skip("screen sharing activates spotlight layout", () => {
```
toger5, line 270: oh wow this was actually skipped so we even had a test for that regression. This is super nice to see!

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -992,7 +991,7 @@ export function createCallViewModel$(
     spotlightExpandedToggle$.pipe(accumulate(false, (expanded) => !expanded)),
   );
 
-  const gridModeUserSelection$ = new Subject<GridMode>();
+  const gridModeUserSelection$ = new BehaviorSubject<GridMode>("grid");
```
toger5, line 994: while at it can we move
```
const setGridMode = (value: ..
``` 
right next to the BehaviorSubject?

toger5 (approved):

Half-Shot (approved):
This makes sense to me 👍 , more readable now too!

PR #3603, opened 2025-12-01
Retain remoteusermedia modifications across sessions

### LINKED ISSUE #3594: Keep user volume persistent between calls

When setting the volume of a user tile, if either the user disconnects or i disconnect in the future and establish a new session, the volume setting is reset for every user. 
It would be nice to have a toggleable setting that switches between the old behaviour and what i'm requesting here.

Users often have varying loudness depending on their environment or setup. 
This feature would allow me to adjust the volume of louder users and maintain that adjustment consistently between calls/sessions, instead of needing to manually adjust it each time.

A setting, when enabled, saves the volume adjustment for each user tile and applies it across different calls/sessions. 
This setting should persist until manually changed, even if the user disconnects and reconnects.

_No response_

### Additional context

_No response_

Half-Shot:
Fixes #3594 

This adds a new settings type that is per-userId, that tracks settings you've applied to another user and retains them for the next call. 

TODO: Need to think about per-userId+deviceId, as you probably only want this to apply to the specific device the user is using (at least for media controls)

### DIFF
--- a/src/state/MediaViewModel.ts
+++ b/src/state/MediaViewModel.ts
@@ -56,6 +56,7 @@ import { platform } from "../Platform";
 import { type MediaDevices } from "./MediaDevices";
 import { type Behavior } from "./Behavior";
 import { type ObservableScope } from "./ObservableScope";
+import { RemoteUserSetting } from "./RemoteUserSettings";
 
 export function observeTrackReference$(
   participant: Participant,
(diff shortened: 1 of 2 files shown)


robintown (commented):
I suspect this change makes it so that if you mute a participant, reload the page, and then try to unmute them by clicking the mute toggle, nothing happens. Storing both `volume` (reflects mute state) and `committedVolume` (non-zero, reflects the "unmuted" value) would be one way to fix this.

fkwp:
@daniellekirkwood  I think this need Product Input

WillNilges:
👋🏻 Wanted to bump this. Really interested in seeing this get merged :) Can I do anything to help?

rwjack:
@fkwp @daniellekirkwood hi, any updates on this? Seems like a pretty down to earth feature.

PR #3605, opened 2025-12-02
Fix: Allow to manually switch back to grid mode when there is a screenshare

BillCarsonFr:
A follow up on #3601 that did fix auto-switching to spotlight when there is a screenshare.
But #3601 also forced staying on spotlight when there is a screen share. 
There is an ongoing [discussion](https://github.com/element-hq/voip-internal/issues/427) with product to properly write down the wanted behavior.

Meanwhile to revert the previous behavior I created this PR, must be reviewed in 2 parts
1- https://github.com/element-hq/element-call/commit/60bc6f1e933fd837ba4173174cf3046185d725b9 Refactor to extract the mode switch code and add proper testing
2- https://github.com/element-hq/element-call/commit/b5d3f3c72a2d31d2bded511bdcffb0adb64901bc Change the behavior to allow switching to grid when there is a remote screenshare


src/state/CallViewModel/Layout.switch.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      cold("-ft-ft", { f: false, t: true }),
+    );
+
+    schedule("---g", {
+      g: () => setGridMode("grid"),
+    });
+
+    const expectation = "ggsg";
+    // If we did not respect manual selection, the expectation would be:
+    // const expectation = "ggsg-s";
```
robintown: Just wanted to note that this is a change in behavior from v0.16.3. I remember originally setting up the `gridMode$` logic so it *would* still be able to auto-switch after manually selecting grid, but this sounds reasonable too (pending the product discussion)

src/state/CallViewModel/LayoutSwitch.test.ts
robintown, line 1: The file name is mis-formatted

robintown (changes requested):
Looking closer, but the remaining regression I see compared to the v0.16.3 behavior is that now only a single auto-switch can happen per call (as there is nothing resetting the `hasAutoSwitched` flag). Here's a test to demonstrate:

```typescript
test("can auto-switch to spotlight again after first screen share ends", () => {
  withTestScheduler(({ cold, behavior, expectObservable }): void => {
    const shareMarble = "ftft";
    const gridsMarble = "gsgs";
    const { gridMode$ } = createLayoutModeSwitch(
      scope,
      behavior("n", { n: "normal" }),
      cold(shareMarble, { f: false, t: true }),
    );

    expectObservable(gridMode$).toBe(gridsMarble, {
      g: "grid",
      s: "spotlight",
    });
  });
});
```

src/state/CallViewModel/Layout.switch.test.ts
BillCarsonFr: Right, I put a comment for that 
```
// XXX If we want to allow switching automatically again after we can
// return hasAutoSwitched: false here instead of keeping the previous value.
```

Was not sure of the previous behavior, will update to keep the old behavior and update the test

BillCarsonFr: Update to maintain previous behavior until we have product decision https://github.com/element-hq/element-call/pull/3605/commits/a93ceeae4ba2402c7fc8ccc3d2c774ab66006210

robintown (changes requested):
Cool, nice to see that that previous test is now working. Should we add it to the test file, since it exercises a repeated auto-switch as opposed to the current tests which are only exercising an auto-switch after a manual switch?

I believe I found one more case in which the user is blocked from manually switching back to grid though:

```typescript
test("can switch manually to grid after screen share while manually in spotlight", () => {
  withTestScheduler(({ cold, behavior, schedule, expectObservable }): void => {
    // Initially, no one is sharing. Then the user manually switches to
    // spotlight. After a screen share starts, the user manually switches to
    // grid.
    const shareMarbles = "  f-t-";
    const setModeMarbles = "-s-g";
    const expectation = "   gs-g";
    const { gridMode$, setGridMode } = createLayoutModeSwitch(
      scope,
      behavior("n", { n: "normal" }),
      cold(shareMarbles, { f: false, t: true }),
    );
    schedule(setModeMarbles, {
      g: () => setGridMode("grid"),
      s: () => setGridMode("spotlight"),
    });

    expectObservable(gridMode$).toBe(expectation, {
      g: "grid",
      s: "spotlight",
    });
  });
});
```

This fails currently with the user being stuck in spotlight for the duration of the screen share.

BillCarsonFr:
> Cool, nice to see that that previous test is now working. Should we add it to the test file, since it exercises a repeated auto-switch as opposed to the current tests which are only exercising an auto-switch after a manual switch?
> 
> I believe I found one more case in which the user is blocked from manually switching back to grid though:
> 
> ```ts
> test("can switch manually to grid after screen share while manually in spotlight", () => {
>   withTestScheduler(({ cold, behavior, schedule, expectObservable }): void => {
>     // Initially, no one is sharing. Then the user manually switches to
>     // spotlight. After a screen share starts, the user manually switches to
>     // grid.
>     const shareMarbles = "  f-t-";
>     const setModeMarbles = "-s-g";
>     const expectation = "   gs-g";
>     const { gridMode$, setGridMode } = createLayoutModeSwitch(
>       scope,
>       behavior("n", { n: "normal" }),
>       cold(shareMarbles, { f: false, t: true }),
>     );
>     schedule(setModeMarbles, {
>       g: () => setGridMode("grid"),
>       s: () => setGridMode("spotlight"),
>     });
> 
>     expectObservable(gridMode$).toBe(expectation, {
>       g: "grid",
>       s: "spotlight",
>     });
>   });
> });
> ```
> 
> This fails currently with the user being stuck in spotlight for the duration of the screen share.

Good catch. Looks like we need to check if the screenshare observed is new, vs an old one that already triggered a change. 
I tried this https://github.com/element-hq/element-call/pull/3605/commits/fdc66a1d62bce8e38b0d89192eba0b70056e2f3c
It is getting more and more heary :/

src/state/CallViewModel/LayoutSwitch.ts
```diff
@@ -59,7 +59,13 @@ export function createLayoutModeSwitch(
...
-          { mode: GridMode; hasAutoSwitched: boolean }
+          {
+            mode: GridMode;
+            /** Remember if the change was user driven or not */
+            hasAutoSwitched: boolean;
+            /** To know if it is new screen share or an already handled */
+            prevShare: boolean;
+          }
         >(
           (acc, [userSelection, hasScreenShares, windowMode]) => {
```
robintown: I would find it slightly more readable to call prevShare "hasScreenShares" (since this is exactly what it reflects) and acc "prev" (since this stateful variable is where the idea of these values being "previous" comes in)

```suggestion
            hasScreenShares: boolean;
          }
        >(
          (prev, [userSelection, hasScreenShares, windowMode]) => {
```

src/state/CallViewModel/LayoutSwitch.ts
```diff
@@ -0,0 +1,130 @@
...
+            // TODO: strange that we do that for flat mode but not for other modes?
+            // TODO: Why is this not handled in layoutMedia$ like other window modes?
+            if (isFlatMode) {
+              logger.debug(`Forcing spotlight mode, windowMode=${windowMode}`);
+              return {
+                mode: "spotlight",
+                hasAutoSwitched: acc.hasAutoSwitched,
+                prevShare: hasScreenShares,
+              };
+            }
```
robintown, line 84: Sorry, just refreshed myself on what this `windowMode` condition was about. It's not that we want to force spotlight mode, as we should in fact support both spotlight and grid:

Screenshot From 2025-12-03 13-24-51
Screenshot From 2025-12-03 13-24-58

Rather, the purpose of the [original condition](https://github.com/element-hq/element-call/blob/768cf6f8ef3cab247b26e0379ad0b815a5782337/src/state/CallViewModel.ts#L1221) was to make the app *default* to spotlight rather than grid for mobile phones held in landscape orientation, as I remember that being what design wanted. So you would rotate your phone and it would immediately make the active speaker or screen share as big as it could be. (This was never explicitly documented, I know…)

From my perspective it's fine if for this PR, we remove this condition entirely and then try to restore that mobile behavior at a later time.

robintown (approved):
Apart from the `windowMode` condition, I think I'm convinced this is correct now. Still would like to add that first test to the suite just in case:

```typescript
test("can auto-switch to spotlight again after first screen share ends", () => {
  withTestScheduler(({ cold, behavior, expectObservable }): void => {
    const shareMarble = "ftft";
    const gridsMarble = "gsgs";
    const { gridMode$ } = createLayoutModeSwitch(
      scope,
      behavior("n", { n: "normal" }),
      cold(shareMarble, { f: false, t: true }),
    );

    expectObservable(gridMode$).toBe(gridsMarble, {
      g: "grid",
      s: "spotlight",
    });
  });
});
```

src/state/CallViewModel/LayoutSwitch.ts
BillCarsonFr: Thx, added a note here to track that https://github.com/element-hq/voip-internal/issues/427#issuecomment-3610896369

PR #3608, opened 2025-12-02
Fix the wrong layout being used until window size changes

robintown:
While looking into what had regressed https://github.com/element-hq/element-call/issues/3588, I found that 28047217b85e2e6f491c887ac7099499662fa46e had filled in a couple of behaviors with non-reactive default values, the "natural window mode" behavior being among them. This meant that the app would no longer determine the correct window mode upon joining a call, instead always guessing "normal" as the value. This change restores its reactivity.

### DIFF
--- a/src/state/CallViewModel/CallViewModel.test.ts
+++ b/src/state/CallViewModel/CallViewModel.test.ts
@@ -502,6 +502,48 @@ describe("CallViewModel", () => {
     });
   });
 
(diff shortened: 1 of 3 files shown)


BillCarsonFr (approved):
Great, fixes the bug on my side, where my iPhone was not starting in the correct layout but fixed itself after rotating

PR #3611, opened 2025-12-02
Unify LiveKit and Matrix connection states

toger5:
From internal discussions about this PR: https://matrix.to/#/!tDLCaLXijNtYcJZEey:element.io/$jPuIV8UAaCVC8IPEe8LaJPDbaZ85o6nnCdxVeQoMcx0?via=element.io&via=matrix.org&via=robin.town

> 
> - We have understood the actual model of the state. (They are NOT simply linear but change how many dimensions they have dependent on the state)
>     - 1D: `[loading transport + create LK Room]`
>     - 3D:`[connectingToLkRoom, createTracks, matrixState]`This one is the thing that makes it complicated. We can create tracks before we are connected to the lk room or after (!! connected to lk room is not the sama as lk room created) so this is the thing that made it feel so wrong with the prev appraoch!
>     - 2D:`[publishing, matrixState]`We can only start the publishing states once both independent prev states `connectingToLkRoom`, `createTracks` are ready.
> - We have found some good names for the states (some are still pending)
> - We have noticed that the `{state: string, error: Error}` structure is not ideal -> a simple "or-Type" is a lot simpler and more elegant `State | Error` (especially with the dimension changes)
>

### DIFF
--- a/playwright/fixtures/widget-user.ts
+++ b/playwright/fixtures/widget-user.ts
@@ -113,7 +113,8 @@ async function registerUser(
   await page.getByRole("button", { name: "Register" }).click();
   const continueButton = page.getByRole("button", { name: "Continue" });
   try {
-    await expect(continueButton).toBeVisible({ timeout: 5000 });
+    await expect(continueButton).toBeVisible({ timeout: 700 });
+    // why do we need to put in the passwor if there is a continue button?
     await page
       .getByRole("textbox", { name: "Password", exact: true })
       .fill(PASSWORD);
@@ -124,6 +125,16 @@ async function registerUser(
   await expect(
     page.getByRole("heading", { name: `Welcome ${username}` }),
   ).toBeVisible();
+
+  // Dismiss incompatible browser toast
+  const dismissButton = page.getByRole("button", { name: "Dismiss" });
+  try {
+    await expect(dismissButton).toBeVisible({ timeout: 700 });
+    await dismissButton.click();
+  } catch {
+    // dismissButton not visible, continue as normal
+  }
+
   await setDevToolElementCallDevUrl(page);
 
   const clientHandle = await page.evaluateHandle(() =>
--- a/playwright/fixtures/widget-user.ts
+++ b/playwright/fixtures/widget-user.ts
@@ -126,7 +126,6 @@ async function registerUser(
     page.getByRole("heading", { name: `Welcome ${username}` }),
   ).toBeVisible();
 
-  await page.pause();
   const browserUnsupportedToast = page
     .getByText("Element does not support this browser")
     .locator("..")
--- a/playwright/fixtures/widget-user.ts
+++ b/playwright/fixtures/widget-user.ts
@@ -111,17 +111,7 @@ async function registerUser(
   await page.getByRole("textbox", { name: "Confirm password" }).click();
   await page.getByRole("textbox", { name: "Confirm password" }).fill(PASSWORD);
   await page.getByRole("button", { name: "Register" }).click();
-  const continueButton = page.getByRole("button", { name: "Continue" });
-  try {
-    await expect(continueButton).toBeVisible({ timeout: 700 });
-    // why do we need to put in the passwor if there is a continue button?
-    await page
-      .getByRole("textbox", { name: "Password", exact: true })
-      .fill(PASSWORD);
-    await continueButton.click();
-  } catch {
-    // continueButton not visible, continue as normal
-  }
+
   await expect(
     page.getByRole("heading", { name: `Welcome ${username}` }),
   ).toBeVisible();
--- a/playwright/fixtures/widget-user.ts
+++ b/playwright/fixtures/widget-user.ts
@@ -126,8 +126,16 @@ async function registerUser(
     page.getByRole("heading", { name: `Welcome ${username}` }),
   ).toBeVisible();
 
+  await page.pause();
+  const browserUnsupportedToast = page
+    .getByText("Element does not support this browser")
+    .locator("..")
+    .locator("..");
+
   // Dismiss incompatible browser toast
-  const dismissButton = page.getByRole("button", { name: "Dismiss" });
+  const dismissButton = browserUnsupportedToast.getByRole("button", {
+    name: "Dismiss",
+  });
   try {
     await expect(dismissButton).toBeVisible({ timeout: 700 });
     await dismissButton.click();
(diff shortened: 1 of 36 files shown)


src/state/CallViewModel/remoteMembers/Connection.ts
```diff
@@ -47,17 +46,31 @@ export interface ConnectionOpts {
...
-    }
-  | { state: "FailedToStart"; error: Error }
-  | { state: "Stopped" };
+export enum ConnectionState {
+  /** The start state of a connection. It has been created but nothing has loaded yet. */
+  Initialized = "Initialized",
+  /** `start` has been called on the connection. It aquires the jwt info to conenct to the LK Room  */
+  FetchingConfig = "FetchingConfig",
+  Stopped = "Stopped",
+  ConnectingToLkRoom = "ConnectingToLkRoom",
```
toger5: remove this?

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -158,20 +161,17 @@ export class Publisher {
...
+    const sub = this.connection.state$.subscribe((state) => {
+      if (state instanceof Error) {
+        const error =
+          state instanceof ElementCallError
+            ? state
+            : new FailToStartLivekitConnection(state.message);
+        reject(error);
+      } else if (state === ConnectionState.LivekitConnected) {
+        resolve();
+      } else {
```
toger5: Not needed anymore when transitioning to higher lvl api.

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -349,140 +297,179 @@ export const createLocalMembership$ = ({
...
       }
     },
   );
 
-  const fatalLivekitError$ = new BehaviorSubject<ElementCallError | null>(null);
-  const setLivekitError = (e: ElementCallError): void => {
-    if (fatalLivekitError$.value !== null)
-      logger.error("Multiple Livkit Errors:", e);
-    else fatalLivekitError$.next(e);
+  const fatalMediaError$ = new BehaviorSubject<ElementCallError | null>(null);
```
toger5: Should this really be a fatal error.

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -349,140 +297,179 @@ export const createLocalMembership$ = ({
...
+          else {
+            return TransportState.Waiting;
+          }
+        },
+      ),
     ),
   );
 
-  // Keep matrix rtc session in sync with localTransport$, connectRequested$ and muteStates.video.enabled$
+  // inform the widget about the connect and disconnect intent from the user.
```
toger5, line 457: where does this come from.

playwright/fixtures/widget-user.ts
```diff
@@ -113,7 +113,8 @@ async function registerUser(
   await page.getByRole("button", { name: "Register" }).click();
   const continueButton = page.getByRole("button", { name: "Continue" });
   try {
-    await expect(continueButton).toBeVisible({ timeout: 5000 });
+    await expect(continueButton).toBeVisible({ timeout: 700 });
+    // why do we need to put in the passwor if there is a continue button?
```
BillCarsonFr: I think this can be removed now? Looks like it was some old onboarding UI from web that is not needed anymore?

playwright/fixtures/widget-user.ts
```diff
@@ -124,6 +125,16 @@ async function registerUser(
   await expect(
     page.getByRole("heading", { name: `Welcome ${username}` }),
   ).toBeVisible();
+
+  // Dismiss incompatible browser toast
```
BillCarsonFr, line 124: Can we add more test to be sure that we really dismiss the popup we think we dismiss (check the title)?

This is fixed by updating playwright, right? is it really needed?

src/room/GroupCallView.tsx
```diff
@@ -444,6 +446,7 @@ export const GroupCallView: FC<Props> = ({
 
   let body: ReactNode;
   if (externalError) {
+    logger.debug("External error occurred:", externalError);
```
BillCarsonFr: I think this is already logged by the GroupCallErrorBoundary

src/room/InCallView.tsx
```diff
@@ -127,6 +127,7 @@ export const ActiveCall: FC<ActiveCallProps> = (props) => {
   const mediaDevices = useMediaDevices();
   const trackProcessorState$ = useTrackProcessorObservable$();
   useEffect(() => {
+    logger.info("START CALL VIEW SCOPE");
```
BillCarsonFr, line 132: add some prefix to the log until we have proper prefixed logger? 
Something like `[InCallView]`

src/room/LobbyView.tsx
```diff
@@ -79,9 +79,9 @@ export const LobbyView: FC<Props> = ({
   waitingForInvite,
 }) => {
   useEffect(() => {
-    logger.info("[Lifecycle] GroupCallView Component mounted");
+    logger.info("[Lifecycle] LobbyView Component mounted");
```
BillCarsonFr, line 82: oups :D 
thx

BillCarsonFr (approved):
We've done a live review of that together, so it is fine by me.
Just some minor comments, but no blockers

playwright/fixtures/widget-user.ts
toger5: cool that makes sense. I was wondering in what cases it would be used.

src/room/GroupCallView.tsx
toger5: okay removed it

src/room/InCallView.tsx
toger5: I added a proper logger.

src/state/CallViewModel/localMember/LocalMember.ts
toger5: this was pulled to make the join rtc session functions sync.

playwright/fixtures/widget-user.ts
toger5: remove all `continueButton` related code.

playwright/fixtures/widget-user.ts
toger5: done

toger5: It helps us save time if we are slow with merging renovate and need to debug why our test fail.

PR #3614, opened 2025-12-03
Simplify computation of analytics ID

robintown:
Since we now bundle a trusted Element Call widget with our messenger applications and this widget reports analytics to an endpoint determined by the messenger app, there is no longer any reason to compute a different analytics ID from the one used by the messenger app.


toger5 (approved):
That makes sense.

PR #3616, opened 2025-12-04
Stop gap example how we can obtain the delay id

toger5:


toger5:
@fkwp closing this as it is just a code example and clutters our PR list.
You can of course still find the diff in the closed PR's

PR #3621, opened 2025-12-05
Add keyboard shortcut disable setting

Half-Shot:
Not *everyone* loves a keyboard shortcut, so allow disabling them if you are not a fan.

### DIFF
--- a/locales/en/app.json
+++ b/locales/en/app.json
@@ -224,6 +224,8 @@
     "preferences_tab": {
       "developer_mode_label": "Developer mode",
       "developer_mode_label_description": "Enable developer mode and show developer settings tab.",
+      "enable_keyboard_shortcuts_description": "Use keyboard shortcuts to control Element Call.",
+      "enable_keyboard_shortcuts_label": "Enable keyboard shortcuts",
       "introduction": "Here you can configure extra options for an improved experience.",
       "reactions_play_sound_description": "Play a sound effect when anyone sends a reaction into a call.",
       "reactions_play_sound_label": "Play reaction sounds",
--- a/src/settings/PreferencesSettingsTab.tsx
+++ b/src/settings/PreferencesSettingsTab.tsx
@@ -1,4 +1,5 @@
 /*
+Copyright (C) 2025 Element Creations Ltd
 Copyright 2024 New Vector Ltd.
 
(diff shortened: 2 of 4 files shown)


locales/en/app.json
```diff
@@ -224,6 +224,8 @@
     "preferences_tab": {
       "developer_mode_label": "Developer mode",
       "developer_mode_label_description": "Enable developer mode and show developer settings tab.",
+      "enable_keyboard_shortcuts_description": "Use keyboard shortcuts to control Element Call.",
```
robintown, line 227: The product name is configurable so it'll need to be templated in

robintown (commented):
There are keyboard shortcuts in ReactionToggleButton.tsx as well (and we should probably hide the shortcut hints from the tooltips if they're disabled)

Half-Shot:
@erikjohnston feedbacks that he's a difficult (reasonable) person who would like to have shortcuts for muting/unmuting and actions but the *numeric* reaction buttons are annoying.

I'm actually wondering if we just drop them, do they feel useful?

fkwp:
> @erikjohnston feedbacks that he's a difficult (reasonable) person who would like to have shortcuts for muting/unmuting and actions but the _numeric_ reaction buttons are annoying.
> 
> I'm actually wondering if we just drop them, do they feel useful?

that is something for product. And note that the shortcut situation is a bit weird as:
if EC widget is in foreground the messenger shortcuts are not working and vice versa.

cc @daniellekirkwood

erikjohnston:
> @erikjohnston feedbacks that he's a difficult (reasonable) person who would like to have shortcuts for muting/unmuting and actions but the _numeric_ reaction buttons are annoying.
> 
> I'm actually wondering if we just drop them, do they feel useful?

For context: I *heavily* rely on desktop shortcuts that involve numerals (either switching chrome tabs or linux workspaces), and I semi-frequently trigger reactions by accident due to mistyping the shortcuts. This is obviously quite bad as it triggers spurious noise on everyone who is on the call (whereas e.g. the mute/unmute buttons only have local affect so it matters less if you accidentally mute/unmute yourself, assuming you notice).

While being able to disable shortcuts for reactions would help me, I think reaction shortcuts should be harder to trigger (e.g. by adding a modifier key), given a) the impact and b) numeral shortcuts are commonly used.

Half-Shot:
I think this is probably the wrong solution, based on feedback gathered internally.

PR #3622, opened 2025-12-05
Bug fix | Race on initial publishing causing mute state to be desync

BillCarsonFr:


toger5 (commented):
I think this is the main place we need to use mute/unmute?
```
**
   * Observe changes in the mute states and update the LiveKit room accordingly.
   * @param scope
   * @private
   */
  private observeMuteStates(scope: ObservableScope): void {
    const lkRoom = this.connection.livekitRoom;
    this.muteStates.audio.setHandler(async (desired) => {
      try {
        await lkRoom.localParticipant.setMicrophoneEnabled(desired);
      } catch (e) {
        this.logger.error("Failed to update LiveKit audio input mute state", e);
      }
      return lkRoom.localParticipant.isMicrophoneEnabled;
    });
    this.muteStates.video.setHandler(async (desired) => {
      try {
        await lkRoom.localParticipant.setCameraEnabled(desired);
      } catch (e) {
        this.logger.error("Failed to update LiveKit video input mute state", e);
      }
      return lkRoom.localParticipant.isCameraEnabled;
    });
  }
  ```

File: src/state/CallViewModel/localMember/Publisher.test.ts [hunk 1]
```diff
@@ -33,8 +39,15 @@ import {
 import { type MuteStates } from "../../MuteStates";
 import { FailToStartLivekitConnection } from "../../../utils/errors";
 
+let scope: ObservableScope;
+
+beforeEach(() => {
+  scope = new ObservableScope();
+});
+
+afterEach(() => scope.end());
```
robintown, line 48: You might also find the `testScope` utility to be convenient

BillCarsonFr:
> I think this is the main place we need to use mute/unmute?
> 


I wonder if I don't prefer that we always use the high level API, it is doing quite a bunch of things.
But given the current high level API we cannot create without publishing. How much of a performance gain it is?

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -144,6 +144,7 @@ export class Publisher {
           this.logger.error("Failed to create tracks", error);
         });
     }
+    // TODO why throw here? should we just do nothing?
```
toger5, line 147: Yes I agree.

If we do the conditional check in here anyways the throw motivates/requests, that we do an additional check to never call this method if both are false. So the throw requests us to do it twice which is odd.

Removing the throw and adding a docstring to the function:
`This function will no-op in case both (audio and video) are not enabled (muted)`
seems like the way to go. We might still want to log in case we end up not creating tracks that could help us debug that case.

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -184,16 +185,36 @@ export class Publisher {
...
-        );
-      });
-      this.logger.info("published track ", track.kind, track.id);
 
+      // XXXX: Patch: Check if the track has been muted manually before publishing
+      // This is only a patch, the proper way would be to use livekit high-level enabled/disabled APIs
+      // or only use the low level create/publish APIs and have our own pending publication protection.
+      // Maybe we could change the livekit api to pre-load tracks without publishing them yet?
+      // Are we sure this is needed at all? What are the gains?
+      let isEnabled: boolean;
```
toger5, line 194: this should probably be `request_enabled` or `should_enable`

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -184,16 +185,36 @@ export class Publisher {
...
+        isEnabled = this.muteStates.audio.enabled$.value;
+      } else if (track.kind === Track.Kind.Video) {
+        isEnabled = this.muteStates.video.enabled$.value;
+      } else {
+        throw new Error("Unsupported track kind " + track.kind);
+      }
+
+      if (!isEnabled) {
+        // TODO should we also drop it?
+        // I believe the high-level LiveKit APIs will recreate a track?
```
toger5, line 205: That definitly needs checking. what does `enableVideo`/`enabledAudio` do if there is a muted track? 
 - unmute
 - create new track

toger5 (commented):
Our race is:
 - publishTrack (`await lkRoom.localParticipant.publishTrack(track)`) 
 - enableAudio

might be called at the same time once at the mute state handler and once in the `createAndSetupTracks` method.

Is this race also fixed by this PR?

toger5:
@BillCarsonFr What is the status of this one. ha this be fixed somehwere else?

BillCarsonFr:
> @BillCarsonFr What is the status of this one. ha this be fixed somehwere else?

We can close, we are using the highlevel API now https://github.com/element-hq/element-call/pull/3632

PR #3624, opened 2025-12-09
Dispel confusion about whether media is remote or local

### LINKED ISSUE #3609: Confusion about whether your media is remote or local after an unclean disconnect
### Steps to reproduce

1. Configure `delayed_leave_event_delay_ms` to a sufficiently large value like 90000 (so that you will be able to observe the confused app state for more than a couple seconds)
2. Join a call
3. Disconnect uncleanly by e.g. disabling your network connection
4. Rejoin the call

#### What did you expect?

I should now see my own user media in the call interface once again, and the app should know that this is local user media.

#### What happened instead?

I see my own user media in the call interface, but the app thinks it is a **remote** media item. This causes a number of things:

- The video is not horizontally mirrored
- Opening the context menu presents local volume & local mute controls, which do nothing
- In the console there are error messages coming from [the line where we try to set the local volume](https://github.com/element-hq/element-call/blob/f05d4b158e8cdad6697babe1cb2bc87153b7b954/src/state/MediaViewModel.ts#L711), because the participant object is actually a LocalParticipant, which does not have this `setVolume` method
- Probably other stuff relating to PiP and one-on-one layouts being broken

This heals itself as soon as the delayed event from the previous session is sent and the membership manager sends a new membership.

The root cause is [this condition in the UserMedia class](https://github.com/element-hq/element-call/blob/f05d4b158e8cdad6697babe1cb2bc87153b7b954/src/state/UserMedia.ts#L71), according to which a faulty cast of `participant$` will be made if the participant would be local but is initially undefined.

To fix this, we should remove the need for casting by properly splitting `MatrixLivekitMember` into two different `LocalMatrixLivekitMember` and `RemoteMatrixLivekitMember` types.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

robintown:
The UserMedia class contained some unsound casts of the participant behaviors. If for any reason your local MatrixRTC membership already existed while joining a call, we would create a UserMedia item for said membership, but the LiveKit participant would start out as `null`, and so UserMedia would mistakenly assume that the media item is for a _remote_ participant.

This confusion is immediately apparent if we run the CallViewModel tests in Matrix 2.0 mode; they fail due to broken layouts. So, I propose we run these tests against all possible MatrixRTC modes.

The fix is to add some more explicit type hints to the data (`type: "local" | "remote"`) so we no longer need to guess at what a `null` participant means. This makes the tests pass again.

You can review this commit-by-commit :)

Closes https://github.com/element-hq/element-call/issues/3609


src/state/UserMedia.ts
```diff
@@ -68,40 +69,46 @@ enum SortingBin {
...
-        this.displayName$,
-        this.mxcAvatarUrl$,
-        this.scope.behavior(this.handRaised$),
-        this.scope.behavior(this.reaction$),
-      )
-    : new RemoteUserMediaViewModel(
-        this.scope,
-        this.id,
-        this.userId,
-        this.participant$ as Behavior<RemoteParticipant | null>,
```
robintown, line 90: This cast specifically was the problem.

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -149,6 +149,8 @@ export interface CallViewModelOptions {
   connectionState$?: Behavior<ConnectionState>;
   /** Optional behavior overriding the computed window size, mainly for testing purposes. */
   windowSize$?: Behavior<{ width: number; height: number }>;
+  /** Optional behavior overriding the MatrixRTC mode, mainly for testing purposes. */
```
BillCarsonFr: nit: I think it would be fine to make it mandatory

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -504,25 +504,23 @@ export function createCallViewModel$(
...
-      localRtcMembership$.pipe(
-        switchMap((membership) => {
-          if (!membership) return of(null);
-          return of(
-            // casting is save here since we know that localRtcMembership$ is !== null since we reached this case.
-            localMatrixLivekitMemberUninitialized as MatrixLivekitMember,
-          );
+  const localMatrixLivekitMember$ = scope.behavior<MatrixLivekitMember | null>(
+    localRtcMembership$.pipe(
+      generateItems(
```
BillCarsonFr: so this is a collection of always one thing? there is only one possible key?

BillCarsonFr (approved):
Thx for breaking down in several commits, and for the explaination. 
Good point for the tests and nice to test all the modes
I just have some questions, but nothing blocking the approval

BillCarsonFr (changes requested):
Wait, I need another look at that. I just looked at a PR based on this and maybe I got it wrong

src/state/CallViewModel/CallViewModel.ts
toger5: I am struggling a bit why we need generateItems here. Or more precisely what part of generateItems do we need?
Since we are not using the scope we really just use it to return null if we dont have a membership.

My bias would be to use the least abstract functions of writing that code to not hide what its actual purpose is.
Right now I am not sure if there is sth else generateItems is doing that is super important but I dont see it.

robintown:
I had to keep the `TaggedParticipants` type. Trying to expand it into `({ type: 'local'; value$: Behavior<LocalParticipant | null> } | { type: 'remote'; value$: Behavior<RemoteParticipant | null> })` was becoming really verbose and more importantly confusing TypeScript into not widening types correctly.

src/state/CallViewModel/CallViewModel.ts
robintown: 00d4b8e985db7b9546d584610a777ef6a4a0e9f5 gets the non-nullable behavior that we want with a more appropriate operator.

src/state/CallViewModel/CallViewModel.ts
robintown: c7e9f1ce1449313adb2f3ef82661b8c70c222d95

BillCarsonFr (approved):
Thx Robin, I prefer it like that without the generateItems, and the tagged types looks sensible to me.
And learned new thing with the `satisfies` 😅
💪 for the stelar 97.38% coverage

PR #3625, opened 2025-12-09
Miscellaneous clean-up

robintown:
Some random things I felt compelled to clean up while investigating and working on the patch for https://github.com/element-hq/element-call/issues/3609.

### DIFF
--- a/src/state/CallViewModel/localMember/LocalMembership.ts
+++ b/src/state/CallViewModel/localMember/LocalMembership.ts
@@ -323,12 +323,14 @@ export const createLocalMembership$ = ({
   //  - overwrite current publisher
   scope.reconcile(localConnection$, async (connection) => {
     if (connection !== null) {
-      publisher$.next(createPublisherFactory(connection));
+      const publisher = createPublisherFactory(connection);
+      publisher$.next(publisher);
+      // Clean-up callback
+      return Promise.resolve(async (): Promise<void> => {
+        await publisher.stopPublishing();
+        publisher.stopTracks();
+      });
     }
-    return Promise.resolve(async (): Promise<void> => {
-      await publisher$?.value?.stopPublishing();
-      publisher$?.value?.stopTracks();
-    });
   });
 
   // Use reconcile here to not run concurrent createAndSetupTracks calls
--- a/src/state/CallViewModel/remoteMembers/ConnectionManager.ts
+++ b/src/state/CallViewModel/remoteMembers/ConnectionManager.ts
@@ -60,11 +60,7 @@ export class ConnectionManagerData {
     transport: LivekitTransport,
   ): (LocalParticipant | RemoteParticipant)[] {
     const key = transport.livekit_service_url + "|" + transport.livekit_alias;
-    const existing = this.store.get(key);
-    if (existing) {
-      return existing[1];
-    }
-    return [];
+    return this.store.get(key)?.[1] ?? [];
   }
   /**
    * Get all connections where the given participant is publishing.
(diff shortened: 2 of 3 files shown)


src/state/CallViewModel/localMember/LocalMembership.ts
```diff
@@ -323,12 +323,14 @@ export const createLocalMembership$ = ({
   //  - overwrite current publisher
   scope.reconcile(localConnection$, async (connection) => {
     if (connection !== null) {
-      publisher$.next(createPublisherFactory(connection));
+      const publisher = createPublisherFactory(connection);
+      publisher$.next(publisher);
+      // Clean-up callback
+      return Promise.resolve(async (): Promise<void> => {
```
Half-Shot, line 329: If this is an async function, I'm not quite sure what the `Promise.resolve(` is actually doing here? Presumably returning the callback directly has the same effect?

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
```diff
@@ -60,11 +60,7 @@ export class ConnectionManagerData {
     transport: LivekitTransport,
   ): (LocalParticipant | RemoteParticipant)[] {
     const key = transport.livekit_service_url + "|" + transport.livekit_alias;
-    const existing = this.store.get(key);
-    if (existing) {
-      return existing[1];
-    }
-    return [];
+    return this.store.get(key)?.[1] ?? [];
```
Half-Shot, line 63: Noting here the logic changes slightly. Before we only checked for `existing` and then returned `existing[1]` which might be undefined (?). This returns the value of `[1]` or [] if it's undefined.

Half-Shot: Oh, this is fine because the types of `this.store` require the array to have a length of 2.

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
```diff
@@ -115,9 +111,6 @@ export function createConnectionManager$({
   logger: parentLogger,
 }: Props): IConnectionManager {
   const logger = parentLogger.getChild("[ConnectionManager]");
-
-  const running$ = new BehaviorSubject(true);
```
Half-Shot, line 119: What was this doing, ooi?

Half-Shot (approved):
Nothing deadly here, asked a few Qs for my knowledge.

src/state/CallViewModel/localMember/LocalMembership.ts
robintown: It's to make the linter happy. We have a lint rule that says async functions need to look visibly async, whether by `await`ing something or returning a promise. It is a little annoying for callback cases like this.

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
robintown: correct

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
robintown: It was used to clear the transports array at the end of the scope. From the commit message:

> [Remove extraneous 'scope running' check](https://github.com/element-hq/element-call/pull/3625/commits/9481dc401c6c06c32085560431552247a56ddb16)
> Semantically, behaviors are only meaningful for as long as their scope is running. Setting a behavior's value to an empty array once its scope ends is not guaranteed to work (as it depends on execution order of how the scope is ended), and subscribers should be robust enough to handle clean-up of all connections at the end of the scope either way.

PR #3626, opened 2025-12-10
Don't show 'waiting for media' on connected participants

### LINKED ISSUE #3627: 'Waiting for media' when other party joins with audio and video disabled
### Steps to reproduce

1. Join a call
2. Have a remote party join the call with both their audio and video disabled

#### What did you expect?

To connect to their transport and see their (lack of) media as normal.

#### What happened instead?

Their media tile shows 'waiting for media' forever, as if we cannot connect.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

robintown:
We would show 'waiting for media' on participants that were connected but had no published tracks, because we were filtering them out of the remote participants list on connections. I believe this was done in an attempt to limit our view to only the participants that have a matching MatrixRTC membership. But that's fully redundant to the "Matrix-LiveKit members" module, which actually has the right information to do this (the MatrixRTC memberships).

After fixing that, I tried to improve some things that had been entangled. The commits have all the details.

Closes https://github.com/element-hq/element-call/issues/3627

### DIFF
--- a/src/state/MediaViewModel.ts
+++ b/src/state/MediaViewModel.ts
@@ -7,8 +7,8 @@ Please see LICENSE in the repository root for full details.
 
 import {
   type AudioSource,
-  type TrackReferenceOrPlaceholder,
   type VideoSource,
+  type TrackReference,
   observeParticipantEvents,
   observeParticipantMedia,
   roomEventSelector,
@@ -33,7 +33,6 @@ import {
   type Observable,
   Subject,
   combineLatest,
-  distinctUntilKeyChanged,
   filter,
   fromEvent,
   interval,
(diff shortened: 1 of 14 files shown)


src/tile/MediaView.tsx
```diff
@@ -129,7 +129,7 @@ export const MediaView: FC<Props> = ({
             />
           )}
         </div>
-        {!video && !localParticipant && (
+        {waitingForMedia && (
```
BillCarsonFr, line 132: 🤩 way more easy to read

BillCarsonFr (approved):
Ok for me

src/state/MediaViewModel.ts
```diff
@@ -596,6 +594,21 @@ export class LocalUserMediaViewModel extends BaseUserMediaViewModel {
...
+   * Whether we are waiting for this user's LiveKit participant to exist. This
+   * could be because either we or the remote party are still connecting.
+   */
+  public readonly waitingForMedia$ = this.scope.behavior<boolean>(
+    combineLatest(
+      [this.livekitRoom$, this.participant$],
+      (livekitRoom, participant) =>
+        // If livekitRoom is undefined, the user is not attempting to publish on
+        // any transport and so we shouldn't expect a participant. (They might
+        // be a subscribe-only bot for example.)
```
toger5, line 607: It would be nice to expose this property in the vm as well (probably not for this PR but to keep it in mind.

Custom rendering for participants without a transport would make recording and other integrations a lot nicer.

toger5 (approved):
Thanks Robin for the nice work on both PR's.
All the changes here gave me a cozy: "ah yea... thats better/how it should be" feeling.

PR #3629, opened 2025-12-11
Support MSC4143 RTC Transport endpoint

Half-Shot:
This implements MSC4143's backend endpoint for transports, and uses https://github.com/matrix-org/matrix-js-sdk/pull/5104 to handle it.

This changes makeTransport to:

 - Try *all* listed transports, not just the first livekit one.
 - Test all backend transports, then all well-known transports, then all config transports.


toger5:
This needs more coverage. 
It looks like throwing in a test that uses the new endpoint would get you almost to 100% since it would run thorugh both "throw levels"

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -116,73 +120,134 @@ export const createLocalTransport$ = ({
 const FOCI_WK_KEY = "org.matrix.msc4143.rtc_foci";
 
 /**
+ * Determine the correct Transport for the current session, including
+ * validating auth against the service to ensure it's correct.
+ * Prefers in order:
+ *
+ * 1. The transports returned via the homeserver.
```
toger5: This should prioritize transports set in the devtools?

toger5: should mention is what I mean

toger5 (approved):
This looks sensible.

If one of them fails we always run into a hard error. Do we want automatically check if we suceed in loading one of the lower prio ones?

Half-Shot:
> This needs more coverage. It looks like throwing in a test that uses the new endpoint would get you almost to 100% since it would run thorugh both "throw levels"

Yup! Working through it

src/state/CallViewModel/localMember/LocalTransport.ts
Half-Shot: Yeah, total miss on my side.

toger5:
In other words: I am not sure if:
> Test all backend transports, then all well-known transports, then all config transports.

is actually happening. Are we really continuing to check well-known transports if backend transports fail?
**edit**
at least in the
`ex instanceof FailToGetOpenIdToken` case.
**edit**
Oh which is an issue with our HS and entirely independent of the sfu/jwt.
Liking this more and more while looking into it :)

Half-Shot:
> Oh which is an issue with our HS and entirely independent of the sfu/jwt.
Liking this more and more while looking into it :)

That was my intention yes, but seeing as it wasn't clear this needs more comments on the PR.

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -116,73 +120,143 @@ export const createLocalTransport$ = ({
...
   if (urlFromDevSettings !== null) {
-    const transportFromStorage: LivekitTransport = {
+    logger.info("Using LiveKit transport from dev tools: ", urlFromDevSettings);
+    // Validate that the SFU is up. Otherwise, we want to fail on this
+    // as we don't permit other SFUs.
+    await getSFUConfigWithOpenID(client, urlFromDevSettings, livekitAlias);
+    return {
       type: "livekit",
       livekit_service_url: urlFromDevSettings,
       livekit_alias: livekitAlias,
```
fkwp: the livekit_alias is derived lk-jwt-service side and passed back to the client

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -116,73 +120,143 @@ export const createLocalTransport$ = ({
...
-  client: Pick<MatrixClient, "getDomain"> & OpenIDClientParts,
+  client: Pick<MatrixClient, "getDomain" | "_unstable_getRTCTransports"> &
+    OpenIDClientParts,
   roomId: string,
   urlFromDevSettings: string | null,
 ): Promise<LivekitTransport> {
-  let transport: LivekitTransport | undefined;
   logger.trace("Searching for a preferred transport");
-  //TODO refactor this to use the jwt service returned alias.
   const livekitAlias = roomId;
```
fkwp: I think we need to tackle this TODO also with this PR

Half-Shot: When I checked the service I couldn't see this in the response struct https://github.com/element-hq/lk-jwt-service/blob/main/main.go#L60?

Both APIs want us to pass in a room, and only provide a URL and JWT?

fkwp: The information is part of the token

```
{
  "exp": 1621657263,
  "iss": "APIMmxiL8rquKztZEoZJV9Fb",
  "sub": "myidentity",
  "nbf": 1619065263,
  "video": {
    "room": "myroom",
    "roomJoin": true
  },
  "metadata": ""
}
```

where `sub` is the LiveKit participant identity and `room` what we call `livekitAlias`

Half-Shot: Right right, in that case I'll make that part of this refactor.

playwright/sfu-reconnect-bug.spec.ts
```diff
@@ -100,5 +95,4 @@ test("When creator left, avoid reconnect to the same SFU", async ({
   // https://github.com/element-hq/element-call/issues/3344
   // The app used to request a new jwt token then to reconnect to the SFU
   expect(wsConnectionCount).toBe(1);
-  expect(sfuGetCallCount).toBe(2 /* the first one is for the warmup */);
```
Half-Shot, line 103: Removed this test because `livekit/jwt/sfu/get` is called whenever we just want to validate a transport, which unfortunately happens every time we instantiate LocalTransport, which seems to happen a few times over this call.

The websocket connection test *should* be enough here as it still tests we aren't making any reconnections.

PR #3630, opened 2025-12-11
Make EC compatible with custom `member.id` value (Pseudonomous users on the SFU)

toger5:
This is the companioning PR for: https://github.com/matrix-org/matrix-js-sdk/pull/5105

It allows to use a custom UUID value in the `member.id` field.
 - compatible with matrix2.0 (it expects the livekit identity to be what we have in member.id)
 - allows us to hide the identity from he SFU.

It also makes sure EC (from v0.17.0 and onwards) will always send the `membershipID` or `member.id` with the expected value.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "livekit-client": "^2.13.0",
     "lodash-es": "^4.17.21",
     "loglevel": "^1.9.1",
-    "matrix-js-sdk": "^39.2.0",
+    "matrix-js-sdk": "git://github.com/matrix-org/matrix-js-sdk.git#5af046f54fc891c62f99b1400aad9964157a87b6",
     "matrix-widget-api": "^1.14.0",
     "normalize.css": "^8.0.1",
     "observable-hooks": "^4.2.3",
--- a/src/state/CallViewModel/remoteMembers/MatrixLivekitMembers.ts
+++ b/src/state/CallViewModel/remoteMembers/MatrixLivekitMembers.ts
@@ -87,8 +87,13 @@ export function createMatrixLivekitMembers$({
         // Each change in the keys (new key, missing key) will result in a call to the factory function.
         function* ([membershipsWithTransports, managerData]) {
           for (const { membership, transport } of membershipsWithTransports) {
-            // TODO! cannot use membership.membershipID yet, Currently its hardcoded by the jwt service to
-            const participantId = /*membership.membershipID*/ `${membership.userId}:${membership.deviceId}`;
+            // membership.membershipID will default to `${membership.userId}:${membership.deviceId}` if undefined.
+            // membership.membershipID will be set to sessionEvent.content.membershipID (stateEvents)
+            // or matrixRTCEvent.content.member.id (stickyEvents)
+            // On top of that session memebrship events from v0.17.0 upwards are going to be sent with `${membership.userId}:${membership.deviceId}`
+            // And sticky event rtc memberships are might be sent with a UUID. This is why this has to be fetched from the memerbship because the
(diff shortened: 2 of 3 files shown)


robintown (changes requested):
I would like us to think through the following scenario: What if Alice joins a call, and then Eve joins a call using the very same membership ID as Alice? The app, SDK, and JWT service all seem at risk of getting confused in certain ways.

- **impact to Element Call**: minimal, sort of.
  - From the perspective of a 3rd party, Bob, we would end up creating two RemoteUserMediaViewModels with the same LiveKit participant. They could get confused about their local volume, since they are both fighting for control over the same participant, though that's not super worrying. We would need to be vigilant in the future to not assume exclusive ownership of a participant in our media code.
  - From the perspective of Alice, she would see Eve as perpetually in a 'waiting for media' state, because there would be no remote participant corresponding to her. It seems not great that Eve's impersonation attempt would remain hidden from Alice in this split-brained way.
- **impact to matrix-js-sdk**: Right now the SDK still uses the hardcoded ID format in [`getParticipantId`](https://github.com/matrix-org/matrix-js-sdk/blob/5af046f54fc891c62f99b1400aad9964157a87b6/src/matrixrtc/utils.ts#L49), which is used by the encryption managers. Assuming we were to change this (as I think we would need to for compatibility), Eve could then DoS Alice's media by sending garbage keys to Bob, which would override Alice's real key. Bob would fail to decrypt Alice's media.
- **impact to the JWT service**: If the JWT service is changed to accept any requested participant ID rather than computing an ID itself, it would no longer be able to block certain DoS or impersonation attempts from others joining a call. If Alice has to do a full LiveKit reconnect for any reason, Eve could jump in and grab the rights to her participant ID.

I know we can't in general guarantee the authenticity of media because SFrame has no authentication mechanism, but now that I'm looking at it more closely I'm not sure custom membership IDs work for the current threat model. If implemented across the stack it gives some undesirable abilities to unprivileged users.

toger5:
@robintown I get the point. I think you are correct. Still need to think through the exact impact.

We have discussed two ways: fully random (UUID) and hashing the expected user + room+ device+ appliaction .... in a defined way.
This would be possible to validate so users could not send state /sticky events that are invalid.
BUT it would allow the opposite attack where I can publish media for a hash i know someone else will compute if they join. (I can already do this with the device id today. the user id is authenticated via the openId token)

robintown:
> hashing the expected user + room+ device+ appliaction .... in a defined way.

Sure, with a salt that could work. I still wonder what the right choice is, generally, between putting computed values in events (which other users must validate and therefore re-compute) versus just asking others to compute the value themselves (no risk of forgotten validation this way).

> BUT it would allow the opposite attack where I can publish media for a hash i know someone else will compute if they join.

Not if the JWT service still verifies the computation (meaning we reveal our identity to the JWT service but not the SFU)

toger5:
> Not if the JWT service still verifies the computation (meaning we reveal our identity to the JWT service but not the SFU)

Yes like it is now. it can do so with our userId but not deviceId application and so on.

fkwp:
> ### Pseudonymous LiveKit Participant Identity
> 
> To protect user privacy, a pseudonymous LiveKit participant identity is used, so the Matrix user ID
> is not exposed to the LiveKit SFU backend. This pseudonymous identity is given by the SHA-256 hash
> of the concatenation of the Matrix `user_id`, a pipe character (`|`) , the `claimed_device_id`,
> a pipe character (`|`) and the `member.id` field, e.g, `SHA256(user_id|claimed_device_id|member.id)`.



is the latest update on the MSC https://github.com/matrix-org/matrix-spec-proposals/pull/4195/commits/ad88a5aaf3b5f6c59d850c1349d40b25b73bbf56

toger5:
Closing in favour of #3636

PR #3632, opened 2025-12-12
Fix: Races that could cause double publication or mute state desyncs

BillCarsonFr:
## Problem:

Livekit's `LocalParticipant` kind of offer two different levels of API.

- A **Low-Level** one: With `createTracks` and `publishTrack`, then accessing directly the tracks to call `mute`/`unmute` etc..
- A **High-Level** one: With `enableCameraAndMicrophone`/`setCameraEnabled`/`setMicrophoneEnabled`. These one will automatically creates the tracks, manage them, remember if there is any pending publications. For example calling `setCameraEnabled(false)` will check if there is already a track or if there is a pending publication and waits for it to complete.

Our `Publisher.ts` wants control on track creation and publication, so it is using the **Low-Level** APIs to createTracks.
**But** it is also using the **High-Level** ones in the mutestates handler with `setCameraEnabled`/`setMicrophoneEnabled`.
This mixed usage of API levels cause problems. For example the `Publisher` will create a track, but meanwhile the mutestate will call  `setCameraEnabled(true)`, the high level is not aware of the tracks that the `Publisher` have created so it will create a new one. Then later the `Publisher` will publish the one it manages.
This can be observed in rage-shake as per this log `publishing a second track with the same source:camera`

## Proposal

We have to stop mixing high-level and low-level.
This PR is an attempt at using only the High-Level API. 
It is sometimes a bit odd because the `Publisher` want to have more control than what the API allows (like create track and wait to publish)
We could eventually create an alternative PR to use the low-level API, but this will then require us to re-built a proper track management system (like tracking pending publications)

### DIFF
--- a/src/state/CallViewModel/localMember/LocalMember.ts
+++ b/src/state/CallViewModel/localMember/LocalMember.ts
@@ -303,7 +303,7 @@ export const createLocalMembership$ = ({
       // Clean-up callback
       return Promise.resolve(async (): Promise<void> => {
         await publisher.stopPublishing();
-        publisher.stopTracks();
+        await publisher.stopTracks();
       });
     }
   });
--- a/src/state/CallViewModel/localMember/LocalMember.ts
+++ b/src/state/CallViewModel/localMember/LocalMember.ts
@@ -74,6 +74,8 @@ export enum PublishState {
   Publishing = "publish_publishing",
 }
 
+// TODO not sure how to map that correctly with the
+// new publisher that does not manage tracks itself anymore
 export enum TrackState {
   /** The track is waiting for user input to create tracks (waiting to call `startTracks()`) */
   WaitingForUser = "tracks_waiting_for_user",
@@ -244,7 +246,7 @@ export const createLocalMembership$ = ({
     if (error) {
       logger.error(`Failed to create local tracks:`, error);
       setMatrixError(
-      // TODO is it fatal? Do we need to create a new Specialized Error?
+        // TODO is it fatal? Do we need to create a new Specialized Error?
         new UnknownCallError(new Error(`Media device error: ${error}`)),
       );
     }
@@ -327,11 +329,11 @@ export const createLocalMembership$ = ({
 
   // Based on `connectRequested$` we start publishing tracks. (once they are there!)
   scope.reconcile(
-    scope.behavior(
-      combineLatest([publisher$, joinAndPublishRequested$]),
-    ),
+    scope.behavior(combineLatest([publisher$, joinAndPublishRequested$])),
     async ([publisher, shouldJoinAndPublish]) => {
-      if (shouldJoinAndPublish) {
+      // Get the current publishing state to avoid redundant calls.
+      const isPublishing = publisher?.shouldPublish === true;
+      if (shouldJoinAndPublish && !isPublishing) {
         try {
           await publisher?.startPublishing();
         } catch (error) {
@@ -339,7 +341,7 @@ export const createLocalMembership$ = ({
             error instanceof Error ? error.message : String(error);
           setPublishError(new FailToStartLivekitConnection(message));
         }
-      } else {
+      } else if (isPublishing) {
         try {
           await publisher?.stopPublishing();
         } catch (error) {
@@ -409,7 +411,9 @@ export const createLocalMembership$ = ({
           // let trackState: TrackState = TrackState.WaitingForUser;
           // if (hasTracks && shouldStartTracks) trackState = TrackState.Ready;
           // if (!hasTracks && shouldStartTracks) trackState = TrackState.Creating;
-          const trackState: TrackState = shouldStartTracks ? TrackState.Ready : TrackState.WaitingForUser;
+          const trackState: TrackState = shouldStartTracks
+            ? TrackState.Ready
+            : TrackState.WaitingForUser;
 
           if (
             localConnectionState !== ConnectionState.LivekitConnected ||
--- a/src/state/CallViewModel/localMember/LocalMember.ts
+++ b/src/state/CallViewModel/localMember/LocalMember.ts
@@ -10,7 +10,6 @@ import {
   ParticipantEvent,
   type LocalParticipant,
   type ScreenShareCaptureOptions,
-  ConnectionState,
   RoomEvent,
   MediaDeviceFailure,
 } from "livekit-client";
@@ -329,11 +328,10 @@ export const createLocalMembership$ = ({
   // Based on `connectRequested$` we start publishing tracks. (once they are there!)
   scope.reconcile(
     scope.behavior(
-      combineLatest([publisher$, tracks$, joinAndPublishRequested$]),
+      combineLatest([publisher$, joinAndPublishRequested$]),
     ),
-    async ([publisher, tracks, shouldJoinAndPublish]) => {
-      if (shouldJoinAndPublish === publisher?.publishing$.value) return;
-      if (tracks.length !== 0 && shouldJoinAndPublish) {
+    async ([publisher, shouldJoinAndPublish]) => {
+      if (shouldJoinAndPublish) {
         try {
           await publisher?.startPublishing();
         } catch (error) {
@@ -341,7 +339,7 @@ export const createLocalMembership$ = ({
             error instanceof Error ? error.message : String(error);
           setPublishError(new FailToStartLivekitConnection(message));
         }
-      } else if (tracks.length !== 0 && !shouldJoinAndPublish) {
+      } else {
         try {
           await publisher?.stopPublishing();
         } catch (error) {
@@ -391,15 +389,6 @@ export const createLocalMembership$ = ({
     combineLatest([
       localConnectionState$,
       localTransport$,
-      // tracks$.pipe(
-      //   tap((t) => {
-      //     logger.info("tracks$: ", t);
-      //   }),
-      // ),
-      // publishing$,
-      connectRequested$,
-      tracks$,
-      publishing$,
       joinAndPublishRequested$,
       from(trackStartRequested.promise).pipe(
         map(() => true),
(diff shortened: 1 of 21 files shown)


src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -152,7 +154,7 @@ export const createLocalMembership$ = ({
   /**
    * This starts audio and video tracks. They will be reused when calling `requestPublish`.
    */
-  startTracks: () => Behavior<LocalTrack[]>;
+  startTracks: () => Behavior<void>;
```
toger5: what is a void behavior?

toger5: It looks like this should be a `undefined`

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -66,17 +68,19 @@ export enum TransportState {
 
 export enum PublishState {
   WaitingForUser = "publish_waiting_for_user",
-  /** Implies lk connection is connected */
-  Starting = "publish_start_publishing",
+  // /** Implies lk connection is connected */
+  // Starting = "publish_start_publishing",
```
toger5, line 74: Comment that this might get added back.

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -152,7 +156,7 @@ export const createLocalMembership$ = ({
   /**
    * This starts audio and video tracks. They will be reused when calling `requestPublish`.
    */
-  startTracks: () => Behavior<LocalTrack[]>;
+  startTracks: () => Behavior<void>;
```
toger5: Why does it return in the first place if it is a void behavior?
Would a void promise or just a void return make more sense?

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -238,16 +268,16 @@ export const createLocalMembership$ = ({
...
+  // );
+  // const publishing$ = scope.behavior(
+  //   publisher$.pipe(switchMap((p) => p?.publishing$ ?? constant(false))),
+  // );
+
+  const startTracks = (): Behavior<void> => {
     trackStartRequested.resolve();
-    return tracks$;
+    return constant(undefined);
   };
```
toger5: Yes lets remove the return type entirely or add a comment that this should return `tracks$`

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -282,40 +312,50 @@ export const createLocalMembership$ = ({
...
         }
       }
+      // XXX Why is that?
+      // else {
+      //   try {
+      //     await publisher?.stopPublishing();
+      //   } catch (error) {
+      //     setLivekitError(new UnknownCallError(error as Error));
+      //   }
+      // }
```
toger5: The commented section now is redundant with `else if (isPublishing)`
But the whole stopPublishing block is for when shouldJoinAndPublish becomes false. (I think we dont need this case now) but i would expect, that setting the subject to false stops the publications.

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -363,16 +401,19 @@ export const createLocalMembership$ = ({
...
         ]) => {
           if (!localTransport) return null;
-          const hasTracks = tracks.length > 0;
-          let trackState: TrackState = TrackState.WaitingForUser;
-          if (hasTracks && shouldStartTracks) trackState = TrackState.Ready;
-          if (!hasTracks && shouldStartTracks) trackState = TrackState.Creating;
+          // const hasTracks = tracks.length > 0;
+          // let trackState: TrackState = TrackState.WaitingForUser;
+          // if (hasTracks && shouldStartTracks) trackState = TrackState.Ready;
+          // if (!hasTracks && shouldStartTracks) trackState = TrackState.Creating;
```
toger5: I think this also can get a generic comment.

But I prefer just removing the core entirely. Maybe linking this PR instead as reference.

src/state/CallViewModel/localMember/Publisher.test.ts
```diff
@@ -5,117 +5,356 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
-        localParticipant: mockLocalParticipant({}),
-      }),
-    } as unknown as Connection;
+let scope: ObservableScope;
+
+beforeEach(() => {
+  scope = new ObservableScope();
+});
+
+// afterEach(() => scope.end());
```
toger5: why do we not want to do this?

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -76,161 +81,178 @@ export class Publisher {
     this.scope.onEnd(() => {
       this.logger.info("Scope ended -> stop publishing all tracks");
       void this.stopPublishing();
+      muteStates.audio.unsetHandler();
+      muteStates.video.unsetHandler();
```
toger5, line 85: oh that is a really good catch!

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -76,161 +81,178 @@ export class Publisher {
...
+    );
   }
 
-  private _tracks$ = new BehaviorSubject<LocalTrack<Track.Kind>[]>([]);
-  public tracks$ = this._tracks$ as Behavior<LocalTrack<Track.Kind>[]>;
-
+  // LiveKit will publish the tracks as soon as they are created
+  // but we want to control when tracks are published.
+  // We cannot just mute the tracks, even if this will effectively stop the publishing,
+  // it would also prevent the user from seeing their own video/audio preview.
```
toger5, line 97: This is good that its commented! Its the immediate question that comes to mind: Why not just muting?

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -76,161 +81,178 @@ export class Publisher {
...
-   *
-   * This will:
-   * wait for the connection to be ready.
-   // * 1. Request an OpenId token `request_token` (allows matrix users to verify their identity with a third-party service.)
-   // * 2. Use this token to request the SFU config to the MatrixRtc authentication service.
-   // * 3. Connect to the configured LiveKit room.
-   // * 4. Create local audio and video tracks based on the current mute states and publish them to the room.
    *
-   * @throws {InsufficientCapacityError} if the LiveKit server indicates that it has insufficient capacity to accept the connection.
-   * @throws {SFURoomCreationRestrictedError} if the LiveKit server indicates that the room does not exist and cannot be created.
```
toger5, line 113: Can we get an updated version of this?

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -76,161 +81,178 @@ export class Publisher {
...
+      if (audio && video) {
+        // Enable both at once in order to have a single permission prompt!
+        await lkRoom.localParticipant.enableCameraAndMicrophone();
+      } else if (audio) {
+        await lkRoom.localParticipant.setMicrophoneEnabled(true);
+      } else if (video) {
+        await lkRoom.localParticipant.setCameraEnabled(true);
+      }
+      return;
+    };
```
toger5:
```
    const enableTracks = ():void => {
      if (audio && video) {
        // Enable both at once in order to have a single permission prompt!
        void lkRoom.localParticipant.enableCameraAndMicrophone();
      } else if (audio) {
        void lkRoom.localParticipant.setMicrophoneEnabled(true);
      } else if (video) {
        void lkRoom.localParticipant.setCameraEnabled(true);
      }
    };
```

We can just explicitly void the promises.

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -76,161 +81,178 @@ export class Publisher {
...
+    const lkRoom = this.connection.livekitRoom;
+    for (const source of [
+      Track.Source.Microphone,
+      Track.Source.Camera,
+      Track.Source.ScreenShare,
+    ]) {
+      const localPub = lkRoom.localParticipant.getTrackPublication(source);
+      if (localPub?.track) {
+        // stops and unpublishes the track
+        await lkRoom.localParticipant.unpublishTrack(localPub!.track, true);
```
toger5, line 260: do we know the exact difference between unpublish and pause upstream?

toger5: why is this not using track.stop?

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -332,17 +354,31 @@ export class Publisher {
...
-        await lkRoom.localParticipant.setMicrophoneEnabled(desired);
+        this.logger.debug(
+          `handler: Setting LiveKit microphone enabled: ${enable}`,
+        );
+        await lkRoom.localParticipant.setMicrophoneEnabled(enable);
+        // Unmute will restart the track if it was paused upstream,
+        // but until explicitly requested, we want to keep it paused.
+        if (!this.shouldPublish && enable) {
+          await this.pauseUpstreams(lkRoom, [Track.Source.Microphone]);
+        }
```
toger5, line 374: This is still nasty. But now that we have dodged the footgun it also really looks like it should work.

src/state/CallViewModel/localMember/Publisher.ts
BillCarsonFr: Pausing will keep the local media stream track (so you see yourself), but it stops sending to the server (and signal a muted)

src/state/CallViewModel/localMember/LocalMember.ts
BillCarsonFr: Thx I had questions about that. Like do we really want to stop / restart publishing? or even start/stop tracks.

Or is it more that we need a `destroy` publisher that should clean itself?

src/state/CallViewModel/localMember/Publisher.test.ts
BillCarsonFr: good catch, did that while debugging the tests

src/state/CallViewModel/localMember/Publisher.ts
BillCarsonFr: right there used to have a `.then() ` block on the enable track but was removed here 610af439a8fbf371cc310af84fe1e96775b1a86e
I guess I can just voids

src/state/CallViewModel/localMember/Publisher.ts
BillCarsonFr: Muting will also stop the local track, i.e you will not see your camera feed. So won't work in a lobby for example. The video would stay black until you publish

BillCarsonFr:
@toger5 I have addressed your comments and they were not blocking so I am merging now as you are OOO and I'd like it to sit a bit on dev.

PR #3637, opened 2025-12-15
Allow user to switch layouts while phone is in landscape

robintown:
This fixes a regression on the development branch: the layout switcher would not respond to input while the window mode is 'flat' (i.e. while a mobile phone is in landscape orientation). See https://github.com/element-hq/element-call/pull/3605#discussion_r2586226422 for more context.
    
I was having a little trouble interpreting the emergent behavior of the layout switching code, so I refactored it in the process into a form that I think is a more direct description of the behavior we want (while not making it as terse as my [original implementation](https://github.com/element-hq/element-call/blob/768cf6f8ef3cab247b26e0379ad0b815a5782337/src/state/CallViewModel.ts#L1205)).

### DIFF
--- a/src/state/CallViewModel/CallViewModel.ts
+++ b/src/state/CallViewModel/CallViewModel.ts
@@ -946,11 +946,12 @@ export function createCallViewModel$(
     ),
   );
 
-  const hasRemoteScreenShares$: Observable<boolean> = spotlight$.pipe(
-    map((spotlight) =>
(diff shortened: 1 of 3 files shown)


toger5 (approved):
I find the refactored version much much easier to understand.

I am not sure that we want the behaviour like that now that I read through it.

If I am making a conscious choice to be in spotlight I am not sure I would be surprised if it switches back to non spotlight automatically once someone starts -> stops a screen-share.

And not really for discussing in this PR.

The code changes in the PR are very nice!

PR #3638, opened 2025-12-17
Pseudonomous identity and use the new jwt service endpoint (with delayed event delegation)

toger5:
This also does all the compatibility work. When to use which endpoint to authenticate agains a jwt service.


toger5:
Looking at the changes it might make more sense to go with a retry system for the jwt so we do not have to pass aroudn the useMatrix2 information with the transport.

src/livekit/openIDSFU.ts
```diff
@@ -52,26 +62,35 @@ export async function getSFUConfigWithOpenID(
...
   const args: [CallMembershipIdentityParts, string, string, IOpenIDToken] = [
     membership,
     serviceUrl,
     livekitRoomAlias,
     openIdToken,
   ];
-  if (matrix2jwt) {
+  try {
+    // we do not want to try the old endpoint, since we are not sending the new matrix2.0 sticky events (no hashed identity in the event)
+    if (forceOldJwtEndpoint) throw new Error("Force old jwt endpoint");
```
BillCarsonFr: I am not really a fan of using an exception as a conditional control flow. 
I'd prefer that this is formatted as a normal conditional

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -62,6 +61,11 @@ interface Props {
  * @prop useOldestMember Whether to use the same transport as the oldest member.
  * This will only update once the first oldest member appears. Will not recompute if the oldest member leaves.
  *
+ * @prop useOldJwtEndpoint$ Whether to set forceOldJwtEndpoint the use the old JWT endpoint.
+ * This is used when the connection manager needs to know if it has to use the legacy endpoint which implies a string concatenated rtcBackendIdentity.
+ * (which is expected for non sticky event based rtc member events)
+ * @returns Behavior<(LivekitTransport & { forceOldJwtEndpoint: boolean }) | null> The `forceOldJwtEndpoint` field is added to let the connection EncryptionManager
```
BillCarsonFr: The @returns doc is outdated

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -142,6 +144,8 @@ const FOCI_WK_KEY = "org.matrix.msc4143.rtc_foci";
  *
  * @param client
  * @param roomId
+ * @param useMatrix2 This implies using the matrix2 jwt endpoint (including delayed event delegation of the jwt token)
```
BillCarsonFr: outdated doc. 
I think we miss some js-doc CI? It owuld help catching those

toger5: there are added now right?

src/livekit/MatrixAudioRenderer.tsx
```diff
@@ -32,7 +31,7 @@ export interface MatrixAudioRendererProps {
    * This list needs to be composed based on the matrixRTC members so that we do not play audio from users
    * that are not expected to be in the rtc session (local user is excluded).
    */
-  validIdentities: ParticipantId[];
+  validIdentities: string[];
```
BillCarsonFr, line 34: This is the rtcBackendIdentity now I guess. 
Would it be worth creating an alias for that in the js-sdK?

src/state/CallViewModel/localMember/LocalTransport.ts
toger5: doc string does not mention this anymore

src/livekit/openIDSFU.ts
```diff
@@ -84,16 +101,44 @@ export async function getSFUConfigWithOpenID(
...
+      sfuConfig = await getLiveKitJWTWithDelayDelegation(
+        membership,
+        serviceUrl,
+        roomId,
+        openIdToken,
+        delayEndpointBaseUrl,
+        delayId,
+      );
+      logger?.info(`Got JWT from call's active focus URL.`);
+    } catch (e) {
```
BillCarsonFr, line 139: Shouldn't we be a bit more specific regarding the error here? What if it is just a transiant network error? Can we check for the exact incompatible error instead?

src/livekit/openIDSFU.ts
```diff
@@ -134,3 +182,53 @@ async function getLiveKitJWT(
...
+      Config.get().matrix_rtc_session?.delayed_leave_event_delay_ms ?? 1000;
+    bodyDalayParts = {
+      delay_id: delayId,
+      delay_timeout: delayTimeoutMs,
+      delay_cs_api_url: delayEndpointBaseUrl,
+    };
+  }
+
+  try {
+    const res = await fetch(livekitServiceURL + "/get_token", {
```
BillCarsonFr: We might want to use something like `doNetworkOperationWithRetry` to have some retry for transiant errors.
And in the catch we should look out for the specific error when the modern endpoint is not supported (what is the error? 404 or something else).
We should then throw a custom UnsupportedEndPointError that can be handled by the caller.

src/state/CallViewModel/remoteMembers/MatrixLivekitMembers.ts
```diff
@@ -93,63 +92,56 @@ export function createMatrixLivekitMembers$({
...
+              participants.find(
+                (p) => p.identity == membership.rtcBackendIdentity,
+              ) ?? null;
             const connection = transport
               ? managerData.getConnectionForTransport(transport)
               : null;
 
             yield {
-              keys: [participantId, membership.userId],
+              keys: [membership.userId, membership.deviceId],
```
BillCarsonFr, line 114: I wonder why the `identity` is not part of the key? or maybe the only key?

dev-backend-docker-compose.yml
```diff
@@ -88,7 +88,7 @@ services:
 
   synapse:
     hostname: homeserver
-    image: docker.io/matrixdotorg/synapse:latest
+    image: ghcr.io/element-hq/synapse:pr-18968-dcb7678281bc02d4551043a6338fe5b7e6aa47ce
```
BillCarsonFr, line 91: Why do we commit that?

BillCarsonFr (commented):
Thx, we are getting close 😅
 I have a few open questions

src/livekit/MatrixAudioRenderer.tsx
toger5: I dont think so. There are no string semantics to that identity... The names should already imply what it is?

But If we have a PR that introduces such a thing I would not object.

src/livekit/openIDSFU.ts
toger5: Added a custom error type. So we can only try with the new endpoint if we fail because the `/get_token` endpoint is not found.

dev-backend-docker-compose.yml
toger5: Good question. It is the newest version with sticky events. So if we want to create playwrights with sticky events we would need it. But we do not so this is just to allow testing it locally with `yarn backend`
Patching the synapse used for that can also be postponed.

toger5: I think it makes sense to keep it as the hash for now until we have a stable version (like we do it with the js-sdk branches)

But a playwright test should be added that uses sticky events and relies on that synapse version.

toger5:
ugh broke playwright. Will dive into that tomorrow.

src/state/CallViewModel/remoteMembers/MatrixLivekitMembers.ts
```diff
@@ -93,63 +92,56 @@ export function createMatrixLivekitMembers$({
...
         values.every((value) => value.epoch === values[0].epoch),
       ),
-      map(
-        ([
-          { value: membershipsWithTransports, epoch },
-          { value: managerData },
-        ]) =>
-          new Epoch([membershipsWithTransports, managerData] as const, epoch),
-      ),
+      map(([x, y]) => new Epoch([x.value, y.value] as const, x.epoch)),
```
robintown:
```suggestion
      map(([ms, data]) => new Epoch([ms.value, data.value] as const, ms.epoch)),
```

I like a little name hint…

src/settings/settings.ts
```diff
@@ -126,7 +126,13 @@ export const alwaysShowIphoneEarpiece = new Setting<boolean>(
...
 export enum MatrixRTCMode {
   Legacy = "legacy",
-  Compatibil = "compatibil",
+  Compatibility = "compatibility",
+  /** This implies using
+   *  - sticky events
+   *  - hashed RTC backend identity
+   *  - the new endpoint for the jwt token on the local membership (remote memberships will always try the new jwt endpoint first -> then the legacy one)
+   *  - use the hashed identity for the local membership
+   */
```
robintown, line 135: We need to be mindful in our dogfooding that if any of us have set our client to the compatibility mode previously, you may need to reset it manually (as EC will no longer recognize the old string value)

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -716,13 +736,13 @@ export function createCallViewModel$(
             participant,
             connection$,
             membership$,
-          } of matrixLivekitMembers) {
-            const participantId = `${userId}:${membership$.value.deviceId}`;
-            if (participantId === localParticipantId) continue;
+          } of matrixLivekitMembers.value) {
+            const userMediaId = `${userId}:${membership$.value.deviceId}`;
```
robintown, line 749: There's another instance of `participantId` just below this in the `factory` callback that should be renamed

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -685,22 +705,22 @@ export function createCallViewModel$(
...
           // add local member if available
           if (localMatrixLivekitMember) {
             const { userId, participant, connection$, membership$ } =
               localMatrixLivekitMember;
-            localParticipantId = `${userId}:${membership$.value.deviceId}`; // should be membership$.value.membershipID which is not optional
-            // const participantId = membership$.value.membershipID;
-            if (localParticipantId) {
+            localUserMediaId = `${userId}:${membership$.value.deviceId}`; // should be membership$.value.membershipID which is not optional
+
+            if (localUserMediaId) {
```
robintown: :thinking: This branch will always be taken, since `localUserMediaId` is necessarily truthy. Don't think we need it.

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -415,11 +420,38 @@ export function createCallViewModel$(
     memberships$,
   );
 
+  const ownMembershipIdentity: CallMembershipIdentityParts = {
+    userId,
+    deviceId,
+    // This will eventually become the salt for the hash endpoint.
+    // For now we keep it as the user+device string since it is expected by non matrix matrixRTCMode === Legacy.
+    memberId: `${userId}:${deviceId}`,
```
robintown: Hmm, I wonder if we should call this change something other than 'pseudonymous identity' then since there are effectively no pseudonymity guarantees yet. Maybe 'support arbitrary identities' or 'prepare for pseudonymous identities' has less risk of misleading?

src/livekit/openIDSFU.ts
```diff
@@ -113,15 +120,21 @@ export async function getSFUConfigWithOpenID(
 
   // If forceOldJwtEndpoint is set we indicate that we do not want to try the new endpoint,
   // since we are not sending the new matrix2.0 sticky events (no hashed identity in the event)
-  if (forceOldJwtEndpoint === false) {
+  if (
+    // we do not force anything. Try with new first (remote connections)
+    !opts?.forceJwtEndpoint ||
```
BillCarsonFr: I find this a bit hard to read `!opts?` can it be simplified

src/livekit/openIDSFU.ts
```diff
@@ -113,15 +120,21 @@ export async function getSFUConfigWithOpenID(
...
+    // we do force the matrix2.0 endpoint
+    (opts?.forceJwtEndpoint &&
```
BillCarsonFr: isn't `opts?.forceJwtEndpoint === JwtEndpointVersion.Matrix_2_0` enough?

src/livekit/openIDSFU.ts
```diff
@@ -235,7 +252,7 @@ export async function getLiveKitJWTWithDelayDelegation(
 
   let bodyDalayParts = {};
   // Also check for empty string
-  if (delayId && delayEndpointBaseUrl && false) {
```
BillCarsonFr: oh wow what was that

toger5: The jwt was returning `invalid json` when not using the most recent version.

toger5: So this was for testing

src/livekit/openIDSFU.ts
toger5: `opts.forceJwtEndpoint = undefined` would resolve to false then.
(with this it will be true) right?

toger5: I have an idea how this can be done in a way that si easier to read.

locales/en/app.json
```diff
@@ -116,6 +116,7 @@
     "matrix_rtc_transport_missing": "The server is not configured to work with {{brand}}. Please contact your server admin (Domain: {{domain}}, Error Code: {{ errorCode }}).",
     "membership_manager": "Membership Manager Error",
     "membership_manager_description": "The Membership Manager had to shut down. This is caused by many consequtive failed network requests.",
+    "no_matrix_2_authorization_service": "Your authorization service for you media server (SFU) is not on the newest version",
```
BillCarsonFr, line 119: I can't find it on localazy

BillCarsonFr (approved):
I'd be ok to merge and continue with follow up PR as long as CI happy and other review request are adressed

locales/en/app.json
toger5: probably will only appear once we merge?

src/settings/settings.ts
toger5: indeed... Lets just keep in in our heads and remember toggling the value in case sth fails.

locales/en/app.json
BillCarsonFr: > probably will only appear once we merge?

We have to do that manually no? First add on localazy then download the files from localazy? no?

src/state/CallViewModel/CallViewModel.ts
toger5: I think the comment is not even true. memberId is not used at all I think.

toger5: I checked and we really only consume this if we use the sticky event membership manager.
So we can just make it a uuid now already!

robintown: I'm checking this myself now as well

robintown: Yes, that appears to be the case. Let's use a UUID!

locales/en/app.json
robintown: Our English strings are automatically uploaded to Localazy by a GitHub Action on every push to `livekit`: https://github.com/element-hq/element-call/actions/workflows/translations-upload.yaml The file on `livekit` is the source of truth.

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -47,6 +47,7 @@ import {
 } from "matrix-js-sdk/lib/matrixrtc";
 import { type IWidgetApiRequest } from "matrix-widget-api";
 import { type CallMembershipIdentityParts } from "matrix-js-sdk/lib/matrixrtc/EncryptionManager";
+import { v4 as uuidv4 } from "uuid";
```
robintown: We should add uuid to our dependencies if we are actually using it outside matrix-js-sdk. (Though the built-in [`crypto.randomUUID`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) function would work just as well here)

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -685,29 +714,28 @@ export function createCallViewModel$(
...
+            localUserMediaId = `${userId}:${membership$.value.deviceId}`;
+            for (let dup = 0; dup < 1 + duplicateTiles; dup++) {
+              yield {
+                keys: [
+                  dup,
+                  localUserMediaId,
+                  userId,
+                  participant satisfies TaggedParticipant as TaggedParticipant, // Widen the type safely
+                  connection$,
+                  membership$.value,
```
robintown: This will be what's causing the issue where the tile reappears after unmuting the camera. `membership$.value` is not a stable key and more to the point, we should design UserMedia to not be tied to a static CallMembership object

toger5: should we fix this as part of that PR?

toger5: > to a static CallMembership object

A user media object is a tile and they are based on the callMemberships. Why do we want them to be non static?

I think we even can think of it as the membership being the only key we need for it?

toger5: I updated this PR to now only pass the `rtcBackendIdenitity`

src/state/CallViewModel/remoteMembers/MatrixLivekitMembers.ts
toger5: I think we can either only do it based on the rtcBackendIdentity or userId+deviceId otherwise its redundant?

src/state/CallViewModel/CallViewModel.ts
toger5: This caused the following error:
`"randomUUID" is not exported by "__vite-browser-external`

I opted for the dedicated `yarn add -D uuid`

src/state/CallViewModel/CallViewModel.ts
robintown: > A user media object is a tile and they are based on the callMemberships. Why do we want them to be non static?

Because a user might update their membership during a call to change the intent, for example. Your fix is good.

PR #3644, opened 2025-12-30
Add script to check that the tsdoc is correct and up-to-date

BillCarsonFr:
I noticed that some docs were out-dated, so adding some script to add some basic checks.
It will not force to add docs on all methods for now, but it will check for missing or dead params for example


package.json
```diff
@@ -18,11 +18,12 @@
     "serve": "vite preview",
     "prettier:check": "prettier -c .",
     "prettier:format": "prettier -w .",
-    "lint": "yarn lint:types && yarn lint:eslint && yarn lint:knip",
+    "lint": "yarn lint:types && yarn lint:eslint && yarn lint:knip && yarn lint:tsdoc",
     "lint:eslint": "eslint --max-warnings 0 src playwright",
     "lint:eslint-fix": "eslint --max-warnings 0 src playwright --fix",
     "lint:knip": "knip",
     "lint:types": "tsc",
+    "lint:tsdoc": "eslint --ext .ts,.tsx src",
```
Half-Shot: isn't that just running eslint against only TS files? Functionally overlapped by `lint:eslint`?

Half-Shot (dismissed):
Seems like a good change, modulo script command?

package.json
BillCarsonFr: oh right, I did some back and forth and now it is unneeded

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -138,7 +138,16 @@ interface Props {
...
+ * @param props The properties required to create the local membership.
+ * @param props.scope The observable scope to use.
+ * @param props.connectionManager The connection manager to get connections from.
+ * @param props.createPublisherFactory Factory to create a publisher once we have a connection.
+ * @param props.joinMatrixRTC Callback to join the matrix RTC session once we have a transport.
+ * @param props.homeserverConnected The homeserver connected state.
+ * @param props.localTransport$ The local transport to use for publishing.
+ * @param props.logger The logger to use.
+ * @param props.muteStates The mute states for video and audio.
+ * @param props.matrixRTCSession The matrix RTC session to join.
```
robintown, line 150: For what it's worth I think we should stick to putting these kinds of comments in the `interface` definition rather than on the function, generally. It's a little easier to pull up the docs for each property while writing a function call that way.

robintown (approved):

src/state/CallViewModel/localMember/LocalMember.ts
BillCarsonFr: Yes it feels a bit strange to me also, but I am not familiar with js doc.
It looks like lint requires them because we are doing `destructuring`/`unpacking` of the arguments in the signature.

PR #3649, opened 2026-01-05
Removed unused `perParticipantE2EE` param.

Half-Shot:
Looking high and low, I can't see any usages of this param. I also can't really tell *what* this would do in Element Call today, as we determine the EncryptionSystem based on whether A) The url params contains a `password` or B) Whether the room is encrypted. 

We seem to have got by for some time without this parameter being used, so I get the feeling it isn't useful to keep around.


toger5 (approved):
There is still the logic in EW for the guest link: https://github.com/vector-im/element-web/blob/fe73a0358c3c12fa764a2b30327ae2eedd4ab6a6/src/models/Call.ts#L757

You are probably right that this is still non breaking since in an encrypted room we will just end up using perParticipantE2EE anyways.

Still would be nice to combine this with an EW PR so that the EW source is not confusing ppl.

toger5:
Before merging I think it also would be very nice to test the guest experience just to make sure that right now everything is working fine without the url param being used.

Half-Shot:
> Before merging I think it also would be very nice to test the guest experience just to make sure that right now everything is working fine without the url param being used.

Totally agree. Let's do this properly and setup a playwright test for it.

BillCarsonFr:
@fkwp to test if the guest mode (ew with lab flag + federating SPA) is still working

PR #3652, opened 2026-01-05
Fix margins in `showControls=false` url param mode

toger5:
Test:
 - add `showControls=false&header=none` to the url params
 - start a call with this netlify build vs call.element.dev
 - Make sure the screen is very large (wide in particular 16:10 should work) so that the video tiles dont get "aspect capped" and end up having a larger margin top and bottom anyways.
 before:

Screenshot 2026-01-09 at 13 52 03
 
 after:
Screenshot 2026-01-09 at 13 51 57
 
 Pulled out of #3638


robintown (commented):
What do you consider broken about the current margins with controls hidden? This is what I see:

Screenshot From 2026-01-08 11-44-11

I'm not sure that we should make the footer take up as much empty space in this mode as it does normally, if the flag is for minimizing extra UI elements.

toger5:
This is what I see.
This is fairly old code I did not want to lose. I was not fully aware of the context when opening the PR.
Its about `showControls=false&header=none`
then it looks like this:
Screenshot 2026-01-09 at 13 40 20

toger5:
@robintown updated the impl. It does not work like this anymore due to the `footer.hidden`
(as i said really old code ;) )

toger5:
This can get squashed merged.

BillCarsonFr:
@toger5 what's up with this PR?

toger5:
I just tested, and this is still a problem.
I updated the description since the reproduction needs to be more specific.

robintown:
Is the issue that the tile runs up against the bottom of the viewport in this mode?

toger5:
I played around with the fix and this is still looking fine.

toger5:
I merged this with the current main branch.
So it can be reviewed again.

robintown (changes requested):
One concern with this approach, if I keep the window at a normal desktop size but call `controls.enablePip()` to force the picture-in-picture layout, the footer is still visible:

Screen Shot 2026-05-04 at 18 47 11

toger5:
Is this sth that can realistically happen?
> but call controls.enablePip() to force the picture-in-picture layout, the footer is still visible

robintown:
Not sure, but I could imagine a tablet which lets you resize a PiP view to be arbitrarily large. Anyway it seems like a similar pattern to the header could be applied, where we differentiate "no footer" (due to the URL param) versus "hidden footer"

PR #3653, opened 2026-01-06
Add context menu and volume adjustment to screensharing

### LINKED ISSUE #2337: We no longer have the ability to set volume of someone else's screen-sharing tile
### Steps to reproduce

1. A screen-shares with audio
2. B watches the stream
3. B cannot change A's volume

#### What did you expect?

A way to change the volume of a screen share

#### What happened instead?

No way to do so

### Operating system

Not relevant, I assume

### Browser information

Not relevant, I assume

### URL for webapp

call.element.io

### Will you send logs?

No

coolsrett3333:
Fixes #2337

For transparency i have used copilot for assistance here, but my local tests show that this works as intended.

image

If this is not acceptable please feel free to use the code in whatever way you think is better to fix this issue.

### DIFF
--- a/src/tile/SpotlightTile.tsx
+++ b/src/tile/SpotlightTile.tsx
@@ -144,7 +144,6 @@ const SpotlightUserMediaItem: FC<SpotlightUserMediaItemProps> = ({
     ? t("microphone_on")
     : t("microphone_off");
 
-  // If we have a localVolume$, read it for the slider
   const localVolume = hasLocalVolume ? useBehavior((vm as any).localVolume$) as number : undefined;
 
   const onSelectFitContain = useCallback(
@@ -174,7 +173,7 @@ const SpotlightUserMediaItem: FC<SpotlightUserMediaItemProps> = ({
     if ("commitLocalVolume" in vm) (vm as any).commitLocalVolume();
   }, [vm]);
 
-  // Menu contents (similar to GridTile, plus slider when no screenshare present)
+  // Menu contents (similar to GridTile when no screenshare present)
   const menuContent = (
     <>
       {"toggleLocallyMuted" in vm ? (
@@ -191,8 +190,7 @@ const SpotlightUserMediaItem: FC<SpotlightUserMediaItemProps> = ({
         checked={cropVideo}
         onSelect={onSelectFitContain}
       />
-      {/* If there is no screenshare in the spotlight and this view model exposes a volume control,
-          show the same slider used elsewhere in the app so styling is consistent. */}
+      {/* If there is no screenshare in the spotlight*/}
       {!hasScreenShare && hasLocalVolume ? (
         <MenuItem as="div" Icon={locallyMuted ? VolumeOffIcon : VolumeOnIcon} label={null} onSelect={null}>
           <Slider
@@ -210,7 +208,7 @@ const SpotlightUserMediaItem: FC<SpotlightUserMediaItemProps> = ({
     </>
   );
 
-  // Overflow button menu (button in corner) — ensure it overlays other buttons
+  // Overflow button menu
   const overflow = (
     <Menu
       title={props.displayName}
@@ -247,7 +245,7 @@ const SpotlightUserMediaItem: FC<SpotlightUserMediaItemProps> = ({
     primaryButton: overflow,
   };
 
-  // Create the tile element (local or remote) then wrap it in a ContextMenu
+
   const tileElement =
     vm instanceof LocalUserMediaViewModel ? (
       <SpotlightLocalUserMediaItem vm={vm} {...baseProps} />
@@ -255,7 +253,6 @@ const SpotlightUserMediaItem: FC<SpotlightUserMediaItemProps> = ({
       <SpotlightRemoteUserMediaItem vm={vm} {...baseProps} />
     );
 
-  // Wrap with ContextMenu to support right-click/contextmenu across the tile
   return (
     <ContextMenu title={props.displayName} trigger={tileElement} hasAccessibleAlternative>
       {menuContent}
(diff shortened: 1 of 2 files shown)


src/tile/SpotlightTile.tsx
```diff
@@ -100,24 +115,148 @@ const SpotlightRemoteUserMediaItem: FC<SpotlightRemoteUserMediaItemProps> = ({
...
+    },
+    [vm],
+  );
+
+  const onVolumeCommit = useCallback(() => {
+    if ("commitLocalVolume" in vm) (vm as any).commitLocalVolume();
+  }, [vm]);
+
+  // Menu contents (similar to GridTile when no screenshare present)
+  const menuContent = (
```
toger5, line 177: I am surprised about the two `const menuContent` sections.
I cannot immediatly tell how they are different.
Also I am wondering if we cannot use the same control as we have for the user media volume.

toger5 (commented):
On first sight this looks like it adds quiet a bit of redundant code (which is not necassarly a bad thing if it makes things much simpler)

But here I am not yet convinced that this is a simple solution where we can work off with little maintainance burden.

Why have you opted for the duplicated code approach? Why was it not possible to reuse the menu component from the user tile?

src/tile/SpotlightTile.tsx
coolsrett3333: as far as i understand copilot separated them due to it being faster to implement instead of rewriting the existing one.
I asked it why, and it responded with:

The duplication happened because screen-share menus and user-media menus had slightly different conditions (screen share sometimes only wants mute + volume, user media wants fit toggle + optional local volume). 
Rather than tying those two flows together inline, I extracted the shared menu markup into a single small component (SharedMenuContent) that accepts the relevant options and callbacks.
This keeps the per-case differences explicit and small (the props) while removing the bulk of the JSX duplication. The behaviour and icons are unchanged; the component simply centralises the menu layout and the Slider/MenuItem usage so we have one place to change later.

So you are likely able to use the same control for both but id assume it would require a larger rewrite.

This was probably the fastest way for me to implement this fix since i have a very basic understanding of the codebase and Type Script to begin with.
However i understand your situation and this code might not fit your requirements, please do use this code as a POC or reference to fix the issue (it is quite an annoying bug)

toger5: Thanks for helping copilot giving this a try.

It definitely looks like it needs a more thorough review than what I have done. The response from copilot is very basic engineering blabla which does not fully convince me that it has done the considerations necessary so we can trust its judgement that this is the way this should be done.
This
> screen share sometimes only wants mute + volume, user media wants fit toggle + optional local volume

seems very doable with optional view props. (but as I said before. haven't looked at that code for some time)

Its very cheap to propose changes compared to the work required to evaluate if they make sense and help keep the codebase maintainable.

coolsrett3333: No problem, thank you for taking the time to look at the pull request.

i agree that it will take some time to evaluate.
How would you want to handle this, should we keep the pull here until you maintainers have a better understanding of the code changes or just close it for now?

I dont want to be a burden so ill leave the decision to you maintainers,  i just want to see this bug fixed :)

toger5: I would like to add another comment here. I hope it is not too frustrating to hear.

I am quiet confident that with the current tools we slow down progress on open source projects by proposing AI generated PR's. It already took a bit of time I could have spend very efficiently on current projects reviewing this (I am convinced not checking what AI can and how it performs would be a mistake as well) PR.

It looks very helpful and powerful to already have the code that fixes the specific issue that makes the experience you have with an app not as enjoyable. But accidentality the opposite is going to happen:
 - The maintainers spend time trying to evaluate what was generated. (which is "inefficient thinking". What I mean by that, is that lot of energy gets put into: "why was it done like that", which could be used for "how should such a thing be done")
 - Its harder to ignore a PR since the gut feeling is "someone put in a lot of work" (so for historic reasons it feels like I am obligated to give it extra attention)
 - Either I go off focus (which is probably the most harmful for an open source project. It is annoying that maintainers often, do not just fix the one thing, that annoys a specific user group the most. But usually they prioritise other things with very good reason) or I close it. -> If I close it the PR gets lost untracked because the PR is closed. In this particular case we are lucky, that there is already a proper issue!

That being said @coolsrett3333 I can tell from your comments that you are very considerate and PPL like you really help projects like element call. And if you put more time into understanding the codebase I am convinced you can help explain how this should be done and actually accelerate features like that. (also AI can help here to let you guide how things work) But as of now we cannot just merge generated code. There are too many known pitfalls when not properly reviewing, especially since there are already some parts that immediately look suspicious to me.
Also it is important to know that the devs also have access to the AI tools. So if I realise that this is an issue where there is little risk to introduce issues or make the codebase less comprehensive I can generate the code.

But for now the best is to help providing as much info as necessary in the issue description.

If you can also spread the word why such PR's are not providing value (maybe even reduce the progress open source projects make). That probably is also helpful to many other projects.

Interesting read about AI + open source productivity that provides a bit of data on my very opinion phrased statements:
https://metr.org/blog/2025-07-10-early-2025-ai-experienced-os-dev-study/
(A study where open source devs are randomly assigned to do tasks with and without AI and how fast they are)

toger5: TLDR: I think we should close this and keep it as reference. (It is already liked to the original issue. Its really nice that you properly marked this PR as a fix PR for that issue!!!)

coolsrett3333: Thanks for taking the time to explain this in detail, i genuinely appreciate the advice.

The concern around review cost and maintainability is understandable, and i see why this PR isnt something you can merge without thinking about the longterm maintenance costs.
im fine with closing it and keeping it as reference linked to the issue.

I clearly leaned too heavily on Copilot to get a working fix rather than fully understanding why that bug exists.
Thats on me, and your explanation makes it clear why thats not helpful for maintainers like you.

PR #3659, opened 2026-01-09
Fix: Ensure voice and audio intent are respected

BillCarsonFr:
This is a regression (not released) following the refactoring
#Fixes https://github.com/element-hq/voip-internal/issues/434

The core logic for choosing the initial mute state is simply extracted to `calculateInitialMuteState`, there was a long history on that. I have done some cleaning and simplified the logic.
It now follows the recommandation here https://github.com/element-hq/voip-internal/issues/394

Also added end to end playwright test for widget call in DM (voice and video)


src/state/MuteStates.ts
```diff
@@ -43,12 +41,6 @@ const defaultHandler: Handler = async (desired) => Promise.resolve(desired);
  * Do not use directly outside of tests.
  */
 export class MuteState<Label, Selected> {
-  // TODO: rewrite this to explain behavior, it is not understandable, and cannot add logging
-  private readonly enabledByDefault$ =
-    this.enabledByConfig && !getUrlParams().skipLobby
-      ? this.joined$.pipe(map((isJoined) => !isJoined))
```
robintown, line 49: https://github.com/element-hq/element-call/issues/2958 was the reason that we were preventing the mute state from automatically becoming `true` when joined. I tested this branch locally, and it does appear to regress https://github.com/element-hq/element-call/issues/2958 in a way. When I plug the webcam in, the video toggle button does change to the 'unmuted' state, but it does not automatically start a media stream. That makes me think that there's some bug with how we sync mute states with LiveKit that is preventing the new web cam from being picked up properly - but not an especially safe situation in general.

[Screencast From 2026-01-09 19-49-18.webm](https://github.com/user-attachments/assets/f9364c6f-4b8c-44b6-b8aa-4cd2b4e0b6fd)

(This shows me plugging and unplugging my webcam repeatedly without interacting directly with the app)

robintown (changes requested):

src/state/MuteStates.ts
BillCarsonFr: > #2958 was the reason that we were preventing the mute state from automatically becoming `true` when joined. I tested this branch locally, and it does appear to regress #2958 in a way. When I plug the webcam in, the video toggle button does change to the 'unmuted' state, but it does not automatically start a media stream. That makes me think that there's some bug with how we sync mute states with LiveKit that is preventing the new web cam from being picked up properly - but not an especially safe situation in general.
> Screencast.From.2026-01-09.19-49-18.webm
> 
> (This shows me plugging and unplugging my webcam repeatedly without interacting directly with the app)

Can we get this PR merged then open a new ticket regarding plugin in a new camera mid call?
So we can get the correct product input and then code and test the behavior.

playwright.config.ts
```diff
@@ -67,6 +67,8 @@ export default defineConfig({
           firefoxUserPrefs: {
             "permissions.default.microphone": 1,
             "permissions.default.camera": 1,
+            // "media.navigator.streams.fake": true,
+            // "media.navigator.permission.disabled": true,
```
Half-Shot: Could you comment why you commented these out, if we still need them?

config/config.devenv.json
```diff
@@ -15,5 +15,6 @@
     "delayed_leave_event_delay_ms": 18000,
     "delayed_leave_event_restart_ms": 4000,
     "network_error_retry_ms": 100
-  }
+  },
+  "trust_localhost_for_mute_state": true
```
Half-Shot: The localhost name is throwing me off a bit. Could we have something a bit more descriptive like start_unmuted_on_skiplobby?
Right, saw from the code below. I'm wondering why this can't just be a setting users have the option of under the dev settings, rather than a hardcoded config option.

playwright/fixtures/widget-user.ts
Half-Shot, line 2: technically I think this needs an update (and so do the other changed files)

src/room/RoomPage.tsx
```diff
@@ -68,15 +68,23 @@ export const RoomPage: FC = () => {
...
+
   useEffect(() => {
     const scope = new ObservableScope();
-    setMuteStates(new MuteStates(scope, devices, joined$));
+    setMuteStates(
+      new MuteStates(
+        scope,
+        devices,
+        calculateInitialMuteState(
+          urlParams,
```
Half-Shot: i'm surprised we don't just pass through the relevant params?

src/state/MuteStates.ts
```diff
@@ -73,76 +65,73 @@ export class MuteState<Label, Selected> {
...
-            return (): void => s.unsubscribe();
+            }
+          };
+
+          const s = desired$.subscribe((desired) => {
+            latestDesired = desired;
+            if (syncing === false) {
+              syncing = true;
+              sync().catch((err) => {
+                // TODO: better error handling
```
Half-Shot, line 122: Are we going to merge without this? What would better look like?

src/state/initialMuteState.ts
```diff
@@ -0,0 +1,52 @@
...
+  const { skipLobby, callIntent } = urlParams;
+
+  logger.debug(
+    `calculateInitialMuteState: skipLobby=${skipLobby}, callIntent=${callIntent}, hostname=${hostname}, trustLocalhost=${trustLocalhost}`,
+  );
+
+  const isTrustedHost =
+    packageType == "embedded" ||
+    // Trust local hosts in dev mode to make local testing easier
+    (hostname == "localhost" && trustLocalhost);
```
Half-Shot: This seems like a poor test. We also test on 127.0.0.1, and a local IP would also be local. Why can't this just be a dev setting?

Half-Shot (changes requested):
Few thoughts from my side. Mainly, I don't love the config+localhost check and wondering if it can just be a developer toggle, where we can add some flavor text to tell them it's not advisable to enable to call untrusted parties with it on.

src/state/MuteStates.ts
BillCarsonFr: We need to nuke this MuteStates thing. It encapsulates lots of behavior in rxjs magic, and it is lacking test describing the expected behavior.
We agree to do that later (there is no change in this PR, just a {} indent change)

playwright.config.ts
BillCarsonFr: Will cleanup. I extracted that as a separate PR to test if it was fixing it https://github.com/element-hq/element-call/pull/3663 but it is not

src/room/RoomPage.tsx
BillCarsonFr: done

src/state/initialMuteState.ts
BillCarsonFr: After discussion, I changed the logic, and will just use isWidget to device on that

Half-Shot (approved):
I'm a happy reviewer 🌞

PR #3664, opened 2026-01-09
Small refactor: User media alternative

toger5:
Alternative implementation for the `userMedia$` observable.
Before this was duplicating the yield block. The hope is that the things this function does become more obvious.


BillCarsonFr (approved):
Ok, I have been able to understand the new code

PR #3669, opened 2026-01-13
Change the livekit alias to just be the room id (for backwards compatibility)

toger5:
compatibility.

The main context can be found in the code comment:
```
        // This looks unintuitive especially considering that `sfuConfig.livekitAlias` exists.
        // Why do we not use: `livekit_alias: sfuConfig.livekitAlias`
        //
        //  - This is going to be used for sending our state event transport (focus_preferred)
        //  - In sticky events it is expected to NOT send this field at all. The transport is only the `type`, `livekit_service_url`
        //  - If we set it to the hased alias we get from the jwt, we will end up using the hashed alias as the body.roomId field
        //    in v0.16.0. (It will use oldest member transport. It is using the transport.livekit_alias as the body.roomId)
        //
        // TLDR this is a temporal fild that allow for comaptibilty but the spec expects it to not exists. (but its existance also does not break anything)
        // It is just named poorly: It was intetended to be the actual alias. But now we do pseudonymys ids so we use a hashed alias.

```

### DIFF
--- a/src/state/CallViewModel/localMember/LocalTransport.ts
+++ b/src/state/CallViewModel/localMember/LocalTransport.ts
@@ -250,7 +250,18 @@ async function makeTransport(
       transport: {
         type: "livekit",
         livekit_service_url: url,
-        livekit_alias: sfuConfig.livekitAlias,
+        // WARNING PLS READ ME!!!
+        // This looks unintuitive especially considering that `sfuConfig.livekitAlias` exists.
+        // Why do we not use: `livekit_alias: sfuConfig.livekitAlias`
+        //
+        //  - This is going to be used for sending our state event transport (focus_preferred)
+        //  - In sticky events it is expected to NOT send this field at all. The transport is only the `type`, `livekit_service_url`
+        //  - If we set it to the hased alias we get from the jwt, we will end up using the hashed alias as the body.roomId field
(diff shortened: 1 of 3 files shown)


src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -250,7 +271,18 @@ async function makeTransport(
...
+        // WARNING PLS READ ME!!!
+        // This looks unintuitive especially considering that `sfuConfig.livekitAlias` exists.
+        // Why do we not use: `livekit_alias: sfuConfig.livekitAlias`
+        //
+        //  - This is going to be used for sending our state event transport (focus_preferred)
+        //  - In sticky events it is expected to NOT send this field at all. The transport is only the `type`, `livekit_service_url`
+        //  - If we set it to the hased alias we get from the jwt, we will end up using the hashed alias as the body.roomId field
+        //    in v0.16.0. (It will use oldest member transport. It is using the transport.livekit_alias as the body.roomId)
+        //
+        // TLDR this is a temporal fild that allow for comaptibilty but the spec expects it to not exists. (but its existance also does not break anything)
```
BillCarsonFr: fild -> field

BillCarsonFr (approved):

PR #3670, opened 2026-01-13
Add livekitAlias debugging info

toger5:
This adds the actual livekit alias (not the one sent via the state event that just is the roomId) to the devtools.

### DIFF
--- a/src/settings/DeveloperSettingsTab.tsx
+++ b/src/settings/DeveloperSettingsTab.tsx
@@ -48,7 +48,12 @@ import { useUrlParams } from "../UrlParams";
 
 interface Props {
   client: MatrixClient;
-  livekitRooms?: { room: LivekitRoom; url: string; isLocal?: boolean }[];
+  livekitRooms?: {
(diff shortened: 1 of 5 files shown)


src/settings/DeveloperSettingsTab.tsx
```diff
@@ -310,6 +315,7 @@ export const DeveloperSettingsTab: FC<Props> = ({
               url: livekitRoom.url || "unknown",
             })}
           </h4>
+          <p>LivekitAlias: {livekitRoom.livekitAlias}</p>
```
BillCarsonFr, line 318: We are inconsistent regarding i18n. Let's use `t()` for all labels

BillCarsonFr (approved):

src/settings/DeveloperSettingsTab.tsx
toger5: I decided to still not add it to LivekitAlias since its the same in all languages.
I added it to remote/local participant

PR #3671, opened 2026-01-13
playwright: End to end test for huddle calls in widget mode

BillCarsonFr:
Adds a new widget test for group calls in widget mode.

This is quite an involving test. Let's see how stable it is.
We can eventually reduce the number of participants if it is too unstable
image

It creates a private room, with 5 members. The call is started and all members joins.
It is checking that every one is seeing the 5 videoTiles, and the camera are on.
image

As a final step, one camera is turned off on one of the participant and we check that the video stream is not visible on the creator side. This to ensure that the asserts we do to ensure they are on are valid.

image

## Next step

This is a first PR of a serie.
The next one will test that Matrix 2_0 calls are also working then that mixed version are compatible

### DIFF
--- a/playwright/widget/huddle-call.test.ts
+++ b/playwright/widget/huddle-call.test.ts
@@ -34,7 +34,7 @@ widgetTest("Create and join a group call", async ({ addUser, browserName }) => {
 
   for (const user of [timo, robin, halfshot, florian]) {
     // Accept the invite
-    // This doesn't super stable to get this as this super generic locator,
+    // This isn't super stable to get this as this super generic locator,
     // but it works for now.
     await expect(
       user.page.getByRole("option", { name: roomName }),
@@ -140,9 +140,6 @@ widgetTest("Create and join a group call", async ({ addUser, browserName }) => {
 
     const videoElements = await frame.locator("video").all();
     expect(videoElements.length).toBe(5);
-    //
-    // // ONLY 4 !!!!!
-    // await expect(frame.locator('video[autoplay]')).toHaveCount(4);
 
     const blockDisplayCount = await frame
       .locator("video")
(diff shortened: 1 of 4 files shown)


playwright/widget/huddle-call.test.ts
```diff
@@ -0,0 +1,161 @@
...
+  await TestHelpers.createRoom(roomName, valere.page, [
+    timo.mxId,
+    robin.mxId,
+    halfshot.mxId,
+    florian.mxId,
+  ]);
+
+  for (const user of [timo, robin, halfshot, florian]) {
+    // Accept the invite
+    // This doesn't super stable to get this as this super generic locator,
```
toger5: isnt?

playwright/widget/huddle-call.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    // There is no other options than to wait for all media to be ready?
+    // Or it is too flaky :/
+    await user.page.waitForTimeout(5000);
+    // No one should be waiting for media
+    await expect(frame.getByText("Waiting for media...")).not.toBeVisible();
+
+    // There should be 5 video elements, visible and autoplaying
+    const videoElements = await frame.locator("video").all();
+    expect(videoElements.length).toBe(5);
+    await expect(frame.locator("video[autoplay]")).toHaveCount(5);
```
toger5, line 109: that looks like a good solution for the "is video running" question

toger5 (approved):
If we make a compatibility testing follow up PR I would like it to minimize the amount of calls we create. So maybe we can just make a couple of us join with different modes.

The only thing that is tricky to test is getting someone to join with v0.16.0. Which is almost the most important test. How would that work?

playwright/widget/huddle-call.test.ts
BillCarsonFr: I think we need better HTML markup and accessibility. Some good aria attribute to indicate that the video is muted, and even a better visual indicator.

BillCarsonFr:
> The only thing that is tricky to test is getting someone to join with v0.16.0. Which is almost the most important test. How would that work?

In the docker config we are downloading some web instances, they have their embedded EC bundled in. We could just get a specific version on a given port?

PR #3672, opened 2026-01-14
Set to fixed matrix-js-sdk 40.0.0

toger5:


robintown (changes requested):
The release doesn't include https://github.com/matrix-org/matrix-js-sdk/pull/5129.

PR #3674, opened 2026-01-14
Fix: Publisher re-enable tracks after room connected in case of timeout

BillCarsonFr:
Fixes https://github.com/element-hq/voip-internal/issues/442

We were calling `LocalParticipant`#`set(Microphone|Camera)Enabled` expecting it to be very resistent to early calls. 
It has support for waiting for the room to be connected in order to publish the track, but it is not unlimited, it will timeout
https://github.com/livekit/client-sdk-js/blob/1744ee4ec7aaed4e71c510c34fd21667364fcce6/src/room/participant/LocalParticipant.ts#L896

We want to support this case a little bit better by ensuring the expected tracks are published once the room is really connected (signal connected).

This is not yet a good "connecting" experience. Because until you are connected to livekit, your video loop back will be lacking, and unmuting would be unresponsive.

### DIFF
--- a/src/state/CallViewModel/localMember/Publisher.ts
+++ b/src/state/CallViewModel/localMember/Publisher.ts
@@ -11,6 +11,7 @@ import {
   LocalVideoTrack,
   ParticipantEvent,
   type Room as LivekitRoom,
+  RoomEvent as LivekitRoomEvent,
   Track,
 } from "livekit-client";
 import {
@@ -89,6 +90,23 @@ export class Publisher {
       ParticipantEvent.LocalTrackPublished,
       this.onLocalTrackPublished.bind(this),
     );
+
+    this.connection.livekitRoom.once(LivekitRoomEvent.Connected, () => {
+      // When `createAndSetupTracks` is called before the connection is established,
+      // there is an internal timeout in livekit that could be triggered if it takes
+      // too long to connect.
+      // It is no-op if tracks are already created, so it is safe to call it again.
+      const audio = this.muteStates.audio.enabled$.value;
+      const video = this.muteStates.video.enabled$.value;
+      this.enableTracks(audio, video, this.connection.livekitRoom).catch(
+        (e) => {
+          this.logger.error(
+            "Failed to enable tracks after connection established",
+            e,
+          );
+        },
+      );
+    });
   }
 
   // LiveKit will publish the tracks as soon as they are created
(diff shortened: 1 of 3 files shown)


toger5:
Do we think this is an actual issue? Or is it more likely that the transport setup sometimes was racing and we ended up with a transport switch -> local connection switch which then ends up in a locked or not cleaned up old transport -> unable to unmute

BillCarsonFr:
> Do we think this is an actual issue? Or is it more likely that the transport setup sometimes was racing and we ended up with a transport switch -> local connection switch which then ends up in a locked or not cleaned up old transport -> unable to unmute

Yes it is, as shown by the test. But it is not a final/big issue given that you can recover by muting/unmuting when it happens. Probably not occuring often but found 5 occurences in RS

toger5:
> you can recover by muting/unmuting when it happens

This definitely implies its another thing. (the publisher bug would make exactly that impossible!!)

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -157,16 +175,29 @@ export class Publisher {
     // We are using the `ParticipantEvent.LocalTrackPublished` to be notified
     // when tracks are actually published, and at that point
     // we can pause upstream if needed (depending on if startPublishing has been called).
+    this.enableTracks(audio, video, lkRoom).catch((e) => {
+      // If it is PublisherTrackError (408), i.e the publishing timed out,
+      // because it took too long to connet to the room, we are safe because
+      // we registered to the RoomEvent.Connected to create the tracks once connected.
+      this.logger.error("Failed to enable tracks", e);
```
toger5, line 182: Do we want to inform the console log about that as well:
`because it took too long to connet to the room, we are safe`

and (very minor) `connet -> `connect`

toger5 (approved):

BillCarsonFr:
> > you can recover by muting/unmuting when it happens
> 
> This definitely implies its another thing. (the publisher bug would make exactly that impossible!!)

In this case it is unresponsive for at least 15s. But it is not final like with the local transport hot swap

toger5:
Let s do this:
> and (very minor) connet -> connect`

and merge?

toger5:
Should we merge this now, that we have the rc branch created?

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -89,6 +90,23 @@ export class Publisher {
       ParticipantEvent.LocalTrackPublished,
       this.onLocalTrackPublished.bind(this),
     );
+
+    this.connection.livekitRoom.once(LivekitRoomEvent.Connected, () => {
```
robintown, line 94: I'm noticing some leaks: both the `LocalTrackPublished` callback and the `Connected` callback should not remain set after the scope ends

robintown (changes requested):

BillCarsonFr:
This is too cold, now. 
Expect to tackle that when refactoring to use low level track management?

PR #3675, opened 2026-01-14
Fix Publisher cleanup procedure

toger5:
Fixes: https://github.com/element-hq/voip-internal/issues/444#event-22090544135

This will get a separate playwright test reproducing the issue, that this fixes (in a seperate PR)
 - [x] test written https://github.com/element-hq/element-call/pull/3681

### DIFF
--- a/dev-backend-docker-compose.yml
+++ b/dev-backend-docker-compose.yml
@@ -3,7 +3,7 @@ networks:
 
 services:
   auth-service:
-    image: ghcr.io/element-hq/lk-jwt-service:0.4.1
+    image: ghcr.io/element-hq/lk-jwt-service:sha-f8ddd00
     pull_policy: always
     hostname: auth-server
     environment:
@@ -25,7 +25,7 @@ services:
       - ecbackend
 
   auth-service-1:
-    image: ghcr.io/element-hq/lk-jwt-service:0.4.1
+    image: ghcr.io/element-hq/lk-jwt-service:sha-f8ddd00
     pull_policy: always
     hostname: auth-server-1
     environment:
--- a/dev-backend-docker-compose.yml
+++ b/dev-backend-docker-compose.yml
@@ -3,7 +3,7 @@ networks:
 
 services:
   auth-service:
-    image: ghcr.io/element-hq/lk-jwt-service:sha-f8ddd00
+    image: ghcr.io/element-hq/lk-jwt-service:0.4.1
     pull_policy: always
     hostname: auth-server
     environment:
@@ -25,7 +25,7 @@ services:
       - ecbackend
 
   auth-service-1:
-    image: ghcr.io/element-hq/lk-jwt-service:sha-f8ddd00
+    image: ghcr.io/element-hq/lk-jwt-service:0.4.1
     pull_policy: always
     hostname: auth-server-1
     environment:
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -435,18 +435,18 @@ export const InCallView: FC<InCallViewProps> = ({
     [vm],
   );
 
-  useEffect(() => {
-    widget?.api.transport
-      .send(
-        gridMode === "grid"
-          ? ElementWidgetActions.TileLayout
-          : ElementWidgetActions.SpotlightLayout,
-        {},
-      )
-      .catch((e) => {
-        logger.error("Failed to send layout change to widget API", e);
-      });
-  }, [gridMode]);
+  // useEffect(() => {
+  //   widget?.api.transport
+  //     .send(
+  //       gridMode === "grid"
+  //         ? ElementWidgetActions.TileLayout
+  //         : ElementWidgetActions.SpotlightLayout,
+  //       {},
+  //     )
+  //     .catch((e) => {
+  //       logger.error("Failed to send layout change to widget API", e);
+  //     });
+  // }, [gridMode]);
 
   useEffect(() => {
     if (widget) {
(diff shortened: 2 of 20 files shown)


src/state/CallViewModel/localMember/LocalTransport.ts
BillCarsonFr: If there is no older member, i.e you are the first this wont work. It will stay stuck in fetching (initial value).
this filter/first should be removed.
Then in switchMap if oldest transport is null, switch the initial fetching to null

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -126,7 +126,9 @@ export const createLocalTransport$ = ({
    * The transport over which we should be actively publishing our media.
    * undefined when not joined.
    */
-  const oldestMemberTransport$ = scope.behavior(
+  const oldestMemberTransport$ = scope.behavior<
+    LocalTransportWithSFUConfig | null | "fetching"
+  >(
     combineLatest([memberships$]).pipe(
```
BillCarsonFr: Why a combineLatest with a one item array?

toger5: Good question!

toger5: This now is an actual combine latest also adding the useOldestMember flag so we can skip the unnecassary jwt request.

src/state/CallViewModel/localMember/LocalTransport.ts
toger5: This was fixed with a couple of more iterations on this PR.

The core issue was a deadlock which is fixed with this PR.

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -47,50 +47,58 @@ export class Publisher {
...
+    this.logger.info("Scope ended -> unset handler");
+    this.muteStates.audio.unsetHandler();
+    this.muteStates.video.unsetHandler();
+
+    this.logger.info(`Start to stop tracks`);
+    try {
+      await this.stopTracks();
+      this.logger.info(`Done to stop tracks`);
+    } catch (e) {
+      this.logger.error(`Failed to stop publishing: ${e}`);
```
BillCarsonFr: maybe keep the task name consistent with the Start xx / Done xx. So `failed to stop tracks`

src/room/InCallView.tsx
```diff
@@ -435,19 +435,6 @@ export const InCallView: FC<InCallViewProps> = ({
     [vm],
   );
 
-  useEffect(() => {
```
BillCarsonFr, line 438: why is this related to current issue?

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
```diff
@@ -189,10 +189,6 @@ export function createConnectionManager$({
           }
         },
         (scope, _data$, serviceUrl, alias, sfuConfig) => {
-          logger.debug(
```
BillCarsonFr, line 192: it was too noisy?

src/widget.ts
```diff
@@ -110,6 +110,7 @@ export const widget = ((): WidgetHelpers | null => {
         EventType.RoomRedaction,
         ElementCallReactionEventType,
         EventType.RTCDecline,
+        EventType.RTCMembership, // Send/Read the membership sticky events
```
BillCarsonFr: related to this PR?

dev-backend-docker-compose.yml
```diff
@@ -3,7 +3,7 @@ networks:
 
 services:
   auth-service:
-    image: ghcr.io/element-hq/lk-jwt-service:sha-f8ddd00
+    image: ghcr.io/element-hq/lk-jwt-service:0.4.1
```
BillCarsonFr: why this change?

BillCarsonFr (changes requested):

dev-backend-docker-compose.yml
toger5: I thought its better to have the stable version. And 0.4.1 is very new and also supports both endpoints. It just does not support delegation. Which we do not really test yet (and also do not have to support. Its optional spec basically.)

In other words: you are fully supporting matrix2.0 even if you do not use the delegation feature.

src/widget.ts
toger5: No. And robin will look into this properly. Its also just the event type and not the new sticky capability. So lets not touch this in this PR.

src/room/InCallView.tsx
toger5: Its not. And also done by this PR: https://github.com/element-hq/element-call/pull/3679

I was hoping it will resolve itself on merge. But this ^ too longer to get approved.

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
toger5: No it was duplicated with the log inside the constructor of the connection.

The one inside the constructor also had better scoping `[Connection <url>]`

src/widget.ts
toger5: Its reverted in this PR now.

BillCarsonFr (approved):
great 👍

dev-backend-docker-compose.yml
BillCarsonFr: I wonder how we could test different versions :/ 
Or maybe we could have renovate to create a PR for new releases?

PR #3676, opened 2026-01-14
playwright: Federated call between legacy and compat

BillCarsonFr:
Create a test for a call between to federated users. One using rtc legacy and one using rtc compat.

**EDIT** working with the latest jwt service

~~CURRENTLY FAILING due to a bug.~~
~~Users cannot see other video.~~

image

### DIFF
--- a/playwright/widget/federated-call.test.ts
+++ b/playwright/widget/federated-call.test.ts
@@ -1,0 +1,71 @@
+/*
+Copyright 2026 Element Creations Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+Please see LICENSE in the repository root for full details.
+*/
+
+import { expect, test } from "@playwright/test";
+
+import { widgetTest } from "../fixtures/widget-user";
+import { HOST1, HOST2, TestHelpers } from "./test-helpers";
+
+widgetTest(
+  "Test federated call default rtc mode",
+  async ({ addUser, browserName }) => {
+    test.skip(
+      browserName === "firefox",
+      "The is test is not working on firefox CI environment. No mic/audio device inputs so cam/mic are disabled",
+    );
+
+    const florian = await addUser("floriant", HOST1);
+    const timo = await addUser("timo", HOST2);
+
+    const roomName = "Call Room";
+
+    await TestHelpers.createRoom(roomName, florian.page, [timo.mxId]);
+
+    await TestHelpers.acceptRoomInvite(roomName, timo.page);
+
+    await florian.page.pause();
+
+    await TestHelpers.setEmbeddedElementCallRtcMode(florian.page, "legacy");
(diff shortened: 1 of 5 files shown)


playwright/widget/federated-call.test.ts
```diff
@@ -0,0 +1,80 @@
...
+      const florian = await addUser("floriant", HOST1);
+      const timo = await addUser("timo", HOST2);
+
+      const roomName = "Call Room";
+
+      await TestHelpers.createRoom(roomName, florian.page, [timo.mxId]);
+
+      await TestHelpers.acceptRoomInvite(roomName, timo.page);
+
+      await florian.page.pause();
```
toger5, line 38: Will it ever succeed with pauses?
I thought they let you interact und one has to manually continue (or is this only in debug mode, not ci)

toger5, line 38: Will it ever succeed with pauses?
I thought they let you interact und one has to manually continue (or is this only in debug mode, not ci)

Either way we might want to remove it?

playwright/widget/federated-call.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        const frame = user.page
+          .locator('iframe[title="Element Call"]')
+          .contentFrame();
+        await expect(frame.getByTestId("videoTile")).toHaveCount(2);
+
+        // There are no other options than to wait for all media to be ready?
+        // Or it is too flaky :/
+        await user.page.waitForTimeout(3000);
+        // No one should be waiting for media
+        await expect(frame.getByText("Waiting for media...")).not.toBeVisible();
```
toger5, line 60: Isnt the second enough?
`await expect(frame.getByText` should wait as long as it takes?

I think sth like a `Promise.race([waitForAllMeida, timeout(5000)])` might be a good idea?
So we do not have this test be blocked on ci forever?

playwright/widget/federated-call.test.ts
```diff
@@ -0,0 +1,80 @@
...
+          .evaluateAll(
+            (videos: Element[]) =>
+              videos.filter(
+                (v: Element) => window.getComputedStyle(v).display === "block",
+              ).length,
+          );
+        expect(blockDisplayCount).toBe(2);
+      }
+
+      // await florian.page.pause();
```
toger5, line 77: can also get removed

toger5 (approved):

playwright/widget/federated-call.test.ts
BillCarsonFr: it is pausing when you run in `--debug` mode. Which let you time to look/record locators

playwright/widget/federated-call.test.ts
BillCarsonFr: No I don't think it is helping. Instead we could use a custom timeout in the `toBeVisible`

toger5: Yes that sounds like a cleaner approach. And still make the waiting race with the clock.

PR #3679, opened 2026-01-16
Fix some error log lines that lead to confusion

toger5:
In the past those log lines often were referenced for issues but they are no real issues.
Either they are just deprecated code running or expected.

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -436,19 +436,6 @@ export const InCallView: FC<InCallViewProps> = ({
   );
 
   useEffect(() => {
-    widget?.api.transport
-      .send(
-        gridMode === "grid"
-          ? ElementWidgetActions.TileLayout
-          : ElementWidgetActions.SpotlightLayout,
-        {},
-      )
-      .catch((e) => {
-        logger.error("Failed to send layout change to widget API", e);
-      });
-  }, [gridMode]);
-
-  useEffect(() => {
     if (widget) {
       const onTileLayout = (ev: CustomEvent<IWidgetApiRequest>): void => {
         setGridMode("grid");
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -436,6 +436,19 @@ export const InCallView: FC<InCallViewProps> = ({
   );
 
   useEffect(() => {
+    widget?.api.transport
+      .send(
+        gridMode === "grid"
+          ? ElementWidgetActions.TileLayout
+          : ElementWidgetActions.SpotlightLayout,
+        {},
+      )
+      .catch((e) => {
+        logger.error("Failed to send layout change to widget API", e);
+      });
+  }, [gridMode]);
+
+  useEffect(() => {
     if (widget) {
       const onTileLayout = (ev: CustomEvent<IWidgetApiRequest>): void => {
         setGridMode("grid");
--- a/src/state/CallViewModel/localMember/LocalMember.test.ts
+++ b/src/state/CallViewModel/localMember/LocalMember.test.ts
@@ -194,8 +194,8 @@ describe("LocalMembership", () => {
       matrixRTCMode: MatrixRTCMode.Matrix_2_0,
     }),
     matrixRTCSession: {
-      updateCallIntent: () => {},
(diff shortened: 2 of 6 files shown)


toger5:
This PR increases the overall coverage and already introduces a not super helpful test (for a log line on error)

So I will merge it with failing CI once approved!

src/room/InCallView.tsx
```diff
@@ -435,46 +434,6 @@ export const InCallView: FC<InCallViewProps> = ({
     [vm],
   );
 
-  useEffect(() => {
```
BillCarsonFr, line 438: This doesn't match the PR description. This will lead to confusion in the future when looking at history

src/state/CallViewModel/localMember/LocalMember.test.ts
```diff
@@ -244,6 +244,37 @@ describe("LocalMembership", () => {
     });
   });
 
+  it("logs if callIntent cannot be updated", async () => {
```
BillCarsonFr, line 247: It is indeed testing the code, but looks like the code is talking to us? why are we trying to update call intent if not connected? I also find it strange that the intent is mutable :/ causes traffic everytime you mute/unmute camera

BillCarsonFr (changes requested):

src/state/CallViewModel/localMember/LocalMember.test.ts
toger5: I also played with checking if we are connected here. But it was somewhat redundant. if we already have a check in the js-sdk if we are connected.

src/room/InCallView.tsx
toger5: Do you want to split this into two PR's? Or update the title.
I think the PR can stay as one cleanup PR. That removes outdated code that causes unwanted logs and also improve logs.

Sth like:
Remove outdated layout widget api messages (reduce and improve error logs)

BillCarsonFr: why is this outdated? This is now directly inside EC?
That means that `ElementWidgetActions.TileLayout` should be deprecated?

BillCarsonFr (changes requested):
It looks to me that this PR is doing 2 different things?
- One is removing deprecated(?) widget actions
- The other about error logs confusion(?)

I'd prefer we ~~merge it~~ split in 2 PRs

toger5:
> I'd prefer we merge it

split it into two PRs?

toger5:
> why is this outdated

It was designed in a way where the layout switch buttons were part of the element web header bar.
Now (as you said) its inside the iframe. So the action is not needed anymore.

BillCarsonFr:
> > I'd prefer we merge it
> 
> split it into two PRs?

YES

toger5:
@BillCarsonFr I have split this:
 - This PR
 - And this one: https://github.com/element-hq/element-call/pull/3689

src/room/InCallView.tsx
```diff
@@ -435,19 +435,6 @@ export const InCallView: FC<InCallViewProps> = ({
...
```
BillCarsonFr, line 438: this is now in https://github.com/element-hq/element-call/pull/3689/changes? no

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -349,17 +335,23 @@ async function makeTransport(
...
+          "Using backend-configured (client.getRTCTransports) SFU",
+          selectedTransport,
+        );
         return selectedTransport;
       }
     } catch (ex) {
       if (ex instanceof MatrixError && ex.httpStatus === 404) {
         // Expected, this is an unstable endpoint and it's not required.
-        logger.debug("Backend does not provide any RTC transports", ex);
+        logger.debug(
```
BillCarsonFr, line 351: This a bit strange for a log no? looks lie it is a user facing string. 
Maybe keep the old log and add that as a comment?

BillCarsonFr (approved):
OK, to merge.
It looks like the line markes as not covered by codecov are actually covered by this test `Creating preferred transport based`?
That is strange, ok to force merge

PR #3681, opened 2026-01-16
Add playwright test for hotswatpping the transport (swap from `oldest_member` (singeSFU) to `multi_sfu`)

toger5:
This playwright test is confirmed to fail locally.
It will be skipped and the PR fixing it will remote the test skip.


BillCarsonFr (approved):
✅

PR #3683, opened 2026-01-16
playwright: Test call in SPA mode with matrix 2.0

BillCarsonFr:


playwright/spa-helpers.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  if (mode) {
+    await setRtcModeFromSettings(page, mode);
+  }
+
+  await page.getByTestId("lobby_joinCall").click();
+  await page.getByRole("radio", { name: "Spotlight" }).check();
+}
+
+async function setRtcModeFromSettings(
```
toger5, line 94: This looks very similar to the one for the widget? is there any difference?

```
public static async setEmbeddedElementCallRtcMode(
    page: Page,
    mode: RtcMode,
  ): Promise<void> {
    const iframe = page.locator('iframe[title="Element Call"]').contentFrame();

    await iframe.getByRole("button", { name: "Settings" }).click();
    await iframe.getByRole("tab", { name: "Preferences" }).click();

    // await iframe.getByText("Developer mode", { exact: true }).click();
    await iframe.getByText("Developer mode", { exact: true }).check(); // Idempotent:  won't uncheck if already checked

    // Move to Developer tab now
    await iframe.getByRole("tab", { name: "Developer" }).click();
    if (mode == "legacy") {
      await iframe.getByText("Legacy: state events").click();
    } else if (mode == "2_0") {
      await iframe.getByText("Matrix 2.0").click();
    } else {
      // compat
      await iframe.getByText("Compatibility: state events").click();
    }
    await iframe.getByTestId("modal_close").click();
  }
```
might be just frame vs page. Which we could add as a porp to `setEmbeddedElementCallRtcMode` (which would nee dot be renamed: `setElementCallRtcMode`

playwright/spa-call-sticky.spec.ts
toger5, line 1: super nice that the actual test is so simple

toger5 (approved):

playwright/spa-helpers.ts
BillCarsonFr: Yes, there is room to factorize all that. I think as a follow-up PR

PR #3686, opened 2026-01-16
Make Element Call widgets compatible with Matrix 2.0 mode

### LINKED ISSUE #3677: Enable Element Call widgets to send sticky RTC membership events

Make it so that Element Call widgets don't necessarily crash when they try to send a sticky event in Matrix 2.0 mode (given that the hosting client supports them)

_No response_

### Additional context

_No response_

robintown:
This gives Element Call widgets the ability to send and receive sticky RTC membership events.

Depends on https://github.com/matrix-org/matrix-js-sdk/pull/5142
Closes https://github.com/element-hq/element-call/issues/3677

To-do:

- [x] Create PRs to Element Web and matrix-rust-sdk auto-approving the new capabilities
  - Element Web: https://github.com/element-hq/element-web/pull/31843
  - matrix-rust-sdk: TODO
- [x] Think about adding an end-to-end test
  - Task passed on to Valere


toger5 (commented):
This looks great.
We cannot yet approve since we need
 - [ ] the correct js-sdk version
 - [ ] Decide what we do with the test coverage (I think a simple test to confirm that we call createRoomWidgetClient with the correct params would be good. Everything on top would just test already tested (in the js-sdk) code)

robintown:
Ah, I was going to override the coverage requirement. It's a 3 line change, and I generally don't see value in tests that are a close copy of what the code already says.

BillCarsonFr:
> Ah, I was going to override the coverage requirement. It's a 3 line change, and I generally don't see value in tests that are a close copy of what the code already says.

What do you mean but close to the code?

Wouldn't it be good to have a test that check what the EC widget is calling the expected list of permissions? when creating the room widget client?

BillCarsonFr (approved):
Thx for all the tests. Maybe in the future would protect us from unoticed removal of some capability ;)

src/widget.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      "org.matrix.rageshake_request",
+      EventType.CallEncryptionKeysPrefix,
+      EventType.Reaction,
+      EventType.RoomRedaction,
+      ElementCallReactionEventType,
+      EventType.RTCDecline,
+      EventType.RTCMembership,
+    ];
+
+    const sendState = [
```
robintown, line 68: @BillCarsonFr For the record this is what I was referring to with the test being a close copy of the code: The obvious way to write it has all the same lists from `widget.ts` duplicated here, with an `expect` at the end rather than a function call.

My take is that writing tests like this is busywork, since they don't tell the reviewer anything that they wouldn't have already known from reading the source file. In general my attitude toward tests - maybe an unpopular opinion :) - is that properties of code that are trivial or self-evident do not need testing.

toger5: I also can get behind that. But there is some hardcoded values in there that we also check.
I was considering just writing down the full list as raw strings. But for the sake of time made it like that.

But what makes this a bit of a maintenance. burden also has the advantage, that if I mess around with the widget file, the test will force me to rethink if that is what I actually want to do.

PR #3687, opened 2026-01-19
Fix delayId delegation. Use correct event reemitting types.

toger5:
When using reemitters, the callback args will change to an array including the originals emitter.

Blocked by: https://github.com/matrix-org/matrix-js-sdk/pull/5144

### DIFF
--- a/src/state/CallViewModel/CallViewModel.ts
+++ b/src/state/CallViewModel/CallViewModel.ts
@@ -142,6 +142,7 @@ import {
 import { Publisher } from "./localMember/Publisher.ts";
 import { type Connection } from "./remoteMembers/Connection.ts";
 import { createLayoutModeSwitch } from "./LayoutSwitch.ts";
+import { IMembershipManager } from "matrix-js-sdk/lib/matrixrtc/IMembershipManager";
 
 const logger = rootLogger.getChild("[CallViewModel]");
 //TODO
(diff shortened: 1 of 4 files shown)


src/state/CallViewModel/CallViewModel.ts
```diff
@@ -445,8 +446,9 @@ export function createCallViewModel$(
         fromEvent(
           matrixRTCSession,
           MembershipManagerEvent.DelayIdChanged,
-        ) as Observable<string | undefined>
-      ).pipe(map((v) => v ?? null)),
+          // The type of reemitted event includes the original emitted as the second arg.
+        ) as Observable<[string | undefined, IMembershipManager]>
+      ).pipe(map(([delayId]) => delayId ?? null)),
```
BillCarsonFr, line 451: So this never worked?

BillCarsonFr (approved):
thx for the description to explain. 
I am surprised though, is this always like that when re-emitting?

toger5:
>  I am surprised though, is this always like that when re-emitting?

I was surprised as well. But i think yes, its always like that.
https://github.com/matrix-org/matrix-js-sdk/blob/b6d297270a163762ea0ba685006eee6c6f2cd51b/src/ReEmitter.ts#L30-L60
```
 public reEmit(source: EventEmitter, eventNames: string[]): void {
        let reEmittersByEvent = this.reEmitters.get(source);
        if (!reEmittersByEvent) {
            reEmittersByEvent = new Map();
            this.reEmitters.set(source, reEmittersByEvent);
        }

        for (const eventName of eventNames) {
            if (reEmittersByEvent.has(eventName)) continue;

            // We include the source as the last argument for event handlers which may need it,
            // such as read receipt listeners on the client class which won't have the context
            // of the room.
            const forSource = (...args: any[]): void => {
                // EventEmitter special cases 'error' to make the emit function throw if no
                // handler is attached, which sort of makes sense for making sure that something
                // handles an error, but for re-emitting, there could be a listener on the original
                // source object so the test doesn't really work. We *could* try to replicate the
                // same logic and throw if there is no listener on either the source or the target,
                // but this behaviour is fairly undesireable for us anyway: the main place we throw
                // 'error' events is for calls, where error events are usually emitted some time
                // later by a different part of the code where 'emit' throwing because the app hasn't
                // added an error handler isn't terribly helpful. (A better fix in retrospect may
                // have been to just avoid using the event name 'error', but backwards compat...)
                if (eventName === "error" && this.target.listenerCount("error") === 0) return;
                this.target.emit(eventName, ...args, source);
            };
            source.on(eventName, forSource);
            reEmittersByEvent.set(eventName, forSource);
        }
    }
```

PR #3688, opened 2026-01-19
Do not end call with error in case of MediaDeviceError

BillCarsonFr:
Fix regressions after CallViewModel refactoring
#Fixes https://github.com/element-hq/voip-internal/issues/440 - Call ends with error if try to enable camera when permission was denied
#Fixes https://github.com/element-hq/voip-internal/issues/439 - Call ends with error if user cancel sceensharing at the window selection screen.

I tried to do some playwright test for that, but could not find a way to **Deny** a permission.
It is possible to not **grant** a permission, but when doing that the playwright browser is in an intermediate state where the permission is nor granted nor denied. Thus it is not possible to reproduce :/
Maybe this could have helped? https://github.com/microsoft/playwright/issues/6488 but it was closed

### DIFF
--- a/src/state/CallViewModel/localMember/LocalMember.ts
+++ b/src/state/CallViewModel/localMember/LocalMember.ts
@@ -266,11 +266,11 @@ export const createLocalMembership$ = ({
 
   mediaErrors$.pipe(scope.bind()).subscribe((error) => {
     if (error) {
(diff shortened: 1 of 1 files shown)


toger5 (approved):
Thanks valere. I am sad and sorry on the not doable tests...

PR #3690, opened 2026-01-19
playwright: intercept send request to ensure sticky events are sent

BillCarsonFr:
Adds route interception to ensure sticky membership are sent

### DIFF
--- a/playwright/spa-call-sticky.spec.ts
+++ b/playwright/spa-call-sticky.spec.ts
@@ -5,7 +5,7 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 Please see LICENSE in the repository root for full details.
 */
 
-import { test } from "@playwright/test";
+import { expect, test } from "@playwright/test";
 
 import { SpaHelpers } from "./spa-helpers";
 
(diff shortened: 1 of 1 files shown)


playwright/spa-call-sticky.spec.ts
```diff
@@ -21,7 +21,19 @@ test("One to One call using matrix rtc 2.0 aka sticky events", async ({
...
   await page.goto("/");
 
-  await SpaHelpers.createCall(page, "John Doe", "HelloCall", true, "2_0");
+  let androlHasSentStickyEvent = false;
+
+  await page.route(
+    "**/_matrix/client/v3/rooms/**/send/org.matrix.msc4143.rtc.member/**",
+    async (route, req) => {
+      androlHasSentStickyEvent = !!new URL(req.url()).searchParams.get(
+        "org.matrix.msc4354.sticky_duration_ms",
```
toger5: This will updated to false again if an event with missing url param is sent?
`org.matrix.msc4354.sticky_duration_ms`.

I think `androlHasSentStickyEvent` is not reflecting that.
Either rename `androlHasSentStickyEvent` to `androlHasLastSentStickyEvent` or change it to:
```
androlHasSentStickyEvent = !!(new URL(req.url()).searchParams.get(
        "org.matrix.msc4354.sticky_duration_ms") || androlHasSentStickyEvent)
```
(or similar)
wdyt?

Half-Shot: Swap the order of your boolean logic, as it's cheaper to compute :)

Half-Shot: Also, noting if the call is encrypted then this would be a m.room.encrypted event.

Half-Shot: we decided that rooms are unencrypted in SPA mode.

playwright/spa-call-sticky.spec.ts
```diff
@@ -21,7 +21,14 @@ test("One to One call using matrix rtc 2.0 aka sticky events", async ({
 
   await page.goto("/");
 
-  await SpaHelpers.createCall(page, "John Doe", "HelloCall", true, "2_0");
+  let androlHasSentStickyEvent = false;
+
+  await interceptEventSend(page, "org.matrix.msc4143.rtc.member", (req) => {
```
toger5:
```suggestion
  // We can assume unencrypted events (`org.matrix.msc4143.rtc.member` vs `m.room.encrypted`) because the SPA will always use unencrypted rooms.
  await interceptEventSend(page, "org.matrix.msc4143.rtc.member", (req) => {
```

toger5 (approved):

PR #3694, opened 2026-01-20
Support for Matrix RTC Slots

Half-Shot:
Placeholder PR to track matrix-js-sdk changes and check things do not break. In the future this will do the required to support slots.


Johennes:
I think this is obsolete by now.

PR #3696, opened 2026-01-21
playwright: Test call running without some permissions

BillCarsonFr:
Follow up on https://github.com/element-hq/element-call/pull/3688

Finally found a way to simulate permissions beeing denied, by simply [mocking some browser API](https://playwright.dev/docs/mock-browser-apis)

Before the 3688 these test were failing the call with an unknown call error. We test that it is not doing that and that the call can continue

### DIFF
--- a/src/state/CallViewModel/localMember/LocalMember.ts
+++ b/src/state/CallViewModel/localMember/LocalMember.ts
@@ -267,6 +267,7 @@ export const createLocalMembership$ = ({
 
   mediaErrors$.pipe(scope.bind()).subscribe((error) => {
     if (error) {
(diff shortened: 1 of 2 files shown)


src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -267,6 +267,7 @@ export const createLocalMembership$ = ({
 
   mediaErrors$.pipe(scope.bind()).subscribe((error) => {
     if (error) {
+      // setMatrixError(new UnknownCallError(error));
```
toger5, line 270: this looks like an artifact?

BillCarsonFr:
I'll look at it

PR #3698, opened 2026-01-22
Forward logs to parent window in widget configuration

Half-Shot:
Uses https://github.com/matrix-org/matrix-widget-api/pull/160.

Still needs the ability to forward prior logs.

### DIFF
--- a/src/settings/rageshake.ts
+++ b/src/settings/rageshake.ts
@@ -27,10 +27,10 @@ Please see LICENSE in the repository root for full details.
 //    actually timestamps. We then purge the remaining logs. We also do this
 //    purge on startup to prevent logs from accumulating.
 
-import EventEmitter from "events";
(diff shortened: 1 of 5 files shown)


src/settings/rageshake.ts
```diff
@@ -94,7 +100,7 @@ class ConsoleLogger extends EventEmitter {
     // http://jsperf.com/concat-vs-plus-vs-join
     this.logs += line;
 
-    this.emit(ConsoleLoggerEvent.Log);
+    this.emit(ConsoleLoggerEvent.Log, level.toString(), args);
```
Half-Shot, line 103: Note to self the log is in numeric form here, we need to translate back into a sensible level.

BillCarsonFr (commented):
Could it be possible to have an example of how it will look like in the web logs?
Will it have a custom prefix or something?
🙏

PR #3701, opened 2026-01-25
[WIP] SDK target improvements

toger5:
This PR does a couple of things:
 - add ci to build and publish the sdk version of EC
 - Fix issues with custom rtc applications (not m.call) related to widget permissions
 - update the widget.tests to be more specific and test for custom rtc applications
 - update sdk/main.ts to include localMember information


fkwp:
GitHub actions look OK to me

sdk/main.ts
```diff
@@ -79,32 +77,54 @@ interface MatrixRTCSdk {
       participant: LocalParticipant | RemoteParticipant | null;
     }[]
   >;
+  /**
+   * flattened local members
+   */
+  localMember$: Behavior<{
```
BillCarsonFr, line 83: would be nice to have current userId, and DeviceId. localMember will be set only when the session is joined. For another PR

src/livekit/openIDSFU.ts
```diff
@@ -146,7 +146,7 @@ export async function getSFUConfigWithOpenID(
       } else {
         logger?.warn(
           `Failed fetching jwt with matrix 2.0 endpoint other issues ->`,
-          `(not going to try with legacy endpoint: forceOldJwtEndpoint is set to false, we did not get a not supported error from the sfu)`,
+          `(not going to try with legacy endpoint if forceMatrix2Jwt is set to false (it is ${forceMatrix2Jwt}), we did not get a not supported error from the sfu)`,
```
BillCarsonFr, line 149: Unrelated to this PR?

src/widget.test.ts
```diff
@@ -52,48 +51,39 @@ describe("widget", () => {
...
     ];
     const sendRecvEvent = [
       "org.matrix.rageshake_request",
-      EventType.CallEncryptionKeysPrefix,
-      EventType.Reaction,
-      EventType.RoomRedaction,
-      ElementCallReactionEventType,
-      EventType.RTCDecline,
-      EventType.RTCMembership,
+      "io.element.call.encryption_keys",
```
BillCarsonFr, line 59: What it the reason for not using constants?

BillCarsonFr (approved):
I havent tested the netlify part. I am not super fan of "catchall" PRs. I am ok with super small PRs. You can stack them

BillCarsonFr:
The playwright test is failing due to 
image

Apparently due to (?):
image

src/widget.test.ts
toger5: this should also be a different PR!

toger5:
Blocked by: https://github.com/element-hq/element-call/pull/3705

And also it has to be split into:
 - update widget capabilities (no backport needed)
 - Move unrelated doc/log changes to new PR (no backport needed)
 - This PR

sdk/main.ts
```diff
@@ -230,6 +251,16 @@ export async function createMatrixRTCSdk(
     }
   };
 
+  const sendRoomMessage = async (message: string): Promise<void> => {
+    const messageString = JSON.stringify(message);
```
BillCarsonFr, line 255: Why stringify? what good could it done?

BillCarsonFr: Oh it is for logging?

package.json
```diff
@@ -13,8 +13,9 @@
     "build:embedded": "yarn build:full --config vite-embedded.config.js",
     "build:embedded:production": "yarn build:embedded",
     "build:embedded:development": "yarn build:embedded --mode development",
-    "build:sdk": "yarn build:full --config vite-sdk.config.js",
     "build:sdk:development": "yarn build:sdk --mode development",
+    "build:sdk": "yarn build:full --config vite-sdk.config.js",
+    "build:sdk:production": "yarn build:sdk",
```
BillCarsonFr, line 18: What is the difference between production and build:sdk? or should it do `yarn build:full` without --config?

toger5: It is the same. I wanted to not do the production part, but that crashed Github CI. so we do need `build:sdk:production`. I think there are also some things we can do (like not generating `.map` files)
But that probably should go in a seperate PR.

sdk/main.ts
toger5: Its not the best. But the `sdk/main.ts` is very much a POC product (also has no real testing yet)

I will add a comment to the file and the readme to make this more clear.

package.json
toger5: I will add a comment and an issue to track this.

toger5: https://github.com/element-hq/element-call/issues/3714

src/widget.test.ts
toger5: Ah wait this is for actually making more sane tests. Thesting that the constants are correct as well. Otherwise we just repeat what we have in the implementation

PR #3706, opened 2026-01-28
The advertised livekit_alias in membership is deprecated

BillCarsonFr:
Since this PR on js-sdk https://github.com/matrix-org/matrix-js-sdk/pull/5165 , the `livekit_alias` is not part anymore
of the membership event. 

Element-Call was broken by the change, because it expected the livekit_alias to be there

The livekit alias should be taken from the jwt service instead. Since this [jwt service PR ](https://github.com/element-hq/lk-jwt-service/pull/147), the alias will be the same regardless of the end-point used, so there should be no problems with compatibility?.

**EDIT** After some test it appears that for backward compat with `16`, we need to put the room_id as the `livekit_alias` in the transports of state events (it is not done in sticky events)

### DIFF
--- a/src/state/CallViewModel/localMember/LocalMember.ts
+++ b/src/state/CallViewModel/localMember/LocalMember.ts
@@ -18,6 +18,7 @@ import {
   Status as RTCSessionStatus,
   type LivekitTransportConfig,
   type MatrixRTCSession,
+  type Transport,
 } from "matrix-js-sdk/lib/matrixrtc";
 import {
   BehaviorSubject,
@@ -735,11 +736,25 @@ export function enterRTCSession(
   const multiSFU =
     matrixRTCMode === MatrixRTCMode.Compatibility ||
     matrixRTCMode === MatrixRTCMode.Matrix_2_0;
+
+  // For backwards compatibility with Element Call versions that do not do Matrix 2.0,
+  // we add the livekit alias to the transport.
+  let backwardCompatibleTransport: Transport;
+  if (matrixRTCMode === MatrixRTCMode.Matrix_2_0) {
+    backwardCompatibleTransport = transport;
+  } else {
+    backwardCompatibleTransport = {
+      livekit_alias: rtcSession.room.roomId,
+      ...transport,
+    };
+  }
+
   // Multi-sfu does not need a preferred foci list. just the focus that is actually used.
   // TODO where/how do we track errors originating from the ongoing rtcSession?
+
   rtcSession.joinRTCSession(
     ownMembershipIdentity,
-    multiSFU ? [] : [transport],
+    multiSFU ? [] : [backwardCompatibleTransport],
     multiSFU ? transport : undefined,
     {
       notificationType,
--- a/src/state/CallViewModel/localMember/LocalMember.ts
+++ b/src/state/CallViewModel/localMember/LocalMember.ts
@@ -16,7 +16,7 @@ import {
 import { observeParticipantEvents } from "@livekit/components-core";
 import {
   Status as RTCSessionStatus,
-  type LivekitTransport,
+  type LivekitTransportConfig,
   type MatrixRTCSession,
 } from "matrix-js-sdk/lib/matrixrtc";
 import {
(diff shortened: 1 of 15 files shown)


src/state/CallViewModel/remoteMembers/ConnectionManager.ts
```diff
@@ -42,24 +42,24 @@ export class ConnectionManagerData {
     }
   }
 
-  private getKey(transport: LivekitTransport): string {
-    return transport.livekit_service_url + "|" + transport.livekit_alias;
+  private getKey(transport: LivekitTransportConfig): string {
+    return transport.livekit_service_url;
   }
```
toger5, line 50: Since I stumbled over this I think it might be helpful, if we give this a comment.

This comment should clarify why it is NOT needed to key by room, application, slotId

toger5 (approved):

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -735,12 +736,26 @@ export function enterRTCSession(
...
+  // we add the livekit alias to the transport.
+  let backwardCompatibleTransport: Transport;
+  if (matrixRTCMode === MatrixRTCMode.Matrix_2_0) {
+    backwardCompatibleTransport = transport;
+  } else {
+    backwardCompatibleTransport = {
+      livekit_alias: rtcSession.room.roomId,
+      ...transport,
+    };
+  }
```
robintown, line 750: I believe we can drop the alias in `Compatibility` mode too.

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -735,12 +736,26 @@ export function enterRTCSession(
   const multiSFU =
     matrixRTCMode === MatrixRTCMode.Compatibility ||
     matrixRTCMode === MatrixRTCMode.Matrix_2_0;
+
+  // For backwards compatibility with Element Call versions that do not do Matrix 2.0,
+  // we add the livekit alias to the transport.
+  let backwardCompatibleTransport: Transport;
```
robintown:
```suggestion
  let backwardCompatibleTransport: LivekitTransport | LivekitTransportConfig;
```

For better type safety

robintown (commented):
Still have a couple more files to review here

src/state/CallViewModel/localMember/LocalMember.ts
toger5: I think you are right. We only need the transport for v0.16.2. For ^v0.17.0 we can always just rely on the room getting passed seperatly and not read from livekit_alias.
Since v0.16. does not support multi sfu compatibility would work.

I still prefer sending it so we can have a simpler model. We now tried to have legacy and compatiblity as similar as possible and do the larger changes only with sticky events.

robintown (commented):
OK, all reviewed now.

src/state/CallViewModel/localMember/LocalMember.ts
BillCarsonFr: We tried live yesterday, if it is not sent in compatibility mode it breaks

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
```diff
@@ -43,6 +43,8 @@ export class ConnectionManagerData {
...
 
   private getKey(transport: LivekitTransportConfig): string {
+    // This is enough as a key because the ConnectionManager is already scoped by room.
+    // The ConnectionManager should also be scoped by slotId when we have multi-slot support.
```
toger5: I think this is already happening due to the rtc session itself being already scoped by slot id. And the connection manager only ever gets the view of the matrix client through one rtc session?

toger5:
```
    // This is enough as a key because the ConnectionManager is already scoped by room.
    // We also do not need to consider the slotId at this point since each `MatrixRTCSession` is already scoped by `slotDescription: {id, appliaction}`.
```

PR #3715, opened 2026-02-04
Revert "[WIP] SDK target improvements"

toger5:
Reverts element-hq/element-call#3701

We would like this PR to be split into sub-PRs.


BillCarsonFr (approved):
Ok, let's break it down

PR #3720, opened 2026-02-04
[MatrixRTC SDK] improvements - compatible with custom rtc application

toger5:
- add local member
 - add optional make sticky
 - default to video not enabled
 - allow sending matrix events


toger5:
We justify not the test coverage
`[codecov/patch](https://github.com/element-hq/element-call/pull/3720/checks?check_run_id=62528550112) — 77.77% of diff hit (target 80.00%)` since this code only mostly the `sdk/` folder which is experimental.

BillCarsonFr:
> We justify not the test coverage `[codecov/patch](https://github.com/element-hq/element-call/pull/3720/checks?check_run_id=62528550112) — 77.77% of diff hit (target 80.00%)` since this code only mostly the `sdk/` folder which is experimental.

The sdk folder is not included in coverage (see `vitest.config.js`). 
The untested lines are regarding the new thing `sendRoomEvents = false`, which I think is worth a test

BillCarsonFr (approved):

PR #3725, opened 2026-02-06
Push docker images to oci.element.io

fkwp:
For https://github.com/element-hq/serverproduct-internal/issues/1153


.github/workflows/build-and-publish-docker.yaml
```diff
@@ -33,19 +33,57 @@ jobs:
...
+      - name: Login to GitHub container registry
         uses: docker/login-action@5e57cd118135c172c3672efd75eb46360885c0ef # v3.6.0
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Connect to Tailscale
+        uses: tailscale/github-action@53acf823325fe9ca47f4cdaa951f90b4b0de5bb9 # v4
+        if: github.event_name != 'pull_request'
```
robintown, line 45: Hm, isn't this workflow only triggered by `workflow_call`?

.github/workflows/build-and-publish-docker.yaml
robintown, line 100: I'm surprised to see that the push step is still entirely the same. For my own understanding: does this action just automatically push to all registries you've logged into, or what's actually directing it to the right place?

fkwp: I was broadly following https://github.com/element-hq/element-web/pull/31734

and yes Ill assume it pushes to all logged in registries...

.github/workflows/build-and-publish-docker.yaml
fkwp: yes, which is triggered by:

build.yaml:    uses: ./.github/workflows/build-and-publish-docker.yaml
pr-deploy.yaml:    uses: ./.github/workflows/build-and-publish-docker.yaml
publish.yaml:    uses: ./.github/workflows/build-and-publish-docker.yaml

Not sure if we should limit the element registry to only releases...

.github/workflows/build-and-publish-docker.yaml
robintown: Seems that must be the case: https://docs.docker.com/build/ci/github-actions/push-multi-registries/ Weird of Docker to work that way tbh!

.github/workflows/build-and-publish-docker.yaml
robintown: Oh, yes, I just read the [GitHub Actions docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_call):

> When a workflow is triggered with the `workflow_call` event, the event payload in the called workflow is the same event payload from the calling workflow.

That makes sense. Now I'm only wondering why https://github.com/element-hq/element-web/blob/develop/.github/workflows/docker.yaml puts this `if` condition on more steps than we do (it puts it on the "login to ghcr.io" and "push" steps as well).

robintown (approved):

PR #3726, opened 2026-02-06
Add debug logs to generateItems

robintown:
It's always worth having logs for when state holders are created or destroyed (these are often the most interesting things happening in the application), so I thought it would be nice to have `generateItems` always log for you when it's doing that.

The logs look like this:

```
13:41:55.520 [MatrixLivekitMembers] Creating item with keys @robin:localhost, BGDDWZJGPB, 18e9e031-50b3-4136-8f4e-e2b8d49700a7, 1iyY8G1LNyEehaR/3qoQTtPtEKHy+nxnaG+jpqs9u2k [rageshake.ts:488:7](https://localhost:3000/src/settings/rageshake.ts)
13:41:57.851 [ConnectionManager connections$] Destroying item with keys http://localhost:8882/, [object Object] [rageshake.ts:488:7](https://localhost:3000/src/settings/rageshake.ts)
13:41:57.852 [MatrixLivekitMembers] Destroying item with keys @robin:localhost, BGDDWZJGPB, 7bce3d2d-d2f6-43ef-b91f-dda81cde18c6, 7LlR9eHgQKvIzIPbOBJt6oAI5QJyggxAN6xBKGnMvZo [rageshake.ts:488:7](https://localhost:3000/src/settings/rageshake.ts)
```

### DIFF
--- a/src/state/CallViewModel/CallViewModel.ts
+++ b/src/state/CallViewModel/CallViewModel.ts
@@ -715,6 +715,7 @@ export function createCallViewModel$(
       // Generate a collection of MediaItems from the list of expected (whether
       // present or missing) LiveKit participants.
       generateItems(
+        "CallViewModel userMedia$",
         function* ([
           localMatrixLivekitMember,
           matrixLivekitMembers,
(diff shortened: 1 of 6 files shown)


BillCarsonFr (approved):
🙏 that's great! I think we need more logs overall. And also do some cleanup on the too verbose ones.
How does these logs looks like? can you add a sample in the PR comment?

PR #3731, opened 2026-02-10
fix: Netlify preview broken

BillCarsonFr:
Fixes https://github.com/element-hq/voip-internal/issues/475

Regression due to https://github.com/element-hq/element-call/pull/3720

`deploy-to-netlify.yaml` was referencing a non-existing config file. 
Causing the required config file to be served with a 200 OK response but a body `404 not found`, thus preventing the app to load

Not sure what was the expected behavior wanted by @toger5 , for now I am adding a custom config for sdk with no analytics or sentry?

### DIFF
--- a/.github/workflows/deploy-to-netlify.yaml
+++ b/.github/workflows/deploy-to-netlify.yaml
@@ -64,6 +64,8 @@ jobs:
       - name: Add config file
         run: |
           if [ "${{ inputs.package }}" = "full" ]; then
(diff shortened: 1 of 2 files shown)


toger5:
Still needs a chmagelog label.
I was going with the assumption that the sdk does not need a config.json. but this might have been wrong... so adding one is the right fix for the 404

toger5 (approved):

PR #3734, opened 2026-02-11
Logically separate the advertised transport from the active transport

### LINKED ISSUE #3716: Legacy mode hits the auth service endpoint whenever memberships change
### Steps to reproduce

1. Use Legacy mode
2. Open network devtools
3. Connect to call
4. Have someone else connect and disconnect repeatedly

#### What did you expect?

Client only hits the auth service endpoint on initial connection setup

#### What happened instead?

Client hits the auth service repeatedly whenever memberships change

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

robintown:
To correctly implement the legacy "oldest membership" mode, we need the code to be more nuanced about the local transport. Specifically, it needs to allow for the transport we advertise in our membership to be different from the transport that we connect to and publish media on. Otherwise, if these two are yoked together, members will resend their memberships whenever an SFU hop occurs, which an attacker could use to cause an amplified wave of state changes.

Closes https://github.com/element-hq/element-call/issues/3716
Closes https://github.com/element-hq/voip-internal/issues/474

### DIFF
--- a/src/state/CallViewModel/CallViewModel.ts
+++ b/src/state/CallViewModel/CallViewModel.ts
@@ -444,35 +444,34 @@ export function createCallViewModel$(
     memberId: uuidv4(),
   };
 
-  const localTransport$ = createLocalTransport$({
-    scope: scope,
-    memberships$: memberships$,
-    ownMembershipIdentity,
-    client,
-    delayId$: scope.behavior(
-      (
-        fromEvent(
-          matrixRTCSession,
-          MembershipManagerEvent.DelayIdChanged,
-          // The type of reemitted event includes the original emitted as the second arg.
-        ) as Observable<[string | undefined, IMembershipManager]>
-      ).pipe(map(([delayId]) => delayId ?? null)),
-      matrixRTCSession.delayId ?? null,
-    ),
-    roomId: matrixRoom.roomId,
-    forceJwtEndpoint$: scope.behavior(
-      matrixRTCMode$.pipe(
-        map((v) =>
-          v === MatrixRTCMode.Matrix_2_0
-            ? JwtEndpointVersion.Matrix_2_0
-            : JwtEndpointVersion.Legacy,
-        ),
+  const localTransport$ = scope.behavior(
+    matrixRTCMode$.pipe(
+      map((mode) =>
+        createLocalTransport$({
+          scope: scope,
+          memberships$: memberships$,
+          ownMembershipIdentity,
+          client,
+          delayId$: scope.behavior(
+            (
+              fromEvent(
+                matrixRTCSession,
+                MembershipManagerEvent.DelayIdChanged,
+                // The type of reemitted event includes the original emitted as the second arg.
+              ) as Observable<[string | undefined, IMembershipManager]>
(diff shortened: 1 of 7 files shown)


src/state/CallViewModel/localMember/LocalMember.ts
toger5, line 1: Always nice to see to only pass what we really need in the function.

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -58,8 +61,8 @@ interface Props {
     OpenIDClientParts;
   // Used by the jwt service to create the livekit room and compute the livekit alias.
   roomId: string;
-  useOldestMember$: Behavior<boolean>;
-  forceJwtEndpoint$: Behavior<JwtEndpointVersion>;
+  useOldestMember: boolean;
+  forceJwtEndpoint: JwtEndpointVersion;
```
toger5, line 65: I do not immediately see how those two can be non observable?
Does this mean I need to rejoin to change from oldest member to mutlisfu?

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -93,23 +96,35 @@ export interface LocalTransportWithSFUConfig {
...
   sfuConfig: SFUConfig;
 }
+
 export function isLocalTransportWithSFUConfig(
   obj: LivekitTransportConfig | LocalTransportWithSFUConfig,
 ): obj is LocalTransportWithSFUConfig {
   return "transport" in obj && "sfuConfig" in obj;
 }
 
+interface LocalTransport {
```
toger5, line 106: I like this

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -444,35 +444,34 @@ export function createCallViewModel$(
...
-        map((v) =>
-          v === MatrixRTCMode.Matrix_2_0
-            ? JwtEndpointVersion.Matrix_2_0
-            : JwtEndpointVersion.Legacy,
-        ),
+  const localTransport$ = scope.behavior(
+    matrixRTCMode$.pipe(
+      map((mode) =>
+        createLocalTransport$({
+          scope: scope,
```
toger5: okay this is how we get away with making useOldestMember non observablie inside createLocalTransport

robintown: Yep, as I mentioned on Wednesday though it is unsafe to reference `scope` in the callback like this, it leads to resource leaks. So I had to replace the `map` with `generateItem`. (6cf859fd9e242501620c3053507d4a669f56004e) I still think this is worth it for how it lets the oldest member and multi-SFU branches be very cleanly separated in LocalTransport.ts.

src/state/CallViewModel/localMember/LocalTransport.ts
robintown: Indeed, the idea is you restart the LocalTransport state entirely when there's this kind of config change.

toger5 (approved):

PR #3747, opened 2026-02-19
Add volume control to screen shares

### LINKED ISSUE #2337: We no longer have the ability to set volume of someone else's screen-sharing tile

JakeTripplJ:
Fixes #2337 

This PR adds volume control and muting capabilities to remote screen shares using the built-in slider wrapper to maintain a cohesive look for the UI. The slider only shows up in spotlight view, and also only shows up when a screen share includes audio. It also remembers the previous volume level upon toggling mute by clicking the speaker icon.

screen share volume slider screenshot
another screen share volume slider screenshot

### DIFF
--- a/src/tile/SpotlightTile.tsx
+++ b/src/tile/SpotlightTile.tsx
@@ -22,6 +22,8 @@ import {
   ChevronRightIcon,
   VolumeOffIcon,
   VolumeOnIcon,
+  VolumeOffSolidIcon,
+  VolumeOnSolidIcon,
 } from "@vector-im/compound-design-tokens/assets/web/icons";
 import { animated } from "@react-spring/web";
 import { type Observable, map } from "rxjs";
@@ -49,7 +51,7 @@ import { type ScreenShareViewModel } from "../state/media/ScreenShareViewModel";
 import { type RemoteScreenShareViewModel } from "../state/media/RemoteScreenShareViewModel";
 import { type MediaViewModel } from "../state/media/MediaViewModel";
 import { Slider } from "../Slider";
-import { constant } from "../state/Behavior";
+import { platform } from "../Platform";
 
 interface SpotlightItemBaseProps {
   ref?: Ref<HTMLDivElement>;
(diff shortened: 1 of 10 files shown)


toger5:
Thanks for the contribution.
Let us know if you need any help on getting the CI happy.

JakeTripplJ:
If the playwright CI test fails again then I'll have to host a backend server to actually properly run the tests locally and see what it's getting hung up on. It seems like playwright tries clicking the spotlight view button after entering a call as part of a test and nothing happens, although it works fine when I perform the same sequence on the Netlify-hosted PR version. Almost all of my local testing was done on the embedded version through element-web, so maybe it's an issue that only shows up on the non-embedded version. Let me know if you have any ideas/suggestions.

src/tile/SpotlightTile.tsx
```diff
@@ -234,6 +238,20 @@ export const SpotlightTile: FC<Props> = ({
   const visibleIndex = media.findIndex((vm) => vm.id === visibleId);
   const canGoBack = visibleIndex > 0;
   const canGoToNext = visibleIndex !== -1 && visibleIndex < media.length - 1;
+  const currentMedia = media[visibleIndex];
+  const isScreenShare = currentMedia instanceof ScreenShareViewModel;
+  const hasAudio$ = useBehavior(currentMedia.audioEnabled$);
```
robintown: In this case the Playwright tests are crashing with: `TypeError: Cannot read properties of undefined (reading 'audioEnabled$')`, which indicates that `currentMedia` can be undefined here. (The test switches to spotlight layout so quickly that even the user's own tile hasn't yet appeared in the interface!)

robintown (commented):
Really glad to have this feature contributed, it's giving us the motivation now to properly work out what the UI for these extra screen share controls should look like. Stay tuned as there are some changes we'd like to request to have it work more similarly to the volume controls on normal user tiles - but I need to finalize that with our design team first.

You don't actually have to run the Playwright tests yourself to see how they failed: You can download the playwright-report.zip artifact from the CI run [here](https://github.com/element-hq/element-call/actions/runs/22206476176?pr=3747), extract it, and then serve it over HTTP to access the "trace viewer" which will show you screenshots, logs, and everything.

Screenshot From 2026-02-20 16-41-34

robintown:
Sorry that the "Publish SwiftPM Library" action is failing by the way, you can ignore it.

JakeTripplJ:
I didn't know that the CI served a report with actual traces, I was just going off of the logs it spat out (which weren't very descriptive). That makes things much easier to debug, thanks. 

As for the actual UI for the slider, I was initially going to put it in its own context menu like the user tiles, but I ended up just going down the easier route of putting it right on the tile itself (which was ironically harder to get working due to finding out the hard way that onPointerUp was required to commit the value). The actual volume control logic for ScreenShareViewModel should still be useful in either case though. Let me know what everyone eventually decides on for the actual UX, I can always tweak it if needed.

robintown:
Hey @JakeTripplJ, I can share the designs now! When a screen share does have an audio track, we would indeed like the volume slider to appear inside a menu, like this:

Screenshot From 2026-03-02 12-37-17
Screenshot From 2026-03-02 12-37-32

To mute, the user can of course drag the slider to zero.

Screenshot From 2026-03-02 12-44-44

I would suggest not trying to make it toggle into the muted state when the icon is clicked, because this complicates the controls.

For the code: sorry about the merge conflicts, but there's now a proper `RemoteScreenShareViewModel` module which you can put the volume logic in. To reduce code duplication, could you also please use the new `createVolumeControls` function that I factored out in https://github.com/element-hq/element-call/pull/3755? I hope that will make this change even simpler. :) Let me know if you have any questions.

JakeTripplJ:
I've changed the UI to match the example you provided, and also rebased it and updated it to be compatible with the refactor. Here are some screenshots:

Screenshot_20260304_041800
Screenshot_20260304_035742
Screenshot_20260304_034024

The only change I made from the screenshots you provided was to make the volume icon within the menu pure white/black instead of gray like in the example, as it matches how the user volume slider from GridTile looks. I ended up making the volume icon within the menu a clickable mute toggle as well, because I personally think a simple two-click muting solution is much faster and easier than having to drag the slider all the way to the left (although that's still an option). Being able to toggle-unmute to the previously set volume is also extremely useful to me (and I assume many other users). Let me know if there are any other issues I have to fix or changes I should make.

JakeTripplJ:
Ok, that last safeguard should hopefully make playwright happy.

JakeTripplJ:
I cleaned up and redid the safety checks I added to hopefully make them more readable and less prone to letting undefined values leak through (which seemed to be causing the playwright tests to fail). Let me know if there's a simpler/better way to do it that you guys prefer.

robintown (approved):
Great stuff! I made some minor changes to ensure this is usable as well on mobile devices.

PR #3755, opened 2026-02-25
Convert media view model classes to interfaces

robintown:
Timo and I agreed previously that we should ditch the class pattern for view models and instead have them be interfaces which are simply created by functions. They're more straightforward to write, mock, and instantiate this way.

The code for media view models and media items is pretty much the last remaining instance of the class pattern. Since I was about to introduce a new media view model for ringing, I wanted to get this refactor out of the way first rather than add to the technical debt.

This refactor also makes things a little easier for https://github.com/element-hq/element-call/pull/3747 by extracting volume controls into their own module.

For https://github.com/element-hq/voip-internal/issues/455


src/state/media/MediaItem.ts
```diff
@@ -0,0 +1,196 @@
...
+export type WrappedUserMediaViewModel = UserMediaViewModel & {
+  /**
+   * All screen share media associated with this user media.
+   */
+  screenShares$: Behavior<ScreenShareViewModel[]>;
+  /**
+   * Which sorting bin the media item should be placed in.
+   */
+  bin$: Behavior<SortingBin>;
+};
```
robintown, line 83: This is the interface that replaces the previous `UserMedia` class. Hopefully its name now makes it a bit clearer where it sits in the hierarchy.

src/state/ScreenShare.ts
```diff
@@ -1,53 +0,0 @@
...
-import { ScreenShareViewModel } from "./MediaViewModel.ts";
-import type { EncryptionSystem } from "../e2ee/sharedKeyManagement.ts";
-import { constant, type Behavior } from "./Behavior.ts";
-
-/**
- * A screen share media item to be presented in a tile. This is a thin wrapper
- * around ScreenShareViewModel which essentially just establishes an
- * ObservableScope for behaviors that the view model depends on.
- */
-export class ScreenShare {
```
robintown, line 24: This class was doing basically nothing, so I've deleted it. The call view model now holds references to `ScreenShareViewModel` directly instead.

BillCarsonFr:
@robintown While we are at it, could it be possible to extract some new files as per [coding style](https://github.com/matrix-org/matrix-js-sdk/blob/develop/code_style.md) rule 30

> There should be approximately one interface, class, or enum per file unless the file is named "types.ts", "global.d.ts", or ends with "-types.ts".
> The file name should match the interface, class, or enum name.

For example split `MediaViewModel.ts`

robintown:
Sure, I can split MediaViewModel.ts.

src/state/media/MemberMediaViewModel.ts
```diff
@@ -0,0 +1,280 @@
...
+
+import { type Behavior } from "../Behavior";
+import { type BaseMediaViewModel, createBaseMedia } from "./MediaViewModel";
+import { type EncryptionSystem } from "../../e2ee/sharedKeyManagement";
+import { type ObservableScope } from "../ObservableScope";
+import { observeTrackReference$ } from "../observeTrackReference";
+import { E2eeType } from "../../e2ee/e2eeType";
+import { observeRtpStreamStats$ } from "./observeRtpStreamStats";
+
+export enum EncryptionStatus {
```
BillCarsonFr, line 43: we should create a task to cleanup that. Given it is not used? and might not even work

src/state/media/MemberMediaViewModel.ts
```diff
@@ -0,0 +1,280 @@
...
+              startWith(EncryptionStatus.Connecting),
+            );
+          }
+        }),
+      ),
+    ),
+  };
+}
+
+function observeInboundRtpStreamStats$(
```
BillCarsonFr: I think this can be moved to observeRtpStreamStats?. 
I could do it after in this [PR](https://github.com/element-hq/element-call/pull/3756), because it is now used for another purpose (find the video orientation)

BillCarsonFr (approved):
Thx, good improvement 👍

src/state/media/MemberMediaViewModel.ts
robintown: Added a TODO comment and opened https://github.com/element-hq/element-call/issues/3760

PR #3756, opened 2026-02-25
Auto fit video in tile depending on orientation of the tile and the source stream

BillCarsonFr:
I moved the logic out to a specific file to make the merging of https://github.com/element-hq/element-call/pull/3755 easier

Fixes https://github.com/element-hq/voip-internal/issues/468

## Before

Mobile video (portrait) are zoomed on desktop (landscape) and the other way around.
It also makes the video looks pixelized

image

## After 

image

The manual option to fit to frame has been removed in favor of the smart default based on orientation
image


src/tile/GridTile.tsx
```diff
@@ -103,18 +104,26 @@ const UserMediaTile: FC<UserMediaTileProps> = ({
...
+  const videoFit = useBehavior(vm.videoFit$);
+
   const rtcBackendIdentity = vm.rtcBackendIdentity;
   const handRaised = useBehavior(vm.handRaised$);
   const reaction = useBehavior(vm.reaction$);
 
+  // We need to keep track of the tile size.
+  // We use this to get the tile ratio, and compare it to the video ratio to decide
+  // whether to fit the video to frame or keep the ratio.
+  const [measureRef, bounds] = useMeasure();
```
robintown: `bounds` will change quite frequently due to animation and impact the performance. Could you instead use the `targetWidth` and `targetHeight` props passed to this component? Those represent the ["width/height this tile will have once its animations have settled"](https://github.com/element-hq/element-call/blob/00f880108e29e629b0c4aae835b06fd55f96eb19/src/grid/Grid.tsx#L171).

src/tile/SpotlightTile.tsx
```diff
@@ -173,7 +174,22 @@ const SpotlightItem: FC<SpotlightItemProps> = ({
...
+  const [measureRef, bounds] = useMeasure();
+
+  // Whenever bounds change, inform the viewModel
+  useEffect(() => {
+    if (bounds.width > 0 && bounds.height > 0) {
+      if (vm.type != "screen share") {
+        vm.setActualDimensions(bounds.width, bounds.height);
+      }
+    }
+  }, [bounds.width, bounds.height, vm]);
```
robintown: This could likewise use `targetWidth` + `targetHeight`

src/utils/videoFit.ts
```diff
@@ -0,0 +1,103 @@
...
+      // If we have a participant, observe their video track stats. If not, return undefined.
+      switchMap((p) => {
+        if (!p) return of(undefined);
+        if (p.isLocal) {
+          return observeOutboundRtpStreamStats$(p, Track.Source.Camera);
+        } else {
+          return observeInboundRtpStreamStats$(p, Track.Source.Camera);
+        }
+      }),
+    )
```
robintown, line 93: A UserMediaViewModel already has a `videoStreamStats$` observable, so rather than creating a second such observable here, could you have this function take the `videoStreamStats$` as a parameter? You'd have to remove [this `showConnectionStats` condition](https://github.com/element-hq/element-call/blob/00f880108e29e629b0c4aae835b06fd55f96eb19/src/state/media/UserMediaViewModel.ts#L138) of course.

(Aside: `audioStreamStats$` and `videoStreamStats$` really ought to both be Behaviors…)

src/utils/videoFit.ts
```diff
@@ -0,0 +1,103 @@
...
+  observeInboundRtpStreamStats$,
+  observeOutboundRtpStreamStats$,
+} from "../state/media/observeRtpStreamStats";
+
+type Size = {
+  width: number;
+  height: number;
+};
+
+export function videoFit$(
```
robintown, line 35: Doc comment? For instance what does an undefined size mean?

robintown (approved):
Yay, excited for this to land

robintown:
Oops, didn't mean to approve it just yet, but anyhow you have my feedback

src/utils/videoFit.ts
BillCarsonFr: Yes it has, but as you said it is used as a devtool option, and most of the time does not emit.
Most of the code is shared in observeRtpStreamStats$ anyhow.
I don't see benefit of doing like so. We have better modularisation like it is now?

src/tile/SpotlightTile.tsx
BillCarsonFr: Not sure it improves readability

src/tile/GridTile.tsx
BillCarsonFr: Thx for the pointer to targetWidth/Height. Was not aware of that, will use it.
Regarding impact on perf with bounds, how much would this be? because at the end videoFit will not change that much and it is the one triggering rendering. Asking because there are other uses of `useMeasure()`, should we consider using sone debounce for those?

robintown: Generally `useMeasure` doesn't create performance problems, only when combined with react-spring animations. The deal with react-spring is that for performance it updates element styles manually (bypassing React) on every animation frame. So if you add `useMeasure` into the mix (which re-renders whenever the element dimensions change) then React would have to re-render the component on every frame, negating react-spring's performance optimizations.

The other usages of `useMeasure` that I can see are in Grid, InCallView, and VideoPreview, none of which are controlled by react-spring, so they shouldn't need a debounce.

src/tile/SpotlightTile.tsx
robintown: I meant for performance reasons (edit: ah, I see you did make this change)

src/utils/videoFit.ts
robintown: Given that we have to do a semi-expensive polling operation to get the track settings, it would make sense to me to have this operation centralized in the view model. That way if we add another non-devtool thing that reads the frame rate or whatever, we're less likely to start polling these things 3+ times over.

Low-temperature feedback though.

robintown (approved):

src/utils/videoFit.ts
BillCarsonFr: Ok, let's see. I keep it like that for now, and will review when we will use another non dev-tool thing based on it

PR #3763, opened 2026-02-28
Add deafen toggle to mute all incoming audio

alisonjenkins:
## Summary
- Add deafen event type and DeafenReader for reading remote deafen state from Matrix room events
- Add local deafen toggle with muteAllAudio$ integration to suppress all incoming audio
- Thread deafened$ observable through CallViewModel and UserMediaViewModel for UI state
- Add DeafenButton component and deafened indicator on participant tiles
- Broadcast deafen state to other participants via ReactionsSenderProvider
- Wire deafen button into the call toolbar with auto-mute mic on deafen (restores previous state on un-deafen), 'd' keyboard shortcut, and translation keys

## Test plan
- [x] Verify deafen button appears in the call toolbar and toggles deafen state
- [x] Confirm all incoming audio is silenced when deafened
- [x] Test that microphone is auto-muted when deafening and restored to previous state on un-deafen
- [x] Verify mic toggle is disabled while deafened
- [x] Test 'd' keyboard shortcut toggles deafen
- [x] Confirm deafened indicator appears on the local participant's tile
- [ ] Verify deafen state is broadcast to and visible by other participants

### DIFF
--- a/locales/en/app.json
+++ b/locales/en/app.json
@@ -61,6 +61,7 @@
     "username": "Username",
     "video": "Video"
   },
+  "deafen_button_label": "Deafen",
   "developer_mode": {
     "always_show_iphone_earpiece": "Show iPhone earpiece option on all platforms",
     "crypto_version": "Crypto version: {{version}}",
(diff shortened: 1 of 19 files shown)


toger5:
Can you create and link this to an issue this PR solves (Or provide some in depth explanation why you would like to see this in element call as part of this PR)

PR #3768, opened 2026-03-02
fix: add id-token permission as its required by tailscale login

fkwp:


robintown (approved):
Git is doing something pretty weird with the diff here, but I can see on further inspection that you're basically just adding the `id-token: write` line.

PR #3780, opened 2026-03-08
Download avatars using the MC4039 Widget API

### LINKED ISSUE #2876: Use Matrix user profile image in avatar when running as a widget

Show the Matrix user profile thumbnail for the user when using embedded mode.

Because it looks nicer and provides a more integrated experience.

Add a new capability to the widget API that would allow the fetching of the media.

We used to have this capability, but it was removed as part of https://github.com/element-hq/element-call/pull/2856 as the is currently no way to implement it whilst also using authenticated media.

An alternative would be to use unauthenticated media, but that is not desirable.

### Additional context

_No response_

JephDiel:
This adds the ability to download avatar images using the MC4039 widget API if we can't directly fetch from the server.

Fixes #2876

### DIFF
--- a/src/Avatar.test.tsx
+++ b/src/Avatar.test.tsx
@@ -14,7 +14,13 @@ import { ClientContextProvider } from "./ClientContext";
 import { Avatar } from "./Avatar";
 import { mockMatrixRoomMember, mockRtcMembership } from "./utils/test";
 import { widget } from "./widget";
-import { WidgetApi } from "matrix-widget-api";
+import { type FC, type PropsWithChildren } from "react";
+import { type WidgetApi } from "matrix-widget-api";
+
+import { ClientContextProvider } from "./ClientContext";
+import { Avatar } from "./Avatar";
+import { mockMatrixRoomMember, mockRtcMembership } from "./utils/test";
+import { widget } from "./widget";
 
 const TestComponent: FC<
   PropsWithChildren<{
--- a/src/Avatar.test.tsx
+++ b/src/Avatar.test.tsx
@@ -17,8 +17,15 @@ import EventEmitter from "events";
 import { widget } from "./widget";
 
 const TestComponent: FC<
-  PropsWithChildren<{ client: MatrixClient; supportsThumbnails?: boolean }>
-> = ({ client, children, supportsThumbnails }) => {
+  PropsWithChildren<{
+    client: MatrixClient;
+    supportsAuthenticatedMedia?: boolean;
+  }>
+> = ({
+  client,
+  children,
+  supportsAuthenticatedMedia: supportsAuthenticatedMedia,
+}) => {
   return (
     <ClientContextProvider
       value={{
@@ -26,7 +33,7 @@ const TestComponent: FC<
         disconnected: false,
         supportedFeatures: {
           reactions: true,
-          thumbnails: supportsThumbnails ?? true,
+          authenticatedMedia: supportsAuthenticatedMedia ?? true,
         },
         setClient: vi.fn(),
         authenticated: {
@@ -81,36 +88,7 @@ test("should just render a placeholder when the user has no avatar", () => {
   expect(client.mxcUrlToHttp).toBeCalledTimes(0);
 });
 
-test("should just render a placeholder when thumbnails are not supported", () => {
-  const client = vi.mocked<MatrixClient>({
-    getAccessToken: () => "my-access-token",
-    mxcUrlToHttp: () => vi.fn(),
-  } as unknown as MatrixClient);
-
-  vi.spyOn(client, "mxcUrlToHttp");
-  const member = mockMatrixRoomMember(
-    mockRtcMembership("@alice:example.org", "AAAA"),
-    {
-      getMxcAvatarUrl: () => "mxc://example.org/alice-avatar",
-    },
-  );
-  const displayName = "Alice";
-  render(
-    <TestComponent client={client} supportsThumbnails={false}>
-      <Avatar
-        id={member.userId}
-        name={displayName}
-        size={96}
-        src={member.getMxcAvatarUrl()}
-      />
-    </TestComponent>,
-  );
-  const element = screen.getByRole("img", { name: "@alice:example.org" });
-  expect(element.tagName).toEqual("SPAN");
-  expect(client.mxcUrlToHttp).toBeCalledTimes(0);
-});
-
-test("should attempt to fetch authenticated media", async () => {
+test("should attempt to fetch authenticated media if supported", async () => {
   const expectedAuthUrl = "http://example.org/media/alice-avatar";
   const expectedObjectURL = "my-object-url";
   const accessToken = "my-access-token";
@@ -142,7 +120,7 @@ test("should attempt to fetch authenticated media", async () => {
   );
   const displayName = "Alice";
   render(
-    <TestComponent client={client}>
+    <TestComponent client={client} supportsAuthenticatedMedia={true}>
       <Avatar
         id={member.userId}
         name={displayName}
@@ -163,7 +141,7 @@ test("should attempt to fetch authenticated media", async () => {
   });
 });
 
-test("should use widget API when unable to authenticate media", async () => {
+test("should attempt to use widget API if authenticate media is not supported", async () => {
   const expectedMXCUrl = "mxc://example.org/alice-avatar";
   const expectedObjectURL = "my-object-url";
   const theBlob = new Blob([]);
@@ -188,7 +166,7 @@ test("should use widget API when unable to authenticate media", async () => {
   );
   const displayName = "Alice";
   render(
-    <TestComponent client={client}>
+    <TestComponent client={client} supportsAuthenticatedMedia={false}>
       <Avatar
         id={member.userId}
         name={displayName}
--- a/src/Avatar.test.tsx
+++ b/src/Avatar.test.tsx
@@ -13,6 +13,8 @@ import { type FC, type PropsWithChildren } from "react";
 import { ClientContextProvider } from "./ClientContext";
 import { Avatar } from "./Avatar";
 import { mockMatrixRoomMember, mockRtcMembership } from "./utils/test";
+import EventEmitter from "events";
+import { widget } from "./widget";
 
 const TestComponent: FC<
   PropsWithChildren<{ client: MatrixClient; supportsThumbnails?: boolean }>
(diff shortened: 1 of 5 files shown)


JephDiel:
Depends on https://github.com/element-hq/element-web/pull/32755

BillCarsonFr:
Thx @JephDiel !

Adding that to my list.
There a couple of things I want to do as part of that.

So my understanding is that it depends on:
- Widget API support of https://github.com/matrix-org/matrix-widget-api/pull/99
- Was implemented here in web https://github.com/matrix-org/matrix-react-sdk/pull/12931 (when it was react sdk)

Then there is this web PR ~~https://github.com/element-hq/element-call/pull/3780~~ https://github.com/element-hq/element-web/pull/32755 that adds the capabilities to the list of auto approved capabilities for Element Call


Can you elaborate quickly on the fallback / compatibility of your PR?
- We need first to merge the web PR (?)
- What happen if the client widget driver does not support it? Like on EX currently?

I'll check if there is any implementation on rust-sdk, if not I'll create the issues so that we had it quickly if posssible

src/Avatar.tsx
```diff
@@ -86,33 +88,17 @@ export const Avatar: FC<Props> = ({
...
-      headers: {
-        Authorization: `Bearer ${token}`,
-      },
-    })
-      .then(async (req) => req.blob())
+
+    getAvatarFromServer(clientState, src, sizePx) // Try to download directly from the mxc://
+      .catch((ex) => {
+        return getAvatarFromWidget(widget?.api, src); // Fallback to trying to use the MSC4039 Widget API
+      })
```
toger5: we should be able to know what to do without letting the "fromServer" approach fail. (via the `widget` singleton)

toger5 (changes requested):

JephDiel:
> Thx @JephDiel !
> 
> Adding that to my list. There a couple of things I want to do as part of that.
> 
> So my understanding is that it depends on:
> 
> * Widget API support of [Implement MSC4039 download_file action matrix-org/matrix-widget-api#99](https://github.com/matrix-org/matrix-widget-api/pull/99)
> * Was implemented here in web [Implement download_file in widget driver matrix-org/matrix-react-sdk#12931](https://github.com/matrix-org/matrix-react-sdk/pull/12931) (when it was react sdk)
> 
> Then there is this web PR #3780 that adds the capabilities to the list of auto approved capabilities for Element Call
> 
> Can you elaborate quickly on the fallback / compatibility of your PR?
> 
> * We need first to merge the web PR (?)
> * What happen if the client widget driver does not support it? Like on EX currently?
> 
> I'll check if there is any implementation on rust-sdk, if not I'll create the issues so that we had it quickly if posssible

Without the web changes the downloadFile function will [fail](https://github.com/matrix-org/matrix-widget-api/blob/1a56b93b78a2065943577a501d7f0ca9e9125658/src/ClientWidgetApi.ts#L994) or [query the user for permission](https://github.com/element-hq/element-web/blob/2e0adc5832ac931cd26324bd032f6bb780866dc5/apps/web/src/stores/widgets/ElementWidgetDriver.ts#L281), and the .catch block would handle an error and continue to use the first letter of the username.

Merging the web PR first would be ideal to mitigate the aforementioned feature permission UI.

src/Avatar.tsx
JephDiel: The `thumbnails` feature apparently represented just that, so I've renamed it to be more apt and used it to conditionally choose the function based on the it and existence of the widget.api object.

In the midst of the changes I also asserted `sizePx` as non-null to remove the unnecessary check.

toger5:
Since we use the embedded version of element call I dont see how EW is blocking this PR.
The PR, that bumps EC to v0.18.0 in EW will be blocked by the permissions PR. But the v0.18.0 release still takes a bit of time. (RC period is 1 week min)

src/Avatar.tsx
```diff
@@ -80,25 +80,37 @@ export const Avatar: FC<Props> = ({
...
+
+    if (
+      supportedFeatures.authenticatedMedia &&
+      authenticated?.client &&
+      sizePx
+    ) {
+      blob = getAvatarFromServer(authenticated.client, src, sizePx);
+    } else if (widget?.api) {
+      blob = getAvatarFromWidgetAPI(widget.api, src);
+    } else {
```
toger5: This is still very hard to read. I need to know that       `supportedFeatures.authenticatedMedia` will be false if I am in widget mode. Without this unintuitive condition one will be left with the question. If I am in widget mode, but all the initial conditions are true, wouldnt I end up with `getAvatarFromServer`?

It looks to me as if `authenticatedMedia` is really only used here.
I dont see why we put it into the supportedFeatures list. Seems like an unecassary abstraction.

Just doing:
```
if (
      widget?.api === undefined &&
      authenticated?.client &&
      sizePx
    ) {
      blob = getAvatarFromServer(authenticated.client, src, sizePx);
    } else if (widget?.api) {
      blob = getAvatarFromWidgetAPI(widget.api, src);
    } else { ...
```
and not having the `authenticatedMedia` all over the place seems sane to me.

JephDiel: That somewhat is the goal; if we could authenticate and download from the server directly that would be ideal for its simplicity and avoiding image size scaling issues mentioned in the msc.

I was trying to leave the code somewhat future proofed for that, but granted that doesn't exist now so an `if (widget?.api)` would be sufficient and the extra complexity can wait untill if/when its useful, if you think that would be better.

JephDiel:
> Since we use the embedded version of element call I dont see how EW is blocking this PR. The PR, that bumps EC to v0.18.0 in EW will be blocked by the permissions PR. But the v0.18.0 release still takes a bit of time. (RC period is 1 week min)

Ah, that makes sense. So this code can be merged but won't be embedded until web explicitly decides to take it/updated the call version, which is the only thing that actually needs to wait for the added permission PR? I assumed web was pulling call from the public url and there wasn't version control.

toger5:
> Ah, that makes sense. So this code can be merged but won't be embedded until web explicitly decides to take it/updated the call version, which is the only thing that actually needs to wait for the added permission PR? I assumed web was pulling call from the public url and there wasn't version control.

We cut an RC yesterday without this PR. So it is now even less controversial to merge this PR. It will be part of the v0.19.0 release.

src/Avatar.tsx
toger5: > That somewhat is the goal; if we could authenticate and download from the server directly that would be ideal for its simplicity and avoiding image size scaling issues mentioned in the msc.

This is a widget msc discussion. It is basically the scoped access token widget api request.
But this is still future dreams as of now.

src/ClientContext.tsx
```diff
@@ -335,7 +336,7 @@ export const ClientProvider: FC<Props> = ({ children }) => {
       }
     } else {
       setSupportsReactions(true);
-      setSupportsThumbnails(true);
+      setSupportsAuthenticatedMedia(true);
```
toger5: This is the only place we call `setSupportsAuthenticatedMedia` which is only checking on `initClientState.widgetApi` so it really is redundant information to `widget !== undefined`.

Lets remove it here. If we actually move to an scoped token based appraoch in the future a much larger PR will be needed. Right now it just makes things more complicated.

JephDiel: Removed the feature, now switching on widget?.api directly.
Also added a quick fix to a potential race condition while I was there.

toger5 (approved):

src/Avatar.test.tsx
```diff
@@ -13,18 +13,21 @@ import { type FC, type PropsWithChildren } from "react";
 import { ClientContextProvider } from "./ClientContext";
 import { Avatar } from "./Avatar";
 import { mockMatrixRoomMember, mockRtcMembership } from "./utils/test";
+import { widget } from "./widget";
+import { WidgetApi } from "matrix-widget-api";
```
toger5:
```suggestion
import { type FC, type PropsWithChildren } from "react";
import { type WidgetApi } from "matrix-widget-api";

import { ClientContextProvider } from "./ClientContext";
import { Avatar } from "./Avatar";
import { mockMatrixRoomMember, mockRtcMembership } from "./utils/test";
import { widget } from "./widget";
```

src/Avatar.tsx
```diff
@@ -16,6 +16,8 @@ import { Avatar as CompoundAvatar } from "@vector-im/compound-web";
 import { type MatrixClient } from "matrix-js-sdk";
 
 import { useClientState } from "./ClientContext";
+import { widget } from "./widget";
+import { WidgetApi } from "matrix-widget-api";
```
toger5:
```suggestion
import { type MatrixClient } from "matrix-js-sdk";
import { type WidgetApi } from "matrix-widget-api";

import { useClientState } from "./ClientContext";
import { widget } from "./widget";
```

src/Avatar.test.tsx
toger5: Here github is a bit confused what it should mark as a diff.
The `import { WidgetApi } from "matrix-widget-api";` needs to be moved up to the previous import block and needs a `type` prefix.

toger5:
Thanks for the changes. I just restarted CI.

JephDiel:
There seems to be an issue with the CI checkout of element-call-swift, there's a few other recent PRs with stopped on the same error ([3781](https://github.com/element-hq/element-call/actions/runs/23005465607/job/66801228618?pr=3781) and [3773](https://github.com/element-hq/element-call/actions/runs/22851379422/job/66280659419?pr=3773
)).

toger5:
> There seems to be an issue with the CI checkout of element-call-swift, there's a few other recent PRs with stopped on the same error ([3781](https://github.com/element-hq/element-call/actions/runs/23005465607/job/66801228618?pr=3781) and [3773](https://github.com/element-hq/element-call/actions/runs/22851379422/job/66280659419?pr=3773)).

I think it will not work when you have a pr from a fork.
We can force merge it once there is a review from @BillCarsonFr

BillCarsonFr:
> > There seems to be an issue with the CI checkout of element-call-swift, there's a few other recent PRs with stopped on the same error ([3781](https://github.com/element-hq/element-call/actions/runs/23005465607/job/66801228618?pr=3781) and [3773](https://github.com/element-hq/element-call/actions/runs/22851379422/job/66280659419?pr=3773)).
> 
> I think it will not work when you have a pr from a fork. We can force merge it once there is a review from @BillCarsonFr

Yes, i am looking at all of that this week. Will also try to add rust-sdk support for EX clients

BillCarsonFr (approved):

Issue #1229, opened 2023-07-10
Ability to pop out video tiles ("picture-in-picture")


SimonBrandner:
One request that we've been getting recently is a way to activate the browser's [picture-in-picture functionality](https://developer.mozilla.org/en-US/docs/Web/API/Picture-in-Picture_API) from within Element Call.

Yet to be discovered:
- What are the use cases for this?
- Would we rather provide a feature that lets you pop out a specific tile only, or one that creates a window always highlighting the active speaker?
- On the implementation side: how does this interact with LiveKit automatically unsubscribing from video streams that are off screen?
- Is this worth implementing given that Firefox doesn't yet support the relevant APIs?
- Design (potential for integrating it with the drag-and-drop interaction, such that you could literally drag a tile and drop it outside the window to pop it out, much like a browser tab? or do we need something less surprising like a button)

robintown:
I've elaborated on this ticket based on what I believe the title was referring to

bblacher:
I'd love to have the ability to pop out specific screen shares. This would mean that one can watch two screen shares at the same time. (This ability should be in element desktop as well)
I think discord does a great job at this.

theraspb3rry:
Yep! Agreed with @bblacher. My friends and I often have multiple screen shares going at the same time and one of the things you can do on discord is pop out the entire video section and also tile all the video and screen share streams. Even if its a hidden feature enabled in the settings this would be ok for our use case.

Use-cases:
- Multiple views of people playing the same game (we play Escape From Tarkov which this situation comes up often)
- People working on a common goal such as programming or debugging an issue live
- Casually watching a whole bunch of friends play different games
- cause you can

I'd imagine it work similar to discord in a way, create a new window and either select unique streams or show all video streams. I think the "double click" to expand UX in discords window is very intuitive in general (not really specific to this issue), but getting it to only show what you want can be a little annoying. I can further elaborate with pictures and more detailed instructions if its helpful.

rriemann:
I think Zoom has it and the picture stays on top of other applications. That's what people know from Webex and Teams. So for feature parity in corporate settings, very relevant in my view.

Super helpful when multi tasking during a call.

Issue #120, opened 2021-11-23
Picture in Picture


jakewb-b:
- As a user in a video call
- I want to see a small display of the speaker when another tab/window/app is active
- So that I can still see who is speaking, see facial expressions, and see any shared screen

The PiP should show the active speaker, or the shared screen if there is a shared screen.

Ideally, the PiP would also offer access to key interactions: enable/disable camera, mute/unmute and end call

SimonBrandner:
I am not sure if this is even possible within the browser

SimonBrandner:
I guess we could make a video element and assign the stream of the speaker every time it changes but I have doubts it will be possible to add a button to the PiP view

Destinyg133:
It should be possible to turn it off somehow so its not in the way

joshua-martius:
Currently the disablePictureInPicture flag is set on the videostream. 

I tested removing the flag locally and I could PiP my 1:1 videocall flawlessly. 

It would be great if that flag could get removed per default, or at least have it saved dynamically based on user preferences.

joshua-martius:
Update: 
I came up with this 
https://gist.github.com/joshua-martius/5b9c80ff09e9f9895681988f8a55ea8b
and Im now injecting it into my element call session. Works like a charm!

daniellekirkwood:
Hiya, thanks for the feedback and comments on this ticket. We have a Pip in the product and so the original purpose of this issue has been met, I'm going to close it. 

Looks like there are other enhancements that you'd like, feel free to open those as issues if they're still a problem for you and we'll take it into consideration

PR #3781, opened 2026-03-08
Allow user to control whether Picture In Picture is enabled or disabled

### LINKED ISSUE #1545: Picture-in-picture button shown on video preview in Firefox
### Steps to reproduce

1. I can't think of a reason to allow the preview webcam feed to be displayed in the picture in picture mode

#### What happened instead?

On Firefox it adds some UI overlay that looks misplaced

Screenshot 2023-09-20 at 10 18 53

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

Yes

beef331:
I realize #1229 has a lot of questions unanswered by this PR, but this is a very functional feature in my testing.

Fixes #1545 #1229
Partially addresses #120

### DIFF
--- a/src/room/VideoPreview.tsx
+++ b/src/room/VideoPreview.tsx
@@ -11,6 +11,8 @@ import { facingModeFromLocalTrack, type LocalVideoTrack } from "livekit-client";
 import classNames from "classnames";
 import { useTranslation } from "react-i18next";
 
+import { allowPipSetting, useSetting, } from "../settings/settings";
+
 import { TileAvatar } from "../tile/TileAvatar";
 import styles from "./VideoPreview.module.css";
 import { type EncryptionSystem } from "../e2ee/sharedKeyManagement";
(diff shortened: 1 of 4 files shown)


toger5:
Can we have a screenshot of the settings (with the pip allow option)

beef331:
image

And in Firefox videos looks like : 
image

toger5:
I like this approach
@robintown what was the reason to disable pip initially? There was some good reason I do not remember.
@daniellekirkwood Is this an acceptable place to put this setting from a product point of view?

Dependent on the answers we might need to move it in the devtools (if there is some technical issues which made us disable it)

beef331:
I have since learned that not many browsers do PIP videos well. So this is only really a stopgap from a proper PIP implementation for select people.

BillCarsonFr:
@beef331 We definitely want to be able to pop-out the call from EW not only tile by tile.

Have you tried to experiment with [Document Picture-In-Picture](https://developer.mozilla.org/en-US/docs/Web/API/Document_Picture-in-Picture_API) like that we could pop-out the app frame?

It is coming soon to Firefox (153?)/

beef331:
I know of users that'd want to be able to pop out both the videos and the entire frame. I did toy with `window.open`, but having React nodes move between two documents is beyond my React knowledge.

PR #3794, opened 2026-03-10
Activate click tests for pip buttons

toger5:
This activates the PIP view click tests that was disabled due to missing docker for EW.
We need: https://github.com/element-hq/element-web/pull/32654

To be published here: https://github.com/element-hq/element-web/pkgs/container/element-web


toger5:
This was merged into the orig. PR

PR #3797, opened 2026-03-11
Fix:  zizmor findings

fkwp:
Fix zizmor findings wrt. GitHub Actions. Add exceptions for dangerous-triggers including comments for reasoning.


.github/workflows/publish-embedded-packages.yaml
```diff
@@ -23,7 +23,16 @@ jobs:
     steps:
       - name: Calculate VERSION
         # We should only use the hard coded test value for a dry run
-        run: echo "VERSION=${{ github.event_name == 'release' && github.event.release.tag_name || 'v0.0.0-pre.0' }}" >> "$GITHUB_ENV"
+        run: |
+          # Die Logik wird nun innerhalb der Shell mit den Variablen ausgeführt
```
toger5: This language is unexpected.

toger5 (changes requested):

.github/workflows/publish-embedded-packages.yaml
fkwp: addressed by https://github.com/element-hq/element-call/pull/3797/commits/3e36891904e90e84223efc7af2c1c7023a8dd181

toger5 (approved):

PR #3801, opened 2026-03-12
Audio Control | Default to using earpiece when doing audio vs speaker for video call (depending on presence of headset or not)

### LINKED ISSUE #3774: Bug | EC should not default voice call to the speaker

I expect a voice call to use the earpiece by default. As I would put the phone to my ear (like I would for a native phone call).
Currently it is starting on speaker

Notice: If user is using BT headset, it should then stay on the headset and not switch to earpiece

BillCarsonFr:
## Content

fix #3774
This PR is a following up the recent addition of [voice call support on android](github.com/element-hq/element-x-android/pull/5995).
When you accept a voice call on android, EC was automatically choosing the speaker audio routing instead of earpiece. Speaker is a good default for video calls, but for voice call you expect the sound out of the earpiece.

We do not handle routing the same way for every platfrom. Mobile has some specificities, as well as some limitations when using only the web API. 
That's why we have some native/custom controls for mobiles.

There are been severl iterations, testing, tweaks for the mobile integration. Part of the code is a bit messy due to that (deprecated code, attributes, outdated behavior, etc..)

I spent some time adding more documentation and logs.
We have more than a difference between mobile and web, there are notable difference between iOS and android.

That's why I decided to extract a specific android audio control class **AndroidControlledAudioOutput**.
It allows me to do extensive testing with real life android values, without riskign regressing any iOS behavior.

The `AndroidControlledAudioOutput` now uses the initial call intent to decide for the best default.
It is also responsible for auto-swtiching to a device if a new device is added or removed. 
If a new device is added it will only switch to it if it is more private. e.g:
- if you are on headset and a BT speaker connects, you will stay on the headset
- If you are on speaker and you connect a bluetooth headset, sound will go to headset

**Notable change**: The android native part used to also do some automatic switching, this was causing selection flickers, as there was 2 different things trying to set a good default. 
This is not the case anymore, EC has all the informations to decide for the best device. So the `AndroidControlledAudioOutput` doesn't listen to native calling the controls to set a device.

=> Need a follow-up task on android, to do some clean-up:
- Stop using controls to auto select a default
- Stop using long deprecated methods
- Stop filtering out hearing aids 
- Mark wired headset as external headsets

## Tests

- Step 1
- Step 2
- Step ...
-

## Checklist


src/controls.ts
```diff
@@ -33,12 +33,36 @@ export interface Controls {
...
 export interface OutputDevice {
   id: string;
   name: string;
+  /**
+   * Reverse engineered:
+   *  - on iOS always true if output is routed to speaker. In other case iOS on declare a `dummy` id device.
+   *    In that case then ElementCalls manually append a earpiece device with id `EARPIECE_CONFIG_ID` anb `{ type: "earpiece" }`
+   *  - on Android this is unused.
+   */
   forEarpiece?: boolean;
```
toger5, line 50:
```suggestion
  /**
   * `forEarpiece` in an iOS only flag, that will be set on the default speaker device.
   * The default speaker device will be used for the earpiece mode by
   * using a stereo pan and reducing the volume significantly. (in combination this is similar to a dedicated earpiece mode)
   *  - on iOS this is true if output is routed to speaker.
   *    In that case then ElementCalls manually appends an earpiece device with id `EARPIECE_CONFIG_ID` and `{ type: "earpiece" }`
   *  - on Android this is unused.
   */
forEarpiece?: boolean;
```

src/controls.ts
```diff
@@ -33,12 +33,36 @@ export interface Controls {
...
+  /**
+   * Reverse engineered:
+   * - on iOS always undefined
+   * - on Android true for the `TYPE_BUILTIN_EARPIECE`
+   */
   isEarpiece?: boolean;
```
toger5, line 56:
```suggestion
  /**
   * Is the device the OS earpiece audio configuration?
   * - on iOS always undefined
   * - on Android true for the `TYPE_BUILTIN_EARPIECE`
   */
isEarpiece?: boolean;
```

src/controls.ts
```diff
@@ -33,12 +33,36 @@ export interface Controls {
...
+  /**
+   * Reverse engineered:
```
toger5:
```suggestion
   * Is the device the OS default speaker:
```

src/controls.ts
```diff
@@ -33,12 +33,36 @@ export interface Controls {
...
+   * - on iOS always true if output is routed to speaker. In other case iOS on declare a `dummy` id device.
+   * - on Android true for the `TYPE_BUILTIN_SPEAKER`
+   */
   isSpeaker?: boolean;
+  /**
+   * Reverse engineered:
```
toger5:
```suggestion
   * Is the device the OR default external headset (bluetooth):
```

src/state/MediaDevices.ts
```diff
@@ -143,19 +175,29 @@ function buildDeviceMap(
 function selectDevice$<Label>(
   available$: Observable<Map<string, Label>>,
   preferredId$: Observable<string | undefined>,
+  defaultPicker: (available: Map<string, Label>) => string | undefined = (
```
toger5: This new field (maybe even all of the fields) could get a docstring.
The `defaultPicker` is the tool i can use to pick a different default in voice vs video right?

src/state/MediaDevices.ts
```diff
@@ -434,7 +486,14 @@ export class MediaDevices {
...
   > = getUrlParams().controlledAudioDevices
-    ? new ControlledAudioOutput(this.usingNames$, this.scope)
+    ? platform == "android"
+      ? new AndroidControlledAudioOutput(
+          controlledAvailableOutputDevices$,
+          this.scope,
+          getUrlParams().callIntent,
+          window.controls,
+        )
+      : new ControlledAudioOutput(this.usingNames$, this.scope)
```
toger5: should we rename this to iOS Controlled AudioOutput? it is only used for iOS from now on right?

src/state/AndroidControlledAudioOutput.ts
```diff
@@ -0,0 +1,358 @@
...
+ */
+export class AndroidControlledAudioOutput implements MediaDevice<
+  AudioOutputDeviceLabel,
+  SelectedAudioOutputDevice
+> {
+  private logger = rootLogger.getChild(
+    "[MediaDevices AndroidControlledAudioOutput]",
+  );
+
+  // STATE stream: the current state of the controller, including the list of available devices and the selected device.
```
toger5: could be a docstring (and remove STATE: stream, as this information is part of the type.)

src/state/MediaDevices.ts
```diff
@@ -275,6 +319,11 @@ class AudioOutput implements MediaDevice<
   }
 }
 
+/**
+ * A special implementation of audio output that allows the hosting application
+ * to have more control over the device selection process. This is used when the
+ * `controlledAudioDevices` URL parameter is set, which is currently only true on mobile.
+ */
 class ControlledAudioOutput implements MediaDevice<
```
toger5: Can we already move this into a file called iOSControlledAudio.ts to by symmetric with the android approach?

src/state/MediaDevices.ts
toger5: This looks unexpected. I would expect the `../control` import for this to happen in the `AndroidControlledAudioOutput.ts` file and then the available should just use the `this.audioOutput` where we have access to the MediaDevices from the controlled branch. The ios specific class should then provide whatever we store in iosDeviceMenu.

toger5: Looking at this again. We should not even allocate this class if we are on android. So `controlledAvailableOutputDevices` should always be empty?

src/state/MediaDevices.ts
```diff
@@ -434,7 +486,14 @@ export class MediaDevices {
     AudioOutputDeviceLabel,
     SelectedAudioOutputDevice
   > = getUrlParams().controlledAudioDevices
-    ? new ControlledAudioOutput(this.usingNames$, this.scope)
+    ? platform == "android"
```
toger5, line 372: Those two together also exclude web. So `ControlledAudioOutput` really is just iOS but it has way too much logic in it if it is just iOS?

src/state/MediaDevices.ts
```diff
@@ -309,8 +358,12 @@ class ControlledAudioOutput implements MediaDevice<
 
         // Create a virtual earpiece device in case a non-earpiece device is
         // designated for this purpose
-        if (iosDeviceMenu && availableRaw.some((d) => d.forEarpiece))
+        if (iosDeviceMenu && availableRaw.some((d) => d.forEarpiece)) {
+          this.logger.info(
+            `IOS Add virtual earpiece device with id ${EARPIECE_CONFIG_ID}`,
+          );
           available.set(EARPIECE_CONFIG_ID, { type: "earpiece" });
+        }
```
toger5: only the `iosDeviceMenu`. branch will ever be executed. Can we remove everything else?

src/state/MediaDevices.ts
```diff
@@ -351,23 +409,17 @@ class ControlledAudioOutput implements MediaDevice<
...
       // whether it should enable the proximity sensor.
       if (device !== undefined) {
-        this.logger.info(
-          "[controlled-output] onAudioDeviceSelect called:",
-          device,
-        );
+        this.logger.info("onAudioDeviceSelect called:", device);
         window.controls.onAudioDeviceSelect?.(device.id);
         // Also invoke the deprecated callback for backward compatibility
         window.controls.onOutputDeviceSelect?.(device.id);
```
toger5: Also not needed since this is just iOS?

toger5 (changes requested):
I would love to also remove all the code that is not used anymore from `ControlledAudioOutput`. Otherwise the state of the PR is a bit confusing since we end up with lots of unreachable code.

src/state/MediaDevices.ts
BillCarsonFr: Will do it in a follow up PR

src/state/MediaDevices.ts
BillCarsonFr: Will clean iOS control with the follow up PR

src/state/MediaDevices.ts
BillCarsonFr: same

src/state/AndroidControlledAudioOutput.ts
```diff
@@ -0,0 +1,360 @@
...
+};
+
+/**
+ * The possible actions that can be performed on the controller,
+ * either by the user or by the system.
+ */
+type ControllerAction =
+  | { type: "selectDevice"; deviceId: string | undefined }
+  | { type: "deviceUpdated"; devices: OutputDevice[] };
+/**
```
toger5, line 51:
```suggestion

/**
```

src/state/AndroidControlledAudioOutput.ts
```diff
@@ -0,0 +1,360 @@
...
+  ): Behavior<SelectedAudioOutputDevice | undefined> {
+    return this.scope.behavior(
+      state$
+        .pipe(
+          map((state) => {
+            if (state.selectedDeviceId) {
+              return {
+                id: state.selectedDeviceId,
+                /** This is an iOS thing, always false for android*/
+                virtualEarpiece: false,
```
toger5, line 228: should we make `virtualEarpiece` optional?

src/state/AndroidControlledAudioOutput.ts
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    // Is there a new added device?
+    // If a device is added, we might want to switch to it if it's more preferred than the currently selected device.
+    const newDeviceWasAdded = availableDevices.some(
+      (device) => !previousDevices.some((d) => d.id === device.id),
+    );
+
+    if (newDeviceWasAdded) {
+      // TODO only want to check from the added device, not all devices.?
```
toger5, line 281: This todo should be resolved?

toger5 (approved):
`maybeDismissBackupChatsToast` seems to cause issues for CI.

I played around a bit with this version and the audio situation on android is really really good. All edge-cases I could come up with just worked flawlessly!
I think a lot of them might have worked before but either way its super great to see how far we have come.
I would go as far as saying on android the whole audio + IFrame situation is a solved problem, especially with the clenaup this PR brings !!

BillCarsonFr:
I am by-passing the codecov requirement. It is just a renaming done for the review, no behavior change. And I am working on a follow-up PR to over the iOS controller that will add tests on all the renamed code

rajil:
@BillCarsonFr This feature showed up in the stable release of EXA today. Unfortunately, it is not working as expected. Initially the call does come out of the earpiece on the receiving phone, but a second later it goes to speaker. I am using a Samsung phone.

BillCarsonFr:
> @BillCarsonFr This feature showed up in the stable release of EXA today. Unfortunately, it is not working as expected. Initially the call does come out of the earpiece on the receiving phone, but a second later it goes to speaker. I am using a Samsung phone.

Thx @rajil can you open an issue on https://github.com/element-hq/element-x-android/issues with a rageshake.
I thought that this was already fixed :/ With the logs I could check the versions

rajil:
Rageshake submitted both from within EC and EXA; and filed bug report [6563](https://github.com/element-hq/element-x-android/issues/6563).

Issue #2385, opened 2024-05-23
Element Call lobby should show call members


bblacher:
#### What would you like to do?
Click on an Element Call powered video room and see who already joined
#### Why would you like to do it?
Because I want to know who is waiting in an video room so I can evaluate if I should join the room.
#### How would you like to achieve it?
It should work like it did in the old lobby before the redesign.

bblacher:
Original issue: https://github.com/element-hq/element-web/issues/27030

jltobler:
I also think this feature would be a very welcomed addition. I wish I was able to see the users currently connected to a video room prior to joining :smile:

rriemann:
This can have data protection implications. Missing some wireframes/screenshot, I may not have the same understanding of the proposition here.

If someone can find a link in the internet and check who is currently in that call, this could reveal internal information (who are the collaborators) and give means to spy on employees.

robintown:
See https://github.com/element-hq/element-call/pull/3803#issuecomment-4960058883; we decided that a solution more fitting of our messenger apps would be to [show status indicators in the People list](https://github.com/element-hq/element-meta/issues/3270).

PR #3803, opened 2026-03-13
feat: show call member presence in pre-join lobby view

alex-vg:
## Summary

Display participant avatars with display names in the lobby when a call is already active, so users can see who's in the call before joining.

### Changes

- **`useCallParticipants` hook** — derives a deduplicated participant list with resolved display names and avatar URLs from `MatrixRTCSession.memberships` + `Room`
- **`CallParticipantRow` component** — renders up to 8 participant avatars in a horizontal row. Overflow participants are shown as a `+N` indicator with a Compound `<Tooltip>` listing the remaining names. Uses Radix `<VisuallyHidden>` for screen reader accessibility.
- **`LobbyView` integration** — the participant row appears above the `VideoPreview` when `callParticipants.length > 0`
- **`GroupCallView`** — calls the hook and passes participants to `LobbyView`
- **`RoomPage`** — passes `callParticipants={[]}` in the knock/waitForInvite state (user can't see room members)
- **i18n** — 7 new translation keys for participant count and overflow labels

### Testing

- 5 unit tests for `useCallParticipants`
- 9 unit tests for `CallParticipantRow`
- All 452 existing tests still pass
- Manually tested with 2 users on local dev backend

### Screenshots

The participant row appears above the video preview in the lobby when another user is already in the call.
image

### DIFF
--- a/locales/en/app.json
+++ b/locales/en/app.json
@@ -156,6 +156,13 @@
     "join_as_guest": "Join as guest",
     "join_button": "Join call",
     "leave_button": "Back to recents",
+    "participants_in_call_one": "{{count}} participant in call: {{names}}",
+    "participants_in_call_other": "{{count}} participants in call: {{names}}",
+    "participants_in_call_overflow_one": "{{count}} participant in call: {{names}}, and {{overflowCount}} other",
+    "participants_in_call_overflow_other": "{{count}} participants in call: {{names}}, and {{overflowCount}} others",
+    "participants_overflow_count": "+{{count}}",
+    "participants_overflow_label_one": "{{count}} more participant",
+    "participants_overflow_label_other": "{{count}} more participants",
     "waiting_for_invite": "Request sent! Waiting for permission to join…"
   },
   "log_in": "Log In",
--- a/src/room/CallParticipantRow.module.css
+++ b/src/room/CallParticipantRow.module.css
@@ -1,0 +1,60 @@
+/*
+Copyright 2026 New Vector Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+Please see LICENSE in the repository root for full details.
+*/
+
+.participantRow {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  align-items: flex-start;
+  gap: var(--cpd-space-4x);
+  padding: 0 var(--cpd-space-4x);
+  max-width: 100%;
+}
+
+.participantItem {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: var(--cpd-space-1x);
+  width: 64px;
+}
+
+.participantName {
+  max-width: 100%;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  text-align: center;
+  font: var(--cpd-font-body-sm-regular);
+  color: var(--cpd-color-text-secondary);
+}
+
+.overflowItem {
+  display: inline-flex;
+  flex-direction: column;
+  align-items: center;
+  gap: var(--cpd-space-1x);
+  width: 64px;
+  cursor: default;
+}
+
+.overflowCircle {
+  width: 48px;
+  height: 48px;
+  border-radius: 50%;
+  background-color: var(--cpd-color-bg-subtle-secondary);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  font: var(--cpd-font-body-md-semibold);
+  color: var(--cpd-color-text-secondary);
+}
+
+.overflowCount {
+  font: var(--cpd-font-body-sm-regular);
+  color: var(--cpd-color-text-secondary);
+}
--- a/src/room/CallParticipantRow.test.tsx
+++ b/src/room/CallParticipantRow.test.tsx
@@ -1,0 +1,136 @@
+/*
+Copyright 2026 New Vector Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+Please see LICENSE in the repository root for full details.
+*/
+
+import { type JSX, type ReactNode } from "react";
+import { describe, expect, test, vi } from "vitest";
+import { render, screen } from "@testing-library/react";
+import { TooltipProvider } from "@vector-im/compound-web";
+
+import { CallParticipantRow } from "./CallParticipantRow";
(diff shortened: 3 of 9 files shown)


bblacher:
fixes: https://github.com/element-hq/element-call/issues/2385

alex-vg:
@toger5 any feedback would be appreciated.

TemoCrevlis:
> @toger5 any feedback would be appreciated.

This would be really useful.
How would scalability work?
E.g. A call with many participants: all the names won't fit with equal spacing of avatars.

alex-vg:
The PR already addresses the scalability concern. The CallParticipantRow component handles large participant counts via a display limit with overflow:

- A displayLimit prop (defaulting to 8) caps how many avatars+names are rendered inline.
- Participants beyond the limit are not rendered in the DOM but are accessible via a tooltip.
- An ellipsis (…) element is shown to signal truncation, followed by a +N badge indicating how many additional participants are hidden (e.g. +2).
- The overflow item is still a proper listitem in the accessible list, so screen readers get the full count.

So for a call with 50 participants, only the first 8 would be shown inline, with +42 displayed at the end. The displayLimit prop makes it easy to tune this threshold (e.g. based on available horizontal space if needed).

TemoCrevlis:
> The PR already addresses the scalability concern. The CallParticipantRow component handles large participant counts via a display limit with overflow:
> 
>     * A displayLimit prop (defaulting to 8) caps how many avatars+names are rendered inline.
> 
>     * Participants beyond the limit are not rendered in the DOM but are accessible via a tooltip.
> 
>     * An ellipsis (…) element is shown to signal truncation, followed by a +N badge indicating how many additional participants are hidden (e.g. +2).
> 
>     * The overflow item is still a proper listitem in the accessible list, so screen readers get the full count.
> 
> 
> So for a call with 50 participants, only the first 8 would be shown inline, with +42 displayed at the end. The displayLimit prop makes it easy to tune this threshold (e.g. based on available horizontal space if needed).

Makes sense.
I tried recreating it in the desktop version to get a better idea of what it would be like and also made some iterations.
Here are a couple:

A participants preview

A participants preview 31

A participants preview 1

A participants preview 6

toger5:
My instinct tells me design would like to see some of the existing avatar group components.
@americanrefugee have you seen this? Do we have designs for this? Do you want changes to how it is done here or do you approve the designs?

This is what I would like to see for reference:
Screenshot 2026-04-22 at 15 10 52

On click i think there should be a modal listing all participants?

americanrefugee:
A few thoughts...

When a group call starts on Web there is a banner asking you to join or ignore. A face pile shown there. If you click "Ignore", then you see a "Join" button in the room and connect directly to the group call if you click it.

Meanwhile, seeing all the participants as a row above the video tile, and then a list to see all participants, feels like overkill. I'm not convinced that there's any current problem to be solved here.

However, for mobile in the lobby views, I'm fine with swapping out the generic person icon with a face pile.

toger5:
I think this PR focuses on Single-Page-App usage (which is a secondary priority since it is not leveraging all the security/privacy/federation features matrix offers)

>However, for mobile in the lobby views, I'm fine with swapping out the generic person icon with a face pile.

This also seems to work very well for the SPA imo.

So:
Screenshot 2026-04-22 at 16 31 48
should become more similar to:
Screenshot 2026-04-22 at 15 10 52
?

alex-vg:
@americanrefugee @toger5

Thanks for taking a look and sharing your thoughts!

Regarding the necessity of this feature: I built this to address an open issue #2385 that has gathered support from the community. The feedback there indicates that users specifically want the ability to see exactly who is in the call before making the decision to join. (As well known from Discord…)

Regarding the design suggestion to use a face pile: If we hide the usernames and overlap half of the profile pictures, would it fulfill this requirement? While a face pile is visually compact, to actually solve the problem of letting users know exactly who is in the room, I think the UI needs to display full profile pictures and display names clearly.

americanrefugee:
The SPA is focused on guest users, in which case they should not have the ability to know anything about the members for privacy reason.

Whereas in the app we already have an existing room list. It would be far less complicated, for example, to simply show a green video icons in the People list. I have no problem changing the User icon Into a face pile, and perhaps tapping/clicking on the face pile simply opens the People panel/screen.

Meanwhile, a carousel of users isn't very accessible to begin with, and is nearly unusable after a certain number of participants. A searchable list would be far better for displaying potentially dozens, hundreds, or even thousands of attendees. In which case this already exists as the People panel/screen :)

austin3410:
Any news on this?

alex-vg:
It’s still active, and from my side this is ready.

I’d like to keep this PR aligned with the original issue: users want to know who is already in a call before deciding to join. A face-pile alone doesn’t fully address that, because it obscures identity and adds an extra interaction each time.

Also, this SPA is not only a guest-only surface, it is used in embedded/in-app client integrations as well (including desktop/web app contexts), so the “guest-only” assumption does not hold for all usage paths.

On privacy, this PR already enforces a clear boundary in code: in knock / wait-for-invite contexts we pass callParticipants={[]} (no member identity exposure). The participant list is shown only in the normal member pre-join lobby flow.

So this is not “show members to guests”; it is “show call presence to users who already have member-level visibility in that room context.”

I’m happy with the current implementation. If we agree on this policy boundary, I think this is ready to move forward; otherwise, the remaining decision is product/policy direction.

robintown:
Hi there, our team reviewed this and we still believe that a participant list in the lobby view is not quite the right fit, as it would duplicate the functionality of other UI elements that we want to have in the Element messenger apps.

First of all, we intend to show a face pile with a preview of the current participants on the call tiles found in the message timeline:

image

(Our app teams have already begun work on these tiles across platforms.)

Then, to reveal the details of who's in the call, we want to direct users to the People list where 📹 icons should be displayed next to the call participants. Perhaps the face pile opens this list when clicked, as @americanrefugee proposed. Here's what that looks like on the various platforms:

|Web|iOS|Android|
|-|-|-|
|image|image|image|

I filed this enhancement as https://github.com/element-hq/element-meta/issues/3270. Contributions implementing this in any of our apps would be very much welcome!

For further context, we generally don't want to be maintaining new features that are specific to the SPA (non-embedded) mode of Element Call, since we are focused on building Element Call for the embedded, in-app use case nowadays. So we're still potentially interested in having the :bust_in_silhouette: icon replaced with a proper face pile, but unfortunately shouldn't take on anything more complex for the SPA case.

PR #3805, opened 2026-03-16
New ringing UI

robintown:
This implements the most important piece of the new ringing UI by showing a placeholder tile for the participant being dialed, rather than an overlay. This change does *not* attempt to implement the custom mobile styles for the spotlight tile or to offer a "call back" button when ringing times out (those are better suited for separate tasks). It's just about the placeholder tile.

Screenshot From 2026-03-16 15-27-25
Screenshot From 2026-03-16 15-27-48

Closes https://github.com/element-hq/voip-internal/issues/455
Closes https://github.com/element-hq/voip-internal/issues/458


toger5:
The transition looks great!

BillCarsonFr (approved):
Thx robin, it looks good. I tested on web and mobile 👍 
It would have been nice to have in the description some informations about the technical details of how you did achieve the task. Like explaining that you introduced a RingingMedia item and how it works with other items, and any other relevant information to share knowledge in the team

PR #3809, opened 2026-03-18
Invert the colors of the camera and microphone buttons

robintown:
So that they use primary color tokens when unmuted, and secondary color tokens when muted. This makes them work like the screen sharing button.

|-|Before|After|
|-|-|-|
|Unmuted (dark theme)|Screenshot From 2026-03-18 11-34-32|Screenshot From 2026-03-18 11-34-14|
|Muted (dark theme)|Screenshot From 2026-03-18 11-34-36|Screenshot From 2026-03-18 11-34-19|
|Unmuted (light theme)|Screenshot From 2026-03-18 13-14-25|Screenshot From 2026-03-18 13-14-51|
|Muted (light theme)|Screenshot From 2026-03-18 13-14-21|Screenshot From 2026-03-18 13-14-54|

Closes https://github.com/element-hq/voip-internal/issues/484


BillCarsonFr (approved):
Thx for taking time for all the screenshots 🙏 
It's fine by me.
I may prefer the old `muted` vs `enabled` naming. Because enabled make me think of a control enable/disable, but it is not exactly what we are doing.
But I think this rather point down another problem in our button; Like they are not easy to test because accessibility is not very good. That would be another (unrelated) PR to maybe see them as toggles, and have an easy way to see if toggle or not

PR #3826, opened 2026-03-30
Possible fix for avatars showing stale image data

robintown:
By giving React a different key for different users' avatars, this hopefully ensures that a new \ element will be created when the avatar component switches to a different user.

Possible fix for https://github.com/element-hq/voip-internal/issues/525 - needs real-world testing.


robintown:
Turns out this was the wrong fix; see https://github.com/element-hq/element-call/pull/3835 instead

PR #3827, opened 2026-03-30
Set up Storybook again

robintown:
As per team discussions, we would like to give Storybook another go. (It was last removed from the project [in 2023](https://github.com/element-hq/element-call/pull/1643).) How to write stories for components backed by view models is still under discussion, but at least the initial setup should be uncontroversial.

I quickly created stories for a couple of simpler components, just to validate the setup.


toger5 (approved):
Thanks for the initial step to add this back!

PR #3828, opened 2026-03-31
Stop calling rtc/transport in widget mode

BillCarsonFr:
In order to discover what livekit transport to use the application will try in order to get it from:
- The new MSC4143 `rtc/transport`
- Then in the homeserver well-known (via the `org.matrix.msc4143.rtc_foci`)
- And last resort to a local config file  (`livekit` key)

When running in widget mode, the `rtc/transport` will always fail with a `401 Missing access token` (needs a new widget API).

This error has caused several red herring reports for MISSING RTC TRANSPORTS, so for now we stop trying it for nothing on widget until it is [properly supported](https://github.com/matrix-org/matrix-js-sdk/issues/5245)

See red herring:
https://github.com/element-hq/element-x-android/issues/6485
https://github.com/element-hq/element-x-ios/issues/5318

## Content

For now we just check if there is a token on the client (only true in SPA mode).
If not we do not even try this end point and go directly to next fallback

## Tests

- Step 1
- Step 2
- Step ...
-

## Checklist

### DIFF
--- a/src/livekit/openIDSFU.ts
+++ b/src/livekit/openIDSFU.ts
@@ -137,6 +137,7 @@ export async function getSFUConfigWithOpenID(
       );
       logger?.info(`Got JWT from call's active focus URL.`);
     } catch (e) {
+      logger?.debug(`Failed fetching jwt with matrix 2.0 endpoint:`, e);
       if (e instanceof NotSupportedError) {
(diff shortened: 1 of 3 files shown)


fkwp (approved):
lgtm

PR #3835, opened 2026-04-01
Avoid redundantly showing the local user in the PiP tile

robintown:
If you are the only participant in the call, the expanded spotlight layout would redundantly show your media in both the spotlight and PiP tiles. This is a regression; in versions 0.16.1 and earlier we would avoid showing the same user twice.

|Before|After|
|-|-|
|Screen Shot 2026-04-01 at 13 25 49|Screen Shot 2026-04-01 at 13 26 08|

Closes https://github.com/element-hq/voip-internal/issues/525

### DIFF
--- a/src/state/CallViewModel/CallViewModel.test.ts
+++ b/src/state/CallViewModel/CallViewModel.test.ts
@@ -750,6 +750,53 @@ describe.each([
     });
   });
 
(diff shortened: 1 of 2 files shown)


BillCarsonFr (approved):
👍

PR #3838, opened 2026-04-02
Refactor LocalTransport discovery and validation 

BillCarsonFr:
Also fixed a bug for delagation of delayed event where the active localTransport was not updated on delayId changes

Should be reviewed commit per commit!

Will also provide better error reporting for discovery by making a clear cut between:
- Missing rtc transport
- and Authentication problems

## Tests

- Step 1
- Step 2
- Step ...
-

## Checklist

### DIFF
--- a/src/livekit/openIDSFU.ts
+++ b/src/livekit/openIDSFU.ts
@@ -137,21 +137,9 @@ export async function getSFUConfigWithOpenID(
     } catch (e) {
       logger?.debug(`Failed fetching jwt with matrix 2.0 endpoint:`, e);
       // Make this throw a hard error in case we force the matrix2.0 endpoint.
-      if (forceMatrix2Jwt) throw new NoMatrix2AuthorizationService(e as Error);
-
-      // if (e instanceof NotSupportedError) {
-      //   logger?.warn(
-      //     `Failed fetching jwt with matrix 2.0 endpoint (retry with legacy) Not supported`,
-      //     e,
-      //   );
-      // } else {
-      //   logger?.warn(
-      //     `Failed fetching jwt with matrix 2.0 endpoint other issues ->`,
-      //     `(not going to try with legacy endpoint: forceOldJwtEndpoint is set to false, we did not get a not supported error from the sfu)`,
-      //     e,
-      //   );
-      //   // NEVER get bejond this point if we forceMatrix2 and it failed!
-      // }
+      if (forceMatrix2Jwt) {
+        throw new NoMatrix2AuthorizationService(e as Error);
+      }
     }
   }
 
--- a/src/livekit/openIDSFU.ts
+++ b/src/livekit/openIDSFU.ts
@@ -5,11 +5,7 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 Please see LICENSE in the repository root for full details.
 */
 
-import {
-  retryNetworkOperation,
-  type IOpenIDToken,
-  type MatrixClient,
-} from "matrix-js-sdk";
+import { type IOpenIDToken, type MatrixClient } from "matrix-js-sdk";
 import { type CallMembershipIdentityParts } from "matrix-js-sdk/lib/matrixrtc/EncryptionManager";
 import { type Logger } from "matrix-js-sdk/lib/logger";
 
@@ -70,6 +66,7 @@ export type OpenIDClientParts = Pick<
   MatrixClient,
   "getOpenIdToken" | "getDeviceId"
 >;
+
 /**
  * Gets a bearer token from the homeserver and then use it to authenticate
  * to the matrix RTC backend in order to get acces to the SFU.
@@ -113,9 +110,6 @@ export async function getSFUConfigWithOpenID(
     );
   }
   logger?.debug("Got openID token", openIdToken);
-
-  logger?.info(`Trying to get JWT for focus ${serviceUrl}...`);
-
   let sfuConfig: { url: string; jwt: string } | undefined;
 
   const tryBothJwtEndpoints = opts?.forceJwtEndpoint === undefined; // This is for SFUs where we do not publish.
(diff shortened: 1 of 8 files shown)


src/livekit/openIDSFU.ts
```diff
@@ -127,57 +121,67 @@ export async function getSFUConfigWithOpenID(
...
         membership,
         serviceUrl,
         roomId,
         openIdToken,
         opts?.delayEndpointBaseUrl,
         opts?.delayId,
       );
-      logger?.info(`Got JWT from call's active focus URL.`);
+
+      return extractFullConfigFromToken(sfuConfig);
```
robintown, line 136: Thanks for using early returns, I find the function much easier to understand that way.

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -137,91 +138,75 @@ export const createLocalTransport$ = ({
...
+  });
+
+  // Get the preferred transport from the current deployement.
+  const discoveredTransport$ = from(
+    transportDiscovery.discoverPreferredTransport(),
+  );
+
+  const preferredConfig$ = customLivekitUrl.value$
+    .pipe(
+      startWith(null),
```
robintown:
```suggestion
  const preferredConfig$ = customLivekitUrl.value$
    .pipe(
```

`customLivekitUrl.value$` is a Behavior, so it'll always have an initial value.

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -268,6 +253,7 @@ export const createLocalTransport$ = ({
             ),
           ),
         ),
+        null,
```
robintown:
```suggestion
```

Likewise, should be safe to do without this initial value since `oldestMemberTransport$` is a Behavior.

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -277,210 +263,93 @@ export const createLocalTransport$ = ({
...
+// Utility to ensure the user can authenticate with the SFU.
+//
+// We will call `getSFUConfigWithOpenID` once per transport here as it's our
+// only mechanism of validation. This means we will also ask the
+// homeserver for a OpenID token a few times. Since OpenID tokens are single
+// use we don't want to risk any issues by re-using a token.
+//
+// If the OpenID request were to fail, then it's acceptable for us to fail
+// this function early, as we assume the homeserver has got some problems.
+async function doOpenIdAndJWTFromUrl(
```
robintown: This would be nice as a doc comment

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -137,91 +138,75 @@ export const createLocalTransport$ = ({
...
+  const preferredConfig$ = customLivekitUrl.value$
+    .pipe(
+      startWith(null),
+      switchMap((customUrl) => {
+        if (customUrl) {
+          return of({
+            type: "livekit",
+            livekit_service_url: customUrl,
+          } as LivekitTransportConfig);
+        } else {
```
robintown, line 167: I also appreciate this custom URL stuff happening explicitly :+1:

robintown (commented):
Nice, this looks significantly cleaner than before.

I still need to convince myself whether the test from fd08489afb8fe58df8b1ed13681913fa67357c56 is really the desired behavior; after all, I wrote [this comment](https://github.com/element-hq/element-call/blob/47cb01ab6f084ed3aee67e79c4f82d569a966ece/src/state/CallViewModel/localMember/LocalTransport.ts#L212) which implies we will see unnecessary reconnects in the middle of a call whenever the delay ID changes, unless those updates are suppressed. I want to do a real-world test of that before approving, but either way I agree it's confusing for the active transport to have stale data.

robintown (changes requested):
> I still need to convince myself whether the test from https://github.com/element-hq/element-call/commit/fd08489afb8fe58df8b1ed13681913fa67357c56 is really the desired behavior; after all, I wrote [this comment](https://github.com/element-hq/element-call/blob/47cb01ab6f084ed3aee67e79c4f82d569a966ece/src/state/CallViewModel/localMember/LocalTransport.ts#L212) which implies we will see unnecessary reconnects in the middle of a call whenever the delay ID changes, unless those updates are suppressed.

Okay, so I managed to reproduce that regression. See the following screencast in which I compare `livekit` and this branch:

[Screencast From 2026-04-03 16-57-10.webm](https://github.com/user-attachments/assets/76e3d510-1703-4621-844c-adc41de7e44c)

Notice how there are two WebSocket connections to the same SFU every time, which results in a brief 'reconnecting' popup at the start of the call.

BillCarsonFr:
> Okay, so I managed to reproduce that regression. See the following screencast in which I compare `livekit` and this branch:

Right, but the way we design this stream is still strange to me.
We switch the delayId stream to a new LocalTransportWIthConfig, just to then ignore any changes with a distinct until change.

That is very odd. If we switchMap we should propagate, then the connectionManager should then put it's own distinct until change if it is already connected.

Or the alternative is to change the design of this.

We don't switch map delayId to LocalTransport.

But instead once we have a LocalTransport we create an effect that will notify the auth server to use a new delay ID.

I think this would make more sense to me. It also appears that our `/jwt` API has also a design flow, because you have to call it twice in a raw for it to work?

robintown:
Either of those designs sound good to me.

> It also appears that our `/jwt` API has also a design flow, because you have to call it twice in a raw for it to work?

I'm also not a huge fan of that. IIRC reusing the same endpoint for delegation was necessary to avoid making the JWT service require a database, for some reason. Let's revisit the design in a future meeting?

BillCarsonFr:
> Either of those designs sound good to me.

In order to unblock this PR I added back the stop gap to avoid the reconnect glitch and I updated the test to ensure the even though we don't emit a transport change we still hit the auth service with the correct delayId.

Hope we can close this and work on the proper fix in a follow up PR?
https://github.com/element-hq/element-call/pull/3838/changes/9fe35ba8220f84bec46b50cd5f5a514a9422444e

robintown (approved):

src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -137,91 +138,75 @@ export const createLocalTransport$ = ({
...
-                areLivekitTransportsEqual(prev.transport, next.transport),
-              ),
-            ),
-          ),
-        ),
-      ),
     );
 
+  const preferredTransport$ = combineLatest([preferredConfig$, delayId$]).pipe(
+    switchMap(async ([transport, delayId]) => {
```
toger5, line 183: Is there another case for when this might change (other than):
 - Oldest member leaves. new oldest member appears
 - User changes legacy -> compatiblity

If not I think it would help listen them here.

PR #3841, opened 2026-04-03
Use synapse API to register instead doing via UI

BillCarsonFr:
## Content

Several changes to make the test more stable and also faster:

- Use the synapse Admin API to register new users instead of browsing the UI (more stable, but no so much faster)
- Use `await Promise.all([])` for test with several users in order to parallelize the wait times.
- Change non-retriable assertion to retriable assertion (use .poll for more exotic tests)
- Based on past runs, adjust the action timeouts on action that was flaky
- Mark some test as slow as they need more than 30s total

It is 2 to 5mn faster?

With this PR it is taking 13mn to run the tests
Without it is in between 15 to 18mn ?

What do you think?

## Tests

- Step 1
- Step 2
- Step ...
-

## Checklist


playwright/spa-helpers.ts
```diff
@@ -119,25 +119,31 @@ async function setRtcModeFromSettings(
...
-  // There are no other options than to wait for all media to be ready?
-  // Or it is too flaky :/
-  await page.waitForTimeout(3000);
   // No one should be waiting for media
-  await expect(page.getByText("Waiting for media...")).not.toBeVisible();
+  await expect(page.getByText("Waiting for media...")).not.toBeVisible({
+    timeout: 10000,
+  });
 
   // There should be 5 video elements, visible and autoplaying
```
robintown:
```suggestion
  // There should be <count> video elements, visible and autoplaying
```

playwright/widget/huddle-call.test.ts
```diff
@@ -47,52 +50,52 @@ widgetTest("Create and join a group call", async ({ addUser, browserName }) => {
...
+      const frame = user.page
+        .locator('iframe[title="Element Call"]')
+        .contentFrame();
+      await expect(frame.getByTestId("videoTile")).toHaveCount(5, {
+        timeout: 15000,
+      });
+      for (const participant of [valere, timo, robin, halfshot, florian]) {
+        // Check the names are correct
+        await expect(frame.getByText(participant.displayName)).toBeVisible();
+      }
```
robintown: Could use `Promise.all` as well

playwright/widget/huddle-call.test.ts
```diff
@@ -108,28 +111,19 @@ widgetTest("Create and join a group call", async ({ addUser, browserName }) => {
...
-            (v: Element) => window.getComputedStyle(v).display === "block",
-          ).length,
-      );
+    await expect(frame.locator("video")).toHaveCount(5, {
+      timeout: 10000,
+    });
 
     // out of 5 ONLY 4 are visible (display:block) !!
     // XXX we need to be better at our HTML markup and accessibility, it would make
     // this kind of stuff way easier to test if we could look out for aria attributes.
```
robintown: Seems like this comment should go into the test helper

robintown (approved):
Thanks for investigating all of this!

Issue #3483, opened 2025-09-10
Multiple Screenshares: Left/Right Button not visible


bblacher:
### Steps to reproduce

Start multiple screen shares in a room.

### Outcome

The buttons for switching between screen shares aren't visible since 0.15.0.
They still work, they're just invisible. Happens on embedded and SPA.

The buttons should be visible in this situation, the first image is missing the "right" button, the second image is missing the "left" button.
<img width="651" height="511" alt="Image" src="https://github.com/user-attachments/assets/f83fc630-1dbf-4c75-b0f3-0da56d1432e9" />
<img width="648" height="489" alt="Image" src="https://github.com/user-attachments/assets/5049d5ee-01e5-403c-bd20-3bbc04d6de06" />

PR #3845, opened 2026-04-05
Fix transparent screen share switch buttons.

ghost:
## Content

the "advance" (next / back) buttons for switching between streams in the spotlight tile were permanently transparent due to `opacity: 0`, making multi-stream navigation more challenging.

## Motivation and context

Unfixed for months #3483 and started to get a bit annoying trying to locate the exact pixels where the button would otherwise be visible and easy to click on.

## Screenshots / GIFs

|Before|After|
|-|-|
|image|image|

## Tests

image

- Investigated with dev tools (Inspect element on firefox)
- Found the buttons location
- Investigated the culprit
- Removed the attribute
- Wrote a proof of concept javascript snippet to unblock it to test in chrome and both seem to work.

## Checklist

- [x] Pull request includes screenshots or videos if containing UI changes

### DIFF
--- a/src/tile/SpotlightTile.module.css
+++ b/src/tile/SpotlightTile.module.css
@@ -40,7 +40,6 @@ Please see LICENSE in the repository root for full details.
 .advance {
   appearance: none;
   cursor: pointer;
-  opacity: 0;
   padding: calc(var(--cpd-space-3x) - var(--cpd-border-width-1));
   border: var(--cpd-border-width-1) solid
     var(--cpd-color-border-interactive-secondary);


BillCarsonFr:
Thx @xynnylol, but the arrow icon will stay forever on now.
Would be better if it is only visible when the tile is hovered (like other controls on the bottom)

there is also a [--] -- -- indicator at the bottom, this should probably be clickable

xynnylol:
Hi, Yes the indicator would be a separate commit, I am still tracking down where in the code that is but have other stuff to do as well. AS for the always-on there should be a way to get them to hide again, Forcing the opacity to 0 by default isnt really the way. Ill look into it more when i have some time.

Thank you and have a Great Day!

robintown:
Hi @xynnylol, thanks for taking a look but we do need the buttons to be hidden by default for hovering and mobile to work as intended here. Continuing this at: https://github.com/element-hq/element-call/pull/3889

PR #3854, opened 2026-04-08
Fix a race condition causing the joiner to never publish media to the call

### LINKED ISSUE #3849: No media when joining quickly
### Steps to reproduce

Join existing call that is on element.io with a matrix.org account.

2. What do you click?
In the lobby immediately click join.

3. More steps…
You cannot mute/unmute, there is no video, you cannot start a screenshare. (feels like there is no publishing LK connection)

Rageshake: https://github.com/element-hq/element-call-rageshakes/issues/15547

**It works if**:
 - ~~You unmute yourself in the lobby.~~ 
 - Wait for about 5s (you just have to be slow, seems like a race)
 - Only then join

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

Yes

BillCarsonFr:
Fixes https://github.com/element-hq/element-call/issues/3849

~~This will not reach the quality gate but~~ I created a playwright test to reproduce the race and proove the fix.

This what is happening, there are 2 users on different home servers. Alice on HS1 and bob on HS2.

Bob starts a call on HS2.
Alice opens the widget. Then 2 things starts at the same time, alice tries to discover and authenticate to it's own SFU (`SFU_HS1`), and tries also to find a authenticate to bob's SFU (`SFU_HS2`).

https://github.com/element-hq/element-call/blob/efd1b42da07a5004098775b7acb13cda27ba5ee7/src/state/CallViewModel/localMember/LocalTransport.ts#L378-L388

The current code will take the first to win the race and mark it as the one to advertised.

In the race condition, it is `SFU_HS1` that wins.

Then localMembership is told to use advertised$
https://github.com/element-hq/element-call/blob/efd1b42da07a5004098775b7acb13cda27ba5ee7/src/state/CallViewModel/CallViewModel.ts#L549-L551

As a consequence the `LocalMember` never [creates a publisher](https://github.com/element-hq/element-call/blob/efd1b42da07a5004098775b7acb13cda27ba5ee7/src/state/CallViewModel/localMember/LocalMember.ts#L316), because it is told to publish to a transport that is not known by the ConnectionManager (and will never be), so connection is null.
https://github.com/element-hq/element-call/blob/efd1b42da07a5004098775b7acb13cda27ba5ee7/src/state/CallViewModel/localMember/LocalMember.ts#L316-L318

The fix is not just to switch from `advertised$` to `active$`, because when you are the first joiner there is no older member so no active$, and in that case you should start publishing on advertised$.

## Tests

- Step 1
- Step 2
- Step ...
-

## Checklist

### DIFF
--- a/playwright/widget/federation-oldest-membership-bug.spec.ts
+++ b/playwright/widget/federation-oldest-membership-bug.spec.ts
@@ -1,0 +1,92 @@
+/*
+Copyright 2026 Element Creations Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+Please see LICENSE in the repository root for full details.
+*/
+
+import { expect, test } from "@playwright/test";
+
+import { widgetTest } from "../fixtures/widget-user";
+import { HOST1, HOST2, TestHelpers } from "./test-helpers";
+
+widgetTest(
+  "Bug new joiner was not publishing on correct SFU",
+  async ({ addUser, browserName }) => {
+    test.skip(
+      browserName === "firefox",
+      "This is a bug in the old widget, not a browser problem.",
+    );
+
+    test.slow();
+
+    // 2 users in federation
+    const florian = await addUser("floriant", HOST1);
+    const timo = await addUser("timo", HOST2);
+
+    // Florian creates a room and invites Timo to it
+    const roomName = "Call Room";
+    await TestHelpers.createRoom(roomName, florian.page, [timo.mxId]);
+
+    // Timo joins the room
+    await TestHelpers.acceptRoomInvite(roomName, timo.page);
+
+    // Ensure we are in legacy mode (should be the default)
+    await TestHelpers.openWidgetSetEmbeddedElementCallRtcModeCloseWidget(
+      florian.page,
+      "legacy",
+    );
+    await TestHelpers.openWidgetSetEmbeddedElementCallRtcModeCloseWidget(
+      timo.page,
+      "legacy",
+    );
+
+    // Let timo create a call
+    await TestHelpers.startCallInCurrentRoom(timo.page, false);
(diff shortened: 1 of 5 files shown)


robintown (changes requested):
I've been staring at this change for a while because something wasn't making sense to me, and I think I've got it now.

The LocalMembership module is responsible for sending the membership events, and we need those events to contain the 'advertised' transport rather than the 'active' transport, otherwise you will resend your membership whenever an SFU hop occurs, (which https://github.com/element-hq/element-call/pull/3734 was intended to prevent). This change tells LocalMembership to send events that contain the active transport, which would be a regression. (As for why the CI didn't catch that: it's only covered by unit tests for LocalTransport.)

What I failed to realize when writing https://github.com/element-hq/element-call/pull/3734 is that LocalMembership also has the task of creating a publisher for the active transport. Therefore we really ought to pass both the advertised and active transports to it.

I believe that's the root cause of this bug then, that LocalMembership does not currently distinguish between the advertised and active transports, and so it waits for a connection to be created for the *advertised* transport, which never happens (as it's the *active* transport that we create a connection to). Would expanding that distinction to LocalMembership make sense to you?

robintown (approved):
Thanks, looks great! I think I mostly get the thing with not publishing if there's an error with the advertised transport now… it's like a precaution that we will never be publishing while at the same time our video tile is invisible?

robintown:
I wonder if this deserves a backport to the 0.19.1 RC. The bug isn't a regression in 0.19.1 as far as I know but still seems relatively high priority to have patched.

BillCarsonFr:
> it's like a precaution that we will never be publishing while at the same time our video tile is invisible?

I think it is some historical decision to only let homeservers that have a working sfu to participate in calls? It only makes sense in oldest membership workd

PR #3856, opened 2026-04-08
Move settings button out of the button bar (space for loud-speaker button)

toger5:
Fixes: https://github.com/element-hq/voip-internal/issues/535

Lots of the actual design/layout parts of this change will be done in the follow up PR: https://github.com/element-hq/element-call/pull/3857

This is just moving the settings button and the loud-speaker button the the right places.
Can be reviewed commit by commit.

This is the first iteration on updating the settings experience.

This does not yet properly add the speaker button to the bottom bar. It will just reuse the button we had in the top right corer.

The logo in the bottom bar is tricky. It was solved by hiding it if there is limited space so we do not hide the settings button super early: at 660px width.
After:

Screenshot 2026-04-08 at 15 52 55
Screenshot 2026-04-08 at 15 53 01
Screenshot 2026-04-08 at 15 53 13

Screenshot 2026-04-08 at 15 03 57
Screenshot 2026-04-08 at 15 52 09
Screenshot 2026-04-08 at 15 52 27
Screenshot 2026-04-08 at 15 52 34


src/AppBar.tsx
```diff
@@ -68,9 +68,12 @@ export const AppBar: FC<Props> = ({ children }) => {
...
             <Tooltip label={t("common.back")}>
-              <IconButton onClick={onBackClick}>
-                <CollapseIcon />
-              </IconButton>
+              <Button
+                kind={"tertiary"}
+                iconOnly
+                Icon={CollapseIcon}
+                onClick={onBackClick}
+              />
```
robintown: With this change the app bar buttons now have more padding which puts them further away from the edge of the screen than they're supposed to be:

Screenshot From 2026-04-13 13-48-23

Screenshot From 2026-04-13 13-45-20

src/room/InCallView.module.css
```diff
@@ -94,14 +97,25 @@ Please see LICENSE in the repository root for full details.
   justify-self: end;
 }
 
+/*First hide the logo*/
 @media (max-width: 660px) {
```
robintown: IMO it makes sense to already hide the logo at about 750px (the screenshot with the logo looks rather cramped)

src/room/InCallView.module.css
```diff
@@ -94,14 +97,25 @@ Please see LICENSE in the repository root for full details.
...
   .footer {
-    grid-template-areas: ". buttons buttons buttons .";
+    grid-template-areas: "buttons buttons buttons";
   }
 
-  .logo {
+  /*.settingsLogoContainer {
     display: none;
-  }
+  }*/
```
robintown, line 123: With this commented out, the footer appears to break onto two lines:

Screenshot From 2026-04-13 13-55-06

src/room/InCallView.module.css
```diff
@@ -73,8 +71,13 @@ Please see LICENSE in the repository root for full details.
   pointer-events: initial;
 }
 
+.settingsLogoContainer {
+  display: flex;
+  gap: var(--cpd-space-4x);
+  flex-direction: row;
```
robintown, line 81:
```suggestion
  flex-direction: row;
  flex-wrap: nowrap;
```

For good measure

src/button/Button.tsx
```diff
@@ -134,8 +136,10 @@ export const SettingsButton: FC<SettingsButtonProps> = (props) => {
     <Tooltip label={t("common.settings")}>
       <CpdButton
         iconOnly
-        Icon={SettingsSolidIcon}
-        kind="secondary"
+        Icon={
+          platform === "android" ? OverflowVerticalIcon : OverflowHorizontalIcon
+        }
+        kind="tertiary"
```
robintown: From the looks of it the Web designs actually want this to be a [secondary icon button](https://compound.element.io/?path=/story/compound-web_button-iconbutton--with-secondary-kind).

https://www.figma.com/design/aOEkaJtaBmPy058V7uoqVr/Element-Call-Updates---Q1-2026--New-?node-id=20-13023&t=NB6R4BXu66aLMail-4

src/utils/test-viewmodel.ts
```diff
@@ -131,6 +133,7 @@ export function getBasicCallViewModelEnvironment(
   members: RoomMember[],
   initialRtcMemberships: CallMembership[] = [localRtcMember, aliceRtcMember],
   callViewModelOptions: Partial<CallViewModelOptions> = {},
+  mediaDevicesOverride?: MediaDevices,
```
robintown, line 135: I would prefer to move the media devices into `CallViewModelOptions` alongside the other testing parameters. Sound okay to you?

src/room/LobbyView.tsx
```diff
@@ -228,6 +228,7 @@ export const LobbyView: FC<Props> = ({
         </div>
         <div className={inCallStyles.footer}>
           {recentsButtonInFooter && recentsButton}
+          <SettingsButton onClick={openSettings} />
```
robintown: On mobile in landscape the button appears to go on the right but not all the way to the right, which is a little odd:

Screenshot From 2026-04-13 14-40-10

I would suggest keeping the button on the left like it is in InCallView and then moving the 'back to recents' button to the right side of the footer.

src/room/InCallView.tsx
```diff
@@ -671,20 +681,16 @@ export const InCallView: FC<InCallViewProps> = ({
...
+          layout.type !== "pip" && (
+            <SettingsButton key="settings" onClick={openSettings} />
+          )}
+
+        {headerStyle !== "none" && logo}
+      </div>
+
       {showControls && <div className={styles.buttons}>{buttons}</div>}
       {showControls && (
         <LayoutToggle
```
robintown, line 701: It turns out that we were relying on a certain breakpoint in CSS to hide the layout switcher on mobile. This PR makes the switcher still visible at certain screen widths when it shouldn't be:

Screenshot From 2026-04-13 14-43-55

(Clicking it has no effect.) Maybe the move here is to make `gridMode$` have a value of `null` whenever `windowMode$` is `"flat"` rather than use CSS.

src/room/InCallView.tsx
```diff
@@ -671,20 +681,16 @@ export const InCallView: FC<InCallViewProps> = ({
...
-          />
-          {/* Don't mind this odd placement, it's just a little debug label */}
-          {debugTileLayout
-            ? `Tiles generation: ${tileStoreGeneration}`
-            : undefined}
-        </div>
-      )}
+      <div className={styles.settingsLogoContainer}>
+        {showControls &&
+          headerStyle !== HeaderStyle.AppBar &&
```
robintown, line 687:
```suggestion
          // Settings button is also shown in the app bar if present
          headerStyle !== HeaderStyle.AppBar &&
```

robintown (changes requested):
Code makes sense. I've noted a few layout cases that I think could be improved:

src/room/InCallView.module.css
toger5: This is addressed in the [follow up PR](https://github.com/element-hq/element-call/pull/3857) that moves the footer to a storyboard component (Trying to keep the diff for this change as small as possible)
I should probably remove the commented code.

src/room/LobbyView.tsx
toger5: This is also addressed (or at least different/better) in the follow up PR: https://github.com/element-hq/element-call/pull/3857

src/AppBar.tsx
toger5: I see. The plan was to unify them so the left and the right button look the same. But then we need to unify in the other direction.

src/room/InCallView.tsx
toger5: This is also part of the actual bottom bar refactor PR.

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -172,6 +172,7 @@ export interface CallViewModelOptions {
   windowSize$?: Behavior<{ width: number; height: number }>;
   /** The version & compatibility mode of MatrixRTC that we should use. */
   matrixRTCMode$?: Behavior<MatrixRTCMode>;
+  mediaDeviceOverride?: MediaDevices;
```
robintown: This value isn't used anywhere and yet the tests are passing, so that's a little funny. I was imagining more that the `mediaDevices` argument to `createCallViewModel$` could basically just become this option BTW.

src/AppBar.tsx
robintown: This is still a tiny bit off, the buttons have 4px of padding + 16px added by the \<header\> for a total of 20px

src/AppBar.tsx
```diff
@@ -68,12 +68,9 @@ export const AppBar: FC<Props> = ({ children }) => {
...
           <LeftNav>
             <Tooltip label={t("common.back")}>
-              <Button
-                kind={"tertiary"}
-                iconOnly
-                Icon={CollapseIcon}
-                onClick={onBackClick}
-              />
+              <IconButton onClick={onBackClick}>
+                <CollapseIcon />
```
robintown:
```suggestion
                <CollapseIcon aria-hidden />
```

src/AppBar.tsx
```diff
@@ -68,12 +68,9 @@ export const AppBar: FC<Props> = ({ children }) => {
         >
           <LeftNav>
             <Tooltip label={t("common.back")}>
```
robintown, line 70: Unrelated to this PR but I think this should now read "Collapse" or "Collapse call" or something to match the icon

src/button/Button.tsx
```diff
@@ -126,22 +130,32 @@ export const EndCallButton: FC<EndCallButtonProps> = ({
...
-      <CpdButton
-        iconOnly
-        Icon={
-          platform === "android" ? OverflowVerticalIcon : OverflowHorizontalIcon
-        }
-        kind="tertiary"
-        {...props}
-      />
+      <IconButton {...props}>
+        <Icon />
```
robintown:
```suggestion
        <Icon aria-hidden />
```

src/button/Button.tsx
```diff
@@ -126,22 +130,32 @@ export const EndCallButton: FC<EndCallButtonProps> = ({
...
+// const SettingsButton: FC<SettingsButtonProps> = (props) => {
+//   const { t } = useTranslation();
+//   const Icon =
+//     platform === "android" ? OverflowVerticalIcon : OverflowHorizontalIcon;
+//   return (
+//     <Tooltip label={t("common.settings")}>
+//       <CpdButton iconOnly Icon={Icon} kind="secondary" {...props} />
+//     </Tooltip>
+//   );
+// };
```
robintown, line 161: Should be fine to just delete.

robintown (approved):
Thanks, looks just about ready then

src/button/Button.tsx
robintown: I see that this is done in the next PR.

src/state/CallViewModel/CallViewModel.ts
toger5: Yes this is just used in: `getBasicCallViewModelEnvironment`.
That is why it was also possible to make it a toplevel parameter just in the test files and not part of the callViewModel.

Now it uses the options to transport the value from `createInCallView` to `createCallViewModel$` but inside the CallViewModel its not even used instead it gets extracted from the options to the callViewModel.
We could remove it from the `createCallViewModel$` constructor but then it is non optional. I think this decision should be made consiously: Do we want it to be possible to construct a callViewModel with a mediaDevices obj?

I gave it another look in: e46be38869ac8d8b2ad33d3748f6693391329178
I think that makes sense for the current situation without diving into a larger refactor?

src/AppBar.tsx
toger5: Lets make this a new issue: https://github.com/element-hq/element-call/issues/3879

src/button/Button.tsx
toger5: I kept it for less confusing merge conflicts with the follow up PR. (it will be readded there so I will keep it in this PR to again make the merge easier for me)

src/AppBar.tsx
toger5: I am not sure I conclude the right approach to fix this:
I think the icon should be 24px but it is 32px so this is what I am changing now.

I think I would like to take the chance and move the header to a storybook to get this sorted for good. Currenlty there is a lot of overhead for me to get it tested (is there a trick to get the app bar on web?)

PR #3857, opened 2026-04-09
Footer component -> Storybook

toger5:
Fixes: https://github.com/element-hq/voip-internal/issues/536
Based on https://github.com/element-hq/element-call/pull/3856

This converts the footer to storybook.
 - Introduce proper loudspeaker button in the bottom bar (for mobile if there is a loudspeaker)
 - create new file for the footer (and stories) move logic out of InCallView
 - move related styles to InCallFooter.module.css
 - Add a second settings button to the bottom bar. Toggle between the two (`display: none/inherit`) when switching between small screens and large screens.
 Screenshot 2026-04-10 at 17 59 40 Screenshot 2026-04-10 at 18 04 38
 - improve when we hide the logo (hide logo on narrow screens so it does not collide with the settings button + also hide logo in all embedded cases)

Additionally this does some CI/testing work:
 - fix a race in the sticky one on one playwright tests
 - add a test to the incallview to make sure there is always one setting button available (for layouts where the app_bar is used and where it isnt (and for the special case where the app bar is configured but the layout is in flat mode (phone landscape) where we also hide the appbar)
Screenshot 2026-04-14 at 14 37 50

The footer could get its own view model to even further move this to our desired architecture.
This PR does intentionally not try this since it needs more discussion and documentation on what approach we want to take.

 - [x] **TODO**
  Find a solution for mobile landscape. We have header set to appbar. But we do not have it shown. So we should set hideSettings to false. **the settings button now dynamically moves into footer when the header bar is not shown** There are dedicated tests for this in the InCallView


src/button/ReactionToggleButton.test.tsx
```diff
@@ -37,7 +37,7 @@ function TestComponent({
         vm={vm}
         rtcSession={rtcSession.asMockedSession()}
       >
-        <ReactionToggleButton vm={vm} identifier={localIdent} />
+        <ReactionToggleButton reactionData={vm} identifier={localIdent} />
```
BillCarsonFr: It is strange to name that a `XXXData` but then to pass it a viewModel.
Maybe you can create a new ReactionButtonViewModel? (can be constructed from the parent view model).
I am not a super fan of the Pick<> thing, l find a bit of a lazy way to do proper abstractions

BillCarsonFr (approved):
I find it quite strange to have changes to unrelated playwright tests in this PR, this could deserve a specific PR with explaination of why it is fixing flakyness.

Great to see the footer as a component and in storybook!

toger5:
> I find it quite strange to have changes to unrelated playwright tests in this PR, this could deserve a specific PR with explaination of why it is fixing flakyness.

I was considering this. But this specific error I have only seen in this PR. It made sense to see if this fixes CI. (after 3 reruns this change made it work)

But now that we have this prove it might make sense to revert this commit and move it into a separate PR.

Note: CI passed after `fix logo in lobby`

src/button/ReactionToggleButton.test.tsx
toger5: Lazy indeed! lets do it properly. Sounds like you are behind the idea of scoping down the VM for the specific component as much as possible.

toger5: I did not do the `ReactionButtonViewModel` yet since we are still in the process of designing a good vm approach.
I am passing the two behaviours manually now. That is already more explicit/less lazy.

src/AppBar.tsx
```diff
@@ -129,6 +132,10 @@ export function useAppBarSecondaryButton(button: ReactNode): void {
     if (setSecondaryButton !== undefined) {
       setSecondaryButton(button);
       return (): void => setSecondaryButton("");
+    } else {
+      logger.warn(
+        "[AppBar] useAppBarSecondaryButton called without AppBarContext provider, this will have no effect",
+      );
```
robintown, line 142: Not sure we should `warn` about this since this branch is designed to be hit on desktop

src/UrlParams.ts
```diff
@@ -618,6 +620,24 @@ export const useRoomIdentifier = (): RoomIdentifier => {
   );
 };
 
+let urlParams$ = undefined as BehaviorSubject<UrlParams> | undefined;
+export const observerUrlParams$ = (
+  scope: ObservableScope,
+): BehaviorSubject<UrlParams> => {
```
robintown: To be honest I have been considering moving URL params in the opposite direction of making them explicitly non-reactive by removing `useUrlParams`. Maybe that influences what you do here.

src/button/Button.tsx
```diff
@@ -130,32 +132,92 @@ export const EndCallButton: FC<EndCallButtonProps> = ({
...
+  return (
+    <Tooltip label={label}>
+      <CpdButton
+        iconOnly
+        Icon={VolumeOnSolidIcon}
+        {...props}
+        kind={enabled ? "primary" : "secondary"}
+        role="switch"
+        aria-checked={enabled}
+      />
```
robintown: The combination of `role="switch"` and a changing `label` is likely to be confusing to screen reader users as they would hear something like "Handset - Off" when in handset mode, which is the opposite of what we need to communicate

src/button/Button.tsx
```diff
@@ -130,32 +132,92 @@ export const EndCallButton: FC<EndCallButtonProps> = ({
...
+  const label = props.isEarpieceTarget
+    ? t("settings.devices.handset")
+    : t("settings.devices.loudspeaker");
+  // if the target is the earpice, we are currently in loudspeaker mode.
+  const enabled = props.isEarpieceTarget;
+  return (
+    <Tooltip label={label}>
+      <CpdButton
+        iconOnly
+        Icon={VolumeOnSolidIcon}
```
robintown: This icon should change when the loudspeaker is off

src/button/Button.tsx
```diff
@@ -130,32 +132,92 @@ export const EndCallButton: FC<EndCallButtonProps> = ({
...
+        {...props}
+        kind={enabled ? "primary" : "secondary"}
+        role="switch"
+        aria-checked={enabled}
+      />
+    </Tooltip>
+  );
+};
+
+function classNamesForScrrenWidth(
```
robintown:
```suggestion
function classNamesForScreenWidth(
```

src/button/Button.tsx
```diff
@@ -130,32 +132,92 @@ export const EndCallButton: FC<EndCallButtonProps> = ({
...
+   * true: currently in loudspeaker mode, pressing will switch to earpiece (rendered as enabled)
+   * false: currently in earpiece mode, pressing will switch to loudspeaker (rendered as disabled)
+   */
+  isEarpieceTarget: boolean;
+}
+export const LoudspeakerButton: FC<LoudspeakerButtonProps> = (props) => {
+  const { t } = useTranslation();
+  const label = props.isEarpieceTarget
+    ? t("settings.devices.handset")
+    : t("settings.devices.loudspeaker");
```
robintown: Are these labels backwards? I think renaming the prop to `enabled` could make things clearer.

src/components/CallFooter.tsx
```diff
@@ -0,0 +1,243 @@
...
+  layoutMode?: GridMode;
+  /** Also controls if the layout button is visible */
+  setLayoutMode?: (mode: GridMode) => void;
+
+  audioEnabled?: boolean;
+  /** Also controls if the audioMute button is disabled */
+  toggleAudio?: () => void;
+  videoEnabled?: boolean;
+  /** Also controls if the videoMute button is disabled */
+  toggleVideo?: () => void;
```
robintown: These should be required properties. Currently the lobby view does not pass `audioEnabled` or `videoEnabled` which causes the buttons to always be in the muted state.

robintown (changes requested):
Haven't reviewed everything but here's what popped out to me so far:

src/components/CallFooter.tsx
toger5: We need to keep the methods optional since they are also used to decide if the buttons are disabled.

robintown: They could be required but of type `boolean | undefined`

src/button/Button.tsx
toger5: Screenshot 2026-04-15 at 16 24 50
Designs for reference.

I am surprised we actually use the crossed icon. On the built in mobile phone app on android it just highlights the speaker.
To me the cross icon resembles "mute".
But this button is a "turn on loudspeaker feature: yes,no" interaction?

We should go with the designs anyways. There might be more to it. (and its more consistent with the other buttons)

@americanrefugee for viz

src/button/Button.tsx
toger5: Depends if we label the target state or the current state (From your question i conclude the current state is expected. I changed it to that now)

Failing to find the Figma ref designs.

@americanrefugee

src/UrlParams.ts
toger5: Interesting. That also does make sense. There is the feature of switching to controls disabled while running a call. But this is not used a lot i think.
We also already lost a bit of the reactivity...

I think I am convinced that this is the better way forward.

I will not do the removal all of the useUrlParam hooks but I will make this behavior static.

src/button/Button.tsx
americanrefugee: I think that the changes I've proposed to the Mic, Camera, and Speaker buttons are causing some headaches. We're currently collecting feedback on whether or not we stick with this change, or revert to the original button styles.

toger5: FIY: we will merge this PR as is.
Once there is a new solution which we prefer we will do a new issue+PR to update the buttons.
Will be a great test to the new storybook component to do some changes on the designs!

src/UrlParams.ts
toger5: This was removed. I updated it so that we just use getUrlParams().

I think the "on the fly updating of the ui" is a bit of a hack. If we really want reactive Layout changes (e.g. show/hide controls) we should add shortcuts or devtools for that.

src/AppBar.tsx
toger5: changed to only log if `platform !== "desktop"`

toger5:
Ugh github scammed me. They were so loud about their new stacked UI experience. I expected this to automatically switch target to livekit once we merge the `[toger5/move-settings-out-of-bottom-bar](https://github.com/element-hq/element-call/tree/toger5/move-settings-out-of-bottom-bar)` branch...
It did that some time back.

I will need to recreate this PR for merging the branch on LK

PR #3851, opened 2026-04-07
Add a shadow to draggable items to prohibit invisible PIP outlines.

toger5:
Fixes: https://github.com/element-hq/voip-internal/issues/516

## Screenshots / GIFs

|Before|After|
|-|-|
|Screenshot 2026-04-07 at 12 59 13|Screenshot 2026-04-07 at 13 00 04|

## Tests

- Step 1
- Step 2
- Step ...
-

## Checklist

- [ ] I have read through [CONTRIBUTING.md](https://github.com/element-hq/element-call/blob/livekit/CONTRIBUTING.md).
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] Tests written for new code (and old code if feasible).
- [ ] Linter and other CI checks pass.
- [ ] I have licensed the changes to Element by completing the [Contributor License Agreement (CLA)](https://cla-assistant.io/element-hq/element-call)

Issue #265, opened 2022-03-08
Configurable noise supression


foxlet:
**Is your feature request related to a problem? Please describe.**
For production and podcast use, a checkbox to disable/enable noise suppression would be beneficial. In our use case we're already using an external program to reduce noise (NoiseTorch), additional noise reduction passes in the browser can degrade audio quality.

**Describe the solution you'd like**
A checkbox to disable noise suppression.

**Additional context**
Jitsi already includes this feature as an URL flag.

Genbuchan:
I've same problem on Discord. If solved this issue I will to encourage switch to Element Call to my friends instead of Discord.

LinuxinaBit:
To my knowledge Element Call doesn’t yet have noise suppression (see https://github.com/element-hq/element-call/issues/714) so I don’t know how useful the ability to disable it is yet…

Sidenote: It has been possible to disable noise suppression in Discord’s audio input/call settings for a while.

xangelix:
> To my knowledge Element Call doesn’t yet have noise suppression (see https://github.com/element-hq/element-call/issues/714) so I don’t know how useful the ability to disable it is yet…
> 
> Sidenote: It has been possible to disable noise suppression in Discord’s audio input/call settings for a while.

I don't have any real evidence to the contrary -- except that I've done testing a few months ago with desktop element call on Linux using it as a loopback adapter for various audio sources. Non-spoken audio like music always had extreme dropouts, that I wouldn't think could be attributable solely to bitrate.

Is there some sort of other audio processing going on other than compression? Because I found the experience with non-speech really quite terrible in quality.

Issue #249, opened 2022-03-05
Configurable media quality


ara4n:
Currently our gUM constraints are hardcoded and very basic (640x360 video; no audio quality controls at all): https://github.com/matrix-org/matrix-js-sdk/blob/96ba061732bc3b5e748719072356d562e52097ba/src/webrtc/mediaHandler.ts#L296

We should let the user pick at least the video resolution & bitrate they want to send (presumably by some sensible defaults).

Also, leaving the bitrate entirely up to the browser feels risky in the extreme.

w3bb:
Framerate is important for gamers, currently there isn't any alternatives to Discord that have framerates high enough to be worthwhile. An option for 30 and 60 FPS would be awesome.

toniocat:
> Currently our gUM constraints are hardcoded and very basic (640x360 video; no audio quality controls at all): https://github.com/matrix-org/matrix-js-sdk/blob/96ba061732bc3b5e748719072356d562e52097ba/src/webrtc/mediaHandler.ts#L296
> 

If element-call is going to be available via Element Synapse server (web browser client) with users **not** installing their own Synapse client, hard-coding limits are preferred.  Perhaps we should have this ability enabled in only certain environments.

terrytw:
I think it is sensible to at least make default resolution and bitrate configurable for homeservers.

darkralts:
> Framerate is important for gamers, currently there isn't any alternatives to Discord that have framerates high enough to be worthwhile. An option for 30 and 60 FPS would be awesome.

i agree, 60fps is very important for me too, i need to watch and screen share movies at 60fps, as lots of media i have is 60fps vs 30, looks choppy on the others end.

unknowntrojan:
The low quality for both audio and video is holding Element back from having more Discord users switch over. I'm not experienced with browser development, but I tried to tinker around with the code and see if I can force the browser to use a higher bitrate and achieve something better than 2kbps dishwater HD. So far I've not managed, but I imagine it can't be that complex, and I'm just missing something. Of course it's not only about the bitrate, being able to choose resolution & framerate is also very important. I'll keep trying, I guess.

a11cf0:
Are there any plans to implement this?

Moresteck:
[...]

PR #3859, opened 2026-04-09
Remove unneeded filtering from screen share audio

JakeTripplJ:
## Content

This PR explicitly disables automatic gain control, noise suppression, and voice isolation effects for screen share audio.

## Motivation and context

Screen share audio currently has signal processing that attempts to suppress all non-voice sounds and whatever else it considers noise. This doesn't make sense for screen share audio, as it removes any ambient noises and makes music sound weird. This PR aims to fix that.

There was an issue/PR that pointed out this problem at one point, or maybe something unrelated contained a discussion about it but I can't find it anymore. If anyone's able to find it and link it, that would be appreciated.

## Tests

- Start a call and listen to a participant's screen share audio while they're playing music or watching a video.

- Current behavior: 
  - Almost all detail from the audio is suppressed except for voices. Ambient sounds or music are considered "noise" and are slowly but surely filtered out.

- PR behavior: 
  - The screen share's music and/or video audio isn't filtered or suppressed anymore, making things sound how they'd be expected to sound.
  
## Checklist

  - Screenshots are inapplicable here.
  - This addition is basically a one-liner and I'm not sure reliable tests can be feasibly made to check perceived audio quality.  


JakeTripplJ:
Does anyone know if the failing test is caused by unlucky/flaky tests or if it's actually from a change I made? I don't understand how the change I made would cause that specific test to fail unless it's some weird Firefox-only weirdness. I thought Firefox wasn't even able to create stream audio in the first place.

ienjoywaffl3s-123:
Thank you for making this! I noticed this issue and was hoping to find a fix. I agree there doesn't seem to be a scenario where users would want these settings when screen sharing.

There are quite a few other issues and PRs that are related to this, but there is a lot of ambiguity on whether the requested video/audio quality settings should apply to the microphone/webcam inputs, screensharing video/audio, or both. I tried skimming over some of these other more complicated PRs and it still wasn't fully clear to me. I also think some users may be confusing the audio filtering in screensharing with low audio bitrate in issue discussions.

The closest issues I found that seem to touch on this are #249 and #265. I can confirm this issue when sharing desktop audio and audio from Chromium tabs, regardless of what existing voice processing settings were toggled in element web.

Hoping this simple fix is merged soon while we wait for those more complex quality setting PRs.

CrazyNicc:
> Does anyone know if the failing test is caused by unlucky/flaky tests or if it's actually from a change I made? I don't understand how the change I made would cause that specific test to fail unless it's some weird Firefox-only weirdness. I thought Firefox wasn't even able to create stream audio in the first place.

hmm you're right, that shouldn't come from this change. Just to be sure though, you should run the playwright tests locally on the `livekit` branch first. If the tests work there though, then this would need some looking into

CrazyNicc:
> Does anyone know if the failing test is caused by unlucky/flaky tests or if it's actually from a change I made? I don't understand how the change I made would cause that specific test to fail unless it's some weird Firefox-only weirdness. I thought Firefox wasn't even able to create stream audio in the first place.

Actually, it looks like the tests broke in #3851, which got fixed in the meantime though. I guess the CI just needs a rerun

JakeTripplJ:
> Thank you for making this! I noticed this issue and was hoping to find a fix. I agree there doesn't seem to be a scenario where users would want these settings when screen sharing.

I agree. I couldn't think of a single situation where someone would ever want these filters applied to their screen share audio. The only situation I can think of where someone would want to change these settings would be if there was some sort of audio pipeline integrated that actively excluded element-call's own audio (like people talking, etc) in the screen share audio. That way, `echoCancellation` could be removed as well as it wouldn't be needed anymore, but that currently isn't the case (and might even have to be included in the logic of whatever would send/split the audio in the first place anyways, completely overriding this).

> There are quite a few other issues and PRs that are related to this, but there is a lot of ambiguity on whether the requested video/audio quality settings should apply to the microphone/webcam inputs, screensharing video/audio, or both.

Yeah, A/B testing this change with the unchanged version seems like a pretty significant change in how screen share audio sounds (especially in the scenarios I listed in the first post). The only "downside" to this approach is that echo cancellation has to be used to remove element-call's own audio, which gives the stream audio a ducking effect when someone talks. It's not really a downside of this PR specifically (because it's present even before this change) but it is something that's still noticeable. I also have no idea if there are any easy ways around that. I'd have to do some more reading and testing to learn about it.

> Hoping this simple fix is merged soon while we wait for those more complex quality setting PRs.

I initially wanted to also make a PR for toggling these effects for voice too, but it seems like there are a few PRs open already dealing with that, so I want to see how those end up before taking a crack at it (if I even have to). I personally think the forced automatic gain control on voice makes having multiple people in a call together unbearable since it can cause weird loud distorted micspam if anyone decides to not talk for a few minutes and the auto-gain gets cranked way up to amplify background noise.

CrazyNicc:
> Does anyone know if the failing test is caused by unlucky/flaky tests or if it's actually from a change I made? I don't understand how the change I made would cause that specific test to fail unless it's some weird Firefox-only weirdness. I thought Firefox wasn't even able to create stream audio in the first place.

Maybe you also need to sync your fork to the `livekit` branch since that should be fixed already

toger5 (approved):
I tested this and the filtering makes it sound noticeable worse.
Thanks for testing/finding this and finding the right place to disable it.

The `SwiftPM` embedded packages don't work if you are creating a PR from a fork. That is a known issue.

I would like to give you the honour of pressing the merge button but I think it might be disabled for you do to the failed SwiftPM situation...

-> I will do the merge an squash the 4 commits.

PR #3860, opened 2026-04-10
Fix a race condition causing the joiner to never publish media to the call

robintown:
Backport of https://github.com/element-hq/element-call/pull/3854 to 0.19.1.


BillCarsonFr (approved):
👍

PR #3871, opened 2026-04-13
Get rid of the 'open in app' mobile prompt

robintown:
We've determined that calls are working fine in mobile browsers these days and we no longer need it.

For https://github.com/element-hq/element-meta/issues/3202


BillCarsonFr (approved):
Ok to waive the quality gate as it fails here because we touch a file (RoomPage) that unfortunatly a zero coverage

PR #3873, opened 2026-04-14
refactor: grug down the connection manager

BillCarsonFr:
## Content

While investigating an issue regarding tracks not beeing created/published, I had a hard time understanding the connection manager code. So I broke it down, extracted some function and remove the generateItems things to do a scan instead.

## Tests

- Step 1
- Step 2
- Step ...
-

## Checklist


src/state/CallViewModel/remoteMembers/ConnectionManager.ts
```diff
@@ -114,150 +126,241 @@ export function createConnectionManager$({
...
+      );
+
+      let addedConnections: Connection[] = [];
+      if (addedTransports.length > 0) {
+        logger.debug("New transports detected", addedTransports);
+
+        addedConnections = addedTransports.map((transport) => {
+          // let's create a connection for each transport
           const connection = connectionFactory.createConnection(
             scope,
```
robintown, line 313: This `scope` now refers to the ConnectionManager's scope, which can live longer than an individual connection - so this likely leads to a resource leak.

If you're set on manually `stop`-ing connections in the connection manager then I suppose you could throw a `new ObservableScope` into the Connection class and `end` it on `stop`. IMO that's a more error-prone pattern (see 717c7420f9a3258c07745ec03820c63f8fbee6ee for context) but if you want to avoid `generateItems` then I think this code is going to be sensitive either way.

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
```diff
@@ -114,150 +126,241 @@ export function createConnectionManager$({
...
-            {
-              type: "livekit",
-              livekit_service_url: serviceUrl,
-            },
+            transport,
             ownMembershipIdentity,
             logger,
-            // TODO: This whole optional SFUConfig parameter is not particularly elegant.
-            // I would like it if connections always fetched the SFUConfig by themselves.
```
robintown, line 198: This comment still feels relevant

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
```diff
@@ -114,150 +126,241 @@ export function createConnectionManager$({
...
     ),
   );
+}
+
+/**
+ * Monitors the list of transports and creates and stops connections accordingly.
+ *
+ * It will automatically:
+ * - Creates new connections when transports are added
+ * - Removes and stops connections when transports are removed;
```
robintown, line 247: If we're going custom here I would like to see a test verifying that all outstanding connections are stopped at the end of the connection manager's scope, since that's one of the trickier things that `generateItems` handles for free

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
```diff
@@ -114,150 +126,241 @@ export function createConnectionManager$({
...
-                data: undefined,
-              };
-            }
+        // stop connections for removed transports
+        removedTransports.forEach((transport) => {
+          const removedCo = state.managedConnections.find((connection) =>
+            areLivekitTransportsEqual(connection.transport, transport),
+          );
+          if (removedCo) {
+            void removedCo.stop();
```
robintown, line 292: Note that Connection is also calling this automatically on the end of its scope. The method is idempotent but this would still double up on "disconnecting from lk room" logs

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
```diff
@@ -114,150 +126,241 @@ export function createConnectionManager$({
...
+};
+/**
+ * All transports currently managed by the ConnectionManager.
+ *
+ * This list does not include duplicate transports.
+ *
+ * It is build based on the list of subscribed transports (`transportsSubscriptions$`).
+ * externally this is modified via `registerTransports()`.
+ */
+function getLocalAndRemoteTransports$(
```
toger5, line 211: I think we never call a derived observable `get` in the current codebase.
`localAndRemoteTransports$()` should also work?

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
```diff
@@ -114,150 +126,241 @@ export function createConnectionManager$({
...
+    localAndRemoteTransports$,
+    ownMembershipIdentity,
+    connectionFactory,
+    logger,
+  );
+
+  const connectionManagerData$ = state$.pipe(
+    switchMap((state) => {
+      // Map each connection to a stream of {connection, participants}
+      const connectionWithParticipants$ = state.managedConnections.map(
```
toger5, line 152: This is not an observable right? its an array so no $

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
```diff
@@ -114,150 +126,241 @@ export function createConnectionManager$({
...
+    managedConnections: [],
+    epoch: -1,
+  };
+
+  return localAndRemoteTransports$.pipe(
+    scan((state: ScannedState, transportsEpoch) => {
+      const transports = transportsEpoch.value;
+
+      // XXX do we need to handle the case where a remote transport is promoted to local?
+      // If so, we could add more info to the state and use that to decide whether to create a new connection or not.
```
toger5, line 269: This is very connected to waht we discussed and what I was doing in the other PR.
I would like to go with: If there is no local transport, we dont yet create any connection. This way we never run into a "promote" case.

robintown (changes requested):

PR #3875, opened 2026-04-14
test: Add some basic test for RoomPage.tsx

BillCarsonFr:
## Content

While reviewing this PR I noticed that there was no test fr RoomPage.tsx. 
So I just gave claude a shot at trying to add some basic tests

## Tests

- Step 1
- Step 2
- Step ...
-

## Checklist

### DIFF
--- a/src/room/RoomPage.test.tsx
+++ b/src/room/RoomPage.test.tsx
@@ -1,0 +1,683 @@
+/*
+Copyright 2026 Element Creations Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+Please see LICENSE in the repository root for full details.
+*/
+
+import {
+  beforeEach,
+  afterEach,
+  describe,
+  expect,
+  test,
+  vi,
+  type Mock,
+} from "vitest";
+import { render, screen, waitFor } from "@testing-library/react";
+import { type MatrixClient, MatrixError } from "matrix-js-sdk";
+import { BrowserRouter } from "react-router-dom";
+import { type FC, type ReactNode } from "react";
+
+import { RoomPage } from "./RoomPage";
+import { ClientContextProvider } from "../ClientContext";
+import { mockConfig, mockMediaDevices } from "../utils/test";
+import * as useRegisterPasswordlessUserModule from "../auth/useRegisterPasswordlessUser";
+import * as UrlParamsModule from "../UrlParams";
+import { HeaderStyle, type UrlParams } from "../UrlParams";
+import * as useLoadGroupCallModule from "./useLoadGroupCall";
+import * as useProfileModule from "../profile/useProfile";
+import * as settingsModule from "../settings/settings";
+import * as MediaDevicesContextModule from "../MediaDevicesContext";
+import * as widgetModule from "../widget";
+import * as ConfigModule from "../config/Config";
+
+// Mock modules
+vi.mock("../auth/useRegisterPasswordlessUser");
+vi.mock("../UrlParams");
+vi.mock("./useLoadGroupCall");
+vi.mock("../profile/useProfile");
+vi.mock("../settings/settings");
+vi.mock("../widget");
+vi.mock("../MediaDevicesContext");
+vi.mock("./GroupCallView", () => ({
+  GroupCallView: vi.fn(() => <div>GroupCallView Mock</div>),
+}));
+vi.mock("./LobbyView", () => ({
+  LobbyView: vi.fn(() => <div>LobbyView Mock</div>),
+}));
+vi.mock("./RoomAuthView", () => ({
+  RoomAuthView: vi.fn(() => <div>RoomAuthView Mock</div>),
+}));
+vi.mock("../home/HomePage", () => ({
+  HomePage: vi.fn(() => <div>HomePage Mock</div>),
+}));
+vi.mock("./AppSelectionModal", () => ({
+  AppSelectionModal: vi.fn(() => <div>AppSelectionModal Mock</div>),
+}));
+
+const mockClient = vi.mocked<MatrixClient>({
+  getUserId: () => "@alice:example.org",
+} as unknown as MatrixClient);
+
+interface WrapperProps {
+  children: ReactNode;
+}
+
+const TestWrapper: FC<WrapperProps> = ({ children }) => {
+  return <BrowserRouter>{children}</BrowserRouter>;
+};
+
+// Helper function to create mock URL params with optional overrides
+// Only includes properties actually used by RoomPage component
+function createMockUrlParams(
+  overrides: Partial<ReturnType<typeof UrlParamsModule.useUrlParams>> = {},
+): ReturnType<typeof UrlParamsModule.useUrlParams> {
+  return {
+    // Properties used by RoomPage
+    confineToRoom: false,
+    preload: false,
+    header: HeaderStyle.Standard,
+    displayName: null,
+    skipLobby: false,
+    callIntent: undefined,
+    ...overrides,
+  } as unknown as UrlParams;
+}
+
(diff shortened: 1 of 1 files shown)


src/room/RoomPage.test.tsx
```diff
@@ -0,0 +1,683 @@
...
+            <RoomPage />
+          </ClientContextProvider>
+        </TestWrapper>,
+      );
+
+      // When not authenticated and no displayName, should show RoomAuthView
+      expect(screen.getByText("RoomAuthView Mock")).toBeInTheDocument();
+    });
+
+    test("shows loading page when registering", async () => {
```
robintown, line 179: This test doesn't do what it says it does (doesn't check for a loading indicator)

src/room/RoomPage.test.tsx
```diff
@@ -0,0 +1,683 @@
...
+    );
+  });
+
+  afterEach(() => {
+    // Reset widget mock to prevent cross-test pollution
+    vi.mocked(widgetModule).widget = null;
+  });
+
+  describe("Loading states", () => {
+    test("shows RoomAuthView when not authenticated without displayName", () => {
```
robintown, line 158: That's not really a loading state

src/room/RoomPage.test.tsx
```diff
@@ -0,0 +1,683 @@
...
+
+      // Should show the call not found error message
+      expect(
+        screen.getByText(/that link doesn't appear to belong/i),
+      ).toBeInTheDocument();
+    });
+  });
+
+  describe("Authentication", () => {
+    test("shows RoomAuthView when not authenticated", () => {
```
robintown, line 271: This is a duplicate of the "shows RoomAuthView when not authenticated without displayName" test

src/room/RoomPage.test.tsx
```diff
@@ -0,0 +1,683 @@
...
+          >
+            <RoomPage />
+          </ClientContextProvider>
+        </TestWrapper>,
+      );
+
+      expect(screen.getByText("RoomAuthView Mock")).toBeInTheDocument();
+    });
+
+    test("auto-registers passwordless user with displayName param", async () => {
```
robintown, line 291: This is a duplicate of the "shows loading page when registering" test

src/room/RoomPage.test.tsx
```diff
@@ -0,0 +1,683 @@
...
+              setClient: vi.fn(),
+              authenticated: undefined,
+            }}
+          >
+            <RoomPage />
+          </ClientContextProvider>
+        </TestWrapper>,
+      );
+
+      expect(mockRegisterPasswordlessUser).not.toHaveBeenCalled();
```
robintown, line 357: This test doesn't do what it says because the registration call happens asynchronously, during an effect (the mock would never be called at this point either way)

src/room/RoomPage.test.tsx
```diff
@@ -0,0 +1,683 @@
...
+
+      vi.mocked(widgetModule).widget = {
+        api: {
+          transport: {
+            send: vi.fn().mockResolvedValue(undefined),
+          },
+        },
+        lazyActions: mockLazyActions,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      } as any;
```
robintown, line 335:
```suggestion
      vi.mocked(widgetModule).widget = {} as unknown as WidgetHelpers;
```

src/room/RoomPage.test.tsx
```diff
@@ -0,0 +1,683 @@
...
+              },
+            }}
+          >
+            <RoomPage />
+          </ClientContextProvider>
+        </TestWrapper>,
+      );
+
+      // The loading state renders with the heading
+      expect(screen.getByRole("heading")).toBeInTheDocument();
```
robintown: Less brittle would be to check for the 'loading' text.

src/room/RoomPage.test.tsx
```diff
@@ -0,0 +1,683 @@
...
+                logout: vi.fn(),
+              },
+            }}
+          >
+            <RoomPage />
+          </ClientContextProvider>
+        </TestWrapper>,
+      );
+
+      expect(setOptInAnalytics).not.toHaveBeenCalled();
```
robintown, line 593: This is another case where the assertion does nothing, as the effect is asynchronous

src/room/RoomPage.test.tsx
```diff
@@ -0,0 +1,683 @@
...
+    });
+  });
+
+  describe("App selection modal", () => {
+    test("shows AppSelectionModal on Android when appPrompt is enabled", () => {
```
robintown, line 598: I just deleted that, so this test can't do what it says.

src/room/RoomPage.test.tsx
```diff
@@ -0,0 +1,683 @@
...
+          </ClientContextProvider>
+        </TestWrapper>,
+      );
+
+      expect(mockRegisterPasswordlessUser).not.toHaveBeenCalled();
+    });
+  });
+
+  describe("Room states", () => {
+    test("shows HomePage when no room is specified", () => {
```
robintown, line 362: The code for this says:

```typescript
// TODO: This doesn't belong here, the app routes need to be reworked
content = <HomePage />;
```

which comes originally from https://github.com/element-hq/element-call/pull/1315. So in other words, this case only exists as a hack to get the user to the home page in case they have a broken /room link. Better would be a test which documents the use case (broken /room link → end up somewhere sensible) without being tied to how this is currently hacked - i.e. by allowing routing actions to happen instead. Sounds more like a Playwright test?

src/room/RoomPage.test.tsx
```diff
@@ -0,0 +1,683 @@
...
+              supportedFeatures: { reactions: true },
+              setClient: vi.fn(),
+              authenticated: {
+                client: mockClient,
+                isPasswordlessUser: false,
+                changePassword: vi.fn(),
+                logout: vi.fn(),
+              },
+            }}
+          >
```
robintown, line 553: All this ClientContextProvider stuff is quite repetitive. Could you store the common values as constants and maybe write a function to remove the JSX boilerplate too? (as in: `renderRoomPage(authedClient);`)

src/room/RoomPage.test.tsx
```diff
@@ -0,0 +1,683 @@
...
+            <RoomPage />
+          </ClientContextProvider>
+        </TestWrapper>,
+      );
+
+      await waitFor(() => {
+        expect(mockRegisterPasswordlessUser).toHaveBeenCalledWith("Bob");
+      });
+    });
```
toger5, line 213: does it actually check if the loading page is shown?

src/room/RoomPage.test.tsx
toger5: Is this actually prove enough that the loading stat e is shown?

toger5 (commented):
I just skimmed through a couple of tests.
Not a full review yet.
They really looks like: lets get coverage up tests.
I am also not sure if a small adaption to the roomPage.tsx might make sense to simplify adding tests?

But its definitly nice to have basic sanity checks in place for the RoomPage.

src/room/RoomPage.test.tsx
BillCarsonFr: @copilot can you look at that and handle it

src/room/RoomPage.test.tsx
BillCarsonFr: @copilot WDYT? if needed update the test

robintown (changes requested):
As you might imagine I'm skeptical of the value of generating tests like this. By the nature of LLMs the output is quite verbose and includes plenty of false reassurances.

BillCarsonFr:
Closing, rejected by reviewers. Claude generated test only add coverage not real value

PR #3877, opened 2026-04-15
Fix race in playwright sticky test

toger5:
I was running into a CI error where `expect(pevaraHasSentStickyEvent).toEqual(true);` was false.
This seemed to be not correct since the logs mention the request that we intercept and should be called.
This change seems to fix it which makes me conclude that the check `expect(pevaraHasSentStickyEvent).toEqual(true);` can be executed to eagerly.

## Tests

- Step 1
- Step 2
- Step ...
-

## Checklist

### DIFF
--- a/playwright/spa-call-sticky.spec.ts
+++ b/playwright/spa-call-sticky.spec.ts
@@ -28,7 +28,7 @@ async function setupTwoUserSpaCall(
   await page.goto("/");
 
   let androlHasSentStickyEvent = false;
-
+  const androlResolver = Promise.withResolvers<void>();
   await interceptEventSend(
     page,
     // This room is not encrypted, so the event is sent in clear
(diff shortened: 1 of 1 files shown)


playwright/spa-call-sticky.spec.ts
```diff
@@ -53,13 +54,15 @@ async function setupTwoUserSpaCall(
 
   let pevaraHasSentStickyEvent = false;
 
+  const paveraResolver = Promise.withResolvers<void>();
```
BillCarsonFr: it is pevara. pevara sedai ;)

robintown (approved):
TIL that [non-retrying assertions](https://playwright.dev/docs/test-assertions#non-retrying-assertions) are a common source of flakes in Playwright. Would be nice to audit our tests for other instances of those.

PR #3885, opened 2026-04-16
[Corrected merge target] Footer component -> Storybook 

toger5:
This is a duplicate of https://github.com/element-hq/element-call/pull/3857 but with the correct target branch.


BillCarsonFr (approved):
Assuming I already reviewed that here https://github.com/element-hq/element-call/pull/3857 I approve

PR #3887, opened 2026-04-16
Transitions Element Call from yarn to pnpm.

### LINKED ISSUE #3884: Bump matrix js-sdk to support latest MSC4195 update
### Your use case

Bump matrix js-sdk to support https://github.com/matrix-org/matrix-js-sdk/pull/5268

_No response_

### Additional context

_No response_

fkwp:
Fixes: https://github.com/element-hq/element-call/issues/3884

## Content

This PR transitions Element Call from Yarn to pnpm.

## Motivation and context

The migration is primarily driven by the need to update matrix-js-sdk to support the latest MSC4195 changes. Since the matrix-js-sdk project recently moved to pnpm, Yarn can no longer correctly resolve or point to specific commits within matrix-js-sdk. 

Switching to pnpm is therefore required to stay up to date with the matrix-js-sdk.

## Note

The playwright tests are currently failing due to lk-jwt-service not being supporting


package.json
```diff
@@ -110,7 +110,7 @@
     "livekit-client": "^2.18.1",
     "lodash-es": "^4.17.21",
     "loglevel": "^1.9.1",
-    "matrix-js-sdk": "matrix-org/matrix-js-sdk#6e3efef0c5f660df47cf00874927dec1c75cc3cf",
+    "matrix-js-sdk": "matrix-org/matrix-js-sdk#ca5655bced1cbb215ac1c18e179384a069d98b0a",
```
robintown: While you're at it, could you please get this pointing back to a branch name rather than a commit hash? It'll still lock to the same commit but that would make it easier to see what sort of matrix-js-sdk version we're using.

vitest.setup.ts
robintown, line 1: Please put test setup code into the existing src/vitest.setup.ts file - but also, why do we need this? Doesn't jsdom implement storage APIs?

.npmrc
```diff
@@ -0,0 +1 @@
+shamefully-hoist=true
```
robintown: Could this get a comment as to broadly what breaks without hoisting so we have an idea of how to get this fixed?

robintown (commented):
With the linker plugin gone, I'd like to still review that linking matrix-js-sdk works as expected (Element Web chose to implement a pnpm linking script of its own, so maybe we also still need something custom like that)

vitest.setup.ts
fkwp: > Doesn't jsdom implement storage APIs?

seems that is not working with node 25

package.json
fkwp: addressed with https://github.com/element-hq/element-call/pull/3887/commits/938a03d0c1e05267b2af4326f073aaa91a5c5a6f

vitest.setup.ts
fkwp: addressed with https://github.com/element-hq/element-call/pull/3887/commits/938a03d0c1e05267b2af4326f073aaa91a5c5a6f

.npmrc
fkwp: I managed to find a way to get rid of that mode: https://github.com/element-hq/element-call/pull/3887/commits/d75f2563c6fa2a8baf9eaae05c2c753bdb19f00d

vite.config.ts
```diff
@@ -150,12 +156,5 @@ export default ({
         "@radix-ui/react-dismissable-layer",
       ],
     },
-    // Vite is using esbuild in development mode, which doesn't work with the wasm loader
-    // in matrix-sdk-crypto-wasm, so we need to exclude it here. This doesn't affect the
-    // production build (which uses rollup) which still works as expected.
-    // https://vite.dev/guide/why.html#why-not-bundle-with-esbuild
-    optimizeDeps: {
-      exclude: ["@matrix-org/matrix-sdk-crypto-wasm"],
-    },
```
robintown, line 159: Can you confirm the app is still working in development mode (`yarn dev`) after removing this?

robintown (approved):

vite.config.ts
fkwp: yes, I can confirm.

toger5 (approved):

toger5:
Confirmed that linking works as before using pnpm links:enable + pnpm links:disalbe BE AWARE that linking.md was updated and we use a different format for the links.yaml (now links.cjs)!

PR #3888, opened 2026-04-16
[Android Audio Control] Reapply preferred device if the OS decides to update the default.

toger5:
## Content
Element X has a custom behaviour we missed:

```kotlin
 /**
 * Selects the default audio device based on the sorted available devices.
 *
 * @param availableDevices The list of available audio devices to select from. If not provided, it will use the current list of audio devices.
 */
private fun selectDefaultAudioDevice(availableDevices: List<AudioDeviceInfo> = listAudioDevices()) {
    val selectedDevice = availableDevices.firstOrNull()
    expectedNewCommunicationDeviceId = selectedDevice?.id
    audioManager.selectAudioDevice(selectedDevice)

    selectedDevice?.let {
        updateSelectedAudioDeviceInWebView(it.id.toString())
    } ?: run {
        Timber.w("Audio: unable to select default audio device")
    }
}

```

```kotlin
override fun onAudioDevicesAdded(addedDevices: Array<out AudioDeviceInfo>?) {
    val validNewDevices = addedDevices.orEmpty().filter { it.type in wantedDeviceTypes && it.isSink }
    if (validNewDevices.isEmpty()) return

    // We need to calculate the available devices ourselves, since calling `listAudioDevices` will return an outdated list
    val audioDevices = (listAudioDevices() + validNewDevices).distinctBy { it.id }.sortedWith(audioDeviceComparator)
    setAvailableAudioDevices(audioDevices.map(SerializableAudioDevice::fromAudioDeviceInfo))
    // This should automatically switch to a new device if it has a higher priority than the current one
    // selectDefaultAudioDevice(audioDevices)
    // This is now done on EC
    // Android just provides the list and we wait until we know what EC wants to use!
}

override fun onAudioDevicesRemoved(removedDevices: Array<out AudioDeviceInfo>?) {
    // Update the available devices
    setAvailableAudioDevices()

    // Unless the removed device is the current one, we don't need to do anything else
    val removedCurrentDevice = removedDevices.orEmpty().any { it.id == currentDeviceId }
    if (!removedCurrentDevice) return

    val previousDevice = previousSelectedDevice
    if (previousDevice != null) {
        previousSelectedDevice = null
        // If we have a previous device, we should select it again
        audioManager.selectAudioDevice(previousDevice.id.toString())
    } else {
        // If we don't have a previous device, we should select the default one
        // selectDefaultAudioDevice()
        // // Android just provides the list and we wait until we know what EC wants to use!
    }
}
```

TLDR: On change of the available devices we update the default device. Which calls: `setAudioDevice` which updates `outputDevice$` which we intentionally ignore on android since we expected it to not change.
But since android might mess with this value (which we probably don't want in the long run)

This PR tackles this from the EC side by emitting the device Id we want after the native OS did a switch.

### DIFF
--- a/src/controls.ts
+++ b/src/controls.ts
@@ -158,12 +158,22 @@ window.controls = {
 
   // wrappers for the deprecated controls fields
   setOutputEnabled(enabled: boolean): void {
+    logger.info(
+      `[MediaDevices controls] setOutputEnabled DEPRECATED method called (enabled: ${enabled}) from native -> forwarding to setAudioEnabled.`,
+    );
     this.setAudioEnabled(enabled);
   },
   setAvailableOutputDevices(devices: OutputDevice[]): void {
+    logger.info(
+      "[MediaDevices controls] setAvailableOutputDevices DEPRECATED method called from native -> forwarding to setAvailableAudioDevices",
+      devices,
+    );
     this.setAvailableAudioDevices(devices);
   },
   setOutputDevice(id: string): void {
+    logger.info(
+      `[MediaDevices controls] setOutputDevice DEPRECATED method called (with id: ${id}) from native -> forwarding to setAudioDevice`,
+    );
     this.setAudioDevice(id);
   },
 };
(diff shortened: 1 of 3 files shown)


toger5:
@BillCarsonFr Closing this in favor of (@BillCarsonFr Can you add the correct PR hash here?)

BillCarsonFr:
> @BillCarsonFr Closing this in favor of (@BillCarsonFr Can you add the correct PR hash here?)

https://github.com/element-hq/element-x-android/pull/6609

Issue #3485, opened 2025-09-10
Multiple Screenshares: When exiting "maximise mode", the shown screenshare switches back to the first one


bblacher:
### Steps to reproduce

1. Have an ongoing call with multiple screenshares
2. Switch to the second screenshare
3. Maximize the screenshare
4. Exit maximize mode

#### What did you expect?
The selected screenshare should be minimized

#### What happened instead?
The shown screenshare switched back to the first screenshare.

Happens on 0.15.0 and previous versions.

bblacher:
Probably related to https://github.com/element-hq/element-call/issues/3484

bblacher:
Another way to reproduce this:

- Switch to the second screenshare
- Switch to a different room in element web, so the screen share gets shown in PiP.
- Observe how the PiP shows the first screenshare. (Often my own)

PR #3889, opened 2026-04-17
Fix visual glitches with multiple screen shares

### LINKED ISSUE #3483: Multiple Screenshares: Left/Right Button not visible

### LINKED ISSUE #3829: screen share labels of multiple users visible at the same time
### Steps to reproduce

1. 2 simultaneous screen shares
2. switch between shares

#### What did you expect?

view 1 name label at a time

#### What happened instead?

Image

### Operating system

_No response_

### Browser information

ungoogled chromium 144.0.7559.132-2

### URL for webapp

call.element.io

### Will you send logs?

No

robintown:
Evidence of it working as intended:

[Screencast From 2026-04-17 12-31-37.webm](https://github.com/user-attachments/assets/f4d27046-7605-4ca2-a96b-cf049f58568c)

See commits for details. No test provided as this would require screenshot tests or similar.

Closes https://github.com/element-hq/element-call/issues/3829
Closes https://github.com/element-hq/element-call/issues/3483

### DIFF
--- a/src/tile/MediaView.module.css
+++ b/src/tile/MediaView.module.css
@@ -11,6 +11,9 @@ Please see LICENSE in the repository root for full details.
   border-radius: var(--media-view-border-radius);
   /*Add spacing for the hover boarder. Otherwise it might get clipped.*/
   margin: var(--hover-space-margin);
+  display: grid;
+  grid-template-areas: "content";
+  place-items: stretch;
 }
 
 .media video {
@@ -37,6 +40,7 @@ Please see LICENSE in the repository root for full details.
 }
 
 .bg {
+  grid-area: content;
   background-color: var(--video-tile-background);
   inline-size: 100%;
   block-size: 100%;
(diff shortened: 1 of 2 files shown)


bblacher:
Does this also fix https://github.com/element-hq/element-call/issues/3485?
It seems to be kind of related.

robintown:
Nope, that bug is still present on this branch.

BillCarsonFr:
What was the visual glitch?

We have no playwrigth test for screenshares, is it possible to had some? Would it be able to detect the "glitch"?

toger5 (approved):

toger5:
It would be even nicer to have storyboard screenshot tests. I think this is exactly what they are useful for

robintown:
> What was the visual glitch?

See the linked issues and commit messages
> We have no playwrigth test for screenshares, is it possible to had some? Would it be able to detect the "glitch"?

I don't believe so since elements with `opacity: 0;` still count as visible as far as Playwright is concerned. And the name tags were visible but in the wrong place, which is not something you can easily test for.

PR #3893, opened 2026-04-19
Draft: Docs: Add MSC4143 feature to Synapse config

### LINKED ISSUE #3643: MSC4143 endpoint fails on non-FQDN Synapse (/transport returns 404)
### Steps to reproduce

Start with a fresh Element Call deployment connected to a Synapse homeserver.
The homeserver server_name is set to a non-FQDN hostname (e.g., myserver).
Synapse is running latest stable release (e.g., v1.144.0).
Element Call is loaded in a browser or client app.
You can see the Element Call interface ready to log in / make a call.
2. What do you click?
Log in to Element Call with a valid account.
Navigate to the “Start a Call” or “Join a Room” feature.
The client attempts to discover the RTC backend (SFU / LiveKit) automatically.
3. More steps…
The client sends a request to:
https://<server_name>/_matrix/client
MSC4143 discovery is not attempted.
If you try to manually request the MSC4143 transport endpoint:
GET /_matrix/client/unstable/org.matrix.msc4143/transport
Synapse responds with:
{"errcode":"M_UNRECOGNIZED","error":"Unrecognized request"}
6. Notes / Additional info
This is essentially the same problem reported in element-call issue #3566.
Related PRs/issues: #3629 (partial MSC4143 support), #3569, #3612.
Even after upgrading Synapse, MSC4143 discovery does not work.

#### What did you expect?
The client should query the MSC4143 endpoint.
Synapse should return valid transport information JSON.
Element Call should successfully discover the RTC backend without falling back to _matrix/client.
Non-FQDN server_name should not prevent MSC4143 from functioning.
#### What happened instead?
MSC4143 endpoint is unavailable / returns 404 or M_UNRECOGNIZED.
Element Call falls back to _matrix/client.
Calls fail because the client cannot locate the RTC backend.
No configuration option exists in Synapse to enable MSC4143.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

Yes

her001:
## Content

It appears that some clients now fail when MatrixRTC transports endpoint isn't available. In Synapse, this requires enabling the `msc4143_enabled` experimental feature. This change adds the necessary config to the self hosting documentation.

## Motivation and context

After figuring this out on my own, I discovered that there was an issue that describes the same behavior: #3643

It may be that this documentation change is considered a fix to that issue, so in the spirit of the contribution guidelines: Fixes #3643

## Tests

- With a complete configuration which doesn't include `msc4143_enabled`, attempt a call from Element X Android. This fails.
- Set `msc4143_enabled` to `true` and restart Synapse. Attempt a call from Element X Android. This succeeds.

## Checklist

### DIFF
--- a/docs/self-hosting.md
+++ b/docs/self-hosting.md
@@ -45,6 +45,14 @@ experimental_features:
   # MSC4222 needed for syncv2 state_after. This allow clients to
   # correctly track the state of the room.
   msc4222_enabled: true
+  # MSC4143: MatrixRTC transports endpoint. Used by some clients to
+  # discover the MatrixRTC Authorization Service for LiveKit.
+  msc4143_enabled: true
+
+# Configure MatrixRTC Transports
+# Must replace <LIVEKIT_URL> with your server's URL.
+matrix_rtc:
+  transports: [{type: livekit, livekit_service_url: "<LIVEKIT_URL>"}]
 
 # The maximum allowed duration by which sent events can be delayed, as
 # per MSC4140.


her001:
It seems I was premature in this, as now when I have the transports endpoint properly set, Element X Android still works, but Element Web/Desktop fails. Disabling the feature has Element X Android still working, but makes Element Web/Desktop work (at least in more cases). I am marking this as a draft, as it may still be needed later.

PR #3896, opened 2026-04-21
post pnpm transition work

### LINKED ISSUE #3904: Post pnpm transition work (pnpm linking) (aka Fix renovate bot)

toger5:
fixes: https://github.com/element-hq/element-call/issues/3904
This does cleanup from the pnpm transition:
- fixup some missing yarn -> pnpm transitions
- remove pnpmfileChecksum and .pnpmfile.cjs
- update linking realted logic
   - remove post-commit hook
   - remove .links.cjs enable/disable (instead just add/rm .pnpmfile.cjs)
   - rename pnpm links:enable -> pnpm links:on
   - rename pnpm links:disable -> pnpm links:off
   - unify doc filenames `-` -> `_`
   - add linking_concept_reasoning.md to provide background why the
     linking is how it is.    
- update CI to ignore pnpmfile
  - This is save since we never run pnpmfiles in ci
  - We can still run it locally
  - It helps us detect if we accidently commit it anyways.
- pnpm v10.33.0 + lock file

---

Here is a writeup of what the linking has now become and why:

### Why do we not enable .pnpmfile.cjs by default
Background: The presence of the `.pnpmfile.cjs` adds a field to the `pnpm-lock.yaml` called: `pnpmfileChecksum`. This field is a checksum of the content of the `.pnpmfile.cjs` file.
`pnpm install --frozen-lockfile` **fails** if there is a `.pnpmfile.cjs` but no `pnpmfileChecksum` or vice versa (or on mismatch).

_TLDR: running with `--ignore-pnpmfile` will fail if `pnpmfileChecksum` is present._

#### `pnpmfileChecksum` + renovate bot
When the renovate bot creates a PR it runs `pnpm install --ignore-pnpmfile`. This means that the `pnpmfileChecksum` in the lockfile will be **empty**.
This breaks builds that **don't** ignore the `.pnpmfile.cjs`-file. (CI that runs on the renovate PR)
From here we have two possible paths:
 - ignore `.pnpmfile.cjs` in all CI builds CI will also fail if we accidently add it locally.
 - fixup the `pnpm-lock.yaml` in the renovate PR to contain the correct `pnpmfileChecksum`.
 
 Ignoring in all CI builds means that CI will always fail if we enable the linking system.
 This is annoying but can be worked around with the git hook we provide that at least lets us know that we are
 commiting with enabled linking.
 Only if we remember setting it back/disbale linking (or let ourselves remember by the git hook) the CI will work.

 #### Summary
  - We will always run into conflicts with the `pnpmfileChecksum` because in renovate prs it will be empty (`--ignore-pnpmfile`)
  - To keep it simple we set `--ignore-pnpmfile` in all of ours CI to see issues immediately.
  - The only solution is to never have a `.pnpmfile.cjs` in the repository when pushing.
    - This way there will never be a commit with `pnpmfileChecksum` in the lockfile.
    - renovate (which uses `--ignore-pnpmfile` which we cannot disable) and other CI will work
  - We are able to use the linking system locally if we `cp` this file from the scripts folder into `./` on demand.
  - `pnpm links:on` and `pnpm links:off` + `./scripts/setup-linking.sh` will help us with this.


src/livekit/BlurBackgroundTransformer.ts
```diff
@@ -29,7 +29,7 @@ interface WasmFileset {
 // MediaPipe and depend on node_modules having this specific structure. It's
 // easy to see this breaking if our dependencies changed and MediaPipe were
 // no longer hoisted, or if we switched to another dependency loader such as
-// Yarn PnP.
+// pnpm PnP.
```
robintown: pnpm PnP isn't a thing, I'd recommend leaving this comment as-is.

robintown (approved):

src/livekit/BlurBackgroundTransformer.ts
toger5: reverted this

PR #3905, opened 2026-04-21
Add Promise.withResolvers polyfill for older WebViews (Chrome < 119)

manfrommedan:
## Problem

Element Call does not work on devices with WebView based on Chromium < 119 (Huawei, some Xiaomi/Redmi devices). The page shows "Join as guest" instead of the call UI.

The bundled JS uses `Promise.withResolvers()` (ES2024, Chrome 119+) in several places via dependencies (`matrix-js-sdk`, `livekit-client`). On older WebViews the ES module silently fails to load because `<script type="module">` errors are not catchable with `window.onerror`. Element Call never initializes.

Huawei ships a system WebView based on Chromium 114 and does not allow users to change the WebView provider. This affects all current Huawei devices.

`tsconfig.json` targets ES2022 but `Promise.withResolvers` is a runtime API that is not downleveled by TypeScript or esbuild.

## Fix

Add a `Promise.withResolvers` polyfill in `index.html` before the module script tag. The polyfill is guarded by `if (!Promise.withResolvers)` so it has zero impact on modern browsers.

The polyfill must be in `index.html` (not in a `.ts` source file) because the module itself fails to load before any application code runs.

## Affected devices

Any device with system WebView older than Chrome 119. Confirmed on:
- Huawei P40 Lite (WebView 114)
- Huawei (WebView 114.05.302)
- Redmi Note 12 Turbo (WebView 118), see element-hq/element-x-android#6298

Related: element-hq/element-x-android#6632

### DIFF
--- a/index.html
+++ b/index.html
@@ -16,6 +16,19 @@
     <script>
       window.global = window;
     </script>
+    <!-- Polyfill for Chrome < 119 (Huawei WebView, etc.) -->
+    <script>
+      if (!Promise.withResolvers) {
+        Promise.withResolvers = function () {
+          var resolve, reject;
+          var promise = new Promise(function (a, b) {
+            resolve = a;
+            reject = b;
+          });
+          return { promise: promise, resolve: resolve, reject: reject };
+        };
+      }
+    </script>
 
     <% if (packageType === "full") { %>
     <!-- Open graph meta tags -->


toger5:
I am wondering if it makes more sense to use an already existing package to share potential fixes:
https://github.com/es-shims/Promise.withResolvers

manfrommedan:
Thanks for the suggestion, good call to think about this more carefully.

The inline `<script>` in `index.html` was chosen initially as the simplest possible fix: zero dependencies, guaranteed execution order, easy to review and reason about. For a single polyfill it felt like a pragmatic choice. But I agree it's not the most maintainable pattern long-term.

I looked into `es-shims/Promise.withResolvers` before going with the inline approach. A few concerns:

The package pulls in `es-abstract`, `define-properties`, and `globalthis` as transitive dependencies. That's a lot of surface area for what is essentially:

```ts
if (typeof Promise.withResolvers !== 'function') {
  Promise.withResolvers = function <T>() { /* 6 lines */ };
}
```

More importantly, it needs to be imported as a side-effect (`import 'promise.withresolvers/auto'`), and Rollup can be aggressive about dropping side-effect-only imports depending on the `sideEffects` field in `package.json`. It would work today, but it's a subtle contract with the bundler that could break silently on a config change.

A cleaner middle ground would be a local `src/polyfills.ts` imported as the first line in the entry point. This is a well-established pattern (`core-js` recommends the same approach, Angular and CRA both use dedicated polyfill entry files). Rollup preserves side-effect imports from local modules by default, and since it's bundled into the entry chunk, execution order is guaranteed.

This also sets up a clean extension point: if another runtime API needs polyfilling down the line (e.g. `Array.fromAsync`, `Set.prototype.union`), there's already a canonical place for it. Given that `tsconfig.json` currently targets ES2022, this is not unlikely as dependencies adopt newer APIs.

Happy to refactor the PR if you'd prefer this approach. And thanks for pushing for a better solution here.

manfrommedan:
Hey, thanks for the feedback and for maintaining Element Call — it's a great project.

If the inline approach feels too manual, here's an alternative that fits better into the existing Vite plugin stack. Since you already use `vite-plugin-node-polyfills` for Node APIs, the same pattern works for ES runtime APIs:

```ts
// src/vite-plugin-es-compat.ts
import type { HtmlTagDescriptor, Plugin } from "vite";

/**
 * Injects runtime API polyfills as a classic script before ES module evaluation.
 *
 * Unlike syntax transforms (handled by esbuild/tsc), runtime APIs like
 * Promise.withResolvers cannot be downleveled at build time. And because
 * <script type="module"> swallows errors silently on parse/eval failure,
 * these polyfills cannot live in application source — the module never
 * loads far enough to execute them.
 *
 * Classic scripts run before module scripts (HTML spec 8.1.5.1),
 * making head-prepend injection the only reliable delivery mechanism.
 */
export function esCompat(): Plugin {
  return {
    name: "es-compat",
    transformIndexHtml(): HtmlTagDescriptor[] {
      return [
        {
          tag: "script",
          injectTo: "head-prepend",
          children: polyfills
            .map(({ guard, impl }) => `${guard}||(${impl})`)
            .join(";"),
        },
      ];
    },
  };
}

const polyfills = [
  {
    // ES2024 — Chrome 119+, Safari 17.2+
    // Enters bundle via: matrix-js-sdk, livekit-client
    guard: "Promise.withResolvers",
    impl: `Promise.withResolvers=function(){let r,j;return{promise:new this(function(a,b){r=a;j=b}),resolve:r,reject:j}}`,
  },
];
```

Then just one line in `vite.config.ts`:

```ts
import { esCompat } from "./src/vite-plugin-es-compat";
```

This keeps `index.html` clean, uses proper Vite hooks (`transformIndexHtml` + `HtmlTagDescriptor`), and the `polyfills` array is easy to extend if another runtime API sneaks in through dependencies later.

Happy to refactor the PR either way. Just wanted to flag that Huawei users are stuck on WebView 114 with no way to update it, so they're completely locked out of Element Call right now. We've tested the polyfill on affected devices and it fully resolves the issue.

maksimkurb:
Can confirm that this PR fixed calls in element-x-android on my 2 Huawei devices I tested:
https://github.com/element-hq/element-x-android/pull/6640#issuecomment-4363533393

index.html
```diff
@@ -16,6 +16,19 @@
     <script>
       window.global = window;
     </script>
+    <!-- Polyfill for Chrome < 119 (Huawei WebView, etc.) -->
```
toger5, line 19:
```suggestion
    <!-- Polyfill for Chrome < 119 (Huawei WebView, etc. EC breaks on those platforms because they do not support Promise.withResolvers) -->
```

toger5:
Thanks for proposing the alternative. We evaluated it and it is a elegant separated way to solve this. But it might end up confusing contributors and maintainers in the long run.
So I am fine with your initial solution. Its a bit noisy to add things to index.html but very primitive enough so no-one will get confused.

toger5 (approved):

PR #3908, opened 2026-04-22
Allow Android native back gesture/button to close settings in EC (only if settings are closed EC will go into pip)

### LINKED ISSUE #3910: Call controls.onBackButtonPressed on esc if nothing has focus

toger5:
Fixes: https://github.com/element-hq/element-call/issues/3910
see https://github.com/element-hq/element-x-android/pull/6642 for a recording


BillCarsonFr (approved):
Thx! the tests makes it easier to understand 👍 
Maybe the linked issue title could be udpated if we want it to be in the changelog?

toger5:
I think in the logs we render the pr title? I updated that instead.

PR #3913, opened 2026-04-23
[Feature] add grace period before reporting the sync loop as disconnected

fkwp:
If, e.g., the homeserver is restarted due to k8s rescaling the CS API is offline for a short amount of time. Currently this results in the "reconnecting overlay" which distracts the call experience. 

This PR adds a grace period before the sync loop is reported as disconnected. With this PR (initially) the grace period defaults to 10 seconds. 

Going forward with https://github.com/element-hq/element-call/pull/3917 this timeout can be even more relaxed up to 60 seconds.


src/config/ConfigOptions.ts
```diff
@@ -168,5 +175,6 @@ export const DEFAULT_CONFIG: ResolvedConfigOptions = {
   features: {
     feature_use_device_session_member_events: true,
   },
+  sync_disconnect_grace_period_ms: 10000,
```
robintown, line 179: This should be added to the `ResolvedConfigOptions` interface as a non-optional field

config/config.sample.json
```diff
@@ -12,6 +12,7 @@
     "feature_use_device_session_member_events": true
   },
   "ssla": "https://static.element.io/legal/element-software-and-services-license-agreement-uk-1.pdf",
+  "sync_disconnect_grace_period_ms": 10000,
```
robintown: IMO we should leave optional values like this one out of the sample config to ensure that new deployments actually trust the default values in practice. That way we can adjust the default values if we think they need changing and have that actually take effect automatically on more deployments.

src/state/CallViewModel/localMember/HomeserverConnected.ts
```diff
@@ -35,21 +46,40 @@ export interface HomeserverConnected {
...
 export function createHomeserverConnected$(
   scope: ObservableScope,
   client: NodeStyleEventEmitter & Pick<MatrixClient, "getSyncState">,
   matrixRTCSession: NodeStyleEventEmitter &
     Pick<MatrixRTCSession, "membershipStatus" | "probablyLeft">,
+  gracePeriodMs?: number,
 ): HomeserverConnected {
+  // Get grace period from parameter or config (default 10000ms)
+  const graceMs =
+    gracePeriodMs ?? Config.get().sync_disconnect_grace_period_ms ?? 10000;
```
robintown: If you add the option to `ResolvedConfigOptions`, then you can remove the `?? 10000` here

src/state/CallViewModel/localMember/HomeserverConnected.test.ts
```diff
@@ -200,3 +201,103 @@ describe("createHomeserverConnected$", () => {
...
+  it("respects gracePeriodMs: stays true during grace period and flips false after", () => {
+    const hsConnected = createHomeserverConnected$(
+      scope,
+      client,
+      session,
+      GRACE_PERIOD,
+    );
+
+    session.setMembershipStatus(Status.Connected);
+    session.setProbablyLeft(false);
```
robintown: These lines are redundant to the test setup, they can probably be removed from each of the tests

robintown (commented):
The new behavior makes sense. I think [marble testing](https://rxjs.dev/guide/testing/marble-testing) would've been perfect for the tests here, as it ensures that your tests aren't ignoring any intermediate values of the behavior, but it doesn't matter that much.

src/state/CallViewModel/localMember/HomeserverConnected.ts
fkwp: addressed with https://github.com/element-hq/element-call/pull/3913/commits/22f16174efebac62f917ab74589a86f47dea6d83

config/config.sample.json
fkwp: addressed with https://github.com/element-hq/element-call/pull/3913/commits/22f16174efebac62f917ab74589a86f47dea6d83

src/config/ConfigOptions.ts
fkwp: addressed with https://github.com/element-hq/element-call/pull/3913/commits/22f16174efebac62f917ab74589a86f47dea6d83

src/state/CallViewModel/localMember/HomeserverConnected.test.ts
fkwp: addressed with https://github.com/element-hq/element-call/pull/3913/commits/88b5db6675a296e5e7a00de451d51b7022d69dc0


needed also a fix in `const rtsSession$ =`...

fkwp:
> The new behavior makes sense. I think [marble testing](https://rxjs.dev/guide/testing/marble-testing) would've been perfect for the tests here, as it ensures that your tests aren't ignoring any intermediate values of the behavior, but it doesn't matter that much.

marble test is here: https://github.com/element-hq/element-call/pull/3913/commits/a081b40fb23087ac7d7d57f71e23d414fb17b000

robintown (approved):
Hope it's not rude of me to have added this commit on your branch: 7916360b520f1a3752bbbb42cfb8011574d66381
I just noticed that the marble test you added could be using the common `withTestScheduler` helper, and then it seemed easy enough to "compress" all the grace period tests into this form.

PR #3922, opened 2026-04-24
Use Compound Switch component

Half-Shot:
Followup from https://github.com/element-hq/compound-web/pull/482

Using https://compound.element.io/?path=/docs/compound-web_switch--docs, this replaces the LayoutToggle switch with the compound web one.


package.json
```diff
@@ -81,7 +81,7 @@
     "@typescript-eslint/parser": "^8.31.0",
     "@use-gesture/react": "^10.2.11",
     "@vector-im/compound-design-tokens": "^10.0.0",
-    "@vector-im/compound-web": "^9.0.0",
+    "@vector-im/compound-web": "element-hq/compound-web#fc2e677326aaefec61ef74fb1d9de3c01eecfa7e",
```
Half-Shot: Release is coming...

src/components/CallFooter.tsx
```diff
@@ -232,10 +237,17 @@ export const CallFooter: FC<FooterProps> = ({
       </div>
       {!hideControls && <div className={styles.buttons}>{buttons}</div>}
       {setLayoutMode && layoutMode && showLayoutSwitcher && (
-        <LayoutToggle
+        <Switch<"spotlight", "grid">
```
robintown, line 240: TIL that you can put type parameters in JSX like this

robintown (approved):

src/components/CallFooter.tsx
Half-Shot: the syntax is a bit wild :laughing: 

to be honest I was a bit disappointed it didn't figure them out automatically

PR #3927, opened 2026-04-27
Refactor of Escape keyboard event bahvior (allow mute/emoji shortcuts after closing a modal with escape)

toger5:
Fixes:https://github.com/element-hq/element-call/issues/3910
 leveraging the fact, things blocking shortcuts are using react portals.
 
 This refactors the Previous PR to use a better approach.
 
 We use the fact that everything that should block shortcuts is displayed via rect portals. 
 So we check if the focus is on a react portal. And if that is the case, we block keyboard shortcuts.


toger5:
The playwright tests fail due to ignore vs decline?

BillCarsonFr:
> The playwright tests fail due to ignore vs decline?

And other related notification toast change => https://github.com/element-hq/element-call/pull/3936

src/room/LobbyView.tsx
```diff
@@ -91,6 +92,11 @@ export const LobbyView: FC<Props> = ({
   const [settingsModalOpen, setSettingsModalOpen] = useState(false);
   const [settingsTab, setSettingsTab] = useState(defaultSettingsTab);
 
+  // This function incorrectly assumes that there is a camera and microphone, which is not always the case.
```
BillCarsonFr, line 95: maybe it would be better to move that comment inside useCallViewKeyboardShortcuts.ts ?

BillCarsonFr (approved):

src/room/LobbyView.tsx
toger5: yes!

PR #3928, opened 2026-04-27
Add speaker pinned.

eleimt:
## Content

Added the ability to fix the window of a specific speaker.

## Motivation and context

In a joint call, where one of the participants is an app service user who broadcasts a movie or any other video clip (we use such calls to watch movies together) to be able to discuss what is happening, there is a problem with the speaker's screens and the video screen flashing. To solve this problem, the possibility of fixing the required window has been added.

## Screenshots / GIFs

image

## Tests

- Step 1
- Step 2
- Step ...
-

## Checklist


robintown:
@eleimt This sounds like a feature that we are unlikely to accept into the app. My suggestion would be to change the app service to publish its video as a screen share track, then it will automatically take precedence in the spotlight. Does that help you?

eleimt:
> @eleimt This sounds like a feature that we are unlikely to accept into the app. My suggestion would be to change the app service to publish its video as a screen share track, then it will automatically take precedence in the spotlight. Does that help you?

I think so, but I don't know how to do it yet.

eleimt:
> @eleimt This sounds like a feature that we are unlikely to accept into the app. My suggestion would be to change the app service to publish its video as a screen share track, then it will automatically take precedence in the spotlight. Does that help you?

I finally found the time for this. As I understand it, it's about [using these values](https://github.com/livekit/protocol/blob/main/livekit/livekit_models.pb.go#L299-L303). Thank you, it suits me. But I'll leave the PR for the story. Maybe someone will need this functionality to focus on a specific user.

eleimt:
@robintown 
Hi. Can you tell me if it is possible to hide user cards during fullscreen viewing?

image

robintown:
@eleimt That's not currently possible, except in the case of 'local' video tiles (your own video tile), which have an 'Always show' option in the menu. 

Closing this PR by the way since screen share tracks worked for your use case.

PR #3944, opened 2026-05-04
Make controls show up on tap on iOS

### LINKED ISSUE #3943: Cannot tap to show controls in landscape on mobile Safari
### Steps to reproduce

1. Join a call with mobile Safari or Element X iOS
2. Rotate the phone into landscape
3. Tap the screen

#### What did you expect?

Footer controls (buttons for audio/video mute, reactions, hanging up) should appear

#### What happened instead?

Nothing

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

robintown:
As documented at https://www.quirksmode.org/blog/archives/2014/02/mouse_event_bub.html, the fact that click events do not bubble by default is a longstanding quirk of iOS Safari.

Assuming the hack works, this closes https://github.com/element-hq/element-call/issues/3943

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -125,6 +125,12 @@ body[data-platform="ios"] {
     -apple-system, BlinkMacSystemFont, "Inter", sans-serif;
 }
 
+body[data-platform="ios"] * {
+  /* This makes click events bubble properly on iOS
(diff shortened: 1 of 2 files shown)


robintown:
Well, this doesn't actually fix anything. (Valere and Florian both tested)

PR #3946, opened 2026-05-05
[Feature] Support delayed event delegation for legacy JWT request

fkwp:
If, e.g., the homeserver is restarted due to k8s rescaling the CS API is offline for a short amount of time. If in this period an Element Call client disconnects (closing browser tab, network issues, etc.) the belonging MatrixRTC membership is not cleaned up. This resolves once the delayed event with the leave event is sent. Typically the delayed event timeout is in the order of 30-90 seconds, which is perceived as a long period of time.

This PR
- adds the ability to add the delayed event delegation information also to the legacy JWT request, which reduces the emission of the delayed leave event to a minimum.
- Increases the `maximumNetworkErrorRetryCount` to NOT fail during (the increased) timeouts. 

Note: The deployed MatrixRTC stack requires a MatrixRTC Authorization Service (lk-jwt-service) to support the delegation feature, which is currently a PR.

### DIFF
--- a/src/config/ConfigOptions.ts
+++ b/src/config/ConfigOptions.ts
@@ -164,6 +164,14 @@ export interface ResolvedConfigOptions extends ConfigOptions {
   };
   sync_disconnect_grace_period_ms: number;
   ssla: string;
+  matrix_rtc_session: {
+    wait_for_key_rotation_ms?: number;
+    delayed_leave_event_delay_ms: number;
+    delayed_leave_event_restart_local_timeout_ms?: number;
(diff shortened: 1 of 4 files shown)


BillCarsonFr (approved):
Looks like this is doing a bit more that delegation? The js-sdk thing about maximumNetworkErrorRetryCount?

fkwp:
> Looks like this is doing a bit more that delegation? The js-sdk thing about maximumNetworkErrorRetryCount?

fair -- however, its very related.

PR #3947, opened 2026-05-05
[fix] nginx config as the missing 50x.html escalates 502 errors to 404

fkwp:
Fix dev nginx config.

### DIFF
--- a/backend/dev_nginx.conf
+++ b/backend/dev_nginx.conf
@@ -42,8 +42,6 @@ server {
         proxy_set_header Host $host;
     }
 
-    error_page   500 502 503 504  /50x.html;
-
 }
 
 # Synapse reverse proxy including .well-known/matrix/client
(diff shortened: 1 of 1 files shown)


toger5:
Are we sure we want to remove all errors?
Can the PR get more context (I am not exactly sure anymore what the issue was? I think we use the errors for a specific check of the jwt to know what is supported?)

toger5 (approved):

fkwp:
> Are we sure we want to remove all errors? Can the PR get more context (I am not exactly sure anymore what the issue was? I think we use the errors for a specific check of the jwt to know what is supported?)

This PR is NOT removing the errors. It's just not presenting an extra error html page. Furthermore, the missing redirected html page is hiding the, e.g., 502 error and instead reporting a 404 -- which is worse.

PR #3953, opened 2026-05-07
Posthog add reconnect event

fkwp:
This PR adds a `CallReconnecting` which is sent every time Element Call is in reconnecting state. The reconnect reason is also provided (syncing, membershipConnected, certainlyConnected, livekit) alongside the reconnection duration.

The `CallEnded` Event is extended by a reconnection summary: callReconnectingCount, callReconnectingCountSyncing, callReconnectingCountMembershipConnected, callReconnectingCountCertainlyConnected, callReconnectingCountLivekit


src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -519,6 +524,22 @@ export const createLocalMembership$ = ({
...
+      filter(Boolean),
+      withLatestFrom(
+        homeserverConnected.disconnectReason$,
+        localConnectionState$,
+      ),
+      scope.bind(),
+    )
+    .subscribe(([_, homeserverReason]) => {
+      const reason = homeserverReason !== null ? homeserverReason : "livekit";
+      PosthogAnalytics.instance.eventCallReconnecting.track(callId, reason);
```
BillCarsonFr: Maybe we could have a single entry point `trackCallReconnection()` that will do all the things? instead of having to do to consecutive calls? (cacheReconnecting could be private in analyticss?(

BillCarsonFr (approved):

src/state/CallViewModel/localMember/LocalMember.ts
fkwp: See https://github.com/element-hq/element-call/pull/3953/commits/7cedc308cad16fec84fd571c6d0f0559a1870b60

src/analytics/PosthogAnalytics.ts
```diff
@@ -421,4 +423,13 @@ export class PosthogAnalytics {
...
   public eventCallConnectDuration = new CallConnectDurationTracker();
+  private readonly _eventCallReconnecting = new CallReconnectingTracker();
+
+  public trackCallReconnecting(
+    callId: string,
+    reason: CallReconnectingReason,
+  ): void {
+    this._eventCallReconnecting.track(callId, reason);
+    this.eventCallEnded.cacheReconnecting(reason);
+  }
```
toger5: Here we are introducing a new concept.

(original concept):
```
  PosthogAnalytics.instance.eventCallEnded.cacheStartCall(new Date());
  PosthogAnalytics.instance.eventCallStarted.track(rtcSession.room.roomId);
```
this is an example from how we do it previously.
I am not sure what I prefer but mixing the two is making things more confusing in the future.

(the two concepts:
 - only ever be able to track things via the `public PosthogAnalytics.instance.event...` objects. 
   - Single point where the events are
   - list of events help as docs
   - auto complete based on `PosthogAnalytics.instance.event` + ctrl+enter
   - if two things need to happen: caching + tracking, we need to do this tranparently. (more code chance to forget it)
 - track functions that limit what makes sense to call and private tracker objects
   - less transparent
   - another layer of abstraction by building the methods that call the cache and track methods
   - impossible to mess up by only tracking and forgetting to chache
)

There is also other ways to achieve the "cache + track atomicity".
Once could pass one tracker into the constructor for the other for instance.

But I like those things to be as far up as possible to make them easy to understand.

Since we only have this one place:
```
 .subscribe(([_, homeserverReason]) => {
      const reason = homeserverReason !== null ? homeserverReason : "livekit";
      PosthogAnalytics.instance.trackCallReconnecting(callId, reason);
    });
```
I am tempted to just call cache and track here like we do for start.

But if you think the atomicity is very important we have to move all other places where we use the cache pattern to also be atomic and make the objects private. For this "touch it every couple of month" code consistency is most important. We dont want to spend time thinking which approach to take.

src/analytics/PosthogEvents.ts
```diff
@@ -249,3 +280,25 @@ export class CallConnectDurationTracker {
...
+  | "syncing"
+  | "membershipConnected"
+  | "certainlyConnected"
+  | "livekit";
+
+interface CallReconnecting extends IPosthogEvent {
+  eventName: "CallReconnecting";
+  callId: string;
+  reason: CallReconnectingReason;
+}
```
toger5, line 311: Wondering if the duration of the reconnect is also of interest?
If we end up with a lot of <2s reconnects we are doing sth wrong. where if they are actually proper disconnect for longer durations there is probably an actual network/hardware issue.

toger5 (changes requested):

src/analytics/PosthogEvents.ts
fkwp: good catch

should be addressed by https://github.com/element-hq/element-call/pull/3953/commits/4ffe3a6d2ac38f177826d184947e8c47cfd81aee

src/analytics/PosthogAnalytics.ts
fkwp: https://github.com/element-hq/element-call/pull/3953/commits/4ffe3a6d2ac38f177826d184947e8c47cfd81aee

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -519,6 +523,31 @@ export const createLocalMembership$ = ({
     false,
   );
 
+  let reconnectStart: { time: number; reason: CallReconnectingReason } | null =
+    null;
+  reconnecting$
+    .pipe(distinctUntilChanged(), scope.bind())
+    .subscribe((reconnecting) => {
+      if (reconnecting) {
+        const homeserverReason = homeserverConnected.disconnectReason$.value;
```
toger5: This looks hairy.
I need to look into how `disconnectReason$` is derived. This might be a race betwen `reconnecting$` and `disconnectReason$`

toger5: (The tests seem to be okay but this might be mocking dependent...)

toger5: I think we need to do this differently:
 - `matrixAndLivekitConnected$` depends on `homeserverConnected.combined$` which has the same deps as `homeserverConnected.disconnectReason$`
 - we end up with unreliable updates.

What we should do instead:
 - matrixAndLivekitConnected$ should become a [boolean, reason] (or a similar obj)
 - it should already extract the reason `homeserverConnected.disconnectReason$` (with a combined latest. Or homeserverConnected.combined$ should already include the reason?). 
 - then we have everything at in one obs update for the reconnecting subscription.

toger5 (changes requested):
This needs to be changed: https://github.com/element-hq/element-call/pull/3953/changes#r3217869921

src/state/CallViewModel/localMember/LocalMember.ts
toger5: My preferred solution is `combined$` already including the reason.

fkwp: addressed https://github.com/element-hq/element-call/pull/3953/commits/6c987eb0d4061f83ea4f2c981c33d0f882c61ffe

src/utils/observable.ts
```diff
@@ -116,6 +116,8 @@ export function getValue<T>(state$: Observable<T>): T {
 /**
  * Creates an Observable that has a value of true whenever all its inputs are
  * true.
+ *
+ * @public
```
toger5, line 120: This can go I think

toger5 (approved):

PR #3960, opened 2026-05-11
Use vite storybook plugin for storybook tests

toger5:
run via `pnpm test:storybook`

## Content

This PR adds the vitest-storybook plugin.
This runs playwright and executes interaction tests and runs smoke tests. All this is connected to our codecov test coverage.

## Motivation and context

We would like to work with a stricter separation between viewModel logic and view.
This is a step towards this:
 - the view will only run storybook tests. (smoke tests that we run with all prop configuration that is of interest for us and that we will review in storybook)
 - interaction tests that only check for the most basic callbacks (click here-> callback A gets called. Nothing more)
 
We will then have every logic that goes beyond this in the viewModel which is easier to test view unit tests.
This saves us time from dom checking (`getByRole...`)

## Tests

- Step 1
- Step 2
- Step ...
-

## Checklist


package.json
```diff
@@ -116,6 +121,7 @@
     "normalize.css": "^8.0.1",
     "observable-hooks": "^4.2.3",
     "pako": "^2.0.4",
+    "playwright": "^1.59.1",
```
robintown: Do we actually need both playwright and @playwright/test? (Seems like more opportunity for versions to get out of sync.)

vitest.config.ts
```diff
@@ -1,20 +1,55 @@
...
+              name: "unit",
+              css: {
+                modules: {
+                  classNameStrategy: "non-scoped",
+                },
+              },
+              setupFiles: ["src/vitest.setup.ts"],
+              environment: "jsdom",
+              // an example of file based convention,
+              // you don't have to follow it
```
robintown: Intentional?

vitest.config.ts
```diff
@@ -1,20 +1,55 @@
...
           },
-        },
-        setupFiles: ["src/vitest.setup.ts"],
-        include: ["src/**/*.test.ts", "src/**/*.test.tsx"],
+          {
+            plugins: [
+              storybookTest({
+                // The location of your Storybook config, main.js|ts
+                configDir: path.join(dirname, ".storybook"),
+              }),
```
robintown, line 36: The recommendation from the docs to join with the project's `dirname` strikes me as odd, is that really correct?

.github/workflows/test.yaml
```diff
@@ -2,11 +2,15 @@ name: Test
...
-    branches: [livekit, full-mesh]
+    branches: [livekit]
 jobs:
   vitest:
     name: Run unit tests
     runs-on: ubuntu-latest
+    container:
+      # Make sure to grab the latest version of the Playwright image
+      # https://playwright.dev/docs/docker#pull-the-image
+      image: mcr.microsoft.com/playwright:v1.59.1-noble
```
robintown, line 13: This is another opportunity for Playwright versions to get out of sync, can we avoid the container use? (I'm surprised that this would be necessary here given that the end-to-end tests run fine in CI without a container, as do Element Web's shared-components tests)

toger5: The idea is, that we can skip the `pnpm exec playwright install --with-deps` which takes approx 20-30s

toger5: Its what storybook recommends in their docs.

vitest.config.ts
toger5: It is more resilient to moving things around. But I guess hardcoding it and updating when moving it is fine too.

package.json
toger5: should be fine to just have playwright test

vitest.config.ts
robintown: I see, I misread `dirname` as the function that would turn `/home/robin/element-call/` into simply `element-call` (that's not what it is at all)

.github/workflows/test.yaml
robintown: Makes sense. We could alternatively try Element Web's [Playwright cache action](https://github.com/element-hq/element-web/blob/develop/.github/actions/setup-playwright/action.yml), which then the end-to-end tests would benefit from as well. Not a blocker for this PR.

robintown (approved):

.github/workflows/test.yaml
toger5: Yes that would very much be ideal.

toger5: Created: https://github.com/element-hq/element-call/issues/3981

PR #3961, opened 2026-05-11
Implement fast switcher (+ ViewModel with snapshot example)

toger5:
Fixes: https://github.com/element-hq/voip-internal/issues/564

## Content

This PR has two purposes:
 - add the fast switcher to EC: https://github.com/element-hq/voip-internal/issues/564 (see also [parent issue](https://github.com/element-hq/voip-internal/issues/489) for screenshots of the figma designs
 - Experiment with a snapshot based view model approach

### Diff size
The diff looks a bit unfortunate

Additional stories
```
src/components/CallFooter.stories.tsx‎
+122-23Lines changed: 122 additions & 23 deletions
```
Seperate tests for the view model
```
src/components/CallFooterViewModel.test.ts‎
+133Lines changed: 133 additions & 0 deletions
```
Snapshot data for the Lobby
```
src/room/__snapshots__/LobbyView.test.tsx.snap‎
+950Lines changed: 950 additions & 0 deletions
```
Snapshot tests for the Lobby
```
src/room/LobbyView.test.tsx‎
+100Lines changed: 100 additions & 0 deletions
```

Once we factor out those changes (Especially the lobby tests are more a sidequest)
which combined have a diff of +1305 -23

Makes the overall diff of the actual changes look more like:
+1.837 -226 -> **+532 -203**
## Screenshots / GIFs
Screenshot 2026-05-12 at 12 55 55

## Tests

The testing is an important part this approach tries to experiment with.
It uses: https://github.com/element-hq/element-call/pull/3960 to run storybook tests.
This allows a 3 layered approch:
 - views:
   - views dont do any logic beyond (no stateful logic)
     - if/and/or to render parts of it
     - if/and/or to make things enabled/disabled allow interaction
   - so we only test interactions of all buttons in all states
   - render the view in all states (here we do not even need snapshot tests. We have the storybook to check if what we see is sane.)
 - ViewModels:
   - We try to do a simple view model for each view. Here we have all the stateful view specific logic. 
   - we do only test the outputs of the view model (no js based testing on this level)
 - The actual integrations tests or model tests

With the storybook approach which this PR experiments with we allow ourselves to do those layers more strictly (it was totally possible to do this before but required discipline) Wit the view model interface we make it harder for us to not follow this direction.

## Checklist


src/components/CallFooter.module.css
```diff
@@ -26,10 +26,6 @@ Please see LICENSE in the repository root for full details.
   );
 }
 
-.footer.hidden {
-  display: none;
-}
```
robintown, line 31: What's the reason for removing this?

src/settings/DeviceSelection.tsx
```diff
@@ -37,12 +30,39 @@ interface Props {
...
+      labelText = label.name;
+      break;
+    case "number":
+      labelText = numberedLabel(label.number);
+      break;
+    case "default":
+      labelText =
+        label.name === null
+          ? t("settings.devices.default")
+          : t("settings.devices.default") + " (" + label.name + ")";
```
robintown: IMO this should still use a separate translation key so that localizations are free to present the parentheses and such as the language requires.

Also (tiniest of nitpicks) would be preferable to pass through the `t` from `useTranslation` to this function rather than importing `t` statically, so the component can react to possible language changes (I don't believe you can trigger that in practice currently)

src/room/InCallView.tsx
```diff
@@ -558,9 +544,28 @@ export const InCallView: FC<InCallViewProps> = ({
...
+  );
+  useEffect(() => {
+    const footerScope = new ObservableScope();
+    setFooterVm(
+      createCallFooterViewModel(
+        footerScope,
+        vm,
+        muteStates,
+        mediaDevices,
+        openSettings,
```
robintown: `openSettings` should live in a state holder rather than React, so there's no risk of it changing more often than we want

src/state/ViewModel.ts
```diff
@@ -0,0 +1,47 @@
...
+import { BehaviorSubject } from "rxjs";
+
+import { useBehavior } from "../useBehavior";
+import { type Behavior } from "./Behavior";
+
+export type ViewModel<Snapshot> = {
+  [K in keyof Snapshot as `${string & K}$`]: Behavior<Snapshot[K]>;
+};
+
+export function useViewModel<Snapshot>(vm: ViewModel<Snapshot>): Snapshot {
```
robintown: Suggestion is to delete this helper function because it isn't safe across changes of the `vm` input (the shape of the VM could in fact change and cause React crashes) and doesn't handle static properties

src/components/CallFooter.module.css
toger5: I did not like the pass a proper to hide the component approach. To me it felt cleaner to just nor render it.
(A fade in/out animation would justify this approach however)

toger5: (The footer is now hidden by not getting rendered in the CallView. -> this ended up being unsused

robintown: I see - the tricky part is that it needs to always be in the DOM in the `asOverlay` case, even when hidden, so that it can fade in and out as expected and also accept keyboard-based interaction. Currently on this branch I can't tab into the footer in the spotlight landscape layout.

src/components/CallFooter.stories.tsx
```diff
@@ -5,18 +5,41 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
+};
 
-function CallFooterWrapper(props: FooterProps): ReactNode {
+/**
+ * A wrapper component that is used for:
+ *  - exposing the snapshot via props so the storybook documents the snapshot properties (basically unpack them form the vm)
+ *  - Add additional react context
+ * The paraeters are all params from the FooterSnapshot,
+ * the Snapshot of the vm, the wrapper will create a mocked vm from it and pass it to the CallFooter.
+ * children used for the "Back to Recents" button in the lobby stories, but can be used for anything really
```
robintown:
```suggestion
 * `children` is used for the "Back to Recents" button in the lobby stories, but can be used for anything really.
```

src/components/CallFooter.stories.tsx
```diff
@@ -121,6 +144,51 @@ export const AudioVideoEnabled: Story = {
...
+    ...Default.args,
+    layoutMode: "spotlight",
+  },
+  play: async ({ args, canvasElement }) => {
+    const canvas = within(canvasElement);
+
+    const spotlightRadio = canvas.getByRole("radio", { name: "Grid" });
+    await userEvent.click(spotlightRadio);
+    await expect(args.setLayoutMode).toHaveBeenCalledWith("grid");
+  },
```
robintown, line 207: Are these `play` functions basically becoming our unit tests, with the other PR that you've based this on? (Rather cool if so)

src/components/CallFooterViewModel.test.ts
```diff
@@ -0,0 +1,133 @@
...
+        /* reactionIdentifier */ undefined,
+      );
+
+      expect(vm.audioOptions$?.value).toEqual([]);
+      expect(vm.videoOptions$?.value).toEqual([]);
+    }
+    it("are empty when both the platform is iOS", () => {
+      checkEmptyFor("ios", gridLayout);
+    });
+    it("are empty when both the layout is pip", () => {
```
robintown:
```suggestion
    it("are both empty when the platform is iOS", () => {
      checkEmptyFor("ios", gridLayout);
    });
    it("are both empty when the layout is pip", () => {
```

src/components/CallFooterViewModel.tsx
```diff
@@ -0,0 +1,288 @@
...
+          disable
+            ? constant([] as MenuOptions[])
+            : mediaDevices.audioInput.available$.pipe(
+                map((available) =>
+                  [...available.entries()].map(([id, label]) => ({
+                    id,
+                    label: mediaDeviceLabelToString(
+                      label,
+                      (n) => "Audio Device " + n,
+                    ),
```
robintown: This `"Audio Device " + n` thing isn't localized - IMO the labeling should happen in the view where we have access to `useTranslation`.

src/components/CallFooterViewModel.tsx
```diff
@@ -0,0 +1,288 @@
...
+        switchMap((disable) =>
+          disable
+            ? constant([] as MenuOptions[])
+            : mediaDevices.videoInput.available$.pipe(
+                map((available) =>
+                  [...available.entries()].map(([id, label]) => ({
+                    id,
+                    label: mediaDeviceLabelToString(
+                      label,
+                      (n) => "Camera " + n,
```
robintown: Also not localized

src/components/CallFooterViewModel.tsx
```diff
@@ -0,0 +1,288 @@
...
+          return (option: string) => {
+            if (option === "blur") {
+              backgroundBlurSettings.setValue(!current);
+            } else {
+              mediaDevices.videoInput.select(option);
+            }
+          };
+        }),
+      ),
+    ),
```
robintown: I feel like the background blur toggle and the selection of a media device should be two separate callbacks.

src/components/CallFooterViewModel.tsx
```diff
@@ -0,0 +1,288 @@
...
+          disable
+            ? constant([] as ToggleOption[])
+            : backgroundBlurSettings.value$.pipe(
+                map((blurActive) =>
+                  supportsBackgroundProcessors()
+                    ? [
+                        {
+                          id: "blur",
+                          enabled: blurActive,
+                          label: "Blur Background",
```
robintown: Not localized

src/components/CallFooterViewModel.tsx
```diff
@@ -0,0 +1,288 @@
...
+  return {
+    ...buildMuteBehaviors(scope, muteStates),
+    ...buildDeviceBehaviors(scope, mediaDevices, disableDeviceSwitcher$),
+
+    hideControls$: constant(!showControls),
+    asOverlay$: scope.behavior(
+      callModel.windowMode$.pipe(map((mode) => mode === "flat")),
+    ),
+    buttonSize$: scope.behavior(
+      isPip$.pipe(map((pip) => (pip ? "md" : "lg") as "md" | "lg")),
```
robintown:
```suggestion
      isPip$.pipe(map<boolean, "md" | "lg">((pip) => pip ? "md" : "lg")),
```

(Nit for safety)

src/components/CallFooterViewModel.tsx
```diff
@@ -0,0 +1,288 @@
...
+          ([isPip, showHeader]) =>
+            openSettings !== undefined &&
+            !isPip &&
+            showControls &&
+            !(headerStyle === HeaderStyle.AppBar && showHeader),
+        ),
+      ),
+    ),
+    showLayoutSwitcher$: scope.behavior(
+      isPip$.pipe(map((l) => !isPip$ && showControls)),
```
robintown:
```suggestion
      isPip$.pipe(map((isPip) => !isPip && showControls)),
```

`!isPip$` would always be `false`

src/components/CallFooterViewModel.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+    openSettings$: scope.behavior(
+      callModel.showHeader$.pipe(
+        map((showHeader) =>
+          headerStyle === HeaderStyle.AppBar && showHeader
+            ? undefined
+            : openSettings,
+        ),
+      ),
+    ),
```
robintown: Maybe instead of `showSettingsButton$` and `openSettings$` button duplicating each other's logic, there should just be `openSettings$` which the view uses for both.

src/components/CallFooterViewModel.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+    reactionIdentifier$: constant(reactionIdentifier),
+    reactionData$: constant(
+      reactionIdentifier !== undefined
+        ? {
+            handsRaised$: callModel.handsRaised$,
+            reactions$: callModel.reactions$,
+          }
+        : undefined,
+    ),
```
robintown, line 208: We should still show others' hand raises and reactions even if we can't send reactions ourselves

src/components/MediaMuteAndSwitchButton.tsx
```diff
@@ -81,66 +77,90 @@ export const MediaMuteAndSwitchButton: FC<MediaMuteAndSwitchButtonProps> = ({
...
+      optionsButtonLabel = t("settings.devices.microphone");
+      break;
+    case undefined:
+      IconOptions = undefined;
+      optionsButtonLabel = "undefined";
+      break;
+    default:
+      IconOptions = iconsAndLabels.IconOptions;
+      optionsButtonLabel = iconsAndLabels.optionsButtonLabel;
+      break;
```
robintown: What's the purpose of these extra `undefined` and `default` cases? Better to enforce that `iconsAndLabels` is always of type `"audio" | "video"` IMO.

src/components/MediaMuteAndSwitchButton.tsx
```diff
@@ -81,66 +77,90 @@ export const MediaMuteAndSwitchButton: FC<MediaMuteAndSwitchButtonProps> = ({
...
+            onMuteClick?.();
+            e.preventDefault();
+            e.stopPropagation();
+          }}
+          disabled={onMuteClick === undefined}
+          data-testid="incall_mute"
+        />
+      );
       break;
     default:
```
robintown: Same with this `default` case

src/room/InCallView.tsx
```diff
@@ -570,34 +575,8 @@ export const InCallView: FC<InCallViewProps> = ({
...
-      // If there is no fn the button will be hidden in the footer.
-      openSettings={settingsButtonInAppBar ? undefined : openSettings}
-      hangup={vm.hangup}
-      //Debug props
-      debugTileLayout={debugTileLayout}
-      tileStoreGeneration={tileStoreGeneration}
-    />
+  const footer = footerVm !== null && (
+    <>{showFooter && <CallFooter ref={footerRef} vm={footerVm} />}</>
   );
```
robintown, line 602: Note that this `footerVm !== null` requirement could cause the footer to blip out of existence briefly when joining a call, so that's another reason to make sure we put the initialization of the footer VM somewhere more suitable eventually.

src/room/LobbyView.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  it("renders without header", () => {
+    const { container } = renderLobbyView({ hideHeader: true });
+    expect(container).toMatchSnapshot();
+  });
+
+  it("renders with waiting for invite state", () => {
+    const { container } = renderLobbyView({ waitingForInvite: true });
+    expect(container).toMatchSnapshot();
+  });
```
robintown: Those snapshots are difficult to review, could we get by with a snapshot for only the first test and targeted assertions for the other two? Also, wouldn't hurt to run `expect(await axe(container)).toHaveNoViolations();` when snapshotting.

robintown (commented):
Alright, I've managed to review the full changes now.

src/components/CallFooter.stories.tsx
toger5: Yes exactly. They will run when calling pnpm test and will count for coverage

src/components/CallFooterViewModel.tsx
toger5: Yes it was a consoderation. I was 50/50. So lets do it liek u propose

src/room/InCallView.tsx
toger5: Yes. Lets see what we conclude here. I think for now i might just add it next to the active call code where we initialize the callviewmodel.

src/settings/DeviceSelection.tsx
toger5: This is tricky since the translation now happens in the view model where the. So we would need to do some useObservable trick to create an observable based on the useTranslation.
I will keep it as the static one for now.

src/state/ViewModel.ts
toger5: removed it and updated it to a list of `useBehavior`s
This has the problem that we now have to rely on vm.behavior$ to be a behavior. But our type lets it be `undefined` as well (not `Behavior<undefined>` but actually `undefined`).

I ended up making all props in snapshot non optional. That seemed the most straight forward solution.

src/components/CallFooterViewModel.tsx
toger5: I think we will. the reaction Data is only used in the Footer. So tiles will still show all the inforamtion. The footer will only be used to send. so we need a `reactionIdentifier`. Otherwise it wont work.

src/room/InCallView.tsx
toger5: moved it to the ActiveCall logic

src/components/CallFooterViewModel.tsx
toger5: done in: https://github.com/element-hq/element-call/commit/d5bebcc3a538d4db5e8caf584ab024951e41be79

src/state/ViewModel.ts
```diff
@@ -0,0 +1,49 @@
...
+      snapshot[key],
+    );
+  }
+  return vm;
+}
+
+export function useStaticViewModel<Snapshot>(
+  snapshot: Snapshot,
+): ViewModel<Snapshot> {
+  const [vm] = useState(createStaticViewModel(snapshot));
```
robintown:
```suggestion
  const [vm] = useState(() => createStaticViewModel(snapshot));
```

For performance

robintown (approved):

PR #3966, opened 2026-05-15
Fix tap-to-show call controls on iOS

fkwp:
iOS Safari doesn't reliably fire `click` on plain `<div>`s, so we listen for `pointerup` instead. Scrolls end in `pointercancel`, not `pointerup`, so this still only fires for taps.

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -8,7 +8,6 @@ Please see LICENSE in the repository root for full details.
 import { type MatrixClient, type Room as MatrixRoom } from "matrix-js-sdk";
 import {
   type FC,
-  type MouseEvent as ReactMouseEvent,
   type PointerEvent as ReactPointerEvent,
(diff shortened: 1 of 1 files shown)


robintown (dismissed):
This generally sounds fine as a fix. I will just test in a couple different browsers and layouts before approving.

We used to use a `touchend` rather than `click` handler for this, and that relied on a timer to determine whether the input was actually a tap rather than a long press. Now that I'm thinking about it again, I don't know why we should care that much about the press length. It might still make sense to listen for `touchend` rather than `pointerup`, for simplicity's sake, but also not a big deal.

toger5 (approved):
This is for a follow up discussion/PR:
We might also want to hide the controls on web if the window is very small:

Image

In this case the PointerUp event is helpful as it also allows to show/hide the controls on web.

For the same reason I approve merging with pointUp instead of touchEnd.

PR #3968, opened 2026-05-15
Posthog: drop $initial_person_info from outgoing events

fkwp:
The $initial_person_info super-property is auto-set by posthog-js. Drop it for increased privacy.

### DIFF
--- a/src/analytics/PosthogAnalytics.ts
+++ b/src/analytics/PosthogAnalytics.ts
@@ -69,7 +69,7 @@ export enum RegistrationType {
 // Sanitize URL / referrer / device fields on a single posthog properties bag.
 // Applied to event.properties and to the person-profile bags ($set / $set_once),
 // since posthog mirrors the same URL fields into those.
-function stripPrivacyFields(
+function stripSensitiveFields(
   obj: Properties | undefined,
   anonymity: Anonymity,
 ): void {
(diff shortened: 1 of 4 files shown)


BillCarsonFr (commented):
We could have make our life simpler by making sanitize an public utility method for testing.

By the way the `sanitize_properties` appears to be deprecated https://github.com/PostHog/posthog-js/blob/5b4eda14c99bed660bc31fd863dfcecaa7cde1e4/packages/types/src/posthog-config.ts#L1393

```
/** @deprecated - use `before_send` instead */
```

Do we have a real life capture showing it is still working?

fkwp:
> We could have make our life simpler by making sanitize an public utility method for testing.
> 
> By the way the `sanitize_properties` appears to be deprecated https://github.com/PostHog/posthog-js/blob/5b4eda14c99bed660bc31fd863dfcecaa7cde1e4/packages/types/src/posthog-config.ts#L1393
> 
> ```
> /** @deprecated - use `before_send` instead */
> ```
> 
> Do we have a real life capture showing it is still working?

fair point. bumped posthog and refactoring is here: https://github.com/element-hq/element-call/pull/3968/commits/a08a561c4ccbc052c6a01b893b377c8fa64d9c95

still added an "end-to-end" posthog test to verify `before_send` does actually what we expect it to do.

src/analytics/PosthogAnalytics.ts
```diff
@@ -65,6 +66,68 @@ export enum RegistrationType {
   Registered,
 }
 
+// Sanitize URL / referrer / device fields on a single posthog properties bag.
+// Applied to event.properties and to the person-profile bags ($set / $set_once),
+// since posthog mirrors the same URL fields into those.
+function stripPrivacyFields(
```
toger5:
```suggestion
function stripSensitiveFields(
```

src/analytics/PosthogAnalytics.ts
```diff
@@ -65,6 +66,68 @@ export enum RegistrationType {
...
+  anonymity: Anonymity,
+): void {
+  if (!obj) return;
+
+  if (anonymity === Anonymity.Anonymous) {
+    // drop referrer information for anonymous users
+    obj["$referrer"] = null;
+    obj["$referring_domain"] = null;
+    obj["$initial_referrer"] = null;
+    obj["$initial_referring_domain"] = null;
```
toger5: wondering if `delete` also makes more sense here? Can we do experiements what posthog will do (if null !== missing field?)

toger5 (approved):
Lets wait for the merge until we see a posthog event form this PR in the dashboard.

PR #3978, opened 2026-05-20
Ensure that foreground elements of media tile do not overlap app bar

### LINKED ISSUE #3976: 'Calling' indicator overlaps the status bar / header on mobile
### Steps to reproduce

1. Start DM call on Element X Android/iOS

#### What did you expect?

'Calling' indicator is below the header

#### What happened instead?

'Calling' indicator is in an unreadable position in the system status bar

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

robintown:
Because the type of header that we use in Element X (an 'app bar') lives in a different place in the document than the other headers, it needs a special branch to propagate the right insets.

[Screencast From 2026-05-20 11-47-08.webm](https://github.com/user-attachments/assets/9bfadf59-0f65-4d56-8002-5e02653c9ef1)

Would require screenshot tests.

Closes https://github.com/element-hq/element-call/issues/3976

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -522,12 +522,19 @@ export const InCallView: FC<InCallViewProps> = ({
           // If edge-to-edge, compute new safe area insets that account for the
           // header and footer.
           "--call-view-safe-area-inset-top":
-            edgeToEdge && header && showHeader
-              ? `calc(env(safe-area-inset-top) + ${headerBounds.height}px)`
+            edgeToEdge && headerStyle !== HeaderStyle.None && showHeader
+              ? // Header has two relevant cases: if it's an app bar, it lives
+                // outside the InCallView and consumes the safe area insets
+                // itself. Otherwise account for the safe area and header size
+                // as part of the InCallView.
+                headerStyle === HeaderStyle.AppBar
+                ? `${bounds.top}px`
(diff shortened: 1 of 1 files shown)


toger5 (commented):
Its hard to tell but all this might be a lot easier to read if we use `classNames(`?

robintown:
> Its hard to tell but all this might be a lot easier to read if we use `classNames(`?

@toger5 Not in this case, these are `style` declarations. I could extract it out into some `let`s and `if`s though if that might help?

toger5 (commented):
I think the right thing to do here would be to introduce the `--call-view-safe-area-inset-top` as behaviors in the call view model. Or sth like this:
```
gridInsetStyling$: {viewSafeAreaInsetTop
viewSafeAreaInsetBottom
gridHeight
gridInsetBlockStart}
```
The issue is, that we have
```
  const [headerRef, headerBounds] = useMeasure();
  const [footerRef, footerBounds] = useMeasure();
```
But I am not even sure why they are computed dynamically.
If they would be static css props, we could just have plain css strings and actually move the computation of gridInsetStyling$ into the callViewModel.

I am not a big fan of how this all reads even though I do understand that this complexity is needed.

PR #3979, opened 2026-05-20
Stop the settings button from appearing while footer is fading out

robintown:
The designs actually never want us to show the settings button in the footer if an app bar is in use (as in Element X mobile apps), so just avoid showing it at all in that case.


toger5 (approved):
This makes sense.
Before we went with the logic: the settings button always has to be visible-> `showHeader=false` implies we need to show it in the bottom bar.

> the settings button always has to be visible

is not true anymore.

PR #3984, opened 2026-05-20
Run unit tests in parallel

toger5:
This got removed in the context of adding storybook tests.


robintown:
I see you re-enabled this already in one of the footer PRs.

PR #3995, opened 2026-05-28
Fix play of second leave sound

### LINKED ISSUE #3923: The "leave-dad-dum" sound is played twice when hanging up in a dm
In DM's we have custom logic to autoleave if the other side hangs up. (like an normal telephones)

EC will not consider that case and plays a leave sound for yourself + the other party.

toger5:
fixes: https://github.com/element-hq/element-call/issues/3923


robintown (approved):
That's a funny workaround… perhaps we would combine the two streams of remote & local leave sound effects, ideally, and debounce the result, but given the lifecycle of the CallViewModel that does sound fairly tricky. This looks reasonable.

PR #4000, opened 2026-06-01
Fix corner radius on screenshare overlay in widget mode

Johennes:
## Content

When in widget mode in Element Web and screensharing, the screenshare overlay appears to be missing the corner radius on its tile container which creates an odd box (I think possibly due to the shadow). This change applies the same corner radius to the tile.

I couldn't find any negative side effects of this but it's possible that I've missed something.

There's a separate issue with the corners and black spacing inside the screenshare overlay. This is not addressed in this change.

## Screenshots / GIFs

|Before|After|
|-|-|
|Screenshot 2026-06-01 at 09 07 51|Screenshot 2026-06-01 at 09 08 02|

## Tests

- Step 1
- Step 2
- Step ...
-

## Checklist

### DIFF
--- a/src/grid/TileWrapper.module.css
+++ b/src/grid/TileWrapper.module.css
@@ -7,7 +7,7 @@ Please see LICENSE in the repository root for full details.
 
 .tile.draggable {
   cursor: grab;
-  box-shadow: var(--big-drop-shadow);
+  --draggable-shadow: var(--big-drop-shadow);
 }
 
 .tile.draggable:active {
--- a/src/tile/GridTile.module.css
+++ b/src/tile/GridTile.module.css
@@ -10,6 +10,7 @@ Please see LICENSE in the repository root for full details.
   --hover-space-margin: var(--cpd-space-1x);
   transition: outline-color ease 0.15s;
   outline: var(--cpd-border-width-2) solid rgb(0 0 0 / 0);
+  box-shadow: var(--draggable-shadow);
 }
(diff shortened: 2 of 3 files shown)


robintown (commented):
Oh of course, it's the shadow…

I don't think this fix is quite right since not all tiles are alike (MediaView accepts a variable border radius via the `--media-view-border-radius` property.) Maybe the `box-shadow` that is set on TileWrapper currently ought to be set somewhere else?

Johennes:
> edia-view-border-radius

Hm, right. I missed the usage in `MediaView`. 😕 

Have change things around to instead pass the shadow down from `TileWrapper` to where the corner radius is applied which seems to work.

src/tile/MediaView.module.css
```diff
@@ -9,6 +9,7 @@ Please see LICENSE in the repository root for full details.
   container-name: mediaView;
   container-type: size;
   border-radius: var(--media-view-border-radius);
+  box-shadow: var(--draggable-shadow);
```
robintown: Looks good except maybe it's more appropriate to put this on GridTile? (a SpotlightTile contains 1 or more MediaViews)

robintown (approved):

src/tile/MediaView.module.css
Johennes: Oh, good catch, thank you. Should be fixed now.

PR #4013, opened 2026-06-04
feat(mute): add syncing state and disable toggle during async mute

### LINKED ISSUE #4006: Sometimes need to press button 3 times to unmute
### Steps to reproduce

1. Be in a call
2. Press the mic button to unmute

(unclear if this is actually a client bug or a LK Cloud thing, but we've seen this last week)

#### What did you expect?

It unmutes

#### What happened instead?

I have to re-mute and re-unmute to get myself to actually show up as unmuted

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

_No response_

### Will you send logs?

No

BillCarsonFr:
## Content

Fixes https://github.com/element-hq/element-call/issues/4006

It is not really that you have to click 3 times to unmute, but rather that unmuting is creating a track that can in some case take some time.

We now give proper feedback when unmuting is in progress and disable the button during that period (also shortcuts).

loading_mute

## Tests

- Step 1
- Step 2
- Step ...

## Checklist


src/components/CallFooter.stories.tsx
```diff
@@ -152,6 +154,28 @@ export const WithAudioAndVideoOptions: Story = {
...
 };
+
+export const AudioBusy: Story = {
+  ...Default,
+  args: {
+    ...Default.args,
+    audioEnabled: true,
+    audioBusy: true,
+    videoEnabled: true,
+    videoBusy: false,
```
Johennes: Nit: These seem to be the same values as in `Default`. Maybe we can omit them here? Some for the audio parameters in the `VideoBusy` story.

src/components/CallFooter.tsx
```diff
@@ -180,8 +185,9 @@ export const CallFooter: FC<FooterProps> = ({ ref, children, vm }) => {
         size={buttonSize}
         key="audio"
         enabled={audioEnabled ?? false}
+        busy={audioBusy ?? false}
         onClick={toggleAudio}
```
Johennes, line 189: When `MicButton` is used inside `MediaMuteAndSwitchButton` it does an early return via `if (isBusy) return;`. Should that also happen here? Same question with `VideoButton` further down.

src/components/MediaMuteAndSwitchButton.tsx
```diff
@@ -83,12 +87,14 @@ export const MediaMuteAndSwitchButton: FC<MediaMuteAndSwitchButtonProps> = ({
       button = (
         <VideoButton
           enabled={enabled ?? false}
+          busy={isBusy}
           onClick={(e) => {
+            if (isBusy) return;
```
Johennes: Are these early returns needed because [Compound routes `disabled` into `aria-disabled` only](https://github.com/element-hq/compound-web/blob/main/src/components/Button/UnstyledButton.tsx#L93)? If so, we could maybe add a comment to indicate that this can be cleaned up once Compound has been fixed.

src/state/CallViewModel/localMember/Publisher.ts
```diff
@@ -112,27 +112,49 @@ export class Publisher {
...
-          e,
-        );
-      });
+      const muteState = this.muteStates.audio;
+      // skip this if a sync is in progress: enabled$ still reflects the old
+      // state while the handler is mid-flight, so the handler itself will apply
+      // the correct mute state once it completes.
+      if (!muteState.syncing$.value) {
+        const enabled = muteState.enabled$.value;
+        if (!enabled) {
```
Johennes, line 127: I'm curious why this is guarded with `if (!enabeld)` now while the old code didn't differentiate. Is it because Livekit publishes the streams enabled by default?

Johennes (approved):
I'm not very deep on most of the areas touched here but the diff looks good to me. I left a small number of cosmetic comments.

I have also tested the PR deployment in various constellations and did not find any muting issues.

src/components/MediaMuteAndSwitchButton.tsx
Johennes: Oh actually, ignore that. It looks like `disabled` is not added intentionally and instead the event handlers are omitted.

https://github.com/element-hq/compound-web/blob/main/src/components/Button/UnstyledButton.tsx#L26

I guess the early return should be unnecessary then?

BillCarsonFr: Right, it is unneeded, added a test for the video that is also ensuring the callback is not called https://github.com/element-hq/element-call/pull/4013/changes/acc5a440c7d9255df6f528288cbe613fe783cb86

src/state/CallViewModel/localMember/Publisher.ts
BillCarsonFr: Yeah, I just did it for clarity (hope it is more clear what is the intention of this code now?). This is the `LocalTrackPublished` callback, so the only thing you can do is to disable the track, the other path is no op.

src/components/CallFooter.tsx
BillCarsonFr: Added test to ensure working as expected https://github.com/element-hq/element-call/pull/4013/changes/a430e87104733da629a59da356fd94b2b970029f

src/components/CallFooter.stories.tsx
BillCarsonFr: done

src/state/CallViewModel/localMember/Publisher.ts
Johennes: Ok yeah, sounds good. 👍

PR #4014, opened 2026-06-04
Add `matrix_rtc_mode` config option

fkwp:
Lets deployments pin the MatrixRTC mode (legacy / compatibility / matrix_2_0) via `config.json`, overriding the per-user Developer Settings choice. Invalid values are dropped at load time with a warning, and the DevSettings radios are disabled when the option is set.

### DIFF
--- a/src/config/Config.test.ts
+++ b/src/config/Config.test.ts
@@ -21,16 +21,15 @@ describe("validateConfig", () => {
     expect(result.matrix_rtc_mode).toBeUndefined();
   });
 
-  it.each([
-    MatrixRTCMode.Legacy,
-    MatrixRTCMode.Compatibility,
-    MatrixRTCMode.Matrix_2_0,
-  ])("keeps a valid matrix_rtc_mode value (%s)", (mode) => {
-    const warnSpy = vi.spyOn(logger, "warn").mockImplementation(() => {});
-    const result = validateConfig({ matrix_rtc_mode: mode });
-    expect(result.matrix_rtc_mode).toBe(mode);
-    expect(warnSpy).not.toHaveBeenCalled();
-  });
+  it.each(Object.values(MatrixRTCMode))(
+    "keeps a valid matrix_rtc_mode value (%s)",
+    (mode) => {
+      const warnSpy = vi.spyOn(logger, "warn").mockImplementation(() => {});
+      const result = validateConfig({ matrix_rtc_mode: mode });
+      expect(result.matrix_rtc_mode).toBe(mode);
+      expect(warnSpy).not.toHaveBeenCalled();
+    },
+  );
 
   it("drops an invalid matrix_rtc_mode value and warns", () => {
     const warnSpy = vi.spyOn(logger, "warn").mockImplementation(() => {});
--- a/src/config/Config.test.ts
+++ b/src/config/Config.test.ts
@@ -1,0 +1,55 @@
+/*
+Copyright 2026 New Vector Ltd.
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
+Please see LICENSE in the repository root for full details.
+*/
+
+import { describe, expect, it, vi, afterEach } from "vitest";
+import { logger } from "matrix-js-sdk/lib/logger";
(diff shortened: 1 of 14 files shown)


src/config/Config.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  it("passes through a missing matrix_rtc_mode unchanged", () => {
+    const result = validateConfig({});
+    expect(result.matrix_rtc_mode).toBeUndefined();
+  });
+
+  it.each([
+    MatrixRTCMode.Legacy,
+    MatrixRTCMode.Compatibility,
+    MatrixRTCMode.Matrix_2_0,
+  ])("keeps a valid matrix_rtc_mode value (%s)", (mode) => {
```
Johennes: I think you can use `Object.values` here to avoid having to update this test if the modes are ever changed.

```suggestionit.each(Object.values(MatrixRTCMode))("keeps a valid matrix_rtc_mode value (%s)", (mode) => {
```

src/config/ConfigOptions.ts
```diff
@@ -104,6 +116,17 @@ export interface ConfigOptions {
...
 
+  /**
+   * Forces the MatrixRTC mode for all clients on this deployment.
+   * When set, this overrides any per-user choice from the Developer Settings.
+   * - "legacy"        — legacy single-SFU + user-keyed memberships + legacy JWT endpoint
+   * - "compatibility" — multi-SFU transport, legacy JWT endpoint, no sticky events
+   * - "matrix_2_0"    — multi-SFU transport, sticky events, new JWT endpoint, hashed identity
+   * If unset, the user's Developer Settings choice (or its default of "legacy") is used.
+   * String values must match `MatrixRTCMode` in src/settings/settings.ts.
+   */
```
Johennes, line 131: I don't like that we're duplicating the documentation of the modes here and on the enum. Maybe the enum should be the canonical source and we should just refer to it here?

src/settings/DeveloperSettingsTab.test.tsx
```diff
@@ -311,4 +314,102 @@ describe("DeveloperSettingsTab", () => {
...
+        legacy: screen.getByRole("radio", {
+          name: /Legacy:/,
+        }) as HTMLInputElement,
+        compatibility: screen.getByRole("radio", {
+          name: /Compatibility:/,
+        }) as HTMLInputElement,
+        matrix20: screen.getByRole("radio", {
+          name: /Matrix 2\.0:/,
+        }) as HTMLInputElement,
+      };
```
Johennes, line 339: We might want to look the radios up by value here to prevent depending on the (translatable) display strings.

```suggestion
      return {
        legacy: screen.getByDisplayValue(MatrixRTCMode.Legacy) as HTMLInputElement,
        compatibility: screen.getByDisplayValue(MatrixRTCMode.Compatibility) as HTMLInputElement,
        matrix20: screen.getByDisplayValue(MatrixRTCMode.Matrix_2_0) as HTMLInputElement,
      };
```

src/settings/DeveloperSettingsTab.tsx
Johennes, line 322: It might not hurt to add a short explanation of why the radios are disabled?

```suggestion
      {matrixRTCModeForced && <p>Your deployment overrides the mode.</p>}
      <Form>
```

Johennes (approved):
Some cosmetic comments but otherwise I think this looks ok.

I thought there could be a problem when the deployment forces Matrix 2.0 mode but the homeserver doesn't support sticky events. It looks like in this case the call just doesn't connect. It might be nice to call out the reason more explicitly but I guess it is ok like this.

Image

fkwp:
> I thought there could be a problem when the deployment forces Matrix 2.0 mode but the homeserver doesn't support sticky events. It looks like in this case the call just doesn't connect. It might be nice to call out the reason more explicitly but I guess it is ok like this.

I do agree. I had a look what we could do here:
- just a log warning (not super helpful especially on mobile)
- A new error class

I think especially in the transition phase a new error class might be a good idea.

My proposal is to merge it now as is. And create the error-screen PR afterwards.

PR #4017, opened 2026-06-05
Show typed error when matrix_2_0 mode is forced without homeserver MSC4354 support

fkwp:
Show typed error when matrix_2_0 mode is forced without homeserver MSC4354 support

depends on https://github.com/matrix-org/matrix-js-sdk/pull/5363


src/room/GroupCallView.tsx
```diff
@@ -164,6 +172,28 @@ export const GroupCallView: FC<Props> = ({
     MatrixRTCSessionEvent.MembershipManagerError,
     (error) => setExternalError(new ConnectionLostError()),
   );
+
+  // If the deployment pins matrix_rtc_mode=matrix_2_0 but this homeserver
+  // doesn't advertise MSC4354 (sticky events), the call will fail to connect
+  // with a generic "Connection lost" message. Surface the real cause up front.
+  useEffect(() => {
```
Johennes: Running this as an effect has the downside that you may still get the device permission popup even though the error is shown.

Image

I'm not sure how to fix this without blocking on the `doesServerSupportUnstableFeature` though.

Johennes (commented):
Looks good overall and appears to work. There is one odd permission popup case though.

src/room/GroupCallView.tsx
fkwp: Refactoring that would involve something like extracting the current body of GroupCallView into an inner component (GroupCallViewBody), and make GroupCallView a thin gate that owns the sticky-events check.

not sure if its worth the effort -- as the frustration about not be able to join the call should not be much increased by the pop up... wdyt?

Johennes: Yeah, I suppose it's a relatively minor issue. Should be ok to fix in a later PR also.

Johennes (approved):

locales/en/app.json
```diff
@@ -120,6 +120,8 @@
     "peer_connection_timeout_description": "Connection to the media server timed out. Try switching to a different network or disabling your VPN. If the problem persists, see our <0>troubleshooting guide</0> or contact your server administrator.",
     "room_creation_restricted": "Failed to create call",
     "room_creation_restricted_description": "Call creation might be restricted to authorized users only. Try again later, or contact your server admin if the problem persists.",
+    "sticky_events_required": "Homeserver does not support Matrix 2.0 calls",
```
BillCarsonFr, line 123: This should also be added to localazy so that it can be translated

src/room/GroupCallView.tsx
BillCarsonFr: I don't think this should leave in the UI component.
Won't the join session already Fail? it will give you the error already?

BillCarsonFr (changes requested):
I don't think this is at the right place.
The js-sdk should throw an error if you use an incompatible mode, cannot we jsut check that?
Eventually we could do it earlier in the publisher or localTransport?

locales/en/app.json
fkwp: wdym?

localazy.json auto-picks up everything in locales/en/*.json

So it should be fine @BillCarsonFr

src/room/GroupCallView.tsx
fkwp: should be addressed by https://github.com/element-hq/element-call/pull/4017/commits/49168d01fa84b6b4b45f1c771fc3fefb75e8d352

fkwp:
> I don't think this is at the right place. The js-sdk should throw an error if you use an incompatible mode, cannot we jsut check that? Eventually we could do it earlier in the publisher or localTransport?

It turned out that the error thrown by the SDK is not typed

```
matrix-js-sdk/src/matrixrtc/MembershipManagerActionScheduler.ts:106:

  } catch (e) {
      throw Error(`The MembershipManager shut down because of the end condition: ${e}`);
  }
```

which leads to the following options:
- error message parsing
```
      172 -        error.name === "UnsupportedStickyEventsEndpointError"
      180 +        error.message.includes("UnsupportedStickyEventsEndpointError")
```
- Let the SDK throw a typed Error

Johennes:
Since we can control the SDK and (I think) this PR isn't critical, I'd lean towards changing the JS SDK to throw a typed error. That'll be less brittle than parsing the string.

locales/en/app.json
BillCarsonFr: I don't think it does? or I missed that step

fkwp:
> Since we can control the SDK and (I think) this PR isn't critical, I'd lean towards changing the JS SDK to throw a typed error. That'll be less brittle than parsing the string.

https://github.com/matrix-org/matrix-js-sdk/pull/5363

fkwp:
> I don't think this is at the right place. The js-sdk should throw an error if you use an incompatible mode, cannot we jsut check that? Eventually we could do it earlier in the publisher or localTransport?

- sdk PR https://github.com/matrix-org/matrix-js-sdk/pull/5363
- pulling the PR into EC https://github.com/element-hq/element-call/pull/4017/commits/b6ab43cdd9b6fa5774b0e6e916178ab750e35004
- using the SDK provided cause https://github.com/element-hq/element-call/pull/4017/commits/5d0047aebe105648dba90f440cbd9eb765fe778b

locales/en/app.json
fkwp: that one is pushing the stuff to localazy

```
❯ cat .github/workflows/translations-upload.yaml
name: Upload translation files to Localazy
on:
  push:
    branches:
      - livekit
    paths-ignore:
      - ".github/**"

jobs:
  upload:
    runs-on: ubuntu-latest
    permissions:
      contents: read

    steps:
      - name: Checkout the code
        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
        with:
          persist-credentials: false

      - name: Upload
        uses: localazy/upload@27e6b5c0fddf4551596b42226b1c24124335d24a # v1
        with:
          write_key: ${{ secrets.LOCALAZY_WRITE_KEY }}
```

BillCarsonFr (approved):
Thx I think it looks better now!
Why do we need to add `github:` to the path in the package.json? 
If it is urgent to have it, I guess we can merge, but ideally we want to wait for some js-sdk release/tag

fkwp:
> Why do we need to add `github:` to the path in the package.json?

On `pnpm update`, pnpm canonicalizes the spec to the explicit github

PR #4027, opened 2026-06-08
Remove deprecated @types/uuid

Johennes:
It looks like this shouldn't be needed anymore according to https://www.npmjs.com/package/@types/uuid.


toger5 (approved):
good finding.
Thanks

PR #4029, opened 2026-06-08
Replace deprecated i18next-parser with i18next-cli

Johennes:
i18next-parser seems to be deprecated according to https://www.npmjs.com/package/i18next-parser.


i18next.config.ts
```diff
@@ -0,0 +1,33 @@
...
+  "extract": {
+    "input": [
+      "src/**/*.{ts,tsx}"
+    ],
+    "output": "locales/{{language}}/{{namespace}}.json",
+    "defaultNS": "app",
+    "keySeparator": ".",
+    "nsSeparator": false,
+    "contextSeparator": "|",
+    "extractFromComments": false,
```
Johennes: This (and the `i18nKey` below) are needed because otherwise the `e2ee_encryption_status.*` strings in commented code get extracted.

package.json
```diff
@@ -24,8 +24,8 @@
     "lint:eslint-fix": "eslint --max-warnings 0 src playwright --fix",
     "lint:knip": "knip",
     "lint:types": "tsc",
-    "i18n": "i18next",
-    "i18n:check": "i18next --fail-on-warnings --fail-on-update",
+    "i18n": "npx i18next-cli extract",
+    "i18n:check": "npx i18next-cli extract --ci",
```
Johennes, line 28: This leaves the working directory dirty, unfortunately, but it didn't look like there was any better flags.

toger5 (approved):
The diff in the pnpm-lock looks very large. Is all this really just caused by the switch from parser->cli?

Johennes:
> The diff in the pnpm-lock looks very large. Is all this really just caused by the switch from parser->cli?

Yeah, it's odd but I just reverified it. It's really just `pnpm remove i18next-parser && pnpm add -D i18next-cli`.

Issue #4011, opened 2026-06-04
A videocall preview can be closed to return to chat


PietG1:
### Steps to reproduce

1. Start a videocall preview through a group chat

### Outcome

\> There is no close button. there is a picture in picture button
\> There is a picture in picture button -> Pressing this closes the videocall preview and there is a X but that end also the call

### Will you send logs?

Yes

### Are you willing to provide a PR?

Yes

bmarty:
Transferring the issue to Element Call project.

I guess that in the lobby, the PiP button should be a closed button:

<img width="306" height="661" alt="Image" src="https://github.com/user-attachments/assets/dd698e1f-3a39-4872-a415-16c8b267548d" />

toger5:
fixed by: https://github.com/element-hq/element-call/pull/4032

PR #4032, opened 2026-06-09
Add primary button icon configuration and set it left arrow in lobby

### LINKED ISSUE #3959: Mobile Lobby Back button (instead of minimise)
Currently the lobby does not allow to go into pip `controls.canEnterPip()` is false in the lobby.

Which is sane. No reason to go in pip. The user has to take an action.

Instead the minimise button closes the call.
It should hence be rendered as a back button.

toger5:
Fixes: https://github.com/element-hq/element-call/issues/3959

## Screenshots / GIFs

|Before|After|
|-|-|
|android|-|
|image|image|
|iOS|-|
|same icon as on android before|image|

## Checklist

### DIFF
--- a/src/room/LobbyView.tsx
+++ b/src/room/LobbyView.tsx
@@ -51,11 +51,7 @@ import { CallFooter, type FooterSnapshot } from "../components/CallFooter";
 import { useCallViewKeyboardShortcuts } from "../useCallViewKeyboardShortcuts";
 import { createLobbyFooterViewModel } from "../components/CallFooterViewModel";
 import { type ViewModel } from "../state/ViewModel";
-import {
-  useAppBarHidden,
-  useAppBarPrimaryButtonIconKind,
-  useAppBarTitle,
-} from "../AppBar";
+import { useAppBarPrimaryButtonIconKind } from "../AppBar";
 
 interface Props {
   client: MatrixClient;
@@ -92,8 +88,6 @@ export const LobbyView: FC<Props> = ({
   const { t } = useTranslation();
   usePageTitle(matrixInfo.roomName);
 
-  useAppBarHidden(false);
-  useAppBarTitle(matrixInfo.roomName);
   useAppBarPrimaryButtonIconKind("back");
   const audioEnabled = useBehavior(muteStates.audio.enabled$);
   const videoEnabled = useBehavior(muteStates.video.enabled$);
--- a/src/room/LobbyView.tsx
+++ b/src/room/LobbyView.tsx
@@ -93,7 +93,7 @@ export const LobbyView: FC<Props> = ({
   const { t } = useTranslation();
   usePageTitle(matrixInfo.roomName);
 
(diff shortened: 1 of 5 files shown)


src/room/LobbyView.tsx
```diff
@@ -87,6 +92,9 @@ export const LobbyView: FC<Props> = ({
   const { t } = useTranslation();
   usePageTitle(matrixInfo.roomName);
 
+  useAppBarHidden(false);
+  useAppBarTitle(matrixInfo.roomName);
```
robintown: Not sure this call to `useAppBarTitle` should be here. That hook is also called in GroupCallView and the docs warn against calling it from multiple sites in a component tree.

(Also, `useAppBarHidden(false)` has no effect.)

robintown (commented):
It looks reasonable to do this with a `useAppBarPrimaryButtonKind` hook. But since the behavior of the back button really depends (in a sense) on whether the widget is always-on-screen, have you considered instead directly calling `setPrimaryButtonKind` depending on the outcome of the `widget.api.setAlwaysOnScreen()` calls in GroupCallView?

I'll let you decide whether you find that more robust or not.

src/room/LobbyView.tsx
toger5: `useAppBar*` indeed can all be removed. I had issues initially of the bar not showing which made me add those.

> widget.api.setAlwaysOnScreen() calls in GroupCallView

They indeed are correlating. The hope with the explicit back button was to reduce the logical combinations the code reader has to do. Also I think the setAlways on screen is not used for the mobile pip check. we use the controls api for this (which is a bit redundant thinking about it like this)
So we would end up with a `alwaysOnScreen` which is desktop exclusive to be used for the icon on mobile which uses the controls api for the PiP control.

robintown: I see, too bad that mobile doesn't support always-on-screen actions yet.

robintown (approved):
Assuming you remove the extra hook calls, LGTM

toger5:
oh no waht is wrong with GH:
Screenshot 2026-06-16 at 18 58 23
```
commit b31825d05b7d42774e3cebdfeffde9baa11751a1 (HEAD -> toger5/lobby-remove-minimise-add-back-button, origin/toger5/lobby-remove-minimise-add-back-button)
Author: Timo K <toger5@hotmail.de>
Date:   Tue Jun 16 16:54:14 2026 +0200

    review
```
Screenshot 2026-06-16 at 19 00 44

toger5:
YAY
this commit:
https://github.com/element-hq/element-call/pull/4032/commits/19fd7f42b1dd73a87a0c8c31b50402f53faf035b
let this one appear:
https://github.com/element-hq/element-call/pull/4032/commits/b31825d05b7d42774e3cebdfeffde9baa11751a1

toger5:
Also fixes: https://github.com/element-hq/element-call/issues/4011

PR #4034, opened 2026-06-10
Improve internal model of ringing, expose ringing intent to call UI

robintown:
I found our code's internal model of ringing a little overgrown (it had superfluous states like 'unknown') and difficult to extend with metadata or callbacks relating to ring attempts. By modeling ringing instead as a stream of ring attempts, where each attempt has an intent, a recipient, and an eventual outcome (accept/decline/timeout), I find it more natural to work with.

This makes room for a future 'try again' callback to allow ringing someone again after a timeout, and also forced me to look for a simpler solution to the [duplicate leave sound effects](https://github.com/element-hq/element-call/issues/3923). I exposed the intent of the ringing attempt to the call UI so I can later [use it in the header](https://github.com/element-hq/voip-internal/issues/527).


src/tile/GridTile.tsx
```diff
@@ -89,11 +88,12 @@ const RingingMediaTile: FC<RingingMediaTileProps> = ({
...
           ? {
               text: t("video_tile.calling"),
-              Icon: videoEnabled ? VideoCallSolidIcon : VoiceCallSolidIcon,
+              Icon:
+                vm.intent === "video" ? VideoCallSolidIcon : VoiceCallSolidIcon,
             }
           : { text: t("video_tile.call_ended"), Icon: EndCallIcon }
       }
-      videoEnabled={videoEnabled}
+      videoEnabled={false}
```
Johennes, line 96: I was going to ask if this should be `vm.intent === "video"` but then noticed that it is only `&&`ed with `video` property which is `undefined`. Do we actually still need the separate `videoEnabled` property?

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -634,6 +622,21 @@ export function createCallViewModel$(
     matrixRoomMembers$,
   );
 
+  // ------------------------------------------------------------------------
+  // callLifecycle
+
+  // TODO if we are in "unknown" state we need a loading rendering (or empty screen)
+  // Otherwise it looks like we already connected and only than the ringing starts which is weird.
```
Johennes, line 629:
```suggestion
  // Otherwise it looks like we already connected and only then the ringing starts which is weird.
```

src/state/CallViewModel/CallNotificationLifecycle.ts
```diff
@@ -84,34 +93,81 @@ export interface Props {
...
+      filter(
+        (
+          notificationEvent: CallNotificationWrapper | null,
+        ): notificationEvent is CallNotificationWrapper =>
+          // only care about new events (legacy do not have decline pattern)
+          notificationEvent?.notification_type === "ring" &&
+          notificationEvent.lifetime > 0,
+      ),
+      switchMap((notificationEvent) => {
+        // We assume that there is only one other user in the room when ringing
```
Johennes, line 129: This is just a limitation of our current implementation, right? The notification event should support ringing more than one person due to `m.mentions`?

Johennes (approved):
I'm only grasping the Rx changes at a very high level so you may want to wait for a second review. Left a few minor comments and tested ringing in widget mode successfully.

src/state/CallViewModel/CallNotificationLifecycle.ts
robintown: Correct, we just don't have designs or a use case for ringing >1 person.

src/tile/GridTile.tsx
robintown: Good question, I guess we can consolidate them

src/state/CallViewModel/CallViewModel.ts
```diff
@@ -879,6 +868,9 @@ export function createCallViewModel$(
     ),
     map(() => {}),
     throttleTime(THROTTLE_SOUND_EFFECT_MS),
+    // Avoid doubling up on any auto-leave sounds (e.g. the decline sound),
+    // which are handled elsewhere
+    takeUntil(autoLeave$),
   );
```
toger5, line 879: I am not seeing how this is enough. the order should be:
other user leaves -> userMedia updates -> play sound
autoleave trigger as a consequence of userMedia update (so might come afterwards. hard to say with rxjs) autoleave call onLeave with "allOthersLeft".

Could be missing sth.

src/state/CallViewModel/CallNotificationLifecycle.ts
```diff
@@ -84,34 +93,81 @@ export interface Props {
...
+        return of({
+          intent: notificationEvent["m.call.intent"] ?? "audio",
+          recipient,
+          outcome$: race(timeout$, accept$, decline$).pipe(
+            take(1),
+            scope.share,
+          ),
+        });
+      }),
+      scope.share,
```
toger5, line 168: This part got really nice.

src/state/CallViewModel/CallNotificationLifecycle.ts
```diff
@@ -122,87 +178,18 @@ export function createCallNotificationLifecycle$({
...
-  );
-
   const autoLeave$ = merge(
     options.autoLeaveWhenOthersLeft === true
       ? allOthersLeft$.pipe(map(() => "allOthersLeft" as const))
       : NEVER,
-    callPickupState$.pipe(
-      filter((state) => state === "timeout" || state === "decline"),
+    ringAttempts$.pipe(
+      switchMap(({ outcome$ }) =>
```
toger5, line 186: We seem to ignore any outcome of the first ringingAttempt if there is a second one (switching to the new ringAttempt outcome$ obs). So even now we require that we only send a new ringingAttempt once the old one has finalised.

src/state/CallViewModel/CallNotificationLifecycle.ts
```diff
@@ -84,34 +93,81 @@ export interface Props {
...
+          return EMPTY;
+        }
+
+        // Ringing times out after lifetime ms have passed
+        const timeout$ = timer(notificationEvent.lifetime).pipe(
+          map(() => "timeout" as const),
+        );
+        // Call is accepted when someone else joins
+        const accept$ = memberships$.pipe(
+          filter((ms) => ms.value.some((m) => m.userId !== localUser.userId)),
```
toger5: We could use recipient here as well? like in decline$?

toger5 (changes requested):
Have you considered the `ack` flow we were thinking about.
Nested observables are not super nice to work with imo. So we should double chellange if we need them. The ringing attempt as a obs mostly adds the feature that we can have multiple rings at the same time. For anything else a simple behavior should be enough? We can still set it back to ringing state after it is done? Do we really need to be able to emit a new ring event while its still ringing?

The only real blocker is to confirm if we actually dont play the leave sound twice with this approach. And i think we need to clarify in comments why this solution does not end up to be a race condition.

robintown:
> Have you considered the `ack` flow we were thinking about.

Sure, `RingAttempt` could be extended to include something like an `acknowledgement$: Observable<void>`, and I think that would already capture the protocol fairly well.

> Nested observables are not super nice to work with imo. So we should double chellange if we need them. The ringing attempt as a obs mostly adds the feature that we can have multiple rings at the same time. For anything else a simple behavior should be enough? We can still set it back to ringing state after it is done? Do we really need to be able to emit a new ring event while its still ringing?

The main advantage of using nested observables for modeling, which I felt would be really nice here, is they give you the language to describe state machines precisely. Conceptually, the state machine for ringing looks more or less like this:

drawing 1

If you translate this into code as a flat `Behavior<CallPickupState>`, you run into the issue that the type doesn't say very much about the underlying state machine. There are all these unintended state transitions that are implied to be possible:

drawing 2

That alone isn't a huge deal, but there's also the challenge of deciding how you would insert the intent or the recipient into the type in a way that's convenient for consumers. Should they go only on the `ringing` state? But what if a consumer still cares about the intent even after a ring attempt has timed out (as `RingingMediaViewModel` does)? You might end up copying those fields across multiple states, but then they're suddenly implied to be dynamic values, potentially changing over the course of a ring attempt…

Anyway, I find it more straightforward to model the ring attempt explicitly, so that it captures the underlying state machine more precisely and there's also a neutral place to store a static intent and recipient.

drawing 3

(You can even imagine a machine extended with some extra acknowledgement states. Consumers would then be forced to decide what they do with the extra transitions. The structure of the type reflects the structure of the state machine, I rather like that.)

drawing 4

That's my pitch for the nested observable model, and hopefully an okay explanation of how I think about these things :) You're free to veto it still. I could stuff the machine back into a flat structure though it's not my preference.

> The only real blocker is to confirm if we actually dont play the leave sound twice with this approach. And i think we need to clarify in comments why this solution does not end up to be a race condition.

Agreed, I need to double check that.

toger5:
Thanks for the evaluation.
Helps to layout the issue more. Its also a very good basis for the ack logic which we will work on (eventually).
I think the "represents the actual state machine better" argument is very compelling.
What I am still a bit hesitant about is that nested observables also introduce the additional degree of freedom to allow multiplying states (basically inner duplications of the state machine). Using observables it is totally possible that the observable spawns multiple ring attempts (which might even be desired). but it can then also spawn multiple outcomes (the nesting still allows for multiple outcome$s or transitions between the outcome states (dependent on the interpretation of a new emission of the outcome$ obs) so on a type level we still have more arrows than your figure suggests, right? accept -> decline for example.)
The point I am making is basically that with ts + rxjs types we always have those undesired state transitions by just looking at the type but in practice cannot happen. But on the other hand that should not stop us from doing as good as we can. So I am convinced by your evaluation, that the tradeoff of more complexity on first sight is worth the stricter typing.

Especially the last picture helped me realise, that there are two timeouts on different levels. One where we wait for the ring to end/timeout and one where we wait the ack to "not"-arrive. Which life on different levels (the no-ack case is outside the acknowledgement box)

toger5 (approved):
Thanks.
I think the double leave sound should still be tested. (there are tests but they might be too dependent on the prev impl to still help... Its hard to have the perfect tests... In this case I am afraid there were too "unit")

src/state/CallViewModel/CallViewModel.ts
robintown: Tested, and it indeed wasn't enough :upside_down_face: Execution order (RxJS's lack of semantics) was a problem as you predicted. I have reverted to your original fix: 8b0f5054ddd7d9446db24dcad432310306e4273d

src/state/CallViewModel/CallNotificationLifecycle.ts
robintown: 05797097c66b867c216c93f5e7ad9fe3ecebf0eb

PR #4035, opened 2026-06-11
Remove deprecated @types/content-type dependency

Johennes:
`@types/content-type` is deprecated as per https://www.npmjs.com/package/@types/content-type. The types seem to now ship with `content-type` itself which we seem to inherit as a dependency from the JS SDK.


robintown (approved):
Nice, thanks

PR #4036, opened 2026-06-11
Fix a minor resource leak with display names and avatars

robintown:
I noticed that calls to `createDisplayNameBehavior$` and `createAvatarUrlBehavior$` were technically leaking resources since they reused the ObservableScope from their outer scope, which in practice lasts for the entire lifetime of the CallViewModel. This would not have had any noticeable effect unless you had other participants leave and rejoin the same call many thousands of times.

### DIFF
--- a/src/state/CallViewModel/CallViewModel.ts
+++ b/src/state/CallViewModel/CallViewModel.ts
@@ -766,11 +766,13 @@ export function createCallViewModel$(
             pretendToBeDisconnected$: localMembership.reconnecting$,
             displayName$: scope.behavior(
               matrixMemberMetadataStore
-                .createDisplayNameBehavior$(userId)
+                .createDisplayNameBehavior$(scope, userId)
                 .pipe(map((name) => name ?? userId)),
             ),
-            mxcAvatarUrl$:
-              matrixMemberMetadataStore.createAvatarUrlBehavior$(userId),
+            mxcAvatarUrl$: matrixMemberMetadataStore.createAvatarUrlBehavior$(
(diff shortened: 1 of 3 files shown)


Johennes:
My RxJS is still infantile but I think I understand what's going on here. Fix looks good to me. 👍

Johennes (approved):

src/state/CallViewModel/CallViewModel.ts
toger5, line 752: I am wondering how we can make the scope footguns less easy to trigger.
Its still one of the things of RxJs where we need to be super careful and there is no tooling helping us to prohibit mistakes.

Do you think it would help to "not allow" (be very alerted and careful) if we use just `scope` as the scope name and try to be disciplined to always call things `${prefix}Scope` with a `prefix` that should make clear what kind of scope you are dealing with?

sth like this:
callViewScope
connectionScope
memberScope



```
        (scope: itemScope, _, dup, mediaId, userId, participant, connection$, rtcId) =>
```

toger5 (approved):
Good job on spotting this.

src/state/CallViewModel/CallViewModel.ts
robintown: That could help, but I think I would rather take the time to create a lint rule. It sounds… fairly doable? (If an ObservableScope is referenced but not defined in the current inner scope, error)

robintown: https://github.com/element-hq/element-call/pull/4040

PR #4038, opened 2026-06-11
Fix Logging indexDB EC/EW collision

toger5:
The issue with the logger is caused by EC using the same domain as EW (develop.element.io) so they end up using the same indexdb scope.
And they also (mistakingly) use the same logs db name.
This has not bothered us before since we were using call.element.io. But now with embedded mode we use the same DB which can lead to collisions with EWs logs.


Johennes (approved):
LGTM :shipit:

toger5:
I bypassed the quality checks for the code coverage:
justification: This just renames one db name and does not introduce additional logic

PR #4040, opened 2026-06-12
Add lint rule to prevent ObservableScope resource leaks

robintown:
The rule of thumb to avoid resource leaks is that you should never call ObservableScope methods in a callback unless the ObservableScope is directly passed to or created inside that callback. I had a go at codifying this as a lint rule.

Off the back of https://github.com/element-hq/element-call/pull/4036#discussion_r3401702662


toger5:
Is this compatible with oxlint?

Johennes:
> Is this compatible with oxlint?

Yes, oxlint supports eslint-compatible JS plugins: https://oxc.rs/docs/guide/usage/linter/js-plugins.html

Johennes (approved):
This looks good to me (provided that you fix the remaining CI error). I'm fine with this landing before https://github.com/element-hq/element-call/pull/4053. We can make the necessary config changes over there.

PR #4042, opened 2026-06-15
Switch to pnpm 11.6.0

Johennes:
This upgrades pnpm to 11.6.0


robintown (approved):
Cool, I like that we can presumably add comments to the `overrides` now that they're in YAML.

Johennes:
> Cool, I like that we can presumably add comments to the `overrides` now that they're in YAML.

Oh, excellent point. I was actually wondering where all of these come from. Have added some comments based on what I could piece together from git history.

PR #4044, opened 2026-06-15
Output the actual command for triggering prettier -w

Johennes:
Maybe this is just me but I occasionally run into prettier errors and am then annoyed because it tells me to invoke prettier with `--write` but doesn't give me the pnpm command.

This change simply outputs our custom command if the check fails for convenience.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "build:sdk": "pnpm build:full --config vite-sdk.config.js",
     "build:sdk:production": "pnpm build:sdk",
     "serve": "vite preview",
-    "prettier:check": "prettier -c .",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
     "build:sdk": "pnpm build:full --config vite-sdk.config.js",
     "build:sdk:production": "pnpm build:sdk",
     "serve": "vite preview",
-    "prettier:check": "prettier -c .",
+    "prettier:check": "prettier -c .; EC=$?; [ $EC -ne 0 ] && printf '[\\033[33mwarn\\033[0m] To fix, run: pnpm run prettier:format\\n' >&2; exit $EC",
```
Johennes, line 20: This won't work on Windows but I suppose that's ok(?)

Johennes:
Abandoning this over https://github.com/element-hq/element-call/pull/4048.

PR #4046, opened 2026-06-16
Move ringing status indicator to header on mobile

robintown:
On mobile, the ringing status indicator is supposed to display in the header rather than on a tile. The exact layout differs between Android and iOS. To get it right I had to refactor AppBar to use CSS grid templates.

| |Before|After|
|-|-|-|
|Android|Screen Shot 2026-06-18 at 17 35 16|Screen Shot 2026-06-18 at 17 32 16|
|iOS|Screen Shot 2026-06-18 at 17 34 53|Screen Shot 2026-06-18 at 17 32 53|

(Also, I changed my mind about the exact ringing data I needed out of CallViewModel - sorry. A little move of the ringtone audio renderer into its own component was necessary to accommodate that.)

Closes https://github.com/element-hq/voip-internal/issues/527


robintown:
I guess this needs a bit more coverage, but should already be ready for review.

src/tile/SpotlightTile.test.tsx
```diff
@@ -177,11 +177,6 @@ test("SpotlightTile displays ringing media", async () => {
...
 
   expect(await axe(container)).toHaveNoViolations();
-  // Alice should be in the spotlight with the right status
+  // Alice should be in the spotlight
   screen.getByText("Alice");
-  screen.getByText("Calling…");
-
-  // Now we time out ringing to Alice
-  act(() => pickupState$.next("timeout"));
-  screen.getByText("Call ended");
```
Johennes, line 186: Does this actually still test what the function claims to test?

> "SpotlightTile displays ringing media"

Johennes (approved):
I think this looks good other than coverage. 👍 

Also did some minor manual testing to verify it.

src/tile/MediaView.module.css
```diff
@@ -121,18 +121,22 @@ unconditionally select the container so we can use cqmin units */
...
 
-.status svg {
-  color: var(--cpd-color-icon-tertiary);
+  span {
+    vertical-align: center;
+  }
+
+  svg {
+    color: var(--cpd-color-icon-tertiary);
+    vertical-align: center;
```
Johennes: Addendum: Isn't this `middle` rather than `center`? It looks like `center` is a Firefox-only thing currently: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/vertical-align

src/components/CallFooterViewModel.tsx
toger5, line 172: why do we still need showHeader here?

src/tile/MediaView.module.css
robintown: Good catch. I tried `middle` but it doesn't look like I want. `text-bottom` seems to be fine, especially since the icon size is tied to `em` units.

src/tile/SpotlightTile.test.tsx
robintown: You're right that it didn't really. I had decided to remove the ringing status from spotlight tiles entirely so that there wouldn't be a "duplicate" status indicator in the interface. But now I've replaced this with a hopefully more sensible way of de-duplicating the status indicators: a456e47796e400c96e28e300682b66116911961c

toger5 (commented):
oh i forgot to ever sent this reveiw. (I was wondering why showHeader still was here..)

PR #4048, opened 2026-06-17
Switch from prettier to oxfmt

Johennes:
oxfmt is also used by the web team since recently. It's about 4 times faster on the element-call repo in my testing. This keeps our previous prettier configuration alive. The only two changes I had to do in code seem like legitimate fixes.


robintown (approved):
LGTM, and I think this is a good move :+1:

PR #4053, opened 2026-06-18
Switch from eslint to oxlint

Johennes:
I used the official migrator but had to disable quite a lot of rules afterwards which apparently fail in our code base (~600 errors). There are also a few rules, that have no replacement yet in oxlint. On the plus side, it is about a factor of 10 faster than eslint on my machine.


Johennes:
@robintown curious what you think of this? Feels a bit odd to have to turn so many rules off.

robintown:
A couple dozen new rules to disable doesn't surprise me *that* much. I had a quick look at a few of them, and I'm seeing

- old eslint rules that oxlint chose to enable by default (eslint/no-await-in-loop)
- rules that oxlint invented but are actually super nitpicky (react/no-object-type-as-default-prop)
- vitest/* of course, which would be a whole new category of rules for us

etc. Nothing I can't explain so far.

Johennes:
Ok, interesting. Well maybe it's not as bad as I thought then.

Johennes:
This will require rebasing and adoption of https://github.com/element-hq/element-call/pull/4040.

Johennes:
Looks like I was mistaken and this doesn't work with the rule from #4040 after all. 😕 

> Not supported yet:
>
> - Custom file formats and parsers (e.g. Svelte, Vue, Angular).
> - Lint rules that rely on TypeScript type-awareness.

Johennes:
@robintown probably best if you double check that I haven't messed the migration of your rule up. 🙈

eslint/NoObservableScopeLeak.js
```diff
@@ -40,19 +40,23 @@ const rule = ESLintUtils.RuleCreator(
           node.parent?.type === "MemberExpression" &&
           node.parent.object === node &&
           node.parent.property.type === "Identifier" &&
-          !safeScopeMethods.includes(node.parent.property.name)
+          !safeScopeMethods.includes(node.parent.property.name) &&
+          node.name.endsWith("cope")
```
robintown: `/(^s|S)cope$/.test(node.name)` maybe?

.oxlintrc.json
```diff
@@ -0,0 +1,197 @@
...
+          {
+            "group": [
+              "matrix-widget-api/src",
+              "matrix-widget-api/src/",
+              "matrix-widget-api/src/**",
+              "matrix-widget-api/lib",
+              "matrix-widget-api/lib/",
+              "matrix-widget-api/lib/**"
+            ],
+            "message": "Please use matrix-widget-api instead"
```
robintown, line 64: Do these `patterns` make the above `paths` redundant?

.eslintrc.cjs
```diff
@@ -1,120 +0,0 @@
...
-  rules: {
-    "element-call/no-observablescope-leak": "error",
-    "jsdoc/no-types": "error",
-    "jsdoc/empty-tags": "error",
-    "jsdoc/check-property-names": "error",
-    "jsdoc/check-values": "error",
-    "jsdoc/check-param-names": "warn",
-    // "jsdoc/require-param": "warn",
-    "jsdoc/require-param-description": "warn",
-    "matrix-org/require-copyright-header": ["error", COPYRIGHT_HEADER],
```
robintown, line 38: It's a little bit of a shame to lose this lint, would you be interested in porting it over from https://github.com/matrix-org/eslint-plugin-matrix-org/blob/main/rules/copyright.js ?

robintown (approved):
Generally seems to work well. I took a moment to set up integration with my editor so I could test it.

.eslintrc.cjs
Johennes: Good call. Added it in d5ce3a955ade3a0d34c1ccbb640e57f0acf94914.

.oxlintrc.json
Johennes: Oh, indeed. Good catch. Fixed in d2cb7e334eb551abea4da28abb267cf249d35ed3.

eslint/NoObservableScopeLeak.js
Johennes: That's better, yes. Fixed in 9d61ca13cb508322511e540d26b2aad34b612060.

.oxlintrc.json
```diff
@@ -29,6 +29,10 @@
     "builtin": true
   },
   "rules": {
+    "element-call/copyright-header": [
+      "error",
+      "/*\nCopyright %%CURRENT_YEAR%% New Vector Ltd.\n\nSPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial\nPlease see LICENSE in the repository root for full details.\n*/\n\n"
```
robintown: This should have `Element Creations Ltd.` as the copyright holder nowadays

robintown (approved):

.oxlintrc.json
Johennes: Argh, good catch. Fixed in 76714ac3ad38441fe2037651fa59cbc174303e87.

PR #4054, opened 2026-06-19
Add PiP border in 1on1 layouts

### LINKED ISSUE #4008: Add a subtle outline to small tile in one-on-one layout
1px, color/border/interactive/secondary, outlining the bounding box as shown:

Image

https://www.figma.com/design/aOEkaJtaBmPy058V7uoqVr/Element-Call-Updates---Q1-2026--New-?node-id=436-34644&t=t5KCbj7fPgVCZZzu-4

This will make the small tile more visually discernible from the background especially in dark mode.

Johennes:
## Content

Fixes: #4008

I'm unsure if this is a particularly elegant or performant way to do this.

## Screenshots / GIFs

|Before|After|
|-|-|
|Screenshot 2026-06-19 at 14 40 35|Screenshot 2026-06-19 at 14 40 27|

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -74,6 +74,7 @@ import {
   useAppBarSubtitle,
 } from "../AppBar.tsx";
 import { useBehavior } from "../useBehavior.ts";
+import { constant } from "../state/Behavior.ts";
 import { Toast } from "../Toast.tsx";
 import overlayStyles from "../Overlay.module.css";
 import { useTrackProcessorObservable$ } from "../livekit/TrackProcessorContext.tsx";
@@ -94,6 +95,7 @@ declare module "react" {
 }
 
 const logger = rootLogger.getChild("[InCallView]");
+const noOutline$ = constant(false);
 
 export interface ActiveCallProps extends Omit<
   InCallViewProps,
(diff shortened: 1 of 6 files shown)


Johennes:
Lacking coverage but otherwise ready for review.

toger5 (changes requested):
We try to become more and more strict in our mvvm approach.
Best case, the inCallView (or any view) does not do any logic.
The desired style should be
```
const a = useBehavior(vm.val$)
const b = useBehavior(vm.sth$)
return <div prop={b}> {a} </div>;
```

Consequently the names should be very descriptive in the view.
So I would like to just call it showOutline instead of pip.
The vm then can use user facing names (like pip, featured...) while the view makes the connection to the actual UI this translates to.
`outlined={isPip} ` or similar.

TLDR:
i think the if logic should live in the vm.
The view should only do `useBehavior` without conditions (if possible)

src/room/InCallView.tsx
```diff
@@ -433,6 +435,9 @@ export const InCallView: FC<InCallViewProps> = ({
         const showSpeakingIndicators = useBehavior(vm.showSpeakingIndicators$);
         const showNameTags = useBehavior(vm.showNameTags$);
         const showRingingStatus = vm.ringingStatusLocation === "tile";
+        const showOutline = useBehavior(
+          model instanceof GridTileViewModel ? model.showOutline$ : noOutline$,
```
toger5:
```suggestion
          model instanceof GridTileViewModel ? model.showOutline$ : constant(false),
```
I think this is clear enough?

toger5: Probably you have done this to optimise for `constant` allocations?
So this is low temp. Performance might even be better like this. Not sure it has a significant impact however

toger5 (approved):
Thanks for updating it.
I was hoping for it to be a little bit simpler.

src/room/InCallView.tsx
Johennes: Yeah, that was my thinking. But I don't feel strongly either.

Johennes: Removed in d534c617d2435acc3dcafd1ff6c15bdd887fd2fd.

PR #4061, opened 2026-06-22
Show the subtitle of the app bar on small screens

robintown:
Previously we were hiding the entire app bar on mobile phones in landscape orientation. However now that the app bar supports a small 'subtitle' element, we should show only the subtitle in this case to match the designs.

||Before|After|
|-|-|-|
|Android|Screen Shot 2026-06-22 at 18 19 00|Screen Shot 2026-06-22 at 18 19 35|
|iOS|Screen Shot 2026-06-22 at 18 18 34|Screen Shot 2026-06-22 at 18 18 08|

The subtitle still hides on tap, just like the footer.

### DIFF
--- a/src/AppBar.module.css
+++ b/src/AppBar.module.css
@@ -67,6 +67,27 @@
     "subtitle subtitle subtitle";
 }
 
+/* Hide everything but the subtitle in small windows */
+@media (max-height: 450px) {
+  .bar {
+    display: none;
+    .title,
+    .primaryButton,
+    .secondaryButton {
+      display: none;
+    }
+  }
+
+  .bar:has(.subtitle) {
+    display: initial;
+    > header {
+      grid-template-columns: 1fr;
+      grid-template-rows: var(--cpd-space-5x) minmax(var(--cpd-space-5x), auto);
+      grid-template-areas: "." "subtitle";
+    }
+  }
+}
+
 .primaryButton {
   grid-area: primaryButton;
   justify-self: start;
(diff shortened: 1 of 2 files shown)


src/AppBar.module.css
```diff
@@ -141,4 +162,26 @@ body[data-platform="ios"] {
       display: none;
     }
   }
+
+  /* Hide everything but the subtitle in small windows */
+  @media (max-height: 450px) {
```
Johennes, line 169: Should some of the styles here such as `.bar { display: none; }` not already be guaranteed by the generic styles further up?

src/AppBar.module.css
```diff
@@ -67,6 +67,27 @@
...
 }
 
+/* Hide everything but the subtitle in small windows */
+@media (max-height: 450px) {
+  .bar {
+    display: none;
+    .title,
+    .primaryButton,
+    .secondaryButton {
+      display: none;
```
Johennes, line 82: If these are element inside a parent styled with `.bar`, does the outer `display: none` not suffice?

Johennes (approved):
LGTM minus two (possibly silly) CSS questions.

toger5 (approved):

src/AppBar.module.css
robintown: I suppose it would be clearer that this is about making the title and buttons not show up even when the bar reverts to `display: initial` by actually putting this rule inside the same block as `display: initial`. be5d8140c90056aae357f4c0d8b9ee59a7f2aa9a

src/AppBar.module.css
robintown: You're right at least about the `display: none` and `display: initial` styles (removed in a4b4a6b655ca3bc7ca8b74ae4eaddcc5e7346fd6). I had to keep the explicit `grid-template-area` style because otherwise it's overridden by the `body[data-platform="ios"] .bar:has(:header) > header` rules.

PR #3736, opened 2026-02-11
feat: advanced media quality settings UI and config-driven defaults

emmick4:
## Why

  Element Call currently hardcodes video/audio parameters with no way for users to tune them. Server admins also have no mechanism to set org-wide defaults. Several community requests have asked for these controls.
  ## What

  This PR adds three new sections to the ~~Settings modal~~ Developer Settings Tab:

  - **Camera quality**: resolution, framerate, bitrate, and codec selection behind an "Advanced camera settings" toggle. Applied at room join.
  - **Screen sharing** : same controls for screen share. Applied when toggling screen share mid-call.
  - **Audio processing**: individual checkboxes for echo cancellation, noise suppression, and automatic gain control.

  All sliders now display their current value inline (e.g. "Framerate: 30 fps").

  A new `media_quality` section in `config.json` lets admins set defaults for video codec, resolution, bitrate, framerate, and simulcast layers. These seed user settings only when the user hasn't explicitly set a preference in localStorage, so user choices always win.

  The static `defaultLiveKitOptions` export is replaced by `buildLiveKitOptions()` / `getLiveKitOptions()` which read from config at runtime.
  ## Testing strategy

  1. Check out the branch and run `yarn && yarn build`
  2. Run `yarn test` — 441 tests should pass
  3. Start the dev server with `yarn dev:full`
  4. Open Settings > Video tab:
     - Enable "Advanced camera settings", pick a non-default codec (e.g. VP9) and resolution (e.g. 1080p), then join a call
     - Open `chrome://webrtc-internals`, find the outbound video track, confirm the codec and resolution match your selection
     - Enable "Advanced screen share settings", start a screen share, confirm encoding in webrtc-internals
  5. Open Settings > Audio tab:
     - Uncheck "Noise suppression", rejoin the call
     - In webrtc-internals, check the audio track constraints
  6. Test config-driven defaults:
     - Add a `media_quality` block to `config.json` (e.g. `{"video_codec": "vp9"}`)
     - Clear localStorage, reload — settings should reflect the config values
     - Change a setting in the UI, reload — the user override should persist
     
 Note: happy to make any adjustments or split this PR up, haven't done much OSS contributing and just wanted to have these features on my own server. it also lays the scaffolding for more configurable audio and setting quality limits by the server admin

dev tab | audio related settings
---| ---
Screenshot 2026-08-10 at 14 08 44 | Screenshot 2026-08-10 at 14 12 59

Fixes: https://github.com/element-hq/element-call/pull/3736

PR #4066, opened 2026-06-25
Default to compatibilty mode (multi sfu) - non breaking, as multi-sfu is compatible with all versions younger 6 month

toger5:
Fixes: https://github.com/element-hq/voip-internal/issues/619

Lets unleash all the SFU's.

This can only get merged once we have a TWIM post:
```
This week we do a small but significant change in Element Call.
From now on, we will be using the multi-sfu approach for our video and voice calls.
This is a noval approach introduced at one of Elements, "How Hard Can It Be" hackathon.
The big advantage is that is mitigates any negotiation on which SFU the users in a call "meet".
Instead everyone connects to the SFU associated with their Homeserver. No negotiation required.

If I want to subscribe to a stream from someone on a different homeserver Element Call will connect to that
SFU in an subscribe only mode.

With the next version of Element Call there will be no more SFU negotiations by default.
You still have the option to switch back to legacy mode using the developer settings tab.

EC is compatible with this mode since 6month already. So the whole ecosystem should be updated by now and this
should not lead to any incompatiblities.

Of course this requires every call participant to have an SFU associated with their homeserver if they want to participante in a session.
Since we already have checks in place so only usres with an SFU can join a call, this should not make a difference for anyone already on the latest versions.
```

### DIFF
--- a/src/settings/settings.ts
+++ b/src/settings/settings.ts
@@ -137,7 +137,7 @@ export const enableExtendedLivekitLogs = new Setting<boolean>(
 
 export const matrixRTCMode = new Setting<MatrixRTCMode>(
   "matrix-rtc-mode",
-  MatrixRTCMode.Legacy,
+  MatrixRTCMode.Compatibility,
 );
 
 export const customLivekitUrl = new Setting<string | null>(
--- a/config/config.sample.json
+++ b/config/config.sample.json
@@ -12,7 +12,7 @@
     "feature_use_device_session_member_events": true
   },
   "ssla": "https://static.element.io/legal/element-software-and-services-license-agreement-uk-1.pdf",
-  "matrix_rtc_mode": "legacy",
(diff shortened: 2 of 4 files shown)


Johennes (approved):
Looks good but we should probably also change this in `config.sample.json`.

src/settings/settings.ts
CrazyNicc, line 1: Will changing the default to Multi-SFU require Sticky Event Support from all participating Homeservers?

toger5: No not yet. This is the next step (moving to sticky events/ full matrix2.0 mode)
Good comment however. I think this should be made particularly clear in the TWIM message.

CrazyNicc: (Also, wholly unrelated to this, I know; Could you guys take a look at #3736 again?)

PR #4067, opened 2026-06-25
Improve the look of voice calls and group calls on mobile

robintown:
This improves the look of calls on mobile by adding a new gradient background and adjusting various visual elements (footer and spotlight tiles) to let the background "shine through" when appropriate.

It also applies a new "sound wave" design to the other participant's avatar when they are speaking. If desired I can split that change out.

|Before|After|
|-|-|
|Screen Shot 2026-06-25 at 19 00 29|Screen Shot 2026-06-25 at 18 59 40|
|Screen Shot 2026-06-25 at 19 08 18|Screen Shot 2026-06-25 at 19 11 35|

You can review this commit-by-commit.

I've marked this as a breaking change because it now applies a gradient background by default even in widget mode, which consumers of the embedded package will need to adapt to. (Element Web will want to disable the gradient, and Element X mobile clients will want to make it look best by forcing a dark theme.)

Closes https://github.com/element-hq/voip-internal/issues/526


src/tile/MediaView.test.tsx
```diff
@@ -129,30 +129,4 @@ describe("MediaView", () => {
       ).toBe(0);
     });
   });
-
-  describe("videoEnabled", () => {
```
robintown, line 129: Removed these tests because they rested on the assumption that `display: none` was set directly in the video's `style` attribute rather than in a style sheet.

src/graphics/mobile-gradient.svg
```diff
@@ -0,0 +1,109 @@
...
+     showgrid="false"
+     inkscape:zoom="0.38508772"
+     inkscape:cx="170.09112"
+     inkscape:cy="257.08428"
+     inkscape:window-width="1920"
+     inkscape:window-height="1131"
+     inkscape:window-x="0"
+     inkscape:window-y="0"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="svg5" />
```
robintown: Note to self, remove the extra Inkscape-specific data

playwright/widget/test-helpers.ts
```diff
@@ -188,6 +188,12 @@ export class TestHelpers {
         break;
       }
     }
+
+    // Also dismiss the release announcement
+    await page
+      .getByRole("dialog", { name: "Introducing Sections" })
+      .getByRole("button", { name: "OK" })
+      .click();
```
Johennes: I _think_ this might be redundant now as I also had to fix it as part of https://github.com/element-hq/element-call/pull/4076.

robintown: Phew, thanks. I will merge that.

src/state/TileStore.ts
```diff
@@ -39,12 +39,29 @@ class SpotlightTileData {
     this.maximised$.next(value);
   }
 
+  private readonly bgStyle$: BehaviorSubject<"solid" | "transparent">;
```
Johennes: Nit: Maybe worth extracting a type given that `"solid" | "transparent"` occurs in multiple places?

src/tile/MediaView.tsx
```diff
@@ -118,6 +120,7 @@ export const MediaView: FC<Props> = ({
       ref={ref}
       data-testid="videoTile"
       data-video-fit={videoFit}
+      data-bg-style={bgStyle}
```
Johennes: I can't help but notice the imparity between `bgStyle` / `data-bg-style` and `background` / `data-background` (used for the page background) which both set the style of the background. This doesn't really matter much; so you may do with this comment whatever you want. 🙈😅

src/room/__snapshots__/InCallView.test.tsx.snap
```diff
@@ -164,7 +164,7 @@ exports[`InCallView > rendering > renders 1`] = `
       class="container"
     />
     <div
-      class="footer"
+      class="footer footer"
```
Johennes: This struck me as odd at first but looks like these are in fact two different classes that just end up using the same string in tests due to our `classNameStrategy` setting. I guess `classNameStrategy == stable` would call this out a little nice but I assume there's no real downsides with the duplicate class names in just tests either?

Johennes (commented):
This looks good to me code-wise. Thanks a ton for breaking it up into commits. That really simplified the review.

I only tested Android. Is this how it is supposed to look without further changes?

|Before|After|
|-|-|
|Image|Image|

src/room/__snapshots__/InCallView.test.tsx.snap
robintown: I don't think there's any particular reason we've set `classNameStrategy` to `non-scoped`, so I like your suggestion as a way to reduce confusion. I will do that in a separate PR.

src/state/TileStore.ts
robintown: bebe79d1094de45a3d37108809768508626ac0e2

src/tile/MediaView.tsx
robintown: 27315a7090079f0527b5a903e4fe4b322103768d

robintown:
Good that you tested Android, I apparently had messed up the SVG data in a way that caused it to render differently on Firefox and Chromium.

And well, it does still render differently on Firefox for Android (the gradient is smaller than intended, as visible in the PR description) but I tried a bunch of different values for `background-size` and `preserveAspectRatio` and none of them did anything. At least Chromium on Android and all desktop browsers I tried now render the gradient as designed.

src/room/__snapshots__/InCallView.test.tsx.snap
robintown: https://github.com/element-hq/element-call/pull/4099

toger5:
I am wondering if the "avatar only" design is also desired in 1:1 non mobile or maybe even for every spotlight wihtout camera?

Here is the pip version:
Screenshot 2026-07-08 at 18 38 52

And here the spotlight version:
Screenshot 2026-07-08 at 18 39 04

I think the spotlight could look really good if it would also just be the "speaking avatar"

robintown:
@toger5 Interesting suggestion, maybe leave a comment for Aaron on the designs and see what he thinks?

The large tile in the landscape one-on-one layout is technically a grid tile rather than a spotlight tile, so it would take some extra changes to turn transparent and have it interact with the footer correctly.

americanrefugee:
> I am wondering if the "avatar only" design is also desired in 1:1 non mobile or maybe even for every spotlight without camera?

The vision is that there is no 1:1 voice call in spotlight mode. If both people have their camera turned off, then you only get a PiP with a speaking avatar. The logic is that if both are only voice, then there's nothing meaningful to look at (especially in spotlight mode).

DM - Voice call started

Johennes (approved):
I think this looks good now. 👍

toger5:
@americanrefugee 
> is no 1:1 voice call in spotlight mode

I see. Currently we still allow to go into non pip for the sake of reaching the element call settings.
Is this sth we should only allow when enabling the camera?
Or is the future vision that you cannot get into non-pip when starting a voice call on web?

americanrefugee:
@toger5 When [starting a voice call](https://www.figma.com/design/aOEkaJtaBmPy058V7uoqVr/Element-Call-Updates---2026?node-id=2-14516&t=lw2zYbVkEGm5JEnN-4), you should go directly to the PiP:

DM - You started a voice call

However, if we feel like users should have access to the mic (or camera) settings, then I need to create a new small button component with the selector UI currently shown in a video call.

Shall I do that and then update the PiP design? cc @daniellekirkwood

americanrefugee:
I already made the change :)

The [mic](https://www.figma.com/design/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=14126-44054&t=ZAngUqliMebduhc3-4) and [camera](https://www.figma.com/design/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=14126-44058&t=ZAngUqliMebduhc3-4) button in the [PiP](https://www.figma.com/design/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=12561-13483&t=ZAngUqliMebduhc3-4) should also display the dropdown arrow to launch the settings:

DM - You started a voice call

PR #4068, opened 2026-06-25
Hide local tile from mobile voice calls in landscape orientation

robintown:
To match the designs. Just like in portrait, it wouldn't contribute anything meaningful.

|Before|After|
|-|-|
|Screen Shot 2026-06-25 at 20 32 47|Screen Shot 2026-06-25 at 20 32 32|


Johennes (approved):
LGTM 👍

PR #4073, opened 2026-06-29
Fix some false positives flagged by ObservableScope leak lint rule

robintown:
The rule should only care about enclosing function/class scopes. For example if an ObservableScope is received as a parameter to a function and then simply used inside an `if` block (technically a different scope), that's not a problem.

### DIFF
--- a/eslint/NoObservableScopeLeak.js
+++ b/eslint/NoObservableScopeLeak.js
@@ -9,7 +9,21 @@ import { ESLintUtils } from "@typescript-eslint/utils";
 
 // These ObservableScope methods will not generally cause resource leaks even if
 // called from a callback
-const safeScopeMethods = ["bind", "end"];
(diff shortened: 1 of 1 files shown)


eslint/NoObservableScopeLeak.js
```diff
@@ -9,7 +9,21 @@ import { ESLintUtils } from "@typescript-eslint/utils";
 
 // These ObservableScope methods will not generally cause resource leaks even if
 // called from a callback
-const safeScopeMethods = ["bind", "end"];
+const safeScopeMethods = new Set(["bind", "end"]);
```
robintown, line 12: Oxlint told me to change it to a set

Johennes: Looks like it's over-optimizing a bit. 😅

Johennes (approved):
This looks good, thank you for following up on it.

PR #4082, opened 2026-07-03
Make renovate track docker-compose files and pin synapse / element-web for playwright tests

Johennes:
Fixes: https://github.com/element-hq/voip-internal/issues/635
We had a regression on our default branch this week caused by an upstream change on element-web's `develop` branch. This was due to our Playwright tests reusing our development docker-compose file which used the `develop` tag for the element-web image.

This pull request:

- Renames all docker-compose files so that Renovate tracks them (default name-matching pattern requires `docker-compose` at the start of the file name)
- Pins the synapse and element-web images to SHAs in the docker-compose file used for Playwright (this allows Renovate to update them on a cycle)
- Leaves synapse and element-web pinned to the `latest` and `develop` tags in the docker-compose file used for development (this prevents Renovate from updating them)
- Pins nginx in the docker-compose file used for development (this allows Renovate to update it on a cycle)
- Removes `playwright-backend-docker-compose.yml` as it just included the dev file and did nothing else


Johennes:
Props to @t3chguy for suggesting to use Renovate to handle this rather than setting up cross-repository CI.

toger5:
This does not seperate the EW updates for playwright into its own renovate "task"

I think it would be handy to have it split out so it can be actioned as quick as possible to detect regressions fast. (It would be a bummer if another update incompatiblity would block EW to be tested against latest EC)

toger5:
> This does not seperate the EW updates for playwright into its own renovate "task"

We decided, that I will look into this.

toger5 (approved):

Johennes:
Looks good to me. Thanks @toger5!

PR #4088, opened 2026-07-07
fix getChild before rageshake setup

toger5:
## Content

Fix missing logs in rageshakes. (yes there was another issue, the first one was using the same rageshake db as EW)
The issue:
 - rageshake loggers are setup on the root logger via the rageshakes file `init()` method.
 - files that used `const logger = rootLogger.getChild("[childscopename]");` often executed before the init call.
 - so the `logger` would be created before the parent logger got configured to use the rageshake log extension.
 - we ended up missing a large amount of logs

## Motivation and context

We need those logs. This Issue revealed the missing logs:
https://github.com/element-hq/element-call/pull/4087
(we first got locally recorded logs from android via logcat and then a rageshake.)

## Tests
uploaded logs before this change:
https://github.com/element-hq/element-call-rageshakes/issues/16750
 - `[CallViewModel]` is missing `[CallViewModel localTransport$] Creating item with keys compatibility ...` does exisint but that is from the `item with keys` logging... not the `getChild` one

after changing some (first commit):
https://github.com/element-hq/element-call-rageshakes/issues/16754

 - `[CallViewModel][1783444309670]` exists!!
 -  but `[InCallView]` is missing

after updating all logs:
https://github.com/element-hq/element-call-rageshakes/issues/16755
 -  we should have all logs now. including `[InCallView]` (only gets logged on fatal errors so not visible in rageshake)
## Checklist

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -34,6 +34,7 @@
       "/*\nCopyright %%CURRENT_YEAR%% Element Creations Ltd.\n\nSPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial\nPlease see LICENSE in the repository root for full details.\n*/\n\n"
     ],
     "element-call/no-observablescope-leak": "error",
+    "element-call/no-top-level-logger-get-child": "error",
     "jsdoc/empty-tags": "error",
     "jsdoc/check-property-names": "error",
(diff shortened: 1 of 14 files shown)


Johennes (approved):
Oh wow, nice find! And amazing that this even adds a lint rule to prevent future regressions. ❤️

PR #4099, opened 2026-07-08
Use vitest's default for scoped class names

robintown:
It reduces confusion when reading snapshots if class names are still scoped to their respective CSS modules. It also discourages the use of class names in tests, which [is a good thing](https://testing-library.com/docs/guiding-principles).

Thanks to Johannes for the suggestion.


src/tile/TileAvatar.tsx
```diff
@@ -17,11 +18,12 @@ interface Props extends AvatarProps {
 }
 
 export const TileAvatar: FC<Props> = ({ size, loading, ...props }) => {
+  const { t } = useTranslation();
   return (
     <div>
       {loading && (
         <div className={styles.loading}>
-          <InlineSpinner size={size / 3} />
```
toger5, line 24: here

src/tile/TileAvatar.test.tsx
```diff
@@ -6,22 +6,22 @@ Please see LICENSE in the repository root for full details.
...
-    expect(container.querySelector(".loading")).toBeInTheDocument();
+    screen.getByLabelText("Loading…");
   });
 
   it("should not show loading spinner when not loading", () => {
-    const { container } = render(
+    render(
       <TileAvatar id="@a:example.org" name="Alice" size={96} loading={false} />,
     );
-    expect(container.querySelector(".loading")).not.toBeInTheDocument();
```
toger5, line 25: here

toger5 (approved):
Thanks!
The commit that fixes some test failures caused by this change raises questions.
I commented "here" on the lines where I am not sure why they caused errors.

If you have time i would be curious why they needed changing. But not blocking (of course. I totally makes sense why the new version works. Just not sure what was wrong with the old)

src/tile/TileAvatar.tsx
robintown: This spinner was setting off an accessibility warning by being neither aria-hidden nor aria-labeled, and I wanted a label anyway for the test you commented on below

src/tile/TileAvatar.test.tsx
robintown: This expectation was looking for a class name (`.loading`) which has changed to something more complicated now that it's scoped. I figured this test should be written in terms of stuff that's present in the accessibility tree anyway, so I preferred that as the fix.

PR #4100, opened 2026-07-09
Stop using colors from previous design system

robintown:
These stop-gap color variables have been largely unused for a long time. This changes the colors of checkboxes and the 'end call' button to align with current designs as well.

|Before|After|
|-|-|
|Screenshot 2026-07-09 at 12-34-15 Element Call The Testing Zone|Screenshot 2026-07-09 at 12-33-26 Element Call The Testing Zone|
|Screenshot 2026-07-09 at 12-34-00 Element Call The Testing Zone|Screenshot 2026-07-09 at 12-33-42 Element Call The Testing Zone|


Johennes (approved):
LGTM

PR #4103, opened 2026-07-10
Add enableClientWellKnownLookups

langleyd:
Adds gating (default on) so a deployment can stop Element Call making client
`.well-known` requests to the homeserver `server_name`. Set via the
`enableClientWellKnownLookups` URL param in widget mode (by the embedder) and
`enable_client_well_known_lookups` in `config.json` in SPA mode; resolution is
URL param → config → default on.

- **Off:** always avoid client `.well-known` — skips the post-login poll
  (`widget.ts` widget mode, `utils/matrix.ts` SPA mode) and the legacy MatrixRTC
  foci fallback.
- **On/unset:** respect the server — only fall back to `.well-known` for
  transports on a `404 M_UNRECOGNIZED`; any other response is authoritative.

Element web embedded:
Complementary to https://github.com/element-hq/element-web/pull/34084.
If remove`element_call.disable` in EW config but keep `enable_client_well_known_lookups` EC will continue to respect it (as long as it passes in the URL Param).

https://github.com/matrix-org/matrix-js-sdk/issues/5245 can then be added on top and will allow EC to function in EW alongside `enable_client_well_known_lookups`.

### DIFF
--- a/src/state/CallViewModel/localMember/RtcTransportAutoDiscovery.ts
+++ b/src/state/CallViewModel/localMember/RtcTransportAutoDiscovery.ts
@@ -8,7 +8,11 @@ import {
   isLivekitTransportConfig,
   type LivekitTransportConfig,
 } from "matrix-js-sdk/lib/matrixrtc";
-import { type IClientWellKnown, type MatrixClient } from "matrix-js-sdk";
+import {
+  type IClientWellKnown,
+  type MatrixClient,
+  MatrixError,
+} from "matrix-js-sdk";
 import { type Logger } from "matrix-js-sdk/lib/logger";
 
 import type { ResolvedConfigOptions } from "../../../config/ConfigOptions.ts";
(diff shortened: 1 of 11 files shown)


Johennes (approved):
This looks good to me. I think we should trim the comments down a little though as they're quite verbose. Ideally they shouldn't describe anything that's obvious from a method's or variables name or anything that refers to plumbing of the item being documented in some other part of the code base.

src/state/CallViewModel/localMember/RtcTransportAutoDiscovery.ts
```diff
@@ -22,13 +26,36 @@ type TransportDiscoveryClient = Pick<
...
+  /** The livekit transport found via the backend endpoint, if any. */
+  transport: LivekitTransportConfig | null;
+  /**
+   * Whether the caller may fall back to the legacy `.well-known` lookup. True
+   * only when the backend endpoint is unavailable, i.e. it was not attempted
+   * (widget mode) or the homeserver does not implement it (404 M_UNRECOGNIZED).
+   * False when the endpoint answered authoritatively (a response without a
+   * livekit transport) or failed with any other error.
+   */
+  mayFallBackToWellKnown: boolean;
```
BillCarsonFr, line 52: Why is this needed? 
Just use a config to disable well-known lookup? Anyhow well-known lookup is deprecated, and it is just blocked on having the widget beeing able to call /rtc/transport via the widget API (because it is now authenticated)

BillCarsonFr (commented):
I don't understand the `mayFallBackToWellKnown` subtility? Just add a config to disable well-known lookup, it is deprecated anyhow

src/state/CallViewModel/localMember/RtcTransportAutoDiscovery.ts
Johennes: The way I understand it this just optimizes when `.well-known` look-ups are required in the first place. When the server implements `/rtc/transports`, we shouldn't fall back to the `.well-known` as that undermines the endpoint's purpose. I guess it's not strictly needed in this PR but it fixes what we're doing right now. So I don't really mind it being in here.

BillCarsonFr:
This could be solved by landing and js and rust implementation and then just remove well-known
https://github.com/matrix-org/matrix-spec-proposals/pull/4515

BillCarsonFr:
I think this should be closed. EC will stop doing any sort of well-known https://github.com/element-hq/element-call/pull/4153

Now EC will call the new WidgetAction for rtc transport discovery [MSC4515](https://github.com/matrix-org/matrix-spec-proposals/pull/4515), and we delegate to the host to decide if it wants to fallback or not (based on some config)

Johennes:
Closing as per the comment above.

langleyd:
Makes sense, thanks

PR #4109, opened 2026-07-14
Create service interruptions view model

### LINKED ISSUE #4098: Create service interruptions view model
Create a view model type for the [service interruptions component](https://www.figma.com/design/aOEkaJtaBmPy058V7uoqVr/Element-Call-Updates---2026?node-id=971-2).

Image

It should account for all the following possible services:

- Encryption keys
- Emoji reactions
- Screen sharing
- Your audio and video
- (One-on-one call) Other person's audio and video
- (Group call) Other people's audio and video

For this task it is **not** necessary to write any RxJS or React code. This is just about agreeing on a type so that the [component](https://github.com/element-hq/element-call/issues/4089) and the [business logic](https://github.com/element-hq/element-call/issues/4091) can then be implemented in parallel.

robintown:
Closes https://github.com/element-hq/element-call/issues/4098

### DIFF
--- a/knip.ts
+++ b/knip.ts
@@ -20,7 +20,12 @@ export default {
     // This is a shell built-in.
     "printf",
   ],
-  ignoreFiles: ["scripts/.pnpmfile.cjs"],
+  ignoreFiles: [
+    "scripts/.pnpmfile.cjs",
+    // Deliberately added prior to any component or business logic
+    // implementation
+    "src/state/ServiceInterruptionsViewModel.ts",
+  ],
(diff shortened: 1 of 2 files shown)


Johennes (approved):
This looks sane to me. I was briefly debating whether `name` should rather be `type` (as "name" could be misunderstood as a display string)but don't feel strongly in either direction.

robintown:
I also considered `type`. Ultimately I don't think there's much of a risk of confusion, so I'm keeping it like this.

Issue #4102, opened 2026-07-09
Default to list view in landscape mobile group calls


robintown:
Currently, when you join a group call from a mobile phone and turn it into landscape orientation, you get put into this fullscreen spotlight view:

<img height="300" alt="Image" src="https://github.com/user-attachments/assets/ae6124ef-7f03-409d-9718-72890329c3a2" />

Now that in-app calls display edge-to-edge, we've decided that the default layout in this configuration should instead be list view, which makes better use of available screen real estate:

<img height="300" alt="Image" src="https://github.com/user-attachments/assets/0145d123-e9a3-473a-8ac0-23da629fc33f" />

https://www.figma.com/design/pDlJZGBsri47FNTXMnEdXB/Compound-Android-Templates?node-id=2255-17677&t=OeMFR378cIbhfNBq-4

PR #4128, opened 2026-07-28
Use a proper one-on-one layout for landscape mobile calls

robintown:
For https://github.com/element-hq/element-call/issues/4102, we want to change the default layout in landscape mobile calls to grid layout. However, this would have the unintended effect that even one-on-one calls start appearing in grid layout on landscape, because these have so far been relying on the default spotlight layout. Therefore, before we can change any defaults, we should change one-on-one calls to use a layout specifically designed for one-on-one calls, even in landscape.

This results in some minor visual changes:

|Before|After|
|-|-|
|Screen Shot 2026-07-28 at 11 22 52|Screen Shot 2026-07-28 at 10 49 55|
|Screen Shot 2026-07-28 at 09 58 38|Screen Shot 2026-07-28 at 09 48 14|

You may review this commit-by-commit.


BillCarsonFr (approved):
LGTM

PR #4129, opened 2026-07-28
Prefer grid mode in landscape on mobile

### LINKED ISSUE #4102: Default to list view in landscape mobile group calls

robintown:
|Before|After|
|-|-|
|Screen Shot 2026-07-28 at 09 54 21|Screen Shot 2026-07-28 at 09 54 15|

Closes https://github.com/element-hq/element-call/issues/4102

### DIFF
--- a/src/state/CallViewModel/CallViewModel.test.ts
+++ b/src/state/CallViewModel/CallViewModel.test.ts
@@ -568,14 +568,14 @@ describe.each([
                 pipSize: "sm",
               },
               b: {
-                type: "spotlight-expanded",
+                type: "spotlight-landscape",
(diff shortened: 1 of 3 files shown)


BillCarsonFr (approved):
👍

PR #4130, opened 2026-07-28
Only show layout switch when it has an effect on the layout

robintown:
Hides the layout switch from the mobile one-on-one layout and the spotlight portrait layout because it has no effect in these modes.

|Before|After|
|-|-|
|||
|Screen Shot 2026-07-28 at 14 53 05|Screen Shot 2026-07-28 at 15 06 29|
|Screen Shot 2026-07-28 at 15 07 15|Screen Shot 2026-07-28 at 15 06 59|

You may review this commit-by-commit. I first extracted the layout switch into its own view + view model so that I could control it more easily from the CallViewModel.


BillCarsonFr (approved):
approved

PR #4146, opened 2026-08-04
Performance: Enable React Compiler

robintown:
With the React Compiler, our component code is transformed at build time to automatically apply various forms of memoization. This changes the runtime semantics of our code a little bit and therefore could surface new bugs in the next release cycle in case any components fail to follow the rules of React. See https://react.dev/learn/react-compiler for more information.

This results in a small bundle size increase and modest performance gains, but I think it's worth it given the potential for larger performance gains once we tune the component structure a bit more.


src/useEvents.test.tsx
```diff
@@ -28,6 +28,10 @@ test("useTypedEventEmitterState reacts to events", async () => {
   const emitter = new TestEmitter();
 
   const Test: FC = () => {
+    // Disable the React Compiler
```
BillCarsonFr: Is it more that it disables some compiler optimisations?  Or does it disable it complitely? Why only here? doesn't happens in the real app?

vite.config.ts
```diff
@@ -29,6 +29,9 @@ export const vitePluginsConfig = ({
 }: Pick<ConfigEnv, "mode">): UserConfig => {
   const env = loadEnv(mode, process.cwd());
   const plugins: PluginOption[] = [
+    babel({
+      presets: [reactCompilerPreset()],
```
BillCarsonFr, line 33: Ok so this is the thing that enables compiler optimisation right?

BillCarsonFr (approved):
Ok to start using it, just some questions but not blocking

vite.config.ts
robintown: Exactly, it's a Babel plugin

src/useEvents.test.tsx
robintown: This stops the compiler from applying its optimizations to the couple of test components that I've marked with `"use no memo"`. The reason is that they are impacted by the linked compiler bug and the tests were broken otherwise. I will clarify the comments

PR #4151, opened 2026-08-05
Performance: Fit video to frame without polling RTP stats

robintown:
It turns out that the timers which repeatedly poll the RTP stats of each video track all run independently of each other and can add up to a small but constant sink of CPU. Meanwhile we can replace these timers with HTMLVideoElement 'resize' event listeners, which is way more efficient and reacts instantly to orientation changes.

There are then a couple of smaller commits improving the performance of stats polling in general, in case the user still has the advanced media statistics developer option enabled, because why not.

### DIFF
--- a/src/utils/videoFit.ts
+++ b/src/utils/videoFit.ts
@@ -15,7 +15,8 @@ export function autoVideoFit(
   tileAspectRatio: number,
 ): "cover" | "contain" {
   if (Number.isNaN(videoAspectRatio) || Number.isNaN(tileAspectRatio)) {
-    // If we have invalid sizes (e.g. width or height is 0), default to cover to avoid black bars.
+    // If we have invalid sizes (e.g. useMeasure returns 0×0 on an initial render),
+    // default to cover to avoid black bars.
     return "cover";
   }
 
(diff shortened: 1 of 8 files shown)


src/utils/videoFit.ts
```diff
@@ -5,107 +5,26 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
...
-          height: stats.frameHeight,
-        };
-      }),
-    );
+export function autoVideoFit(
+  videoAspectRatio: number,
+  tileAspectRatio: number,
+): "cover" | "contain" {
+  if (Number.isNaN(videoAspectRatio) || Number.isNaN(tileAspectRatio)) {
+    // If we have invalid sizes (e.g. width or height is 0), default to cover to avoid black bars.
```
Johennes: Nit: A zero width and non-zero height would not enter this `if`. Maybe we should clarify the comment slightly?

Johennes (approved):
LGTM 🚀

src/utils/videoFit.ts
robintown: 64a5b2ee5c791974461a40472f6b25f55fd1accf

PR #4152, opened 2026-08-05
Performance: Disable animations when there are a large number of tiles

robintown:
I don't believe this affects the cost of running the useSprings hook, but it at least skips the animation frames that would otherwise follow, so it's worth a try.

I can't easily write tests for anything involving useSprings, so I would like to bypass the coverage requirement here.


Johennes (approved):
:shipit:

PR #4158, opened 2026-08-10
Performance: Avoid re-rendering entire spotlight layout so often

robintown:
This removes a lot of unnecessary jank in spotlight layout when swapping individual speakers in and out of the spotlight tile.

**Before**

Screenshot From 2026-08-06 14-42-28

This shows a test scenario with 250 tiles, where two of the tiles alternate between being the speaker every second. The red segments indicate jank, which happens because the Grid components entirely re-render on each speaker switch.

**After**

Screenshot From 2026-08-06 14-42-25

No visible jank. Only the exact two tiles involved in the speaker switch re-render.


toger5 (approved):
It makes sense that it removes the sub dependency from the `SpotlightLandscapeLayout`.
That react cannot skip rendering if `withIndicators` will update to the same val as it already executes the rendering method...

PR #4172, opened 2026-08-17
Add key rotation participant limit to devtools and config.json

toger5:
The actual limit logic is done in the js-sdk.
This PR adds it to the developer tab and allows configuring the limit value via config.json

Most of the diff is adding the key rotation status to the developer tab.


config/config.sample.json
```diff
@@ -18,6 +18,7 @@
     "membership_event_expiry_ms": 180000000,
     "delayed_leave_event_delay_ms": 18000,
     "delayed_leave_event_restart_ms": 4000,
-    "network_error_retry_ms": 100
+    "network_error_retry_ms": 100,
+    "key_rotation_participant_limit": null
```
robintown: `null` isn't a valid value according to the ConfigOptions.ts types, can't we omit this?

src/room/InCallView.tsx
```diff
@@ -632,6 +632,7 @@ export const InCallView: FC<InCallViewProps> = ({
             onDismiss={(): void => setSettingsOpen(false)}
             tab={settingsTab}
             onTabChange={setSettingsTab}
+            vm={vm}
```
robintown: It's somewhat of an anti-pattern that we're passing the entire CallViewModel into these smaller "organism" components, we should avoid doing more of that if possible

src/settings/DeveloperSettingsTab.test.tsx
```diff
@@ -412,4 +413,79 @@ describe("DeveloperSettingsTab", () => {
...
+            client={client}
+            env={{} as unknown as ImportMetaEnv}
+            vm={mockVm}
+          />
+        </TooltipProvider>,
+      );
+
+      await waitFor(() =>
+        expect(client.doesServerSupportUnstableFeature).toHaveBeenCalled(),
+      );
```
robintown, line 440: This expectation doesn't seem relevant to the test, can the `screen.getByText` expectation be wrapped in `waitFor` instead?

src/state/CallViewModel/localMember/LocalMember.test.ts
```diff
@@ -136,6 +136,67 @@ describe("LocalMembership", () => {
         }),
       );
     });
+
+    it("passes keyRotationParticipantLimit from config to joinRTCSession", () => {
+      const focusFromOlderMembership = {
+        type: "livekit",
+        livekit_service_url: "http://my-oldest-member-service-url.com",
+        livekit_alias: "my-oldest-member-service-alias",
+      };
```
robintown: Since we just removed oldest membership mode, you shouldn't need this

src/state/SessionBehaviors.test.ts
```diff
@@ -0,0 +1,60 @@
...
+import { MatrixRTCSessionEvent } from "matrix-js-sdk/lib/matrixrtc";
+import { EventEmitter } from "events";
+
+import { createKeyRotationSuppressed$ } from "./SessionBehaviors";
+import { ObservableScope } from "./ObservableScope";
+
+describe("SessionBehaviors", () => {
+  describe("createKeyRotationSuppressed$", () => {
+    it("emits initial value from isKeyRotationSuppressed", () => {
+      const scope = new ObservableScope();
```
robintown: If you use `testScope()` instead, that gets you a scope which ends even if the test fails.

src/state/SessionBehaviors.test.ts
```diff
@@ -0,0 +1,60 @@
...
+      const keyRotationSuppressed$ = createKeyRotationSuppressed$(
+        scope,
+        mockSession as any,
+      );
+
+      expect(keyRotationSuppressed$.value).toBe(false);
+      scope.end();
+    });
+
+    it("updates when KeyRotationSuppressedChanged event is emitted", () => {
```
robintown: IMO these could be one test

src/room/InCallView.tsx
toger5: The idea was, that for the devtools tab it might be useful in the future to just be able to access the full vm.
But I agree its a rather large change. So maybe should get its own PR and proper evaluation if that it the right direction.

robintown: Okay well, your call.

toger5: I decided to try to fully use our mvvm pattern. That is a bit unnecassary for now but at least its one more component transitioned. (Missing the storybook part... but still a step in the right direction)

src/settings/DeveloperSettingsTab.test.tsx
toger5: Yes. Was copied from the matrix_rtc_mode tests. Thanks for spotting it!

Johennes (approved):
It looks like we want to get this out today. Given Robin's previous review I only did a superficial skim and a smoke test on the deployment. Judging from these this looks fine to me.

Johennes (approved):

PR #4178, opened 2026-08-18
Performance: Isolate tile store debug info in its own component

robintown:
It's very uncommon to have this debug option enabled, and yet it currently causes the footer to re-render on every layout update, which is a small but avoidable cost.

### DIFF
--- a/src/components/CallFooter.tsx
+++ b/src/components/CallFooter.tsx
@@ -26,6 +26,7 @@ import {
   MediaMuteAndSwitchButton,
   type MenuOptions,
 } from "./MediaMuteAndSwitchButton";
+import { type Behavior } from "../state/Behavior";
 import { type ViewModel } from "../state/ViewModel";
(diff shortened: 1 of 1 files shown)


toger5 (approved):

src/components/CallFooter.tsx
```diff
@@ -283,7 +283,9 @@ export const CallFooter: FC<FooterProps> = ({
           />
         </>
       )}
-      {debugTileLayout ? `Tiles generation: ${tileStoreGeneration}` : undefined}
+      {debugTileLayout ? (
+        <TilesDebugInfo generation$={vm.tileStoreGeneration$} />
+      ) : undefined}
```
toger5, line 288:
```suggestion
      {debugTileLayout !== undefined &&
        (<TilesDebugInfo generation$={vm.tileStoreGeneration$} />)
      }
```
Isnt this the more common pattern?

PR #4180, opened 2026-08-18
Performance: Use PNGs for the background gradients

robintown:
It turns out that for a blurry gradient background, low-res raster images look very convincing when scaled up and interpolated. That means that we can rasterize our gradient SVGs for improved paint performance when resizing the window, while still keeping them quite lightweight (just a couple of kilobytes each).

In my testing in Firefox 153.0.4, where I joined a call and repeatedly maximized and minimized the browser window, this regularly saves a full 120ms of painting time.

**Before (red = jank, light green = paint)**
Screenshot From 2026-08-18 17-07-19

**After**
Screenshot From 2026-08-18 17-07-28

Chromium appears to be quite well optimized to deal with blurred SVGs already, so I couldn't measure any performance improvements using it. Safari may be interesting though, if someone wants to test that for me.

Below is a visual comparison of the SVGs and the PNGs in Firefox.

|SVG|PNG|
|-|-|
|Screenshot 2026-08-18 at 17-14-30 Element Call The Testing Zone before|Screenshot 2026-08-18 at 17-14-16 Element Call The Testing Zone|
|Screenshot From 2026-08-18 17-18-31|Screenshot From 2026-08-18 17-23-45|
|Screenshot From 2026-08-18 17-20-09|Screenshot From 2026-08-18 17-23-07|


src/index.css
```diff
@@ -74,20 +74,23 @@ body {
   -webkit-tap-highlight-color: transparent;
 }
 
-@media (min-height: 330px) {
+@media (min-height: 33px) {
```
Johennes: Was this an intentional change? Realistically we shouldn't really experience such low heights?

Johennes (approved):
I tried this out in Safari against another PR deployment but I couldn't really get many insights from the numbers. Also page loads performance was quite random and erratic, sometimes cached, which complicated the analysis. 😕

The maximum CPU load during loads was generally lower by 10-15% with the PNG though. Also rendering performance when resizing the window feels noticeably smoother with the PNG (even if that in itself is not a very meaningful use case).

Visually I'm unable to spot much of a difference between the PNG and SVG variants.

So in total, it looks like we save _something_ here without compromising the visual appearance. So I'd say let's land this. :shipit:

It'd be really nice if we had some form of page load / rendering performance test that we could use to prevent future regressions. But that can be a separate endeavor.

src/index.css
robintown: Oops, that's left over from me testing things on mobile

PR #4208, opened 2026-08-26
Remove MatrixRTC legacy mode

robintown:
This is the mode in which we sent membership events with the 'oldest membership' transport selection algorithm, which stopped being the default back in version 0.21.0. Users will no longer be able to select this mode in developer settings, and admins will no longer be able to select legacy mode through the config either (hence this is a breaking change). The app will still continue to support *receiving* membership events with the 'oldest membership' transport selection algorithm from others, however.


src/state/CallViewModel/localMember/LocalTransport.ts
```diff
@@ -331,85 +258,6 @@ async function doOpenIdAndJWTFromUrl(
   };
 }
 
-function observeLocalTransportForOldestMembership(
```
toger5, line 334: this is really wonderful to see go away!

toger5 (approved):

PR #4209, opened 2026-08-26
Delegate delayed events using dedicated endpoint

### LINKED ISSUE #4198: Delegate delayed events using dedicated endpoint
The JWT service has a dedicated `/delegate_delayed_event` endpoint nowadays which should be the preferred way to delegate an event.

robintown:
Closes https://github.com/element-hq/element-call/issues/4198


toger5 (commented):
Part of me was wondering if we even want to clean up this path as we plan to also switch to the cs-api based endpoints.

But seeing the changes it is probalby very valuable to first make things less confusing to then introduce the new cs based work.

PR #4232, opened 2026-09-03
Surface screen share failures instead of silently doing nothing

ara4n:
## Content

`toggleScreenSharing` had `.catch(logger.error)` as its only handling, which a never-settling `getDisplayMedia` defeats: the button does nothing and the log says nothing.

- Every toggle now logs when it is requested and when it completes or fails, with the elapsed time (`Screen share start requested` / `completed in 812 ms` / `failed after ...`), so a hanging `getDisplayMedia` is visible in the logs.
- Explicit failures other than the user cancelling the picker (`NotAllowedError`) are exposed on the view model as `screenShareError$` and shown as a non-modal, auto-dismissing "Could not start screen sharing" toast.
- Nothing is inferred from a toggle taking a long time: the user may just be choosing what to share.

## Motivation and context

element-call-rageshakes#17152 (Element Desktop 1.12.26 on Windows, EC 0.22): the user pressed the screen share button 14 times in 25 seconds. The log has nothing but the `toggleScreenSharing called. Switching On` lines and livekit-client's `waiting for pending publication promise timed out`, so the first `getDisplayMedia` never settled, presumably a source picker that never resolved. We cannot investigate that from the current logs, and the user got no feedback. This gives both.

## Screenshots / GIFs

Toast reuses the existing non-modal `Toast` component (same as the reconnecting toast), with the new string `error.screen_share_failed`.

## Tests

- New unit tests for `watchScreenShareToggle`: completes (no report), rejects (report), rejects with `NotAllowedError` (no report).
- `InCallView` render snapshot updated: only React `useId` counters shifted because of the extra Toast.
- `pnpm lint` and `pnpm i18n:check` clean.

## Checklist

- [ ] A linked, pre-approved issue exists for this feature or UI change. (small UI addition attached to a bug fix; happy to file an issue if wanted)
- [x] Pull request includes screenshots or videos for any UI changes. (see above)
- [x] Tests written for new code (and existing touched code where feasible).

(fable genned)

### DIFF
--- a/src/room/InCallView.tsx
+++ b/src/room/InCallView.tsx
@@ -270,6 +270,7 @@ export const InCallView: FC<InCallViewProps> = ({
   const audioParticipants = useBehavior(vm.livekitRoomItems$);
   const participantCount = useBehavior(vm.participantCount$);
   const reconnecting = useBehavior(vm.reconnecting$);
+  const screenShareError = useBehavior(vm.screenShareError$);
   const layout = useBehavior(vm.layout$);
   const edgeToEdge = useBehavior(vm.edgeToEdge$);
   const overflowing = useBehavior(vm.overflowing$);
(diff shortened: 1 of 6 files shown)


src/room/InCallView.tsx
```diff
@@ -399,6 +400,21 @@ export const InCallView: FC<InCallViewProps> = ({
       );
   }
 
+  const onDismissScreenShareToast = useCallback(
+    () => vm.dismissScreenShareError(),
+    [vm],
+  );
+  const screenShareToast = (
+    <Toast
```
Johennes, line 408: Should this use the same background overlay that the reconnecting toast below employs?

Johennes (changes requested):
Looks sane to me. If we can add more coverage, that'd be nice. Otherwise, just a small question / nit.

src/room/InCallView.tsx
Johennes: Ah, actually on second thought we don't need to block the entire screen in this case. 🤦‍♂️

Johennes (approved):

PR #4233, opened 2026-09-03
ElementCall as a react component: refactor + a local dev harness

BillCarsonFr:
# Element Call as an embeddable React component (M1)

**Experimental.** This makes Element Call importable as a React component, so a host application can render a call inside its own page instead of loading Element Call in an iframe as a widget. The standalone app and the widget build keep working unchanged; they now sit on the same seams the component uses.

image

## What a host gets

```tsx
import {
  ElementCall,
  initializeElementCall,
  UserIntent,
  supportedLanguages,
  type ElementCallHandle,
  type ElementCallHostBridge,
} from "@element-hq/element-call-component";
import "@element-hq/element-call-component/style.css";

await initializeElementCall(configJson); // once, before rendering

const call = useRef<ElementCallHandle>(null);
<ElementCall
  ref={call}
  client={matrixClient}
  roomId={roomId}
  intent={UserIntent.JoinExistingCall}
  config={{ hideScreensharing: true }}
  hostBridge={bridge}
  theme="dark"
  language="de"
/>;
await call.current?.setDeviceMute({ audio_enabled: false });
```

- `client`: the host's `MatrixClient`. Element Call authenticates nobody and manages no session of its own. The room must already be known to the client when the component mounts.
- `roomId`: the room to call in.
- `intent` (optional, default `JoinExistingCall`): what the user pressed. Element Call turns it into the same defaults the widget gets from `?intent=` (`configurationForIntent`).
- `config` (optional): `ElementCallConfiguration`, the widget's URL configuration plus `background`, layered over the intent. Compared by value (`useStableValue`), so it can be written inline without restarting the call.
- `hostBridge` (optional): `ElementCallHostBridge`, plain async callbacks for what Element Call tells the host: `contentLoaded`, `notifyJoined`, `notifyHungUp`, `notifyDeviceMute`, `setAlwaysOnScreen`, `close`, plus two facts about the host, `supportsReactions` and `allowJoinUnmutedViaIntent`. Every one is optional and the latest object passed is the one called, so it too can be inline. `close` is what makes Element Call offer a close button on error screens and skip its own post-call screens; without it Element Call behaves as if nobody can dismiss it.
- `hostBridge.allowJoinUnmutedViaIntent` (optional, default false): whether the user may start unmuted when the intent skips the lobby, so they never see their devices before joining. Left out, they start muted and unmute themselves. A host that chose the intent on the user's behalf and knows they expect to be heard at once sets it to true, which is what the widget build does for the Matrix client hosting it; the standalone app answers no.
- `ref`: an `ElementCallHandle` for what the host asks of Element Call: `join(devices)` (for `preload`), `hangUp()`, `setDeviceMute(request)`. Each resolves once Element Call has acted and rejects when nothing is in a position to act.
- `theme` (optional): `light` or `dark`; changes apply at once.
- `language` (optional): a BCP 47 tag from `supportedLanguages`, or something that falls back to one. Defaults to the browser's language. Translations are shared by every Element Call on the page, so the last language set wins for all of them.

The component itself has no rxjs, router or widget API in its public surface: `component/host.ts` adapts the callbacks and the handle to the `HostBridge` the rest of Element Call works with (`useComponentHostBridge`). The theme travels the same `themeChange$` channel a widget's theme does, replayed so late subscribers see it.

Built with `pnpm build:component` into `component/dist` (ES bundle `element-call.js`, one stylesheet `element-call.css`, a chunk per locale, workers under `assets/`, `.d.ts` under `types/`). `react`, `react-dom`, `matrix-js-sdk` and `livekit-client` are externals the host supplies. The package is not published; a host installs it as a git dependency on `component/` (`github:element-hq/element-call#<ref>&path:/component`) and its `prepare` script builds it, which needs `allowBuilds` in the host's pnpm config (see README).

## How Element Call got there

Everything Element Call used to take from *the page it is on* now comes from a provider, with a default that keeps the standalone and widget builds behaving as before.

| Was | Now | Default when no provider |
| --- | --- | --- |
| `widget` module global, `widget.api.transport.send(...)` everywhere | `HostBridge` interface (`src/HostBridge.ts`): `createWidgetHostBridge(widget)` for the widget build, `nullHostBridge` standalone, `useComponentHostBridge` for the component. Carries `close`, `supportsProfileChanges`, `supportsReactions`, `allowJoinUnmutedViaIntent`, `downloadMedia` and the host's requests as observables | `nullHostBridge` |
| `getUrlParams()` reading `window.location` from view models, `MuteStates`, `MediaDevices`, `Publisher`, `PosthogAnalytics`, `sharedKeyManagement`, `useTheme` | `UrlParamsProvider` + `useUrlParams()`; the app derives them from the router (`useUrlParamsFromLocation`), the component builds them from `componentProperties` + intent + config. View models take the values as options (`callViewModelOptionsFromParams`, `AudioOutputOptions`, `CallViewModelOptions.hostBridge`) | parse `window.location` |
| `useNavigate("/")`, `<Link to="/">` in `CallEndedView`, `LobbyView`, `Header`, `ErrorView`, `ClientContext` | `LeaveToHomeContext` + `useLeaveToHome()` / `LeaveToHomeLink`; the app shell supplies the navigation, the component supplies nothing and the call offers no way out. This is what lets the call render without a router | `null` (no way home) |
| `document.body` as the element to theme, decorate (`data-platform`, `data-background`, `no-scroll-body`), portal modals into and listen for keyboard shortcuts on | `RootElementProvider` + `useRootElement()`; `useTheme` stamps `data-element-call-root` and `data-platform` on it; stylesheets key off `[data-element-call-root]` instead of `body` | `document.body` |
| `window.innerWidth/Height` and `useMediaQuery` deciding the layout mode and header size | `windowSize$` is a required `CallViewModelOptions` entry, fed by `observeElementSize$(rootElement)` (ResizeObserver); `useRootSizeMatches(predicate)` for components | none, callers must pass it |
| `@media (max-width: …)` breakpoints, `vw`/`vh` units inside the call | `@container element-call (…)` queries against the root (`container: element-call / size`); `cqw`/`cqh` and percentages | n/a, the root is the page |
| `ClientProvider` finding a client itself (widget or SPA session) | `ClientProvider client={…}` accepts a supplied client and follows a swapped one; only falls back to `initSPA` when none is given | finds one itself |
| global `i18next` singleton, `initReactI18next`, `t` from `i18next` | dedicated instance `src/utils/i18n.ts`, supplied via `<I18nextProvider>`; non-React code uses `i18n.t` | n/a |
| `Config.init()` fetching `config.json` relative to the page | `Config.initWith(config)` for an embedder; a later `init()` is a no-op | fetch |
| `PosthogAnalytics` discovering config from `Config`, the URL and `widget` | `PosthogAnalytics.configure(AnalyticsConfig)`, called by the `Initializer` (`analyticsConfigFromEnvironment`) | analytics off |
| `Initializer.initBeforeReact()` assigning the `widget` global | returns `WidgetHelpers \| null`, passed to `<App widget>` | n/a |

Along the way:

- `GroupCallView` is now `CallView`: a call from lobby to post-call screen, with no routing, auth or alias resolution. It owns `joined` and builds its own `MuteStates` through the new `useMuteStates()` hook, so it no longer takes `widget`/`joined`/`setJoined`/`muteStates` props. `RoomPage` renders it standalone, the component renders it directly.
- The knock / wait-for-invite lobby moved into `KnockLobbyView`, which also owns its own mute state. `usePageTitle` moved up to `RoomPage`: the page title is the page's to set, not the call's.
- `InCallView`, `SettingsModal`, `ErrorView`, `Avatar`, `useTheme`, `MuteStates`, `LocalMember`, `CallViewModel` and the SDK entry point talk to the host only through `HostBridge`. The `ErrorView`/`ErrorPage`/`GroupCallErrorBoundary` `widget` props are gone. Whether the profile may be edited is stated outright (`supportsProfileChanges`) rather than inferred from being a widget.
- `MuteStates` takes the host bridge as a fourth argument and syncs mute state both ways through it. Whether a call that skips the lobby may start unmuted is now the bridge's `allowJoinUnmutedViaIntent` (widget: true; component: the host's choice, default false; standalone: an explicit false) rather than `widget !== null`, so `calculateInitialMuteState` no longer knows about widgets. `Publisher` takes `controlledAudioDevices`; `MediaDevices`/`AppViewModel` take `AudioOutputOptions`; `createCallFooterViewModel` takes `{ showControls, header }`; `createLocalMembership$` takes `hideScreensharing` and `hostBridge`; `enterRTCSession` takes `sendNotificationType`/`callIntent`. Nothing in the state layer reads `getUrlParams()` any more.
- Keyboard shortcuts are listened for on the root element rather than `window`, so two components on one page each hear their own and a key pressed elsewhere on the host's page is ignored. Whether focus is "claimed" is now judged by what has focus (a dialog, a text field) rather than whether it sits inside `#root`.
- `index.css` is split: `base.css` is what Element Call needs wherever it is drawn (tokens, fonts, normalize, Compound, typography, the `[data-element-call-root]` container declaration and background rules); `index.css` keeps what is about owning a page.
- `HeaderLogo` is a `<button>` when there is a home to go to and a plain `<div>` otherwise. `LeaveToHomeLink` renders nothing when there is nowhere to go.
- The component wraps the call in its own Sentry `ErrorBoundary` rendering `ErrorPage`, so a crash shows inside the container instead of unmounting the host's tree, and asks the host to stop keeping it on screen.
- `.root` in `component/ElementCall.module.css` uses `isolation: isolate` and `contain: layout paint` so `position: fixed` modals and the scrim resolve against the container and nothing paints outside it. `InCallView` exposes `data-layout` for tests.
- The component creates its `MediaDevices` in an effect so the device observers end on unmount.

## Component build

- `vite-component.config.ts`: lib build, `cssCodeSplit: false`, `publicDir: false`, externals listed per subpath (rolldown ignores the pattern/callback forms), `react/compiler-runtime` external, `use-sync-external-store/shim` aliased to `react`. `scripts/check-component-externals.mjs` (`pnpm lint:externals`, part of `pnpm lint`) fails if source imports a `react`/`react-dom`/`matrix-js-sdk`/`livekit-client` subpath the list does not name.
- `component/build/scopeStylesToRoot.ts`: PostCSS plugin rewriting every selector so `html`/`body`/`:root` become the root element and everything else gains `:where([data-element-call-root], [data-element-call-root] *)` (zero specificity, so the cascade is unchanged). Runs after `@import` inlining so normalize and Compound are scoped too. CSS modules only have their bare element selectors touched.
- `component/localization.ts`: `import.meta.glob` splits every locale into its own chunk, loaded by an i18next backend the first time a language is asked for; English is bundled in. `supportedLanguages` is derived from the glob.
- `component/tsconfig.build.json` emits declarations under `component/dist/types`; `tsconfig.json` type-checks `component/**`.
- `component/package.json` + a one-line `component/pnpm-workspace.yaml` make `component/` its own pnpm project so a git dependency runs `prepare` (`cd .. && pnpm install --frozen-lockfile && pnpm build:component`).
- `vitePluginsConfig({ html: false })` lets the lib and harness builds share plugins without the app's HTML entry.

## Development harness and tests

- `pnpm dev:component` serves `component/dev` on port 3001: logs in twice against the dev backend and shows two calls side by side in resizable containers with host chrome around them, a theme and language switcher, and a log of both directions of the host bridge (`DevHostBridge`). Stands in for a host; nothing in it is shipped.
- Playwright `playwright/component/component-call.spec.ts` (second `webServer` entry) asserts: two components on one page each show a lobby and, once joined, each sees two tiles; the settings dialog and reaction picker stay inside a 900×640 container; the host's `html`/`body` keep browser defaults while Compound's font features apply inside the root; the bridge reports `contentLoaded`, `notifyJoined`, `setAlwaysOnScreen(true)` and `notifyDeviceMute`; a 300×300 container switches `data-layout` to `pip` and back while the window does not change; and a 300×300 container renders the same shape (layout, tile/footer height, footer buttons) as a 300×300 window.
- Unit tests for `component/host.ts`, `component/localization.ts`, `scopeStylesToRoot`, `HostBridge`, `ClientContext` with a supplied client, `observeElementSize$`, `useRootSize`, `useStableValue`, `MediaDevices` options, `Config.initWith`, `PosthogAnalytics.configure`, `sharedKeyManagement`, the error classes on the dedicated i18n instance, `useCallViewKeyboardShortcuts` scoping, and `UrlParams.isWidget`/`configurationForIntent`. `vitest` runs `component/**/*.test.ts` in the jsdom project; the setup adds a no-op `ResizeObserver` and gives `document.body` a size.

## Behaviour changes to be aware of

- As a widget, `App` now waits for `widget.client` before rendering `ClientProvider` (was: `ClientProvider` awaited it internally). `LoadingPage` shows meanwhile, as before.
- `data-platform`, the theme class and `data-element-call-root` are set by `useTheme` (layout effect on first render) instead of `Initializer.initBeforeReact`. All stylesheets key off `[data-element-call-root]`; `body` still starts `no-theme`.
- `PosthogAnalytics` must be configured before first use; unconfigured it is off. The standalone/widget app configures it from the `Initializer`. The component does not configure it, so analytics are off when embedded.
- The widget bridge's `close()` always stops the transport, whether or not the host acknowledged the `Close` request (before, `GroupCallView` only stopped after a successful send). The SDK's `leave()` now goes through the same `close()`, so it also stops the transport, which it never did before.
- `CallViewModelOptions.windowSize$` is required; `MuteStates` requires a `HostBridge`. The SDK measures `document.body` instead of `window.inner*`.
- `.no-scroll-body` only applies to `body` now (`body.no-scroll-body`), so a host's container is not taken out of its layout.
- Component default background is `solid` (the gradient is a `position: fixed` pseudo-element that would escape the container); the standalone default stays `gradient`.
- A component host without `close` gets the standalone behaviour after a call: `CallEndedView` if analytics are on or the user is passwordless, otherwise nothing.
- The knock lobby and the call each own a `MuteStates` now, one after the other, instead of `RoomPage` holding one across both.
- Whether a lobby-skipping call starts unmuted is decided by `HostBridge.allowJoinUnmutedViaIntent` instead of `widget !== null`. Widget: true, standalone: false, both as before. The component starts muted unless its host opts in with `allowJoinUnmutedViaIntent: true`, in which case it behaves like the widget.
- `Config.initDefault()` (tests) delegates to `initWith({})`, so `Config.init()` in tests no longer fetches.
- Keyboard shortcuts in the component only fire while focus is inside the container, since the listener is on the root rather than the window.

## Known limitations (follow-ups)

- Component analytics, Sentry, rageshake and `seedSettingsFromConfig` are not wired.
- `hostBridge.close` doubles as "the host owns our lifetime" (post-call screens, instant PostHog flush). Marked in `CallView`.
- The stylesheet is about 1.6 MB, most of it Fontsource fonts inlined as data URIs by lib mode.
- Menus near the edge of a small container are clipped by `contain: paint` rather than overflowing into the host.
- `UrlParams` is a misnomer now that params need not come from a URL; renamed later to keep this diff reviewable.
- `useRoomEncryptionSystem` passes an already-prefixed key to `useRoomSharedKey`; pre-existing, preserved and noted.
- The component logs an error and renders an empty container if the client does not know the room at mount; it does not wait for the room to arrive.

> [!IMPORTANT]
> **Features and UI changes require a pre-approved issue.**
> Every PR must have a linked issue
> that a maintainer has reviewed and approved **before you started writing code**.
> PRs that don't meet this requirement will not be reviewed.
> See [CONTRIBUTING.md](https://github.com/element-hq/element-call/blob/livekit/CONTRIBUTING.md) for ElementCall decided for this approach.

## Checklist


toger5 (approved):
This has gotten a couple of LLM reviews, is tested with EW (see: https://github.com/element-hq/element-web/pull/34935) and got refactors to improve the local api.

See commit history for a list of changes.

PR #4235, opened 2026-09-03
Log remote participant and track events per LiveKit connection

ara4n:
## Content

Log remote participant and track lifecycle events from livekit-client on each `Connection`: participant connected/disconnected, track published/unpublished, subscribed/unsubscribed, subscription failed, muted/unmuted (remote participants only, the `Publisher` already logs local ones) and stream state changes (paused/active). Listeners are removed when the connection scope ends.

Example lines, all under the connection's `[RemoteTracks]` child logger:

```
Participant connected: @bob:example.org:DEV111 (PA_xxx)
Published: audio microphone TR_xxx of @bob:example.org:DEV111 muted=false
Subscribed: video camera TR_yyy of @bob:example.org:DEV111 muted=false
Muted: video camera TR_yyy of @bob:example.org:DEV111
Stream paused: video camera TR_yyy of @bob:example.org:DEV111
```

Also log the two derived states a tile depends on, so that a tile stuck on "Waiting for media" can be correlated with the LiveKit events above:

```
[RemoteMatrixLivekitMembers] @bob:example.org:DEV111: LiveKit participant matched (PA_xxx)
[RemoteMatrixLivekitMembers] @bob:example.org:DEV111: LiveKit participant missing
[RemoteUserMedia @bob:example.org:DEV111:0] waitingForMedia=false
```

Finally, `ObservableScope.behavior` now warns (with a stack trace) when a subscriber synchronously re-enters the same behavior while it is delivering a value. It reports the nesting **depth** and logs each new depth once (not just the first re-entry): a single re-entry can only strand a contiguous run of subscribers, so stranding a subscriber in the *middle* of the list requires a nested (deeper) re-entry, and the depth makes that visible. Behaviors created by `splitBehavior` are tagged with their field name (e.g. `participant`) so the warning names which behavior tore. rxjs delivers the nested value to all subscribers and then resumes delivering the older value to the remaining ones, so later subscribers are left permanently stale. That is the only mechanism found so far that can produce a tile which is both "speaking" and "waiting for media" at the same time (see the rageshake below); the stack trace names the re-entrant path. The fix for the hazard itself is in a separate PR.

Also, for two further hypotheses (a member being matched against a participant from the wrong connection, and a scope being torn down synchronously from inside one of its own behavior emissions):

- `ConnectionManagerData.add` warns if it ever merges participants from a second connection to the same SFU URL (upstream dedup should make this unreachable).
- The member generator warns when more than one LiveKit participant matches a member's backend identity, and the per-member match log now names the connection the participant came from.
- `ObservableScope.end()` warns with a stack trace when called while one of the scope's behaviors is mid-delivery, since that unsubscribes the remaining bound subscribers before the value reaches them.

## Motivation and context

Rageshakes currently contain nothing about the state of remote tracks: livekit-client only logs at info level by default, and EC logs nothing about remote publications. A user reporting "I can hear X but their tile shows a muted mic and no video" therefore produces bundles from which the tile's inputs cannot be reconstructed (element-hq/element-call-rageshakes#17279 to #17281). These lines make published/subscribed/muted/paused state per remote member visible without enabling the trace-level LiveKit logging setting.

## Screenshots

N/A (logging only).

## Tests

Unit test added to `Connection.test.ts`, plus an assertion on the participant-match line in `MatrixLivekitMembers.test.ts` and a re-entrancy test in `ObservableScope.test.ts` (which also documents the stale-subscriber hazard), which emits the LiveKit room events on the mocked room and asserts the logged lines, that local mute events are skipped, and that listeners are removed when the scope ends. `pnpm vitest run --project=unit src/state/CallViewModel/remoteMembers/`, `pnpm exec tsc`, `pnpm exec oxlint src` pass.

- [x] New or updated `public`/`exported` symbols have accurate [TSDoc](https://tsdoc.org/) documentation.

## Checklist

- [ ] Pre-approved issue exists for this change.
- [x] PR has a `PR-*` label.

(fable genned)

### DIFF
--- a/src/state/CallViewModel/remoteMembers/MatrixLivekitMembers.ts
+++ b/src/state/CallViewModel/remoteMembers/MatrixLivekitMembers.ts
@@ -111,13 +111,17 @@ export function createRemoteMatrixLivekitMembers$({
             const participants = transport
               ? managerData.getParticipantsForTransport(transport)
               : [];
-            const participant =
-              participants.find(
-                (p) => p.identity == membership.rtcBackendIdentity,
-              ) ?? null;
+            const matches = participants.filter(
+              (p) => p.identity == membership.rtcBackendIdentity,
+            );
+            const participant = matches[0] ?? null;
             const connection = transport
               ? managerData.getConnectionForTransport(transport)
               : null;
+            if (matches.length > 1)
+              logger.warn(
+                `[RemoteMatrixLivekitMembers] ${membership.rtcBackendIdentity}: ${matches.length} LiveKit participants match (sids ${matches.map((p) => p.sid).join(", ")}), using ${participant?.sid}`,
+              );
 
             yield {
               // This could just be the backend identity without the other keys.
@@ -139,8 +143,9 @@ export function createRemoteMatrixLivekitMembers$({
           // Log whether the member could be matched to a LiveKit participant,
           // since a tile shows "waiting for media" for as long as it cannot.
           participant$.pipe(scope.bind()).subscribe((p) => {
+            const url = data$.value.connection?.transport.livekit_service_url;
             logger.info(
-              `[RemoteMatrixLivekitMembers] ${rtcBackendIdentity}: LiveKit participant ${p ? `matched (${p.sid})` : "missing"}`,
+              `[RemoteMatrixLivekitMembers] ${rtcBackendIdentity}: LiveKit participant ${p ? `matched (${p.sid})` : "missing"} on ${url ?? "no connection"}`,
             );
           });
           // will only get called once per backend identity.
--- a/src/state/CallViewModel/remoteMembers/MatrixLivekitMembers.ts
+++ b/src/state/CallViewModel/remoteMembers/MatrixLivekitMembers.ts
@@ -11,6 +11,7 @@ import {
   type LivekitTransportConfig,
 } from "matrix-js-sdk/lib/matrixrtc";
 import { combineLatest, filter, map } from "rxjs";
+import { logger } from "matrix-js-sdk/lib/logger";
 
 import { type Behavior } from "../../Behavior";
 import { type IConnectionManager } from "./ConnectionManager";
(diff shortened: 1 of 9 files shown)


src/state/CallViewModel/remoteMembers/MatrixLivekitMembers.ts
```diff
@@ -133,8 +134,15 @@ export function createRemoteMatrixLivekitMembers$({
...
-        (scope, data$, userId, _deviceId, _memberId, _rtcBackendIdentity) => {
+        (scope, data$, userId, _deviceId, _memberId, rtcBackendIdentity) => {
           const { participant$, ...rest } = scope.splitBehavior(data$);
+          // Log whether the member could be matched to a LiveKit participant,
+          // since a tile shows "waiting for media" for as long as it cannot.
+          participant$.pipe(scope.bind()).subscribe((p) => {
+            logger.info(
+              `[RemoteMatrixLivekitMembers] ${rtcBackendIdentity}: LiveKit participant ${p ? `matched (${p.sid})` : "missing"}`,
+            );
+          });
```
robintown, line 150: I took a moment to confirm that this will not cause logspam. `ObservableScope.splitBehavior` uses `scope.behavior` internally which creates "calm" behaviors, so this should be fine.

src/state/ObservableScope.test.ts
```diff
@@ -237,3 +237,36 @@ describe("Reconcile", () => {
...
+    expect(messages).toEqual([
+      expect.stringContaining("Behavior (tagged) re-entered at depth 2"),
+      expect.stringContaining("Behavior (tagged) re-entered at depth 3"),
+    ]);
+    warn.mock.calls.forEach((c) => expect(c[1]).toEqual(expect.any(String)));
+    // The behavior settles on the newest value while the last subscriber is
+    // left stranded on the oldest.
+    expect(behavior$.value).toBe(3);
+    expect(seen.at(-1)).toBe(1);
+    scope.end();
```
robintown: Best to create these scopes with `testScope` so that they clean up even in a failing test

robintown (approved):
Thank you, this all looks helpful.

src/state/ObservableScope.test.ts
robintown: 054913d17d1ad8e13b0aec221c0a9a2d9f315153

PR #4236, opened 2026-09-03
Queue re-entrant emissions in scope.behavior

ara4n:
## Content

`ObservableScope.behavior` now queues emissions that arrive while it is already delivering a value to its subscribers, so a synchronous re-entrant emission is delivered to every subscriber after the current one, in order. It also logs the first re-entry per behavior with a stack trace (the same warning as element-hq/element-call#4235) so remaining re-entrant paths can be found and cleaned up.

## Motivation and context

A `BehaviorSubject` walks its subscribers synchronously. If a subscriber synchronously causes the same subject to emit again, rxjs delivers the nested value to all subscribers first and then resumes delivering the outer, older value to the remaining subscribers. Those subscribers are left permanently on the stale value while `.value` holds the new one, and nothing re-emits afterwards.

Every EC view model is built on `scope.behavior`, and several derived behaviors read the same source (for a remote tile: media, speaking, video, waiting-for-media all derive from `participant$`). This hazard is the only mechanism found so far that explains a tile in element-hq/element-call-rageshakes#17279 that showed the speaking indicator (participant present) and "Waiting for media" (participant missing) at the same time.

The change is confined to the one primitive. The only observable difference is for the re-entrant emitter itself: its nested value is now delivered after the outer delivery completes rather than in the middle of it, which is the ordering every other subscriber already saw.

## Screenshots

N/A.

## Tests

`ObservableScope.test.ts` gains a test where a subscriber synchronously re-emits; it asserts that both subscribers see `[0, 1, 2]` and that the warning is logged. Without the queue the second subscriber ends on `1`. Full unit suite passes.

- [x] New or updated `public`/`exported` symbols have accurate [TSDoc](https://tsdoc.org/) documentation.

## Checklist

- [ ] Pre-approved issue exists for this change.
- [x] PR has a `PR-*` label.

### DIFF
--- a/src/state/ObservableScope.ts
+++ b/src/state/ObservableScope.ts
@@ -20,6 +20,8 @@ import {
   takeUntil,
 } from "rxjs";
 
+import { logger } from "matrix-js-sdk/lib/logger";
+
 import { type Behavior } from "./Behavior";
 
 type MonoTypeOperator = <T>(o: Observable<T>) => Observable<T>;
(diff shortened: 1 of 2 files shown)


src/state/ObservableScope.ts
```diff
@@ -73,9 +75,34 @@ export class ObservableScope {
...
+            );
+          }
+          return;
+        }
+        delivering = true;
+        try {
+          while (pending.length > 0) subject$.next(pending.shift()!);
+        } finally {
+          delivering = false;
+          pending.length = 0;
```
robintown, line 104: Wild… first time I've seen that trick for clearing an array.

robintown (commented):
That is indeed a thing that can go subtly wrong with RxJS, however I think I would like to see the smoking gun of a user reproducing the bug and logging the re-entry warning from #4235 before we try to paper over recursion issues like this. The reason being:

- There are a great number of other places in the code where we use `BehaviorSubject`s
- If we paper over this one thing, and it does fix the immediate issue, we won't get user reports of the bug anymore
- Therefore we're less likely to ever learn the root cause of the re-entry, which could still be affecting other usages of `BehaviorSubject` in subtle ways
- and ultimately, it's probably a better call to try to use RxJS in a more boring way, because it's not entirely well-defined what should happen in case of recursive expressions, simultaneous events, etc. Short of writing a whole new FRP library with carefully chosen semantics, that way lies madness.

Meanwhile some other suspects for the bug could be:

- `useBehavior` (is it really using `useSyncExternalStore` as React intended? Our "stores" are deliberately not monolithic)
- react-spring (I have seen tiles getting stuck mid-animation in the middle of the screen a number of times, clearly that library is confusing either itself or React or both)

tl;dr not necessarily a red herring, but I'd prefer to capture evidence of problematic recursion before deciding on a fix.

PR #4241, opened 2026-09-07
Log track encryption flag and cryptor events for remote tracks

ara4n:
Follow-up to #4235.

Adds to the `[RemoteTracks]` logging in `Connection.ts`:
- `encrypted=<bool>` (`TrackPublication.isEncrypted`) on every published/subscribed/muted/stream-state line
- `RoomEvent.ParticipantEncryptionStatusChanged` (info)
- `RoomEvent.EncryptionError` (warn; livekit-client already throttles these per cryptor)

Why: LiveKit frame encryption is AES-GCM, so a wrong key can only drop frames (silence / PLC stutter), never produce corrupt audio. The one path that puts ciphertext into the decoder is the subscriber treating a publication as `Encryption_Type.NONE` while the publisher encrypts: the cryptor is bypassed and Opus decodes random payload behind a plaintext TOC byte, which sounds like loud noise bursts. That flag and the cryptor errors are exactly what rageshakes for "only loud sounds" reports (element-hq/element-call-rageshakes#17371) lack today.

Test extended to cover the flag and both events.

### DIFF
--- a/src/state/CallViewModel/remoteMembers/Connection.ts
+++ b/src/state/CallViewModel/remoteMembers/Connection.ts
@@ -243,7 +243,7 @@ export class Connection {
         `Encryption status of ${p?.identity ?? "unknown participant"}: encrypted=${encrypted}`,
       );
     // livekit-client throttles these per cryptor; they indicate frames being
-    // dropped (missing/invalid key), which is the other half of the picture.
+    // dropped (missing/invalid key).
     const onEncryptionError = (error: Error, p?: Participant): void =>
       log.warn(
         `Encryption error for ${p?.identity ?? "unknown participant"}:`,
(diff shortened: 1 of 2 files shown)


src/state/CallViewModel/remoteMembers/Connection.ts
```diff
@@ -232,6 +235,20 @@ export class Connection {
...
     ): void => log.info(`Stream ${state}: ${track(pub, p)}`);
+    const onEncryptionStatusChanged = (
+      encrypted: boolean,
+      p?: Participant,
+    ): void =>
+      log.info(
+        `Encryption status of ${p?.identity ?? "unknown participant"}: encrypted=${encrypted}`,
+      );
+    // livekit-client throttles these per cryptor; they indicate frames being
+    // dropped (missing/invalid key), which is the other half of the picture.
```
Johennes:
```suggestion
    // livekit-client throttles these per cryptor; they indicate frames being
    // dropped (missing/invalid key).
```

It seems hard to related this to the comment 40 lines further up outside of this diff. I think we can just drop this part.

Johennes (approved):

PR #4242, opened 2026-09-07
Adapt delayed leave timings when delegation is available

### LINKED ISSUE #4189: Adapt delayed event timings when delegation is available
This task is to split the current delayed event timing options from Config.ts into two presets: one for when delegation is available, one for when it's not. Then we need to apply the right one based on the responses from the JWT service.

robintown:
Splits the config options for the timings of a delayed leave event into two sets: one for when delegation is available (as you can relax the timings and get more stable calls this way), and another for when it's unavailable (as we must continue to gracefully downgrade even after Matrix 2.0 is fully rolled out).

This works by bluntly hitting the delegation endpoints without auth before joining to check for a 404.

Closes https://github.com/element-hq/element-call/issues/4189

### DIFF
--- a/src/livekit/openIDSFU.ts
+++ b/src/livekit/openIDSFU.ts
@@ -19,7 +19,7 @@ import {
 } from "../utils/errors";
 import { doNetworkOperationWithRetry } from "../utils/matrix";
 import { Config } from "../config/Config";
-import { JwtEndpointVersion } from "../state/CallViewModel/localMember/LocalTransport";
+import { MatrixRTCMode } from "../config/ConfigOptions";
 
 /**
  * Configuration and access tokens provided by the SFU on successful authentication.
@@ -80,11 +80,10 @@ export type OpenIDClientParts = Pick<
  * @param serviceUrl The URL of the livekit SFU service
  * @param roomId The room id used in the jwt request. This is NOT the livekit_alias. The jwt service will provide the alias. It maps matrix room ids <-> Livekit aliases.
  * @param opts Additional options to modify which endpoint with which data will be used to acquire the jwt token.
- * @param opts.forceJwtEndpoint This will use the old jwt endpoint which will create the rtc backend identity based on string concatenation
- * instead of a hash.
+ * @param opts.matrixRTCMode Determines which version of the JWT endpoint to use, which affects whether the
+ * RTC backend identity is based on string concatenation (legacy) or a hash (Matrix 2.0).
  * This function by default uses whatever is possible with the current jwt service installed next to the SFU.
  * For remote connections this does not matter, since we will not publish there we can rely on the newest option.
- * For our own connection we can only use the hashed version if we also send the new matrix2.0 sticky events.
  * @param opts.delayEndpointBaseUrl The URL of the matrix homeserver.
  * @param opts.delayId The delay id used for the jwt service to manage.
  * @param logger optional logger.
@@ -97,7 +96,7 @@ export async function getSFUConfigWithOpenID(
   serviceUrl: string,
   roomId: string,
   opts?: {
-    forceJwtEndpoint?: JwtEndpointVersion;
+    matrixRTCMode?: MatrixRTCMode;
     delayEndpointBaseUrl?: string;
     delayId?: string;
   },
@@ -116,10 +115,9 @@ export async function getSFUConfigWithOpenID(
   logger?.debug("Got openID token", openIdToken);
   let sfuConfig: { url: string; jwt: string } | undefined;
 
-  const tryBothJwtEndpoints = opts?.forceJwtEndpoint === undefined; // This is for SFUs where we do not publish.
+  const tryBothJwtEndpoints = opts?.matrixRTCMode === undefined; // This is for SFUs where we do not publish.
 
-  const forceMatrix2Jwt =
-    opts?.forceJwtEndpoint === JwtEndpointVersion.Matrix_2_0;
+  const forceMatrix2Jwt = opts?.matrixRTCMode === MatrixRTCMode.Matrix_2_0;
 
   // We want to start using the new endpoint (with optional delay delegation)
   // if we can use both or if we are forced to use the new one.
--- a/src/livekit/openIDSFU.ts
+++ b/src/livekit/openIDSFU.ts
@@ -126,9 +126,9 @@ export async function getSFUConfigWithOpenID(
   if (tryBothJwtEndpoints || forceMatrix2Jwt) {
     try {
       logger?.info(
-        `Trying to get JWT with delegation for focus ${serviceUrl}...`,
+        `Trying to get JWT via default endpoint for focus ${serviceUrl}...`,
       );
-      const sfuConfig = await getLiveKitJWTWithDelayDelegation(
+      const sfuConfig = await getLiveKitJWT(
         membership,
         serviceUrl,
         roomId,
@@ -154,7 +154,7 @@ export async function getSFUConfigWithOpenID(
     logger?.info(
       `Trying to get JWT with legacy endpoint for focus ${serviceUrl}...`,
     );
-    sfuConfig = await getLiveKitJWT(
+    sfuConfig = await getLiveKitJWTLegacy(
       membership.deviceId,
       serviceUrl,
       roomId,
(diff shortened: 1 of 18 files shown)


src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -236,11 +247,58 @@ export const createLocalMembership$ = ({
...
+  // The transport that we will advertise in our membership, paired with info as
+  // to whether delayed event delegation is supported
+  const joinParams$ = localTransport$.pipe(
     switchMap((lt) => lt.advertised$),
     catchError(handleTransportError),
     distinctUntilChanged(areLivekitTransportsEqual),
+    switchMap((transport) => {
+      if (transport === null) return of(null);
+      const transportSupportsDelegation = checkDelegationSupport(
+        transport.livekit_service_url + "/delegate_delayed_leave",
```
Johennes, line 302: Since `livekit_service_url` is now deprecated, do we need to also handle it missing entirely?

src/livekit/openIDSFU.ts
```diff
@@ -204,11 +204,10 @@ async function getLiveKitJWT(
   let bodyDalayParts: IDelayParams = {};
   // Also check for empty string
   if (delayId && delayEndpointBaseUrl) {
-    const delayTimeoutMs =
-      Config.get().matrix_rtc_session?.delayed_leave_event_delay_ms;
     bodyDalayParts = {
```
Johennes, line 205: Nit: Not related to your changes but this typo makes me think of the Dalai Lama. 🙈 😅

src/livekit/openIDSFU.ts
Johennes: Nit: Also not your work but the naming is slightly confusing because both `getLiveKitJWT` and `getLiveKitJWTWithDelayDelegation` add the delegation parameters. Maybe we could call these `getLiveKitJWTLegacy` and `getLiveKitJWT` or something similar instead?

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -936,18 +998,15 @@ export function enterRTCSession(
...
-      makeKeyDelay: matrixRtcSessionConfig?.wait_for_key_rotation_ms,
-      membershipEventExpiryMs:
-        matrixRtcSessionConfig?.membership_event_expiry_ms,
+        delayedLeaveTimings.restart_timeout_ms,
+      networkErrorRetryMs: sessionConfig?.network_error_retry_ms,
+      makeKeyDelay: sessionConfig?.wait_for_key_rotation_ms,
+      membershipEventExpiryMs: sessionConfig?.membership_event_expiry_ms,
       keyRotationParticipantLimit:
-        matrixRtcSessionConfig?.key_rotation_participant_limit,
+        sessionConfig?.key_rotation_participant_limit,
```
Johennes: Nit: Looks like this is now non-optional, so I think we can drop the `?`.

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -617,16 +675,20 @@ export const createLocalMembership$ = ({
...
       if (!shouldConnect) return;
+      const sessionConfig = Config.get().matrix_rtc_session;
 
       try {
-        joinMatrixRTC(transport);
+        joinMatrixRTC(
+          joinParams.transport,
+          joinParams.delegationSupported
+            ? sessionConfig.delegated_delayed_leave
+            : sessionConfig.delayed_leave,
```
Johennes, line 697: I wonder if we should thread the result of` joinParams.delegationSupported` through here as well. The delay parameters are currently added to the request inside an `if (delayId && delayEndpointBaseUrl) { ... }`. This makes it seem as if we could end up believing delegation is disabled here (and using the non-delegated timeouts) but then still include the delegation parameters on the request (with the delegated timeouts).

I _think_ this should not happen in practice because delegation in both the old endpoints and the stand-alone endpoint were, I believe, released in the same version of the service. So maybe there's no actual issue here. Just wanted raise the disparity.

Johennes (approved):
I think this looks good with minor comments.

src/livekit/openIDSFU.ts
robintown: Absolutely, I forgot that I did that rename in https://github.com/element-hq/element-call/pull/4209 before abandoning the PR

src/state/CallViewModel/localMember/LocalMember.ts
robintown: I haven't given much thought to that deprecation yet, but I was thinking we should handle it holistically in another PR.

src/livekit/openIDSFU.ts
robintown: My mind went to Art Vandelay :')

src/state/CallViewModel/localMember/LocalMember.ts
robintown: Making sure I've got your comment straight: you mean threading `delegationSupported` through the openIDSFU.ts functions? (Or equivalently keeping `delayId$` `null` in CallViewModel if delegation is not supported?)

I agree that would be cleanest, so I will check if it does not complicate things too much.

src/state/CallViewModel/localMember/LocalMember.ts
Johennes: Ah ok. That sounds fine then.

src/state/CallViewModel/localMember/LocalMember.ts
Johennes: Yeah, exactly that.

src/livekit/openIDSFU.ts
robintown: 09782a4d8497e6c3a6047a9eb06d5c14ac5d28c3

src/state/CallViewModel/localMember/LocalMember.ts
robintown: d38daef56736779b06e1f5f51cf560b2e2306bee

src/state/CallViewModel/localMember/LocalMember.ts
robintown: Sorry, this diff got a good bit larger after unpicking this. The issue is that the delegation calls previously lived in LocalTransport, which LocalMember reads from, so if I had made LocalTransport also read `delegationSupported` from LocalMember, that'd create a nasty cyclic dependency between the two state holders.

In cff8766aa618f8350eb98a71aa49df6c19719847 I move the delegation calls into LocalMember, which I think simplifies things considerably. It also prepares the code a little better for the future in which we use the dedicated delegation endpoint from the CS API.

If you'd prefer to review that in a separate PR, ignore it for now and I'll split it out.

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -168,6 +174,7 @@ interface Props {
  * @param props.muteStates The mute states for video and audio.
  * @param props.matrixRTCSession The matrix RTC session to join.
  * @param props.baseUrl Base URL of the homeserver.
+ * @param props.delayId$ ID of the delayed leave event to delegate to the SFU.
```
Johennes, line 177: Nit: I guess they are obvious but is there any value in documenting the other added parameters here as well?

src/state/CallViewModel/localMember/LocalMember.ts
```diff
@@ -711,6 +722,34 @@ export const createLocalMembership$ = ({
...
 
+  // Delegate delayed leaves to the SFU
+  scope.reconcile(
+    scope.behavior(combineLatest([joinParams$, delayId$])),
+    async ([joinParams, delayId]) => {
+      if (joinParams?.delegationSupported && delayId !== null) {
+        try {
+          // This will technically cause the service to issue a new JWT token,
+          // but it's safe to discard. We're only interested in triggering
+          // delegation.
```
Johennes, line 733: This doesn't feel great but I guess it'll be auto-fixed once we switch to the dedicated delegation endpoint. So seems fine, I think.

Johennes (approved):
This looks fine to me. Thanks a lot for making the changes despite them being more involved than expected.

src/state/CallViewModel/localMember/LocalMember.ts
robintown: Yeah, I found this awkward about the original design too. We were discarding the token before in LocalTransport, just in a different & subtler way.

src/state/CallViewModel/localMember/LocalMember.ts
robintown: I looked at that and then considered them too obvious, yeah. (Aside: I feel we should put docstrings on the props interface rather than doing them like this, for a more consistent IDE experience)

PR #4244, opened 2026-09-08
Prefer the browser default audio output when no output has been chosen

ara4n:
## Content

On browsers that have no `default` pseudo-device (Firefox, Safari), `AudioOutput.available$` appended the virtual default entry (`""`) *after* the physical outputs, so `selectDevice$` picked the first physical device whenever the user had no saved preference, and every remote `<audio>` element got pinned to it with `setSinkId`. This PR lists the virtual default first, so it is the fallback both when nothing has been chosen and when the chosen output disappears, and drops the misleading "Default (first device)" label (the browser default is not necessarily that device; it now reads plain "Default").

Two new tests cover the Firefox-style device list (no `default`): initial selection is the browser default, and the selection falls back to it when the chosen device goes away. `AudioOutput.test.ts` now resets the persisted preference between tests.

## Motivation and context

Rageshake element-hq/element-call-rageshakes#17320: Firefox 154 / Linux / PipeWire with Bluetooth earbuds. EC pinned the output to the earbuds' A2DP sink at connect time; the user's first unmute opened the headset mic, PipeWire switched the earbuds to HFP, the pinned sink id vanished and Firefox left all remote audio elements silent (no error, no fallback, no `devicechange`). After a page refresh, output was on browser default and worked. The system default sink was the laptop, so the explicit pin was the only reason audio was on the earbuds at all.

Pinned sinks are never re-routed by the browser, so EC should not opt users into them by default; Chrome already behaves this way because its `default` device comes first in `enumerateDevices`.

### DIFF
--- a/src/state/AudioOutput.test.ts
+++ b/src/state/AudioOutput.test.ts
@@ -10,6 +10,7 @@ import * as ComponentsCore from "@livekit/components-core";
 
 import { ObservableScope } from "./ObservableScope";
 import { AudioOutput } from "./MediaDevices";
+import { audioOutput as audioOutputSetting } from "../settings/settings";
 import { withTestScheduler } from "../utils/test";
 
 const BT_SPEAKER = {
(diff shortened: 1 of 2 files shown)


Johennes (approved):
Looks sane to me.

PR #4245, opened 2026-09-08
Use the browser default microphone when none has been chosen

ara4n:
## Content

`AudioInput.available$` now gets the same virtual browser-default entry (`""`) that `AudioOutput.available$` has on browsers without a `default` pseudo-device (Firefox, Safari, iOS), listed first so it is what gets selected when the user has not chosen a microphone. The consumers translate `""` into "no deviceId constraint":

- `ConnectionFactory` / `LobbyView`: `deviceId: undefined`, so livekit-client uses its `{ ideal: "default" }` constraint and the browser captures from the OS default input.
- `Publisher.syncDevice`: asks LiveKit for `"default"` as a *non-exact* constraint (browsers that have no such device ignore it), and tracks the ID it last requested because LiveKit reports the physical device it ended up with, which never equals `""`. No switch is issued at join when the room was already created without a deviceId.
- `MediaMuteAndSwitchButton` renders the new `default` label ("Default", existing i18n key).

An explicit choice still pins that device (unchanged), and the selection falls back to the browser default if the chosen device disappears. Tests: new `AudioInput.test.ts` (Firefox/iOS-style list, Chrome-style list, explicit choice + fallback) and a `Publisher` test for the sync path.

## Motivation and context

On iPhone, EC always captured from the built-in microphone even with a wired or Bluetooth headset connected: `selectDevice$` fell back to the first enumerated device and pinned it with an exact `deviceId`, so iOS could neither pick the headset at join nor re-route the capture when a headset was plugged in mid-call. Leaving the constraint out lets the OS route the audio session, which is what every native app gets. Desktop Firefox and Safari benefit the same way (OS default instead of "first device"). Chrome is unaffected because its `default` device already comes first.

Companion to #4244, which does the same for audio output.

### DIFF
--- a/src/state/CallViewModel/remoteMembers/ConnectionFactory.ts
+++ b/src/state/CallViewModel/remoteMembers/ConnectionFactory.ts
@@ -174,7 +174,9 @@ function generateRoomOption({
     publishDefaults,
     audioCaptureDefaults: {
       ...liveKitOptions.audioCaptureDefaults,
-      deviceId: devices.audioInput.selected$.value?.id,
+      // "" is the virtual browser default: leave the constraint out so the
+      // browser captures from the OS default input.
+      deviceId: devices.audioInput.selected$.value?.id || undefined,
       echoCancellation: echoCancellationSetting.getValue(),
       noiseSuppression: noiseSuppressionSetting.getValue(),
       autoGainControl: autoGainControlSetting.getValue(),
--- a/src/state/MediaDevices.ts
+++ b/src/state/MediaDevices.ts
@@ -36,11 +36,14 @@ export type DeviceLabel =
   | { type: "name"; name: string }
   | { type: "number"; number: number };
 
-export type AudioOutputDeviceLabel =
+export type AudioInputDeviceLabel =
   | DeviceLabel
-  | { type: "speaker" }
-  | { type: "earpiece" }
   | { type: "default"; name: string | null };
+
+export type AudioOutputDeviceLabel =
+  | AudioInputDeviceLabel
+  | { type: "speaker" }
+  | { type: "earpiece" };
 
 /**
  * Base selected-device value shared by all media kinds.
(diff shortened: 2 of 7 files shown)


Johennes (approved):
I have added two fixes for things that looked like bugs / gaps to me. @robintown could you maybe give this a second look to ensure I didn't screw up?

The conditional nesting is also quite subtle. I'm not sure if we're missing an obvious / rx way of doing this in a cleaner way?

src/state/MediaDevices.ts
```diff
@@ -203,7 +209,29 @@ class AudioInput implements MediaDevice<DeviceLabel, SelectedAudioInputDevice> {
...
+        // to be enumerated first, which on iOS is the built-in microphone.
+        // Unlike Chrome's "default" device, a stream opened this way does
+        // not follow later changes of the OS default on desktop browsers;
+        // iOS re-routes the audio session itself.
+        if (available.size && !available.has("") && !available.has("default"))
+          return new Map<string, AudioInputDeviceLabel>([
+            ["", { type: "default", name: null }],
+            ...available,
+          ]);
+        return available;
```
robintown, line 232: I think this long-winded language model comment amounts to:

- Chrome already provides a "default" device
- iOS does not, but if you remove the deviceId constraint then it will track the correct input
- With the same trick, Firefox and Safari will not react to changes in the system default (so it's arguably broken)

Therefore it sounds like we should only enable this trick on iOS?

src/state/CallViewModel/remoteMembers/ConnectionFactory.ts
```diff
@@ -174,7 +174,9 @@ function generateRoomOption({
     publishDefaults,
     audioCaptureDefaults: {
       ...liveKitOptions.audioCaptureDefaults,
-      deviceId: devices.audioInput.selected$.value?.id,
+      // "" is the virtual browser default: leave the constraint out so the
+      // browser captures from the OS default input.
+      deviceId: devices.audioInput.selected$.value?.id || undefined,
```
robintown, line 179: Special casing `""` like this seems a bit brittle, maybe the MediaDevices module could return an `undefined` to begin with

PR #4246, opened 2026-09-08
Re-apply a remote participant's playback volume when an audio element is attached

ara4n:
## Content

`RemoteUserMediaViewModel` now watches the participant's microphone track for `TrackEvent.ElementAttached` and pushes the current playback volume to the participant again at that point. It also logs `playbackMuted` changes and each re-application, since "mute for me" left no trace in rageshakes until now. One new test in `MediaViewModel.test.ts`: mute a participant before they have a track, then simulate the track being subscribed and an element attached, and expect `setVolume(0)` again.

## Motivation and context

element-hq/element-call-rageshakes#17381: a user had "mute for me" on a colleague in the same physical room; when the colleague unmuted for the first time, the audio came through at full volume (feedback) while the tile still showed them as muted for me.

Mechanism: the colleague joined muted, so EC had not published a microphone track yet and `RemoteParticipant.setVolume(0)` only recorded the volume in livekit-client's `volumeMap`. On unmute the track was published and subscribed; livekit-client applied `setVolume(0)` to the new `RemoteAudioTrack` before any element was attached (so it only set `elementVolume = 0`), and `RemoteAudioTrack.attach()` then skipped re-applying it because of `if (this.elementVolume)`, which is false for 0. The same happens for any new subscription of a muted-for-me participant's audio (reconnect, rejoin). The upstream one-liner is livekit/client-sdk-js `RemoteAudioTrack.attach` using `!== undefined` (introduced by livekit/client-sdk-js#371, still on their `main`); this PR makes EC correct regardless, and `ElementAttached` is emitted after that check so the re-application lands on the element.

### DIFF
--- a/src/state/media/RemoteUserMediaViewModel.ts
+++ b/src/state/media/RemoteUserMediaViewModel.ts
@@ -6,8 +6,18 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 Please see LICENSE in the repository root for full details.
 */
 
-import { type RemoteParticipant } from "livekit-client";
-import { combineLatest, map, of, switchMap } from "rxjs";
+import { type RemoteParticipant, Track, TrackEvent } from "livekit-client";
+import { observeParticipantMedia } from "@livekit/components-core";
+import {
+  combineLatest,
+  distinctUntilChanged,
+  fromEvent,
+  map,
+  NEVER,
+  of,
+  startWith,
+  switchMap,
+} from "rxjs";
 import { logger } from "matrix-js-sdk/lib/logger";
 
 import { type Behavior } from "../Behavior";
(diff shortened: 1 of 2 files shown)


src/state/media/RemoteUserMediaViewModel.ts
```diff
@@ -60,14 +70,54 @@ export function createRemoteUserMedia(
...
+    pretendToBeDisconnected$,
+    sink$: scope.behavior(
+      combineLatest([
+        inputs.participant$,
+        audioElementAttached$.pipe(startWith(null)),
+      ]).pipe(
+        map(([p, attached]) => (volume) => {
+          if (attached !== null)
+            logger.info(
+              `[RemoteUserMedia ${inputs.id}] re-applying playback volume ${volume} after audio element attach`,
```
robintown, line 107: This runs not only on changes in `attached` but also on changes to `p` or simply `volume`, so logging 're-applying after attach' on every call could be misleading

robintown (commented):
Two suggestions here:

- We also have volume controls for screen shares (see RemoteScreenShareViewModel), perhaps they're affected as well?
- Doing this in the user media module makes this a bit clunky, it would be more idiomatic to locate the re-applying code in the volume controls module, like so:

```diff
diff --git a/src/state/VolumeControls.ts b/src/state/VolumeControls.ts
index beb7ae00..33ed6abc 100644
--- a/src/state/VolumeControls.ts
+++ b/src/state/VolumeControls.ts
@@ -5,7 +5,19 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-Element-Commercial
 Please see LICENSE in the repository root for full details.
 */
 
-import { combineLatest, map, merge, of, Subject, switchMap } from "rxjs";
+import {
+  combineLatest,
+  map,
+  merge,
+  NEVER,
+  type Observable,
+  of,
+  startWith,
+  Subject,
+  switchMap,
+  tap,
+} from "rxjs";
+import { type Logger } from "matrix-js-sdk/lib/logger";
 
 import { type Behavior } from "./Behavior";
 import { type ObservableScope } from "./ObservableScope";
@@ -35,6 +47,13 @@ interface VolumeControlsInputs {
    * requested volume.
    */
   sink$: Behavior<(volume: number) => void>;
+  /**
+   * Emits whenever the volume should be reapplied to the {@link sink$}, even if
+   * unchanged. Set this to deal with e.g. audio elements getting reattached.
+   * @default NEVER
+   */
+  reapply$?: Observable<void>;
+  logger: Logger;
 }
 
 /**
@@ -43,7 +62,12 @@ interface VolumeControlsInputs {
  */
 export function createVolumeControls(
   scope: ObservableScope,
-  { pretendToBeDisconnected$, sink$ }: VolumeControlsInputs,
+  {
+    pretendToBeDisconnected$,
+    sink$,
+    reapply$ = NEVER,
+    logger,
+  }: VolumeControlsInputs,
 ): VolumeControls {
   const toggleMuted$ = new Subject<"toggle mute">();
   const adjustVolume$ = new Subject<number>();
@@ -76,18 +100,29 @@ export function createVolumeControls(
     ),
   );
 
-  // Sync the requested volume with the audio playback module
-  combineLatest([
-    sink$,
-    // The playback volume, taking into account whether we're supposed to
-    // pretend that the audio stream is disconnected (since we don't necessarily
-    // want that to modify the UI state).
-    pretendToBeDisconnected$.pipe(
-      switchMap((disconnected) => (disconnected ? of(0) : playbackVolume$)),
+  // The playback volume, taking into account whether we're supposed to pretend
+  // that the audio stream is disconnected (since we don't necessarily want that
+  // to modify the UI state).
+  const effectivePlaybackVolume$ = pretendToBeDisconnected$.pipe(
+    switchMap((disconnected) => (disconnected ? of(0) : playbackVolume$)),
+  );
+
+  // Stream of requests to sync the volume with the audio playback module
+  const syncRequests$ = combineLatest([sink$, effectivePlaybackVolume$]).pipe(
+    switchMap((syncRequest) =>
+      reapply$.pipe(
+        map(() => syncRequest), // Reapply whenever requested
+        tap(([, volume]) =>
+          logger.info(
+            `Reapplying playback volume ${volume} after audio element attach`,
+          ),
+        ),
+        startWith(syncRequest), // But also sync immediately
+      ),
     ),
-  ])
-    .pipe(scope.bind())
-    .subscribe(([sink, volume]) => sink(volume));
+  );
+
+  syncRequests$.pipe(scope.bind()).subscribe(([sink, volume]) => sink(volume));
 
   return {
     playbackVolume$,
```

Then you can just set `reapply$` to the right event source in the two media VMs.

src/state/media/RemoteUserMediaViewModel.ts
robintown: (my proposed diff above would incidentally fix that)

robintown:
Actually, this is really trivial to fix upstream. I suggest we do that instead: https://github.com/livekit/client-sdk-js/pull/2111

PR #4247, opened 2026-09-09
Fix call.element.dev not updating after branch rename

robintown:


.github/workflows/build.yaml
```diff
@@ -6,7 +6,7 @@ on:
       - opened
       - labeled
   push:
-    branches: [livekit, full-mesh]
+    branches: [main]
```
robintown, line 9: `full-mesh` I consider too dead to keep

Johennes (approved):
Good catch, thanks!

PR #4249, opened 2026-09-09
Stop MatrixAudioRenderer warning on every render

ara4n:
### Content

`LivekitRoomAudioRenderer` now skips the local participant's tracks outright and only logs the "has no matching matrix call member" warning once per unexpected identity (re-armed if that identity later becomes valid), instead of on every render.

### Motivation and context

The renderer re-renders on every active speaker update, and the local participant's own microphone track is never in `validIdentities`, so every EC log was accumulating a warning roughly every 400ms for the whole call (~4.7k lines/hour in the rageshakes looked at today). That is pure noise, buries real "unexpected publisher" warnings, and bloats rageshakes. A genuine unexpected publisher still gets warned about, just once.

https://claude.ai/code/session_015KVLCMTTepFiHw7cBdA9e8


robintown (approved):
Yeah okay, we had something like this before and it beats me why it was removed.

PR #4250, opened 2026-09-09
Don't show the reconnecting overlay for signal-only reconnects

ara4n:
### Content

`connectionDisconnectReason$` in `LocalMember` now treats `ConnectionState.LivekitSignalReconnecting` as connected, so `reconnecting$` (and hence `pretendToBeDisconnected$`, the reconnecting toast, the reconnect analytics event and the hands/reactions pause) only fires for a full `Reconnecting`, a disconnect, or a homeserver problem. `mediaState$` is unchanged.

### Motivation and context

When the SFU drops the signalling WebSocket (seen today across every client on an SFU during what looks like a rolling restart), livekit-client resumes the signal connection while the peer connection stays CONNECTED and media keeps flowing. EC nonetheless flipped `reconnecting$`, which zeroes playback volume on every remote tile, hides remote video and shows the non-dismissable "Reconnecting" overlay, so a 2s blip silenced the whole call. If the peer connection actually goes down livekit moves to `Reconnecting`, which still triggers the overlay as before.

https://claude.ai/code/session_015KVLCMTTepFiHw7cBdA9e8

### DIFF
--- a/src/state/CallViewModel/localMember/LocalMember.test.ts
+++ b/src/state/CallViewModel/localMember/LocalMember.test.ts
@@ -930,6 +930,55 @@ describe("LocalMembership", () => {
       scope.end();
     });
 
+    it("ignores signal-only reconnects", async () => {
+      const scope = new ObservableScope();
+      const trackSpy = vi.spyOn(
+        PosthogAnalytics.instance.eventCallReconnecting,
+        "track",
+      );
+
+      const connectionState$ = new BehaviorSubject<ConnectionState>(
+        ConnectionState.LivekitConnected,
+      );
+      const mutableConnection = {
+        ...connectionTransportAConnected,
+        state$: connectionState$,
+      } as unknown as Connection;
+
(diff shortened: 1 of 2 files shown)


robintown (dismissed):

robintown (commented):
Actually, upon trying it in practice (Firefox 155 on NixOS), not sure this is an improvement. If you pull the plug on your WiFi, LiveKit seems to go a lot more eagerly into the `signalReconnecting` state than into the `reconnecting` state, and in fact, I can't get it to enter `reconnecting` at all until I turn my network connection back on, at which point it bounces erratically between `reconnecting` and `signalReconnecting`. So the result is EC sits there with frozen video but no UI acknowledgement for the most part.

I totally get the desire to have media artificially blocked less, we have some designs which are fixing that by turning the 'reconnecting' toast into a 'service interruptions' component which sits out of the way while providing more detailed info. (https://github.com/element-hq/voip-internal/issues/668)

robintown:
Turning down this suggestion as per above

PR #4254, opened 2026-09-10
Audio quick menu: speakers, level meter and sound-effect volume in the microphone chevron

fkwp:
## Content

The chevron beside the microphone button opens an **Audio controls** menu with three groups: the microphones with a live level meter, the audio outputs, and the sound-effect volume. One commit per slice of the spec, each building and passing the unit suite on its own:

1. **Speaker group in the menu.** Outputs are listed below a separator, the active one marked; selecting switches output and keeps the menu open. With one output or none, as on Safari, the group still names the output in use as a non-selectable row. The outputs come from the same `MediaDevices` helper as the device lists, gated like them: desktop and not PiP during a call, ungated before joining.
2. **Level meter and capture hook.** `useMicrophoneLevel` holds a capture of the selected microphone only while enabled and tears it down on disable, device change or unmount, including mid-acquisition. `AudioLevelMeter` renders the level as bars, focusable and announcing its state to a screen reader only while focused; a denied permission shows a hint, an unopenable microphone a greyed-out meter.
3. **Meter in the menu.** Capture starts when the menu opens and stops when it closes, muted or not; the meter follows the selected microphone and sits at the foot of the microphone group. Keyboard focus gets a border, pointer hover a background only.
4. **Sound-effect volume and scrolling.** A slider reading and writing the same setting as the one in Settings, so the next effect plays at the new level. The menu is bounded to the height Radix reports and only the device lists scroll; heading and slider stay put. Every control is reachable by keyboard: Radix swallows Tab inside menus, so the audio menu keeps Tab from it and lets the browser move focus from the device rows to the meter and the slider.
5. **Surface coverage.** The checks that hold the menu where it belongs: offered before joining on desktop, Android and iOS, and no chevron at all once a call starts on mobile.

Telling pointer use from keyboard use is six lines of local state rather than anything reused, because there is nothing to reuse: Compound exports no modality helper, the repo has no react-aria or focus-visible polyfill, and Radix's `data-highlighted` marks both modalities alike. A CSS-only `:focus-visible:not(:hover)` was rejected: it hides the border under the cursor but still leaves keyboard users with no border at all in Firefox, which never calls a programmatic focus keyboard-driven. The problem belongs to Compound's `Menu`, which focuses items as the pointer moves, and would be better solved there.

No new shared component beyond `AudioLevelMeter`: nothing in Compound or the repo renders a level, and it is reused nowhere else yet. The speaker rows are Compound `MenuItem`s; the read-only row is a plain row styled like one because a disabled `MenuItem` would read as a broken control rather than as information. The slider is the shared `Slider`.

## Motivation and context

Spec: `FEATURES_SPEC/2026-09_Audio_Quick_Menu.md` (status `implement`). One PR by owner decision, recorded in the spec's drift log. Acceptance criteria and their checks:

| AC | Check | Result |
| --- | --- | --- |
| AC1 [FR-001] | `pnpm vitest run --project=unit -t "audio menu is headed Audio controls"` | CI |
| AC2 [FR-002] | `-t "audio menu switches microphone and stays open"` | CI |
| AC3 [FR-003] | `-t "audio menu switches audio output and stays open"` | CI |
| AC4 [FR-004] | `-t "audio menu separates microphone speaker and sound effects groups"` | CI |
| AC5 [FR-005] | `-t "audio menu shows single audio output as non-selectable"` | CI |
| AC6 [FR-005] | manual, Safari: join a call, open the microphone chevron, confirm the speaker group names the output in use and no row responds to a click | _reviewer_ |
| AC7 [FR-006] | `-t "audio output selection is shared between menu and settings"` | CI |
| AC8 [FR-007] | `-t "level indicator follows the microphone signal level"` | CI |
| AC9 [FR-008] | `-t "level indicator responds while muted"` | CI |
| AC10 [FR-009] | `-t "audio menu starts capture on open and stops on close"` | CI |
| AC11 [internal] | `-t "menu capture is released when closed during a device switch"` | CI |
| AC12 [FR-010] | `-t "level indicator follows a microphone change"` | CI |
| AC13 [FR-011] | `-t "level indicator stays idle for a silent microphone"` | CI |
| AC14 [FR-018] | `-t "audio menu hints when microphone permission is denied"` | CI |
| AC15 [D7] | `-t "level indicator greys out when the microphone is unavailable"` | CI |
| AC16 [FR-012] | `-t "sound effects volume from the menu applies to the next effect"` | CI |
| AC17 [FR-013] | `-t "sound effects volume is shared between menu and settings"` | CI |
| AC18 [FR-014] | `-t "audio menu is present pre-join on every platform"` | CI |
| AC19 [FR-015] | `pnpm test:playwright --project=mobile -g "no audio menu during a call on mobile"` | CI |
| AC20 [FR-016] | `-t "audio menu is fully operable from the keyboard"` | CI |
| AC21 [FR-017] | manual, VoiceOver: open the menu, tab to the level indicator, speak, confirm the announced state changes with the signal | _reviewer_ |
| AC22 [D11] | `pnpm vitest run --project=storybook -t "With Many Devices"` | CI |
| AC23 [D12] | `-t "level indicator stays with the microphone list rather than the speakers"` | CI |
| AC24 [D13] | `pnpm test:playwright --project=chromium --project=firefox -g "the focus border follows the keyboard and not the pointer"` | CI |
| AC25 [SC-001] | `pnpm test:playwright --project=chromium -g "audio menu leaves participants visible while switching output"` | CI |
| AC26 [SC-002] | `pnpm test:playwright --project=chromium --project=firefox -g "level indicator moves with microphone input"` | CI, Chromium; skipped on Firefox (see below) |
| AC27 [SC-003] | `pnpm test:playwright --project=chromium --project=firefox -g "audio menu is keyboard operable in a real browser"` | CI, Chromium; skipped on Firefox (see below) |

Firefox reads the meter as zero on the CI runner, which has no audio backend for it. Neither the test device nor the code is at fault: its fake stream carries a 1 kHz tone, and the same test drives the meter to 0.62 against Firefox off the runner, including against the Docker image over plain HTTP that CI serves. AC26 is therefore skipped on Firefox, a deviation from its check. AC27 is skipped there too, because headless Firefox drives Tab unreliably, as `reconnect.spec.ts` records. AC25 skips it because that stream enumerates no audio outputs, which its Chromium-only check does not ask for. All three deviations are in the spec's drift log.

While chasing that, one real defect surfaced: Firefox starts an `AudioContext` suspended, which feeds the analyser silence for a few seconds until it resumes on its own. The capture is now resumed on creation, so the meter answers while the person is still speaking, as D10 requires.

Build targets: nothing here reads the page, and no viewport unit is used. Checked standalone and, through the component harness, as a component. As a component every chevron menu, this one and the untouched camera one alike, renders outside the root element and so outside the scoped stylesheet: no background, no border, and this menu's height bound inert. That is a pre-existing consequence of Compound's `Menu` portalling to `document.body`, needs a portal container Compound does not expose, and is raised separately rather than fixed here. It is in the spec's drift log. Widget mode is an iframe with a document of its own, so the scoping does not apply there.

The first commit is the spec itself, including the drift log this implementation appended, so the claims below can be checked against it in the same diff.

This PR targets `fkwp/feature/audio_quick_menu` rather than `main`, because it builds on two documentation PRs that are in review separately: #4256 (the repo agent contract) and #4255 (the FEATURES_SPEC process). The base branch carries only those two, so this diff is the spec plus the feature. Once they land, the base rebases onto `main` and this retargets there.

## Screenshots / GIFs

Storybook: `CallFooter › With Audio Menu`, `CallFooter › With Single Audio Output`, `CallFooter › Lobby`, and the five `AudioLevelMeter` states.

|Before|After|
|-|-|
|||

## Tests

- Unit: `MediaMuteAndSwitchButton.test.tsx` (audio menu suite), `CallFooter.test.tsx` (heading, selection and volume shared with settings), `CallFooterViewModel.test.ts` (menu behaviours per platform, layout and surface), `useMicrophoneLevel.test.tsx`, `AudioLevelMeter.test.tsx`. Whole suite: 99 files, 784 passing, 9 skipped.
- Storybook: `AudioLevelMeter` (five states) and `CallFooter` (`WithAudioMenu`, `WithSingleAudioOutput`, `WithManyDevices`, lobby stories) with play functions. `WithManyDevices` measures in the browser that the heading and the slider stay on screen while the device lists scroll, which jsdom cannot.
- E2E: `playwright/audio-menu.spec.ts` (four tests), `playwright/mobile/audio-menu-mobile.spec.ts` (mobile project). The fourth test checks at a short window that the menu, its heading and its slider stay inside the viewport, which holds however many devices a browser reports; the deterministic overflow case is the Storybook story, since Chromium and Firefox invent different numbers of fake devices.
- `InCallView` and `LobbyView` snapshots updated: the microphone button now sits in the chevron container wherever the audio menu exists.
- Not measured: the CPU cost of the meter's second capture during a call (spec `## Measurements`).

## Checklist

- [x] Tests written for new code (and existing touched code where feasible).

### DIFF
--- a/FEATURES_SPEC/2026-09_Audio_Quick_Menu.md
+++ b/FEATURES_SPEC/2026-09_Audio_Quick_Menu.md
@@ -1,0 +1,304 @@
+---
+status: implement
+blocked_from: null
+title: Audio quick menu
+owner: fkwp
+signoff:
+  eng: ok
+  design: ok
+  product: ok
+experiment_ref: fkwp/feature/audio_quick_menu_exploration
+base: 029622f07b44f58f8a84e7e167e5b66374479b63
+repos: [element-call]
+specs: []
+supersedes: []
+superseded_by: null
+---
+
+## Problem
+
+- `MediaMuteAndSwitchButton` populates its menu from `audioInput.available$` only. On
+  desktop, audio output is reachable only through `SettingsModal` → audio tab, which covers
+  the call view; the footer's `audioOutputSwitcher$` is an earpiece/loudspeaker toggle that
+  only materialises for controlled (mobile) devices.
+- No input-level indicator exists anywhere in the app: a participant cannot tell whether the
+  selected microphone captures sound without unmuting and asking another participant.
+- `soundEffectVolume` is adjustable only in `SettingsModal`, which replaces the call view.
+- `DeviceSelection` renders nothing when `available.size <= 1`, and `AudioOutput.available$`
+  is emptied on Safari, so those users get no indication of where call audio is going.
+- The in-call chevron is suppressed off desktop and in PiP via `disableDeviceSwitcher$`
+  (`CallFooterViewModel`), while `createLobbyFooterViewModel` passes `constant(false)`; the
+  two surfaces already disagree about when the menu exists.
+
+## Decisions
+
+- D1 — The level indicator captures from the microphone only while the menu is open, muted or
+  not — bounds the capture and the microphone-in-use indicator to a deliberate user action
+  while still answering "is my microphone working?" before unmuting.
+- D2 — The menu appears both before joining and during a call — pre-join is where a device
+  check matters most, in-call is where the settings detour costs most.
+- D3 — Before joining, the full menu appears on every platform, including mobile; during a
+  call it stays web/desktop only — the pre-join chevron already exists everywhere, so
+  suppressing sections there would be new work for no user gain.
+- D4 — Sound-effects volume appears in the menu and stays in settings, both reading and
+  writing one stored value — no migration, and neither audience loses the control it knows.
+- D5 — Where the audio output cannot be changed, the menu still shows the active output as a
+  non-selectable row — knowing where audio goes is useful even when it cannot be redirected.
+- D6 — The menu carries no disclosure that the microphone is sampled while muted — mute state
+  and the operating system's microphone-in-use indicator are independent of each other.
+- D7 — A microphone producing no signal shows an idle indicator; one held exclusively by
+  another application shows a greyed-out indicator — two distinguishable states, the second
+  rare on current operating systems.
+- D8 — A denied microphone permission shows a hint rather than a silent idle indicator —
+  silence is otherwise indistinguishable from a broken microphone.
+- D9 — The level indicator is focusable and announces its state to a screen reader while
+  focused — it is the one control in the menu with no non-visual equivalent.
+- D10 — No numeric latency target for the level indicator; it must move visibly while the
+  person is still speaking — perceived latency is not critical for a presence check.
+- D11 — Only the device lists scroll; the menu's heading and the sound-effect slider stay
+  in place — a list longer than the screen otherwise carries both out of view.
+- D12 — The level indicator belongs to the microphone group and stays visible while that
+  group is on screen, rather than sitting with the output controls — pinned next to the
+  sound-effect slider it reads as one of them.
+- D13 — Keyboard navigation marks the current item with a border, pointer use marks it with
+  a background and no border — the menu moves focus to whatever the pointer is over, so
+  without the distinction both appear at once.
+
+## Acceptance criteria
+
+Test names are the anchor; tests are created with exactly these names.
+
+- AC1 [FR-001] — The chevron beside the microphone button opens a menu headed "Audio controls".
+  - check: `pnpm vitest run --project=unit -t "audio menu is headed Audio controls"`
+- AC2 [FR-002] — The menu lists every available microphone, marks the active one, and selecting
+  another switches input without closing the menu.
+  - check: `pnpm vitest run --project=unit -t "audio menu switches microphone and stays open"`
+- AC3 [FR-003] — The menu lists every available audio output, marks the active one, and
+  selecting another switches output without closing the menu.
+  - check: `pnpm vitest run --project=unit -t "audio menu switches audio output and stays open"`
+- AC4 [FR-004] — The menu renders a separator between the microphone, speaker and
+  sound-effects groups.
+  - check: `pnpm vitest run --project=unit -t "audio menu separates microphone speaker and sound effects groups"`
+- AC5 [FR-005] — With exactly one audio output available, the speaker group still names the
+  active output and offers no selectable row.
+  - check: `pnpm vitest run --project=unit -t "audio menu shows single audio output as non-selectable"`
+- AC6 [FR-005] — In a browser that does not permit choosing an audio output, the speaker group
+  names the active output and offers no selectable row.
+  - check: manual, on Safari: join a call, open the microphone chevron, confirm the speaker
+    group names the output in use and that no row responds to a click.
+- AC7 [FR-006] — An output selected in the menu is shown as active in settings, and an output
+  selected in settings is shown as active in the menu.
+  - check: `pnpm vitest run --project=unit -t "audio output selection is shared between menu and settings"`
+- AC8 [FR-007] — While the menu is open, the level indicator's reported level follows the
+  signal level of the selected microphone.
+  - check: `pnpm vitest run --project=unit -t "level indicator follows the microphone signal level"`
+- AC9 [FR-008] — The level indicator responds to the microphone signal while the participant
+  is muted.
+  - check: `pnpm vitest run --project=unit -t "level indicator responds while muted"`
+- AC10 [FR-009] — Opening the menu starts capturing from the selected microphone; closing it
+  stops the capture.
+  - check: `pnpm vitest run --project=unit -t "audio menu starts capture on open and stops on close"`
+- AC11 [internal] — No microphone capture started by the menu outlives the menu, including
+  when the menu is closed while a device switch is in flight.
+  - check: `pnpm vitest run --project=unit -t "menu capture is released when closed during a device switch"`
+- AC12 [FR-010] — Selecting a different microphone re-points the level indicator at it without
+  closing the menu.
+  - check: `pnpm vitest run --project=unit -t "level indicator follows a microphone change"`
+- AC13 [FR-011] — A selected microphone that produces no signal leaves the indicator in its
+  idle state and produces no error.
+  - check: `pnpm vitest run --project=unit -t "level indicator stays idle for a silent microphone"`
+- AC14 [FR-018] — With microphone permission denied, the menu shows a hint in place of the
+  idle indicator.
+  - check: `pnpm vitest run --project=unit -t "audio menu hints when microphone permission is denied"`
+- AC15 [D7] — When the selected microphone cannot be opened because another application holds
+  it, the indicator renders in a greyed-out state.
+  - check: `pnpm vitest run --project=unit -t "level indicator greys out when the microphone is unavailable"`
+- AC16 [FR-012] — Moving the sound-effects slider changes the level at which the next sound
+  effect plays.
+  - check: `pnpm vitest run --project=unit -t "sound effects volume from the menu applies to the next effect"`
+- AC17 [FR-013] — Sound-effects volume set in the menu is shown in settings and vice versa.
+  - check: `pnpm vitest run --project=unit -t "sound effects volume is shared between menu and settings"`
+- AC18 [FR-014] — The full menu is present before joining on every platform, and during a call
+  on web and desktop.
+  - check: `pnpm vitest run --project=unit -t "audio menu is present pre-join on every platform"`
+- AC19 [FR-015] — During a call on mobile, the microphone chevron offers no menu.
+  - check: `pnpm test:playwright --project=mobile -g "no audio menu during a call on mobile"`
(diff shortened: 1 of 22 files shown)


src/components/useMicrophoneLevel.ts
```diff
@@ -0,0 +1,154 @@
...
+        const samples = new Float32Array(analyser.fftSize);
+
+        let smoothed = 0;
+        const tick = (): void => {
+          if (disposed) return;
+          analyser.getFloatTimeDomainData(samples);
+          const level = amplitudeToLevel(rms(samples));
+          smoothed +=
+            (level - smoothed) * (level > smoothed ? ATTACK : RELEASE);
+          setState({ type: "active", level: smoothed });
```
Johennes: This makes the meter re-render the entire menu ~60×/second, even in silence. Here `setState({ type: "active", level: smoothed })` is called with a fresh object on every animation frame, and the hook lives in `MediaMuteAndSwitchButton`, so the whole menu subtree (device rows, speaker group, slider) reconciles every frame while the menu is open — including when the level hasn't changed at all (silence keeps smoothed at 0 but the object identity still changes).

The spec already flags the capture's CPU cost as unmeasured; this multiplies it with per-frame React work during a live call.

Resolution: bail out when the displayed value wouldn't change — quantize smoothed to bar granularity `(Math.round(level * BAR_COUNT)` (mirrors what `AudioLevelMeter.tsx` renders) and use the functional `setState(prev => …)` form to return the previous object when equal; or move the hook call into a small wrapper component so only the meter subtree re-renders.

FEATURES_SPEC/2026-09_Audio_Quick_Menu.product.md
```diff
@@ -0,0 +1,100 @@
...
+
+1. **Given** the menu is open, **When** the participant moves the sound-effects slider, **Then** the next sound effect plays at the new level.
+2. **Given** the participant changed the level in the menu, **When** they open settings, **Then** the settings slider shows the same level, and changing it there is reflected back in the menu.
+
+---
+
+### Edge Cases
+
+- The menu is opened while muted: the operating system's "microphone in use" indicator turns on even though the participant is muted. No disclosure in the menu is required — that indicator is expected to behave the same whether the participant is muted or not, and mute state and microphone sampling are independent of one another.
+- The active microphone or speaker is unplugged while the menu is open — the menu must settle on a device that exists rather than showing a stale selection.
```
Johennes, line 65: This isn't covered in the functional requirements and thus also not in the ACs. Might be intentional though.

FEATURES_SPEC/2026-09_Audio_Quick_Menu.product.md
```diff
@@ -0,0 +1,100 @@
...
+- No microphone is present at all: the microphone section and the level indicator have nothing to show.
```
Johennes: This is also not covered in FRs and ACs and I think it's actually violated in the implementation. With no mic, the menu still renders a meter announcing "No sound detected" — indistinguishable from a working-but-silent microphone.

src/components/MediaMuteAndSwitchButton.tsx
```diff
@@ -239,3 +342,154 @@ export const MediaMuteAndSwitchButton: FC<MediaMuteAndSwitchButtonProps> = ({
...
+      case "number":
+        return t("settings.devices.speaker_numbered", { n: label.number });
+      case "speaker":
+        return t("settings.devices.loudspeaker");
+      case "earpiece":
+        return t("settings.devices.handset");
+      case "default":
+        return label.name === null
+          ? t("settings.devices.default")
+          : `${t("settings.devices.default")} (${label.name})`;
```
Johennes: This should use `settings.devices.default_named` rather than string concatenation for proper translatability.

src/components/useMicrophoneLevel.ts
```diff
@@ -0,0 +1,154 @@
...
+        };
+        setState({ type: "active", level: 0 });
+        frame = requestAnimationFrame(tick);
+      })
+      .catch((e: unknown) => {
+        if (disposed) return;
+        rootLogger
+          .getChild("[useMicrophoneLevel]")
+          .warn("Could not open microphone for level metering", e);
+        setState(classifyError(e));
```
Johennes, line 153: If new `AudioContext()` or `createMediaStreamSource` throws after `getUserMedia` resolved, this sets the state to `unavailable` but never stops the already-assigned stream, so the OS mic-in-use light stays on behind a greyed meter until the effect cleans up. Bounded by menu lifetime, hence low but would still be good to stop the stream (and close context) in the catch as well.

FEATURES_SPEC/2026-09_Audio_Quick_Menu.md
```diff
@@ -0,0 +1,304 @@
...
+- AC20 [FR-016] — Every control in the menu — device rows, level indicator and sound-effects
+  slider — is reachable and operable from the keyboard alone.
+  - check: `pnpm vitest run --project=unit -t "audio menu is fully operable from the keyboard"`
+- AC21 [FR-017] — The level indicator takes focus and announces its current state to a screen
+  reader while focused.
+  - check: manual, with VoiceOver active: open the menu, tab to the level indicator, speak,
+    confirm the announced state changes with the signal.
+- AC22 [D11] — With more devices than fit on screen, the heading and the sound-effect slider
+  stay visible while the device lists scroll.
+  - check: `pnpm vitest run --project=unit -t "audio menu keeps its title and volume slider out of the scrolling area"`
```
Johennes: The drift log itself records that the jsdom test passes regardless of layout and that the real verification moved to the `WithManyDevices` story and the "audio menu stays inside a short window" e2e — and leaves "owner to decide whether AC22's check names it" unresolved. This should be updated to name the story test (and optionally the e2e) so the anchor points at the test that would actually fail.

src/components/useMicrophoneLevel.ts
```diff
@@ -0,0 +1,154 @@
...
+        context.createMediaStreamSource(acquired).connect(analyser);
+        const samples = new Float32Array(analyser.fftSize);
+
+        let smoothed = 0;
+        const tick = (): void => {
+          if (disposed) return;
+          analyser.getFloatTimeDomainData(samples);
+          const level = amplitudeToLevel(rms(samples));
+          smoothed +=
+            (level - smoothed) * (level > smoothed ? ATTACK : RELEASE);
```
Johennes, line 127: This appears to be frame-rate dependent so the decay runs twice as fast on a 120 Hz display compared to a 60 Hz display. Could either be fixed by scaling with elapsed time or accepted as a compromise and documented accordingly in the code.

FEATURES_SPEC/2026-09_Audio_Quick_Menu.md
```diff
@@ -0,0 +1,304 @@
...
+  sound-effect slider it reads as one of them.
+- D13 — Keyboard navigation marks the current item with a border, pointer use marks it with
+  a background and no border — the menu moves focus to whatever the pointer is over, so
+  without the distinction both appear at once.
+
+## Acceptance criteria
+
+Test names are the anchor; tests are created with exactly these names.
+
+- AC1 [FR-001] — The chevron beside the microphone button opens a menu headed "Audio controls".
```
Johennes, line 71: This doesn't align with the title casing used for the camera menu which uses "Camera **S**ource"

Johennes (approved):
Having done a time-boxed review, I think this looks and works in a sane way except for a few comments.

src/components/useMicrophoneLevel.ts
fkwp: I think that is audio device sample rate dependent which is 44.1kHz or 48kHz, I think we do not need to special case those...

fkwp: The other question is here since its FFT frames if we want to apply a high pass filter to remove rumbling from the meter volume

Johennes: Is it really based on the audio frame rate? I thought `requestAnimationFrame` ran on the display frame rate.

fkwp: true -- for now Id say lets not over engineer it.. its an indicator about volume level not a precise measurement jig.

src/components/useMicrophoneLevel.ts
fkwp: https://github.com/element-hq/element-call/pull/4254/commits/eac200a727987326b86644565cdccd15a96aa707

src/components/MediaMuteAndSwitchButton.tsx
fkwp: https://github.com/element-hq/element-call/pull/4254/commits/888d66332521f4f9b5e6d27838809a25d561a723

src/components/useMicrophoneLevel.ts
fkwp: https://github.com/element-hq/element-call/pull/4254/commits/fb9fd5880c0bd7160b63e6ab6e6ff521f339af13

fkwp:
superseeded by https://github.com/element-hq/element-call/pull/4275

PR #4255, opened 2026-09-10
Add the FEATURES_SPEC process for turning explorations into specs and PRs

fkwp:
## Content

Adds the `FEATURES_SPEC/` process: how an exploration is distilled into a feature spec, and how a spec is turned into PRs against a `main` nobody has seen yet.

- `FEATURES_SPEC/AGENTS.md`: the spec format, its state machine, traceability to the product spec, the drift check, slicing rules and hard rules.
- `FEATURES_SPEC/_TEMPLATE.product.md`: the product spec sidecar that a brainstorming session fills in.
- `FEATURES_SPEC/` is excluded from `oxfmt`: a product spec is immutable once a spec leaves `draft`, and the process documents are prose written in their own style.

The first spec using it is the audio quick menu, #4254.

## Motivation and context

Documentation and process only. The root `AGENTS.md` (separate PR) points here for "implement from a feature spec".

## Screenshots / GIFs

None, documentation.

## Tests

None, documentation. `pnpm format:check` passes with the exclusion.

## Checklist

### DIFF
--- a/FEATURES_SPEC/AGENTS.md
+++ b/FEATURES_SPEC/AGENTS.md
@@ -1,0 +1,415 @@
+# FEATURE_SPEC — Agent Instructions
+
+This folder holds **feature specs distilled from experiments**. A spec is the durable
+artifact of an exploration; the exploration branch is not. Each spec must contain
+enough information for an agent to produce mergeable PRs against a *current* `main`
+without access to the original branch.
+
+## What this folder is not
+
+- Not a design-doc archive. Specs that are not intended to be implemented do not
+  belong here.
+- Not a place for narrative. Prose that does not constrain the implementation or its
+  verification should be cut.
+- Not a substitute for the PR description. The spec explains *why* and *what must be
+  true*; the PR explains *this diff*.
+
+## File naming
+
+`YYYY-MM_Snake_Case_Title.md` — e.g. `2026-09_Raise_Hand_Improvements.md`
+
+The filename is immutable once committed. Renames break links from PRs, issues and
+the drift log. If scope changes fundamentally, write a new spec and set
+`supersedes:` / `superseded_by:`.
+
+A sidecar `YYYY-MM_Snake_Case_Title.product.md` holds the product spec — user stories,
+functional requirements, success criteria. Copy `_TEMPLATE.product.md`. Product owns
+it; see "Brainstorming: the product spec".
+
+An optional sidecar `YYYY-MM_Snake_Case_Title.notes.md` holds lessons learned,
+narrative, dead-end logs and anything else that is worth keeping but does not
+constrain the implementation. Rules:
+
+- The sidecar is **never load-bearing**. Agents do not read it to derive
+  implementation, and nothing in it may contradict or extend the spec. If a note
+  constrains the work, it belongs in `## Decisions`, `## Rejected alternatives` or
+  `## Measurements` instead.
+- Agents may append to it (an observation from a drift check, a wrong turn during
+  implementation) and must never rewrite or delete existing content.
+- An agent that is `blocked` may read the sidecar to draft an open question, but
+  quotes what it found rather than acting on it.
+- No sidecar is required. An empty one is worse than none.
+
+## Style: facts only
+
+Applies to the spec file only. The product spec is prose by design and the notes
+sidecar is unconstrained. The spec is re-read on every regeneration, where prose costs context
+and dilutes the contract, so:
+
+- Bullets, not paragraphs. Declarative present tense. No motivation, no background,
+  no summary of what the code currently does beyond locating the change.
+- Name things instead of describing them — a path, a symbol, an event type, a command,
+  an event id. A reference is shorter and more precise than a sentence about it. Which
+  kind of reference is allowed depends on the section; see "Reference durability".
+- One line per decision plus at most one sentence of rationale. If the rationale needs
+  more, the extra goes in the sidecar.
+- No judgement adjectives (clean, elegant, proper, obviously) and no hedging (should
+  probably, might be nice). Either it is a constraint or it is not in the spec.
+- `## Problem` is at most five bullets.
+- Target under 200 lines for the whole spec; over 300 lines, split the feature or move
+  material to the sidecar.
+
+The test for any line: does it constrain the implementation, constrain its
+verification, or record a measured fact? If not, delete it or move it to the sidecar.
+
+## Reference durability
+
+The spec must stay implementable against a `main` nobody has seen yet, so which
+references are allowed depends on how fast they rot. Most to least durable:
+
+1. Protocol identifiers — `m.reaction`, `m.call.member`, `m.relates_to`, a reaction
+   key. MSC-bound; effectively permanent.
+2. Observable behaviour. Permanent.
+3. Concept names — membership, participation, session.
+4. Test names. Survive file moves and renames, because the test moves with the code.
+5. Exported symbols. Survive refactors sometimes.
+6. File paths. Rot fast.
+7. Line numbers. Never.
+
+Per section:
+
+- `## Problem`, `## Measurements`, `## Slicing plan`, `## Drift log` — any reference,
+  paths and internal symbols included. These describe the codebase **as of `base`**. A
+  reference going stale here is a finding for the drift check, not a defect, and
+  removing the concrete references would leave the drift check with nothing to verify.
+- `## Decisions`, `## Acceptance criteria` — levels 1 to 4 only. No file paths, no
+  internal symbols. This is the part that has to hold against a future `main`.
+- An acceptance criterion that is not user-observable is tagged `[internal]` and
+  anchored by **its test name**, not by the symbol under test. The test name moves with
+  a refactor; the symbol does not.
+
+`[internal]` ACs are legitimate and expected: an invariant that no current consumer can
+observe is exactly what protects the consumer that does not exist yet. Tagging them
+keeps visible which ACs defend a product requirement and which defend a code
+invariant.
+
+## Required frontmatter
+
+```yaml
+---
+status: draft            # see state machine below
+blocked_from: null       # state to return to when leaving `blocked`
+title: Raise hand improvements
+owner: <human handle>    # a human, always
+signoff:                 # join condition for leaving exploration.signoff
+  eng: pending
+  design: pending
+  product: pending
+experiment_ref: <branch name or commit sha of the spike, if any>
+base: <sha of main the spec was validated against>
+repos: [matrix-js-sdk, element-call]    # in dependency order
+specs: [MSC4143]         # protocol documents touched, if any
+supersedes: []
+superseded_by: null
+---
+```
+
+`base` is mandatory and is what makes drift detectable. `owner` is a human even when
+every section was drafted by an agent.
+
+## Brainstorming: the product spec
+
+The usual entry point is a human handing an agent a one-line feature idea. The output
+of that session is `<stem>.product.md`, not code and not a feature spec. Copy
+`_TEMPLATE.product.md` and fill it.
+
+Ownership: an agent may create and fill this file during `brainstorm`. From `draft`
+onwards **product owns it and agents do not edit it** — not to fix wording, not to
+resolve a marker, not to reconcile it with the code. Agents raise mismatches in the
+feature spec's `## Open questions`.
+
+If product's source of truth lives outside the repo (a tracker, a design doc), put its
+URL in the header `**Source**:` line and keep the requirement text and ids in the repo
+anyway. The ids are what the feature spec references; a link alone cannot be traced.
+
+Agent behaviour during the session:
+
+- **Ask, do not assume.** An unanswered question is recorded as
+  `[NEEDS CLARIFICATION: <the question>]`, inline where the answer would go. Filling a
+  gap with a plausible guess is the main failure mode of this state — the marker is a
+  deliverable, not a defect.
+- Ask in one round of at most five questions, ordered by how much the answer changes
+  the shape of the feature. Do not interview.
+- **No implementation in the product spec.** No file paths, symbols, libraries, event
+  types, API shapes or framework names. If a requirement cannot be stated without them,
+  it is not a product requirement.
+- Every user story must be independently deployable and demonstrable on its own. A
+  story that only makes sense once another story ships is not a story; merge it.
+- More than four priority levels means more than one feature. Split before `draft`.
+- Never invent an SC number. Ask for the target, or mark it
+  `[NEEDS CLARIFICATION]`.
+
+For protocol and infrastructure work, success criteria may be expressed as observable
+client behaviour or a wire-level invariant ("a client joining mid-session converges
+within one sync"), but still never as a code path or an internal API.
+
+`FR-###` and `SC-###` ids are immutable once assigned. Product may revise the document
+at any time, including after `in_production`, but a changed requirement gets a **new
+id** and the old one is marked superseded — never edited in place. Renumbering silently
+breaks every AC that references it, and there is no way to detect that afterwards.
+
+A new or changed FR after `implement` does not amend the current spec's scope. It is
+either a new feature spec or the next iteration; the mapping table simply shows an
+unmapped FR, which is the intended signal.
+
+## Traceability
+
+The product spec states intent; the feature spec states what must be true and how it is
+checked. They are kept in sync by ids, not by duplication.
+
+- Every `FR-###` maps to at least one acceptance criterion. Tag the AC with its
+  origin: `AC2 [FR-003] — <observable behaviour>`.
+- Every `SC-###` maps to an entry in `## Measurements` with a method. An SC with no
+  measurement blocks `exploration.signoff`.
+- Every user story maps to at least one slice. **Slice order follows story priority**
+  (P1 first) unless a technical dependency overrides it, and the override is recorded
+  in the slicing plan with its reason.
+- An AC with no `FR` origin is allowed and expected — implementation-level invariants
+  have no counterpart in the product spec. The converse is not: an unmapped FR is an
+  incomplete spec.
+- Agents do not resolve a mismatch between the two. The product spec wins on intent,
+  the feature spec wins on verification; a genuine contradiction is `blocked`.
+
+## Status state machine
+
+Transitions are explicit; an agent may perform only the ones marked (agent).
+
+```
+brainstorm ─► draft ─► exploration ─► implement ─► qa ─► in_production
+                           │              │        │
+                           └──────────────┴────────┴──► blocked ──► (blocked_from)
+                                                           │
+                                                           └──► abandoned
+```
+
+Exploration has sub-states, written dotted in `status`:
+
+```
+exploration.scoping ─► exploration.iterating ─► exploration.signoff
+        ▲                       │                      │
+        └───────────────────────┴──────────────────────┘
+```
+
+- `brainstorm` → `draft`: human. Requires `<stem>.product.md` to exist with at least one
+  P1 story carrying acceptance scenarios, every FR and SC id assigned, and every
+  remaining `[NEEDS CLARIFICATION]` copied into the feature spec's `## Open questions`.
+- `draft` → `exploration.scoping`: human. Requires `## Problem` and `## Out of scope`.
+  Scoping before the branch exists is what keeps exploration from becoming the
+  long-lived branch this folder replaces.
+- `exploration.scoping` → `.iterating`: human. `experiment_ref` must name the branch.
+- `.iterating` → `.signoff`: human. Requires `## Decisions`, `## Acceptance criteria`,
+  `## Rejected alternatives` and `## Measurements` to be complete, no
+  `[NEEDS CLARIFICATION]` left in the product spec, and full traceability (below).
+- `.signoff` → `implement`: **only when all three sign-offs in `signoff:` are `ok`.**
+  This is a join, not a sequence — a single reviewer cannot advance it. On entering
+  `implement`, the exploration branch stops being a base: it is reference material,
+  and implementation branches off current `main` (see "Turning a spec into PRs").
+- `.signoff` → `.iterating`: human, on any sign-off returning changes. Reset the
+  rejecting role to `pending`; leave the others as they are.
+- `implement` → `qa`: (agent) when every acceptance criterion is covered, as recorded
+  in `## PRs`, and the feature flag is enabled in the QA environment.
+- `qa` → `in_production`: human. The flag is on by default in production.
+- `qa` → `implement`: (agent) when a QA finding needs code. See "QA feedback".
+- any active state → `blocked`: (agent) when drift or an unresolved question prevents
+  progress. Requires a drift-log entry and `blocked_from: <state>`. Do not guess and
+  continue. Leaving `blocked` returns to `blocked_from`, never to a fixed state.
+- any state → `abandoned`: human only.
+
+`in_production` is terminal. A production regression does not reopen the spec: it
+either is an ordinary bug fix with no spec, or it is a new spec. The acceptance
+criteria remain as the regression contract (see "QA feedback").
+
+## QA feedback
+
+The point of the `qa` state is not to record that testing happened. It is to make the
+spec a better regeneration source than it was before implementation. So:
+
+- Every reproduced QA finding becomes **either** a new acceptance criterion **or** a
+  new entry in `## Rejected alternatives`. Never a prose note, never only a bug
+  tracker link.
+- (agent) may **append** ACs during `qa`, with provenance:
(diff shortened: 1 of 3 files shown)


FEATURES_SPEC/_TEMPLATE.product.md
```diff
@@ -0,0 +1,117 @@
...
+[Describe this user journey in plain language]
+
+**Why this priority**: [Explain the value and why it has this priority level]
+
+**Independent Test**: [Describe how this can be tested independently - e.g., "Can be fully tested by [specific action] and delivers [specific value]"]
+
+**Acceptance Scenarios**:
+
+1. **Given** [initial state], **When** [action], **Then** [expected outcome]
+2. **Given** [initial state], **When** [action], **Then** [expected outcome]
```
Johennes, line 36: How do these two differ? They seem more or less equivalent to me just that one uses free text and the other a fixed grammar?

FEATURES_SPEC/_TEMPLATE.product.md
```diff
@@ -0,0 +1,117 @@
...
+
+---
+
+### User Story 2 - [Brief Title] (Priority: P2)
```
Johennes, line 40: Do we need to repeat user story 2 and 3 given the "[Add more user stories as needed, each with an assigned priority]" below them?

FEATURES_SPEC/_TEMPLATE.product.md
```diff
@@ -0,0 +1,117 @@
...
+---
+
+[Add more user stories as needed, each with an assigned priority]
+
+### Edge Cases
+
+<!--
+  ACTION REQUIRED: The content in this section represents placeholders.
+  Fill them out with the right edge cases.
+-->
```
Johennes, line 75: It feels like that applies to all sections but it's only called out in three of them. In face, isn't the whole point of a template to contain structure and placeholders? I wonder if we need to spell this out at all. It feels clear enough given the `[placeholder]` syntax.

FEATURES_SPEC/_TEMPLATE.product.md
```diff
@@ -0,0 +1,117 @@
...
+
+*Example of marking unclear requirements:*
+
+- **FR-006**: System MUST authenticate users via [NEEDS CLARIFICATION: auth method not specified - email/password, SSO, OAuth?]
+- **FR-007**: System MUST retain user data for [NEEDS CLARIFICATION: retention period not specified]
+
+### Key Entities *(include if feature involves data)*
+
+- **[Entity 1]**: [What it represents, key attributes without implementation]
+- **[Entity 2]**: [What it represents, relationships to other entities]
```
Johennes, line 103: Stupid non-PM question: Could it be that the order is wrong here? I think you would first define the key entities of the problem, then follow on with functional requirements for what needs to be achieved with these entities and then build user stories out of these? But the template has these three sections in exactly the opposite order.

FEATURES_SPEC/_TEMPLATE.product.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Success Criteria *(mandatory)*
+
+<!--
+  ACTION REQUIRED: Define measurable success criteria.
+  These must be technology-agnostic and measurable.
+-->
+
+### Measurable Outcomes
```
Johennes, line 112: I guess we should always have measurable outcomes as otherwise we cannot define success. So we might just fold the two nested headings into a single one a la "## Measurable success criteria".

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+
+- The sidecar is **never load-bearing**. Agents do not read it to derive
+  implementation, and nothing in it may contradict or extend the spec. If a note
+  constrains the work, it belongs in `## Decisions`, `## Rejected alternatives` or
+  `## Measurements` instead.
+- Agents may append to it (an observation from a drift check, a wrong turn during
+  implementation) and must never rewrite or delete existing content.
+- An agent that is `blocked` may read the sidecar to draft an open question, but
+  quotes what it found rather than acting on it.
+- No sidecar is required. An empty one is worse than none.
```
Johennes, line 41: These seems superfluous because the sidecar is already deemed "optional" above.

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+
+## Style: facts only
```
Johennes, line 43: Should we maybe also combine this with something like https://github.com/juliusbrussee/caveman?

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+
+## Reference durability
+
+The spec must stay implementable against a `main` nobody has seen yet, so which
+references are allowed depends on how fast they rot. Most to least durable:
+
+1. Protocol identifiers — `m.reaction`, `m.call.member`, `m.relates_to`, a reaction
+   key. MSC-bound; effectively permanent.
+2. Observable behaviour. Permanent.
+3. Concept names — membership, participation, session.
```
Johennes, line 73: Do we need to define those somewhere? Otherwise every agent would have to start inferring them from the code again?

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+  deliverable, not a defect.
+- Ask in one round of at most five questions, ordered by how much the answer changes
+  the shape of the feature. Do not interview.
+- **No implementation in the product spec.** No file paths, symbols, libraries, event
+  types, API shapes or framework names. If a requirement cannot be stated without them,
+  it is not a product requirement.
+- Every user story must be independently deployable and demonstrable on its own. A
+  story that only makes sense once another story ships is not a story; merge it.
+- More than four priority levels means more than one feature. Split before `draft`.
+- Never invent an SC number. Ask for the target, or mark it
```
Johennes, line 149: What is SC?

Johennes: I guess it's "success criteria"? We might want to make it obvious here.

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+  `[NEEDS CLARIFICATION]`.
+
+For protocol and infrastructure work, success criteria may be expressed as observable
+client behaviour or a wire-level invariant ("a client joining mid-session converges
+within one sync"), but still never as a code path or an internal API.
+
+`FR-###` and `SC-###` ids are immutable once assigned. Product may revise the document
```
Johennes, line 156: Likewise FR (functional requirement?) also appears sort of out of context here.

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+## Traceability
+
+The product spec states intent; the feature spec states what must be true and how it is
+checked. They are kept in sync by ids, not by duplication.
+
+- Every `FR-###` maps to at least one acceptance criterion. Tag the AC with its
+  origin: `AC2 [FR-003] — <observable behaviour>`.
+- Every `SC-###` maps to an entry in `## Measurements` with a method. An SC with no
+  measurement blocks `exploration.signoff`.
+- Every user story maps to at least one slice. **Slice order follows story priority**
```
Johennes, line 174: What is a "slice"?

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+   - map to at least one acceptance criterion, and
+   - land behind a flag or be inert if it is not the last slice of a user-visible
+     change.
+4. **Implement one slice at a time**, branch `spec/<spec-slug>/<n>-<short>` off the
+   pinned `head`. Tests for the slice's ACs go in the same PR as the code. A PR merges
+   only when every check of every AC it lists passes on that PR: automated checks in
+   CI, manual checks performed by a human reviewer with the result recorded in the PR
+   description. An AC is *covered* once such a PR has merged.
+5. **Open the PR** linking the spec path and listing the ACs it covers. Update
+   `## PRs` and `status`.
```
Johennes, line 309: Should this be using stacked PRs?

Johennes (approved):
I guess since we're trying this out and will most definitely iterate on it, I have no reason to push back on it. Other than the lot of small comments I left, this does feel like _quite_ a large and complicated process for a first shot to me.

FEATURES_SPEC/_TEMPLATE.product.md
Johennes: Or maybe I just have it the wrong way around in my head..

FEATURES_SPEC/_TEMPLATE.product.md
daniellekirkwood: i think also there's a point in the agents.md that says too many user stories highlights it should be broken up into smaller features?

FEATURES_SPEC/_TEMPLATE.product.md
daniellekirkwood: The Key Entities part is standard with the SpecKit file and I've always removed it in the Web specs -- we probably don't need it at all. Capturing a "Terminology" document though might be handy -- maybe that sits in a different document and can be referenced

FEATURES_SPEC/_TEMPLATE.product.md
daniellekirkwood: Agreed, the measurable outcomes section is always quite hard because we're limited on what we can measure. Maybe instead we require that the success criteria has measurable outcomes as a part of that section?

I'd give an example but I'm lost with GH files so can't see the SC of the feature_spec linked in the description...

FEATURES_SPEC/AGENTS.md
daniellekirkwood: Success Criteria -- might be worth typing it out so there's no doubts :)

FEATURES_SPEC/_TEMPLATE.product.md
daniellekirkwood: You're right that I don't think there's a lot in it but imo the purposes are different... The independant test can be used to ensure we've got the user story right and to draft up the design. The AC is the QA validation step ... happy to discuss

FEATURES_SPEC/_TEMPLATE.product.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Functional Requirements
+
+- **FR-001**: System MUST [specific capability, e.g., "allow users to create accounts"]
+- **FR-002**: System MUST [specific capability, e.g., "validate email addresses"]
+- **FR-003**: Users MUST be able to [key interaction, e.g., "reset their password"]
+- **FR-004**: System MUST [data requirement, e.g., "persist user preferences"]
+- **FR-005**: System MUST [behavior, e.g., "log all security events"]
+
+*Example of marking unclear requirements:*
```
daniellekirkwood, line 95: I think it would be better to create a "Decisions" section rather than have FRs that are not confirmed... In the Brainstorming stage the AI puts all unknowns into Decisions, the decisions are made and results are crafted into FRs. This also keeps the tracking for 'why' in the future if we want to revisit that and removes the possibility to miss anything in the FR section

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+- Bullets, not paragraphs. Declarative present tense. No motivation, no background,
+  no summary of what the code currently does beyond locating the change.
+- Name things instead of describing them — a path, a symbol, an event type, a command,
+  an event id. A reference is shorter and more precise than a sentence about it. Which
+  kind of reference is allowed depends on the section; see "Reference durability".
+- One line per decision plus at most one sentence of rationale. If the rationale needs
+  more, the extra goes in the sidecar.
+- No judgement adjectives (clean, elegant, proper, obviously) and no hedging (should
+  probably, might be nice). Either it is a constraint or it is not in the spec.
+- `## Problem` is at most five bullets.
```
daniellekirkwood, line 58: I'm not sure we need to build this constraint at 5 maybe "Problem is clear and concise"

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+- Target under 200 lines for the whole spec; over 300 lines, split the feature or move
```
daniellekirkwood, line 59: I like the idea to limit the lines and anything that's too big should be separated -- but i wonder if there's a way we say this thats not lines based as you can still have a complicated spec at 200 lines... 
- The spec should be easy to read and grok by anyone in the team, including our sales team members
- The spec should not be endless, it should be complete but concise. Where lots of ideas are merging to cause complications, split these to a different feature

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+- `## Decisions`, `## Acceptance criteria` — levels 1 to 4 only. No file paths, no
+  internal symbols. This is the part that has to hold against a future `main`.
+- An acceptance criterion that is not user-observable is tagged `[internal]` and
+  anchored by **its test name**, not by the symbol under test. The test name moves with
+  a refactor; the symbol does not.
+
+`[internal]` ACs are legitimate and expected: an invariant that no current consumer can
+observe is exactly what protects the consumer that does not exist yet. Tagging them
+keeps visible which ACs defend a product requirement and which defend a code
+invariant.
```
daniellekirkwood, line 94: I don't really understand this section

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+superseded_by: null
+---
+```
+
+`base` is mandatory and is what makes drift detectable. `owner` is a human even when
+every section was drafted by an agent.
+
+## Brainstorming: the product spec
+
+The usual entry point is a human handing an agent a one-line feature idea. The output
```
daniellekirkwood, line 122: 'human handing an agent a feature idea.' 

I do not intend to hand over 1 line -- that's not going to be good enough imo. We start with the regular brainstorming session <-- happy to share what this looks like for me!

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+of that session is `<stem>.product.md`, not code and not a feature spec. Copy
+`_TEMPLATE.product.md` and fill it.
+
+Ownership: an agent may create and fill this file during `brainstorm`. From `draft`
+onwards **product owns it and agents do not edit it** — not to fix wording, not to
+resolve a marker, not to reconcile it with the code. Agents raise mismatches in the
+feature spec's `## Open questions`.
```
daniellekirkwood, line 129: I think we found in Experiment 1 that if the product changes the 'draft' then we must start over?

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+
+If product's source of truth lives outside the repo (a tracker, a design doc), put its
+URL in the header `**Source**:` line and keep the requirement text and ids in the repo
+anyway. The ids are what the feature spec references; a link alone cannot be traced.
+
+Agent behaviour during the session:
+
+- **Ask, do not assume.** An unanswered question is recorded as
+  `[NEEDS CLARIFICATION: <the question>]`, inline where the answer would go. Filling a
```
daniellekirkwood, line 138: it would be good to cpature these in one place -- the Decisions area -- then they are clearly captured and can be viewed by others easily. when clarified the content is copied to the appropriate place

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+  gap with a plausible guess is the main failure mode of this state — the marker is a
+  deliverable, not a defect.
+- Ask in one round of at most five questions, ordered by how much the answer changes
```
daniellekirkwood, line 141: we found in practice that it asks 5 questions, then stops and puts the rest in the document as 'needs clarification' . I think we either have to have all the 'needs clarification' questions asked, or none asked but limiting to 5 seems strange imo

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+  the shape of the feature. Do not interview.
+- **No implementation in the product spec.** No file paths, symbols, libraries, event
```
daniellekirkwood, line 143: if we already know we want to use a particular design component, should this be named? that feels like it might be helpful?

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+id** and the old one is marked superseded — never edited in place. Renumbering silently
+breaks every AC that references it, and there is no way to detect that afterwards.
+
+A new or changed FR after `implement` does not amend the current spec's scope. It is
+either a new feature spec or the next iteration; the mapping table simply shows an
+unmapped FR, which is the intended signal.
+
+## Traceability
+
+The product spec states intent; the feature spec states what must be true and how it is
```
daniellekirkwood, line 167: what's the difference between the product spec and the feature spec? are these 2 documents?

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+checked. They are kept in sync by ids, not by duplication.
+
+- Every `FR-###` maps to at least one acceptance criterion. Tag the AC with its
```
daniellekirkwood, line 170: this feels like overkill -- there will be AC for each user story that helps us (humans) to check that we all understand the user story and there's high level ACs. The Functional requirements cover all of those + the edge cases and items such as a11y... one AC may have several FRs

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+  origin: `AC2 [FR-003] — <observable behaviour>`.
+- Every `SC-###` maps to an entry in `## Measurements` with a method. An SC with no
```
daniellekirkwood, line 172: We may not be able to measure everything. I think it would be better to craft the SC with any measurements directly and remove the Measure section

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+
+Exploration has sub-states, written dotted in `status`:
+
+```
+exploration.scoping ─► exploration.iterating ─► exploration.signoff
+        ▲                       │                      │
+        └───────────────────────┴──────────────────────┘
+```
+
+- `brainstorm` → `draft`: human. Requires `<stem>.product.md` to exist with at least one
```
daniellekirkwood, line 203: if i understand correctly from our first experiment this week once the product spec is locked to draft any further changes would require us redoing the brainstorming phase, if that is accurate then we cannot allow any open questions in the spec when it moves to draft.

I'm not sure this is the right flow though as we will find things in the exploration phase that will cause the spec to change...

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+  P1 story carrying acceptance scenarios, every FR and SC id assigned, and every
+  remaining `[NEEDS CLARIFICATION]` copied into the feature spec's `## Open questions`.
+- `draft` → `exploration.scoping`: human. Requires `## Problem` and `## Out of scope`.
+  Scoping before the branch exists is what keeps exploration from becoming the
+  long-lived branch this folder replaces.
+- `exploration.scoping` → `.iterating`: human. `experiment_ref` must name the branch.
+- `.iterating` → `.signoff`: human. Requires `## Decisions`, `## Acceptance criteria`,
+  `## Rejected alternatives` and `## Measurements` to be complete, no
```
daniellekirkwood, line 211: i think we remove Measurements here as a must-have and make it a nice-to-have

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+- `qa` → `in_production`: human. The flag is on by default in production.
+- `qa` → `implement`: (agent) when a QA finding needs code. See "QA feedback".
+- any active state → `blocked`: (agent) when drift or an unresolved question prevents
+  progress. Requires a drift-log entry and `blocked_from: <state>`. Do not guess and
+  continue. Leaving `blocked` returns to `blocked_from`, never to a fixed state.
+- any state → `abandoned`: human only.
+
+`in_production` is terminal. A production regression does not reopen the spec: it
+either is an ordinary bug fix with no spec, or it is a new spec. The acceptance
+criteria remain as the regression contract (see "QA feedback").
```
daniellekirkwood, line 230: this might be covered later in the document but do we mark specs as 'retired' or 'superceded' when there's a new spec that builds on top of the current one?

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+
+## QA feedback
+
+The point of the `qa` state is not to record that testing happened. It is to make the
+spec a better regeneration source than it was before implementation. So:
+
+- Every reproduced QA finding becomes **either** a new acceptance criterion **or** a
```
daniellekirkwood, line 237: I don't understand...

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+appended with `[prod: <ref>]` provenance without any state change.
+
+## Required sections
+
+Order matters — agents read top-down and the first three sections are the contract.
+
+1. **`## Problem`** — up to five bullets. What breaks today, for whom, observably,
+   with the file or symbol where it breaks.
+2. **`## Decisions`** — the choices the exploration settled, each with a one-line
+   rationale. **Human-owned. An agent must never add, remove or reword a decision.**
```
daniellekirkwood, line 258: If we rely on the agent to write all the other parts of the document, i don't think we should then rely on the humans to remember to add decisions here -- i think the agent can help us

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+   this section is empty for a spec with an `experiment_ref`, the spec is incomplete.
+5. **`## Measurements`** — numbers from the experiment with the method used to obtain
+   them (command, hardware, load profile). Numbers without a method are not usable as
+   regression baselines and must be marked `unverified`. A baseline states its direction: a
+   count that unrelated work grows — tests passing, test files — is a floor (`≥ n`); a
+   cost — latency, CPU, bundle size — is a ceiling (`≤ n`). An exact count is stale as
+   soon as `main` moves and is not a drift finding.
+6. **`## Out of scope`** — explicit non-goals, so agents stop expanding the diff.
+7. **`## Slicing plan`** — ordered, independently mergeable PRs (see below).
+8. **`## Drift log`** — append-only, agent-written.
```
daniellekirkwood, line 277: what's a drift log?

FEATURES_SPEC/AGENTS.md
```diff
@@ -0,0 +1,415 @@
...
+exploration.
+
+## Hard rules
+
+- Never fill a `[NEEDS CLARIFICATION]` marker with a guess, in any state.
+- Never put file paths, symbols or library names in the product spec.
+- Never reference a file path or an internal symbol in `## Decisions` or
+  `## Acceptance criteria`. Behaviour, protocol identifiers, concepts and test names
+  only.
+- Never edit `<stem>.product.md` after `draft`, and never renumber an FR or SC id.
```
daniellekirkwood, line 358: the exploration phase will highlight spec changes so i'm not sure this can be a hard rule?

FEATURES_SPEC/AGENTS.md
Johennes: IIUC all of these come from SpecKit. But nothing seems to currently explicitly reference SpecKit. If we had a link to it somewhere, that might suffice, too.

FEATURES_SPEC/_TEMPLATE.product.md
Johennes: Ok, sounds fine then, I guess.

FEATURES_SPEC/AGENTS.md
fkwp: I agree. However, Id say this is something for the 2nd iteration... because we also need a place where those terms are living :-)

FEATURES_SPEC/AGENTS.md
fkwp: The spec is locked after the exploration

FEATURES_SPEC/AGENTS.md
fkwp: a QA finding is either 
- a missing AC otherwise the (QA would have found nothing) 
- or it is missing functionality / another flow -> rejected alternatives

FEATURES_SPEC/AGENTS.md
fkwp: superceded is possible, retired is something new

FEATURES_SPEC/AGENTS.md
fkwp: which is fine -- then the FR-## point to the very same AC

PR #4273, opened 2026-09-21
Simplify local transport code

robintown:
As of the removal of 'Legacy' mode, there can no longer be a difference between the transport you advertise in your membership and the transport you publish media on, so the local transport code can be simplified considerably. For instance, the function which gets the local transport can simply return a promise rather than being reactive.

I kept the local transport as an Observable in other modules so that they could easily be tested with existing marble tests.


src/state/CallViewModel/CallViewModel.ts
```diff
@@ -599,17 +599,7 @@ export function createCallViewModel$(
...
     scope: scope,
     connectionFactory: connectionFactory,
-    localTransport$: scope.behavior(
-      localTransport.active$.pipe(
-        catchError((e: unknown) => {
-          logger.info(
-            "could not pass local transport to createConnectionManager$. localTransport$ threw an error",
-            e,
-          );
-          return of(null);
```
Johennes, line 609: This appears to have swallowed any errors. I think the new code passes them through unhandled now?

src/state/ObservableScope.ts
```diff
@@ -67,6 +68,17 @@ export class ObservableScope {
...
+  private readonly shareReplayImpl: MonoTypeOperator = shareReplay({
+    bufferSize: 1,
+    refCount: false,
+  });
+  /**
+   * Shares (multicasts) the Observable as a hot Observable, replaying the most
+   * recently emitted value upon subscription.
+   */
+  public readonly shareReplay: MonoTypeOperator = (input$) =>
+    input$.pipe(this.bindImpl, this.shareReplayImpl);
```
Johennes: This doesn't seem to be used anywhere.

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
```diff
@@ -170,7 +178,7 @@ export function createConnectionManager$({
         "ConnectionManager connections$",
         function* (transports) {
           for (const transport of transports) {
-            if (isLocalTransportWithSFUConfig(transport)) {
+            if (isLocalTransport(transport)) {
               // This is the local transport; only the `LocalTransportWithSFUConfig` has a `sfuConfig` field.
```
Johennes: Nit: Still using the old name here.

src/state/ObservableScope.ts
robintown: 0211d5c367be4955cbd2a436033249c025ae3f6c

src/state/CallViewModel/remoteMembers/ConnectionManager.ts
robintown: 3a189d155bc9b5c8c13e300833135bb8d4d938b3

src/state/CallViewModel/CallViewModel.ts
robintown: Good point, fixed in c9f911b7c97788835ecaf97276eaec02169edc53

Johennes (approved):
🚢

